PR #2, opened 2019-08-28
BLUEBUTTON-1086: restructure Ansible roles

karlmdavis:
Rename and relocate the two main shared Ansible roles, so that they can be reused throughout the project.

https://jira.cms.gov/browse/BLUEBUTTON-1086


karlmdavis:
Not submitting monorepo migration changes like this for review, as the renames make effective review pretty much impossible. Merging, though I'll note that this hasn't (and can't, really) be tested yet, so it's probably a bit off the mark.

PR #3, opened 2019-08-28
BLUEBUTTON-1086: renamed Ansible playbook directories

karlmdavis:
Just shortening & simplifying the names, to make things less verbose.

https://jira.cms.gov/browse/BLUEBUTTON-1086


karlmdavis:
We're outside of the normal review processes for the duration of this monorepo migration, so merging.

PR #8, opened 2019-09-03
BLUEBUTTON-1218 Migration DNS Zone

RickHawesUSDS:
**Why**

To gradually migrate partners between the HealthApt and CCS environments. The plan is contained in https://confluence.cms.gov/display/BB/Traffic+routing+between+HealthApt+and+CCS.

**What Changed**
- A new env/migration folder deploys the migration DNS zone per environment. The main.tf contains the weights.  
- A new migration module drives the weight policy
- The DNS resource has a bunch more features. 
- See https://gist.github.com/RickHawesUSDS/9669dc8683b23ff40861badd44ce80f3 for the Terraform plan

**Decisions Made**
- One zone per environment with the apex record resolving to the CCS environment
- One record per partner to allow complete flexibility per partner
- Use the Route53 alias feature to allow for health checks 
- Registered bfdcloud.net domain to allow us to start. Started the process to get bfd.cmscloud.gov.

**Work to Do**
BLUEBUTTON-1247 Create an equivalent to hhsdevcloud.us.

### DIFF
--- a/ops/terraform/modules/resources/dns_pairs/main.tf
+++ b/ops/terraform/modules/resources/dns_pairs/main.tf
@@ -1,0 +1,37 @@
+# Create weighted CNAME record pairs with 0-100 for weights. 
+#
+resource "aws_route53_record" "record_a" {
+  for_each  = var.weights
+  zone_id   = var.zone_id
+  name      = each.key
+  type      = "A"
+
+  alias {
+    name                    = var.a_alias
+    zone_id                 = var.a_zone_id
+    evaluate_target_health  = true
+  }
+
+  set_identifier            = var.a_set
+  weighted_routing_policy {
+    weight                  = min(100, each.value)
+  }
+}
+
+resource "aws_route53_record" "record_b" {
+  for_each  = var.weights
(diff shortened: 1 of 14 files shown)


ops/terraform/modules/resources/dns_pairs/main.tf
```diff
@@ -0,0 +1,37 @@
+# Create weighted A-record pairs with 0-100 for weights. 
+#
+resource "aws_route53_record" "record_a" {
+  for_each  = var.weights
+  zone_id   = var.zone_id
+  name      = each.key
+  type      = "A"
```
jeskew-gov: Do these need to be CNAME records since the alias is an FQDN instead of an IP address?

RickHawesUSDS: Good question. Can you ping me offline about this. Since the target is an Alias, I think A records work. At least they do in my testing. It saves a step in DNS resolution.

jeskew-gov: My mistake, R53 will resolve the alias to an IP before responding.

PR #10, opened 2019-09-03
Remove RDS storage autoscaling, implement RDS import mode, tweak test/prod RDS params for import

njdister:
This PR does 3 things:  
* Removes RDS storage autoscaling completely by removing the `max_allocated_storage` parameter in favor of directly setting `allocated_storage`.  We have a large existing dataset and to import it properly storage needs to be allocated up front.  
* Implements an RDS import mode on master nodes.  This is a feature that when enabled allows for immediate application of RDS parameters, and sets several RDS parameters tuned for import per this guide: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL.Procedural.Importing.html#PostgreSQL.Procedural.Importing.Copy  
* I have also tweaked RDS instance size, storage size, and IOPS parameters based feedback, though I think these are still subject to change (definitely post-import).  

Let me know if you think this could be structured better, or if any of the parameters seem off.

### DIFF
--- a/ops/terraform/env/prod/stateful/main.tf
+++ b/ops/terraform/env/prod/stateful/main.tf
@@ -10,11 +10,10 @@ provider "aws" {
 module "stateful" {
   source = "../../../modules/stateful"
 
-  # Large DB
   db_config = { 
     instance_class    = "db.r5.24xlarge"
-    iops              = 5000
-    allocated_storage = 10000
+    iops              = 32000
+    allocated_storage = 16000
   }
 
   env_config = {
--- a/ops/terraform/env/prod/stateful/main.tf
+++ b/ops/terraform/env/prod/stateful/main.tf
@@ -12,7 +12,7 @@ module "stateful" {
 
   db_config = { 
     instance_class    = "db.r5.24xlarge"
-    iops              = 32000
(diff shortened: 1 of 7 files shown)


njdister:
PROD-SBX terraform plan: https://gist.github.com/njdister/301104115e8cb18c52c87022c1276b3b  
TEST terraform plan: https://gist.github.com/njdister/49b7be718e81bf59b781a1da33a84977  
PROD terraform plan: https://gist.github.com/njdister/8c763bebf5f5ab857f1f67d8742ee199

ops/terraform/env/prod/stateful/main.tf
```diff
@@ -10,11 +10,15 @@ provider "aws" {
 module "stateful" {
   source = "../../../modules/stateful"
 
-  # Large DB
   db_config = { 
     instance_class    = "db.r5.24xlarge"
-    iops              = 5000
-    allocated_storage = 10000
+    iops              = 16000
+    allocated_storage = 16000
```
RickHawesUSDS: I've had issues where the iops/storage ratio wasn't accepted. I'd be curious if this will work.

PR #13, opened 2019-09-03
Fix RDS import mode issues

njdister:
Quick fix for two things observed:  
* Two parameters required `pending-reboot` rather than `immediate` apply method  
* It makes more sense if the read replicas are in lock-step with master during import mode so changes like IOPS and storage size can be propagated immediately  

I have already applied this in PROD-SBX to get going.

### DIFF
--- a/ops/terraform/modules/stateful/main.tf
+++ b/ops/terraform/modules/stateful/main.tf
@@ -185,6 +185,7 @@ resource "aws_db_parameter_group" "import_mode" {
   parameter {
     name  = "max_wal_size"
     value = "256"
+    apply_method = "pending-reboot"
   }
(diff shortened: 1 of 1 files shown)


ops/terraform/modules/stateful/main.tf
```diff
@@ -243,7 +245,7 @@ module "replica1" {
 
   vpc_security_group_ids = local.db_sgs
 
-  apply_immediately    = false
+  apply_immediately    = var.db_import_mode.enabled ? true : false
```
RickHawesUSDS: Could be written as `apply_immediately = var.db_import_mode.enabled`.

PR #19, opened 2019-09-06
BBrooks/BLUEBUTTON-1215: Adding AWS SSM Parameter Store Support

cthulhuplus:
Fixes BLUEBUTTON-1215
Adds AWS SSM Parameter Store Support. Pushing parameters to the Parameter store using Jenkins, and pulling values from the Parameters store using Ansible.

Change Details
Added code to allow Jenkins to push parameters to the AWS parameter store; changes made to terraform for a KMS key, IAM role, etc..., created a Jenkins file for doing the pushing, added an Ansible role with multiple strategies for pulling values from the Parameter store.

Security Implications
There are no security implications other than adding KMS key and modifying the Jenkins IAM role to have Parameter store permissions.

Acceptance Validation
I know this is done when: 

a single AMI (per application) can be used to launch our applications across all environments. 

Feedback Requested
Please review


README.md
```diff
@@ -16,6 +16,99 @@ Run the following commands to perform a release:
...
+The parameter names are made from the following:
+* Project Name
+* Environment (dev/prod/sandbox/etc)
+* Date that the value was added in YYYYMMDD format
+* Parameter key
+Example: /weapon-x/dev/20190130/subject
+
+Log into Jenkins and upload your AWS SSM Parameter Store json file to the Managed Files section of Manage Jenkins.
+
+The format should look something like this:
```
whytheplatypus: Suggest using terraform or manual entry for things like this. Eventually kms may give another automated solution.

README.md
```diff
@@ -16,6 +16,99 @@ Run the following commands to perform a release:
...
+## Parameter Store
+
+Parameter naming
+
+The parameter names are made from the following:
+* Project Name
+* Environment (dev/prod/sandbox/etc)
+* Date that the value was added in YYYYMMDD format
+* Parameter key
+Example: /weapon-x/dev/20190130/subject
```
whytheplatypus: I'd remove the date aspect. That would allow the params to be discover-able by the system. For example on system spin up it could identify it's "folder" `weapon-x/dev` and load all variables under that as config or env variables, allowing an admin to simply update, remove, or add configuration without a redeploy.

cthulhuplus: There was a reason that the date was being included, I think for versioning, but it's been awhile since I wrote this. I know there was discussion around it and what format that we should be using for the date.

PR #26, opened 2019-09-10
Bluebutton 165 add new relic

cthayes:
This adds New Relic configuration to the BFDS. 

BLUEBUTTON-165


StewGoin:
`Add become root`

I'm assuming this doesn't mean become the root user, but I just wanted to check and make sure -- not a New Relic config guru personally.

cthayes:
> `Add become root`
> 
> I'm assuming this doesn't mean become the root user, but I just wanted to check and make sure -- not a New Relic config guru personally.

It's an ansible connection setting. Basically equivalent to adding 'sudo' to the command it runs on the server.

PR #27, opened 2019-09-11
BLUEBUTTON-1171: SSH static keys

njdister:
This PR makes public keys available for each environment via file and terraform.  Private keys have been uploaded to keybase.  This is fairly barebones and does not update existing terraform resources since that will likely trigger a rebuild.


njdister:
Closing as this needs to be re-scoped.

PR #29, opened 2019-09-12
BLUEBUTTON-1171: SSH users role

njdister:
Looking for feedback on a few things before I get ready to PR this:  
- I'm using `000_cross_env_vars.yml` since it was there, but unsure if this is specific to CCS or a port from HealthApt  
- If there is any intention to dump our public keys elsewhere it may be good to write those out to a file rather than embedding them in a var file  
- I don't foresee this being env specific but welcome to opinions  
- I could see this being run either at AMI bake or during user data, more opinions needed on where to insert this in the build step


njdister:
> Where is the sudoer's list going to be kept?
I have it so sudoers is just a toggle in the ssh_users list rather than being a separate list.

jzulim:
@njdister 000_cross_env_vars.yml feels like the right spot. Each env has a symlink to this file and the variables are loaded during both the pre and post bake step via tags.

jzulim:
> Look's good. Just a question, why the 000 in 000_env_vars.yml?

@RickHawesUSDS - My understanding with how this works.. 

When loading var files from a directory, specifying 000 will lower the priority so that if the same variable is defined for a specific environment, it takes precedence over those found in 000_*

PR #30, opened 2019-09-12
BLUEBUTTON-1163 Instance Snapshots

RickHawesUSDS:
**Why**
For security audits and other security purposes, retain daily snapshots of BFD's instance volumes. 

**What Changed**
- A new global data lifecycle manager policy that takes snapshots of volumes and retains them for 21 days.
- Since DLM policy runs on tag matching, tag all ec2 instance volumes. 
- In order to tag volumes, switched from Launch Configurations to Launch Templates to specify the instances created by Autoscale Groups

The terraform plan is here:
https://gist.github.com/RickHawesUSDS/dad91f312d9ec5ae1b755c00466a6df1

**Choices Made**
- The use of DLM to do manage the snapshots. Could of done this with a Lambda function.
- Use Launch Templates to tag snapshots. Found articles that recommended using Lambdas to do this task.

**Whats Left**
This finishes BLUEBOTTON-1163.

### DIFF
--- a/ops/terraform/env/global/dlm/backend.tf
+++ b/ops/terraform/env/global/dlm/backend.tf
@@ -1,0 +1,11 @@
+terraform {
+  backend "s3" {
+    bucket         = "bfd-tf-state"
+    key            = "global/dlm/terraform.tfstate"
+    region         = "us-east-1"
+    dynamodb_table = "bfd-tf-table"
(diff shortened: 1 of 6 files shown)


RickHawesUSDS:
Added a TF plan in the explanation

PR #38, opened 2019-09-14
BLUEBUTTON-850: Resolve non-unique bucket name test errors

karlmdavis:
It's _almost_ the same as before, except now the "get a random bucket name" code is centralized, and improved:

1. Added a couple decimal places to the random number.
2. Added usernames to the generated strings, so that one person failing to clean up their mess shouldn't impact someone else.

There are still several other things that can cause intermittent test failures, but this was definitely the biggest offender.

https://jira.cms.gov/browse/BLUEBUTTON-850

### DIFF
--- a/apps/bfd-pipeline/bfd-pipeline-rif-extract/src/main/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetTestUtilities.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rif-extract/src/main/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetTestUtilities.java
@@ -19,6 +19,7 @@ import java.net.URL;
 import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Random;
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Marshaller;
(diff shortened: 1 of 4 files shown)


apps/bfd-pipeline/bfd-pipeline-rif-extract/src/main/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetTestUtilities.java
```diff
@@ -30,6 +31,29 @@
...
+    if (username == null || username.isEmpty()) username = "anonymous";
+    username.replaceAll("@", "-");
+    username.replaceAll("\\\\", "-");
+    int randomId = new Random().nextInt(100000);
+    String bucketName = String.format("bb-test-%s-%d", username, randomId);
+
+    Bucket bucket = s3Client.createBucket(bucketName);
+    /*
+     * Note: S3's API is eventually consistent, so any calls to use this new bucket will
+     * intermittently fail for a brief period of time. The only solution to this is to retry all of
```
jeskew-gov: You could also use a [waiter](https://aws.amazon.com/blogs/developer/waiters-in-the-aws-sdk-for-java/) here to pause until the bucket is available. I think it would be:

```
s3Client.waiters()
        .bucketExists()
        .run(new HeadBucketRequest().withBucketName(bucketName));
```

karlmdavis: Interesting!

Does that cover the eventual consistency issue, or will it just verify that whichever endpoint it happens to connect to has the bucket?

jeskew-gov: It will cover DNS and S3 metadata service eventual consistency.

PR #42, opened 2019-09-16
BLUEBUTTON-1135: Increase user level file descriptor limits

jcgeer:
Ansible tasks to increase the user level file descriptor limits (both
hard and soft) to resolve the "too many open files" errors.


jcgeer:
I'm pretty sure this is still an issue with the File Descriptor limits not being large enough so this code is useful, but I think that is only part of the picture, since upping it did not resolve the issue.

njdister:
This is not an issue with CCS as far as I have been aware over the last several months, closing.

PR #50, opened 2019-09-17
BLUEBUTTON-888: Fix cleanup of fda-data temp files.

karlmdavis:
Honestly not sure why the previous shutdown hook wasn't working: it
should have? Regardless, this will hopefully be more reliable.

https://jira.cms.gov/browse/BLUEBUTTON-888
https://jira.cms.gov/browse/BLUEBUTTON-889


karlmdavis:
Note: this is a lot of changed lines, but basically, I just extracted the previous code to a new method, wrapped it in a `try-finally` block, and changed the cleanup method to something a bit more concise.

PR #56, opened 2019-09-18
Migrate from CloudWatchLogs agent to CloudWatch unified agent

jeskew-gov:
This PR updates the CCS ansible playbooks to forgo baking the CloudWatchLogs agent into the platinum AMI and to include the CloudWatch unified agent in the bfd-pipeline, bfd-server, and Jenkins build targets.

The move to the unified agent was motivated by a lack of visibility yesterday into our disk usage on the Jenkins server. Neither EC2 nor EBS offer vended metrics on disk usage, so we need to collect this using an installed agent. I've configured the agent to collect [everything in the "standard" metrics set](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/create-cloudwatch-agent-configuration-file-wizard.html) as well as the log that were previously collected by the CloudWatchLogs agent.


jeskew-gov:
@jzulim Tweaked a few things based on tests in Jenkins. Everything looks to be building correctly in the latest version.

jzulim:
> @jzulim Tweaked a few things based on tests in Jenkins. Everything looks to be building correctly in the latest version.

@jeskew-gov LGTM. Curious, why install / apply the cloudwatch agent for each instance type instead of within the platinum ami? I'm sure you have a reason for this suggested change.

jeskew-gov:
> @jeskew-gov LGTM. Curious, why install / apply the cloudwatch agent for each instance type instead of within the platinum ami? I'm sure you have a reason for this suggested change.

I didn't think the metrics would be useful without an `env` dimension attached, and the platinum AMI is used for all environments. I think you have to overwrite the full config to add a new dimension to all metrics, which would lead to some repetition for the base level of metrics collected. 

Definitely open to moving the install to the platinum AMI if you have a strong opinion there.

PR #58, opened 2019-09-20
BLUEBUTTON-1157: Make deployed commit available to user_data script

jzulim:
# What

This PR supplies the specific git branch and git commit used during the build pipeline and makes it available to terraform to pass along to user_data 

# Why

... so that during post bake of the AMI this commit is checked out to finalize the deployment thus keeping all stages of the pipeline in sync with a specific branch / commit. 

# Testing

Built and Deployed this PR to TEST with successful results. Jumped on instance to confirm the branch was cloned and commit was checked out and used to complete the user_data script. 

# Improves

BLUEBUTTON-1157

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -146,7 +146,9 @@ stage('Deploy to TEST') {
 	milestone(label: 'stage_deploy_test_start')
 
 	node {
-		scriptForDeploys.deploy('test', amiIds, appBuildResults)
+		gitBranchName = env.BRANCH_NAME
+		gitCommitId = env.GIT_COMMIT
+		scriptForDeploys.deploy('test', gitBranchName, gitCommitId, amiIds, appBuildResults)
 	}
 
 }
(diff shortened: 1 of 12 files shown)


karlmdavis:
@jeskew-gov: LGTM, but should we be using an autoscaling group (with a min and max capacity of 1) for the ETL server instead of a single instance? It could reduce some of the duplication showing up here (where branch and commitId need to be passed into the EC2 terraform module and the ASG terraform module) and would also automatically bring failing ETL instances back up.

Not unless we think **very carefully** about what will happen when that leads to two of the instances running at the same time (i.e. one unhealthy (but maybe still working) and one healthy one). I've not had a chance to sit down and really think about how it might go sideways, and I recommend against this change until that review has occurred.

karlmdavis:
FYI: I don't have time to review today, but feel free to proceed if you've tested it, given the other reviews.

PR #64, opened 2019-09-23
Startup smoke testing for data server

njdister:
Getting this out there for feedback.  The main issue right now is there are a few values that are hardcoded outside of the scope of the bfd-server role (local testing cert and wildfly version) so I can't think of a good way of grabbing them.

### DIFF
--- a/ops/ansible/roles/bfd-server/tasks/appserver_install.yml
+++ b/ops/ansible/roles/bfd-server/tasks/appserver_install.yml
@@ -20,6 +20,18 @@
   tags: 
     - pre-ami
 
+- name: Create App Server Startup Wrapper Script
+  template:
+    src: bluebutton-appserver-start.sh.j2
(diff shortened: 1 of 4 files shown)


ops/ansible/roles/bfd-server/tasks/appserver_install.yml
```diff
@@ -20,6 +20,18 @@
...
 
+- name: Create App Server Startup Wrapper Script
+  template:
+    src: bluebutton-appserver-start.sh.j2
+    dest: "{{ data_server_dir }}/bluebutton-appserver-start.sh"
+    owner: "{{ data_server_user }}"
+    group: "{{ data_server_user }}"
+    mode: u=rwx,g=rwx,o=rx
+  become: true
+  when: data_server_startup_wrapper
```
RickHawesUSDS: What happens when this is false? or undefined?

PR #65, opened 2019-09-24
Fix-delete-on-termination

RickHawesUSDS:
**Why**
Nathan noticed that volumes where accumulating a very fast rate. 

**What Changed**
Change the delete on termination flag on ETL and FHIR servers.


njdister:
> I'm just curious: why are we setting this in two different places? I understand what the setting on the ASG is doing, but I'm less clear about what the other one is used for.

The EC2 resource is a different deployment type used by the ETL server.

karlmdavis:
> The EC2 resource is a different deployment type used by the ETL server.

Oh, of course -- duh. Thanks!

PR #67, opened 2019-09-24
Refactor how Jenkinsfile populates branch name and commit id vars. 

jzulim:
# Why

When building PR's we need the BRANCH NAME passed to the user_data script, and not the JOB NAME. Added a step to check if the build is a PR, if so switching to the env.CHANGE_BRANCH var and if not, using env.BRANCH_NAME. 

I tested this PR and this branch to confirm both methods work as expected.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -163,7 +163,6 @@ stage('Deploy to TEST') {
 	milestone(label: 'stage_deploy_test_start')
 
 	node {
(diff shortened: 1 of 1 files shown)


karlmdavis:
One other comment: I would have just put this in the existing "Prepare" stage. NBD, but changing the stages resets the history view on https://builds.bfd-management.oeda.cmscloud.local/jenkins/job/test/job/master/

Sorry I didn't catch that before merging.

PR #69, opened 2019-09-25
Fix CW permissions and build CW agent into platinum AMI

jeskew-gov:
Verified in the AWS console that logs from the FHIR server began showing up in CloudWatch once I manually attached the `CloudWatchAgentServerPolicy` managed IAM policy to the relevant instance profile role in the `test` environment.

Based on feedback from @jzulim on a previous PR, I've moved the agent installation tasks to be part of the platinum AMI build and generalized the initial config to just collect `/var/log/messages` and `/var/log/secure`. Metric collection strategies will vary by server role (e.g., we want `netstat` metrics from the servers terminating SSL connections but not from the ETL server), so those have been moved to the role-specific configuration files.


njdister:
@jzulim Can this be closed or do you think this can be reconciled with the current state of infra code?

jzulim:
Closing this PR, created Jira Ticket BLUEBUTTON-1828

PR #70, opened 2019-09-25
Fix to always pull last AMI ID from packer manifest

njdister:
I think this bug occurred because there was an assumption that packer would overwrite the old manifest on every run.  However packer is "intelligent" in that it actually appends to an existing manifest if detected.  This ensures that if the manifest has multiple entires, we always pull the last addition.

Tested in Jenkins script console.


karlmdavis:
Merging without a successful build, as I might as well test it on my Jetty branch.

PR #78, opened 2019-09-27
BLUEBUTTON-1315: Encrypt AMIs

njdister:
This does two things:
- Encrypts the platinum AMI with the mgmt KMS key
- Ensures that at launch time, the root volume of an instance in an ASG is then encrypted with the per environment KMS key, rather than inheriting encryption from the mgmt key

Alternately we could apply the per environment KMS key at application AMI build, but this solution makes more sense (and is simpler) if we are planning to move to environment agnostic AMIs after implementing SSM.


njdister:
Need to test this PR in Jenkins since it affects the build pipeline.

njdister:
Tested successfully, merging.

PR #83, opened 2019-10-01
BLUEBUTTON-1271: Startup smoke testing; Jetty edition

njdister:
This conditionally injects a backgrounded function into the startup routine that will make a number of attempts (with various configurable timeouts) to validate that one metadata endpoint and one database backed endpoint return expected values.  If not, it will attempt to safely stop the process as to prevent a successful deploy.

Need some feedback on a few things before it's readier to merge:
* Are the default timeouts sane?
* We probably want to check for a larger/more complete value, however I'm not sure what will be consistent between all environments.
* Should we attempt a more drastic process kill if safe kill fails?
* Any other error handling I'm missing?

### DIFF
--- a/ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml
+++ b/ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml
@@ -46,7 +46,3 @@ data_server_ssl_client_certificates:
     certificate: "{{ lookup('file', 'files/client_data_server_performance_tester_certificate.pem') }}"
 
 data_server_startup_testing: true
-data_server_startup_testing_retries: 3
-data_server_startup_testing_boot_timeout: 30
-data_server_startup_testing_req_timeout: 15
-data_server_startup_testing_req_backoff_timeout: 10
--- a/ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml
+++ b/ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml
@@ -44,5 +44,3 @@ data_server_ssl_client_certificates:
     certificate: "{{ lookup('file', 'files/client_data_server_bluebutton_frontend_test_certificate.pem') }}"
   - alias: client_performance_tester
     certificate: "{{ lookup('file', 'files/client_data_server_performance_tester_certificate.pem') }}"
-
-data_server_startup_testing: true
--- a/ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml
+++ b/ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml
@@ -44,3 +44,9 @@ data_server_ssl_client_certificates:
     certificate: "{{ lookup('file', 'files/client_data_server_bluebutton_frontend_test_certificate.pem') }}"
   - alias: client_performance_tester
     certificate: "{{ lookup('file', 'files/client_data_server_performance_tester_certificate.pem') }}"
+
+data_server_startup_testing: true
+data_server_startup_testing_retries: 3
+data_server_startup_testing_boot_timeout: 30
+data_server_startup_testing_req_timeout: 15
+data_server_startup_testing_req_backoff_timeout: 10
--- a/ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
+++ b/ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
@@ -33,7 +33,6 @@ export NEW_RELIC_ENVIRONMENT='{{ data_server_new_relic_environment }}'
 {% endif %}
 {% endif %}
 
-{% if data_server_startup_testing is defined and data_server_startup_testing is sameas true %}
 # Begin smoke testing startup routine
 STARTUP_TESTING_RETRIES='3'
 STARTUP_TESTING_BOOT_TIMEOUT='30'
@@ -102,7 +101,6 @@ stop_service_if_failing() {
 
 (stop_service_if_failing >>"{{ data_server_dir }}/bfd-server-startup.log" 2>&1) &
 # End smoke testing startup routine
-{% endif %}
 
 # Set some additional variables.
 JVM_ARGS='{{ data_server_appserver_jvmargs }}'
--- a/ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
+++ b/ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
@@ -74,7 +74,7 @@ stop_service_if_failing() {
     STARTUP_TESTING_CHECK_METADATA=$(check_endpoint "https://localhost:7443/v1/fhir/metadata?_format=application%2Fjson%2Bfhir" "status")
     STARTUP_TESTING_CHECK_METADATA_EXIT=$?
 
-    log "Checking database endpoint"
+    log "Checking coverage resource endpoint for bene $STARTUP_TESTING_BENE_ID"
     STARTUP_TESTING_CHECK_DATABASE=$(check_endpoint "https://localhost:7443/v1/fhir/Coverage?beneficiary=$STARTUP_TESTING_BENE_ID&_format=application%2Fjson%2Bfhir" "id")
     STARTUP_TESTING_CHECK_DATABASE_EXIT=$?
 
--- a/ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
+++ b/ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
@@ -45,35 +45,35 @@ check_endpoint() {
 }
 
 stop_service_if_failing() {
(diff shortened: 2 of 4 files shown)


ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
```diff
@@ -33,6 +33,55 @@ export NEW_RELIC_ENVIRONMENT='{{ data_server_new_relic_environment }}'
...
 
+{% if data_server_startup_testing is defined and data_server_startup_testing is sameas true %}
+# Begin smoke testing startup routine
+STARTUP_TESTING_RETRIES='{{ data_server_startup_testing_retries | default(3) }}'
+STARTUP_TESTING_BOOT_TIMEOUT='{{ data_server_startup_testing_boot_timeout | default(30) }}'
+STARTUP_TESTING_REQ_TIMEOUT='{{ data_server_startup_testing_req_timeout | default(15) }}'
+STARTUP_TESTING_REQ_BACKOFF_TIMEOUT='{{ data_server_startup_testing_req_backup_timeout | default(10) }}'
+
+check_endpoint() {
+  curl --max-time $STARTUP_TESTING_REQ_TIMEOUT --silent --insecure --cert-type pem --cert "{{ data_server_dir }}/bluebutton-backend-test-data-server-client-test-keypair.pem" "$1" | grep "$2"
```
karlmdavis: Does this need Bash's `set -o pipefail` in order to work as expected?

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
```diff
@@ -33,6 +33,55 @@ export NEW_RELIC_ENVIRONMENT='{{ data_server_new_relic_environment }}'
...
+    STARTUP_TESTING_CHECK_METADATA_EXIT=$?
+
+    echo "$(date): Checking database endpoint"
+    STARTUP_TESTING_CHECK_DATABASE=$(check_endpoint 'https://localhost:7443/v1/fhir/Coverage?beneficiary=-201&_format=application%2Fjson%2Bfhir' 'id')
+    STARTUP_TESTING_CHECK_DATABASE_EXIT=$?
+
+    if [[ $STARTUP_TESTING_CHECK_METADATA_EXIT == 0 ]] && [[ $STARTUP_TESTING_CHECK_DATABASE_EXIT == 0 ]]; then
+      echo "$(date): Server started properly"
+      return 0
+    elif [[ $I != $STARTUP_TESTING_RETRIES ]]; then
```
karlmdavis: Think you had a bug here?
```suggestion
    elif [[ $I >= $STARTUP_TESTING_RETRIES ]]; then
```

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
```diff
@@ -33,6 +33,55 @@ export NEW_RELIC_ENVIRONMENT='{{ data_server_new_relic_environment }}'
...
+STARTUP_TESTING_RETRIES='{{ data_server_startup_testing_retries | default(3) }}'
+STARTUP_TESTING_BOOT_TIMEOUT='{{ data_server_startup_testing_boot_timeout | default(30) }}'
+STARTUP_TESTING_REQ_TIMEOUT='{{ data_server_startup_testing_req_timeout | default(15) }}'
+STARTUP_TESTING_REQ_BACKOFF_TIMEOUT='{{ data_server_startup_testing_req_backup_timeout | default(10) }}'
+
+check_endpoint() {
+  curl --max-time $STARTUP_TESTING_REQ_TIMEOUT --silent --insecure --cert-type pem --cert "{{ data_server_dir }}/bluebutton-backend-test-data-server-client-test-keypair.pem" "$1" | grep "$2"
+}
+
+stop_service_if_failing() {
```
karlmdavis:
```suggestion'
##
# Runs test queries to verify that the server is working as expected. If not, issues a 'kill -INT' on the server's process to shut it down.
##
stop_service_if_failing() {
```

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
```diff
@@ -33,6 +33,55 @@ export NEW_RELIC_ENVIRONMENT='{{ data_server_new_relic_environment }}'
...
+  echo "$(date): Waiting $STARTUP_TESTING_BOOT_TIMEOUT seconds for initial startup"
```
karlmdavis: Can you replace all of these `echo`s with a simple logging function like this one?

https://github.com/CMSgov/beneficiary-fhir-data/blob/master/ops/ansible/playbooks-healthapt/dev/stress-test.sh#L80

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
```diff
@@ -33,6 +33,55 @@ export NEW_RELIC_ENVIRONMENT='{{ data_server_new_relic_environment }}'
 {% endif %}
 {% endif %}
 
+{% if data_server_startup_testing is defined and data_server_startup_testing is sameas true %}
```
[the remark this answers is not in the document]
njdister: This was to prevent or easily disable this feature from HealthApt if needed

ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml
```diff
@@ -44,3 +44,9 @@ data_server_ssl_client_certificates:
     certificate: "{{ lookup('file', 'files/client_data_server_bluebutton_frontend_test_certificate.pem') }}"
   - alias: client_performance_tester
     certificate: "{{ lookup('file', 'files/client_data_server_performance_tester_certificate.pem') }}"
+
+data_server_startup_testing: true
+data_server_startup_testing_retries: 3
+data_server_startup_testing_boot_timeout: 30
+data_server_startup_testing_req_timeout: 15
+data_server_startup_testing_req_backoff_timeout: 10
```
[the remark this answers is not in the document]
njdister: I was thinking different environments might behave differently and might need parameters tweaked but perhaps thats premature optimization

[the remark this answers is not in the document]
karlmdavis: Makes sense!

I'd say it's a bit premature, though, and (until it's needed) will make it harder to maintain this code.

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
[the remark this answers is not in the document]
karlmdavis: Interesting! I'm pretty sure that won't be necessary, but I'll push this branch out there and verify.

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
njdister: Oh right, I did this because I wasn't sure if bash considered this value a string?  Looking it up, seems like I just need to change my comparison operator to ensure its evaluated as an integer:

https://www.tldp.org/LDP/abs/html/comparison-ops.html

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
njdister: Technically this should be ok without it, since this is such a short pipe and grep should always fail when curl fails.  Wouldn't hurt to add it though and would catch a weird edge case where the curl fails, but its output still matches the pattern we're looking for.

njdister:
Ran through shellcheck, only errors/warnings are related to jinja templating language.

Need to run this through deploy testing once more.

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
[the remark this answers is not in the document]
karlmdavis: Want to yank this now?

ops/ansible/roles/bfd-server/templates/bfd-server.sh.j2
```diff
@@ -33,6 +33,77 @@ export NEW_RELIC_ENVIRONMENT='{{ data_server_new_relic_environment }}'
...
+  log "Waiting $STARTUP_TESTING_BOOT_TIMEOUT seconds for initial startup"
+  sleep $STARTUP_TESTING_BOOT_TIMEOUT
+
+  for I in $(seq $STARTUP_TESTING_RETRIES); do
+    log "Checking metadata endpoint"
+    STARTUP_TESTING_CHECK_METADATA=$(check_endpoint "https://localhost:7443/v1/fhir/metadata?_format=application%2Fjson%2Bfhir" "status")
+    STARTUP_TESTING_CHECK_METADATA_EXIT=$?
+
+    log "Checking database endpoint"
+    STARTUP_TESTING_CHECK_DATABASE=$(check_endpoint "https://localhost:7443/v1/fhir/Coverage?beneficiary=$STARTUP_TESTING_BENE_ID&_format=application%2Fjson%2Bfhir" "id")
```
RickHawesUSDS: This is testing the Coverage resource end-point, which includes the database connection, but has more. I suggest updating the log comment.

PR #84, opened 2019-10-02
BLUEBUTTON-1233: Enabling Hikari Logging

jcgeer:
Enabling additional logging to help with debugging Hikari

### DIFF
--- a/apps/bfd-server/bfd-server-launcher/src/main/resources/logback.xml
+++ b/apps/bfd-server/bfd-server-launcher/src/main/resources/logback.xml
@@ -56,6 +56,11 @@
 			</layout>
 		</encoder>
 	</appender>
+	
+	<!-- Enabling DEBUG level logging for hikari -->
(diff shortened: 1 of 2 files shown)


apps/bfd-server/bfd-server-launcher/src/main/resources/logback.xml
```diff
@@ -56,6 +56,11 @@
 			</layout>
 		</encoder>
 	</appender>
+	
+	<!-- Enabling DEBUG level logging for hikari -->
```
karlmdavis: Do you know how much log spam this will generate? If it's a lot, please add a comment here explaining when it can be removed, e.g. "TODO: disable this once BLUEBUTTON-1234 is resolved"

jcgeer: It looks like there will be quite a bit of spam. I've added this comment. Thanks.

PR #85, opened 2019-10-02
RFC0000: Since parameter support

RickHawesUSDS:
See the markdown document


rfcs/0000-since-parameter.md
```diff
@@ -0,0 +1,141 @@
...
+
+### ETL Corner Case
+
+Every export job completion record contains a `transactionTime` field that bulk-export clients use as the `_since` parameter in a subsequent export job. The `transactionTime` is the time that the export job starts. The FHIR bulk export specification states that an export SHALL only contain resources updated before the `transactionTime` time. The specification further says that an exporter should delay a job until all pending writes have finished to satisfy this constraint. The ELT feed allows bulk-export implementors to know when the BFD ETL finishes. Since both the bulk export operation and the BFD ETL process take several hours, the delay may be significant. 
+
+To avoid this delay, a bulk-export implementor may take an optimistic approach by immediately starting an export job, but monitoring if it receives a resource updated after the job start time. If it does, the it should delay and restart the job.
```
whytheplatypus: Would requesting resources where `lastUpdated` is between the time specified in the `_since` parameter and the export `transactionTime`?

rfcs/0000-since-parameter.md
```diff
@@ -0,0 +1,141 @@
...
+2. The ETL feed could contain more ETL metadata, such as the FHIR resource touched, the DB table updated, or the care team involved. In the end, it was decided not to put this information into the
+feed because it was not general enough or universally useful. 
+
+## Future Possibilities
+
+This proposal should scale as BFD, and it's partners serve more beneficiaries and clients. It should continue to work as BFD adds more partners and data sources. 
+
+In future releases, BFD may receive claim data faster than the current delay. If the ETL process runs daily instead of weekly as it does today, the algorithms in this proposal should continue to work. If one day, the BFD receives claim data continuously, we should revisit the algorithms of this proposal. 
+
+In discussions with DPC customers, they have asked for notification when the DPC has new beneficiary data. Instead of polling for updates, they would like to have the ability for a push update.  Similarly, FHIR is developing a subscription model that supports webhooks \[[6](#ref6)\]. If a BFD partner wants to develop these features, they can use the ETL feed as a source of information for this feature.
```
whytheplatypus: I think this is a great reason for the feed, and use of s3.

rfcs/0000-since-parameter.md
```diff
@@ -0,0 +1,141 @@
...
+
+![Bulk ](https://www.websequencediagrams.com/files/render?link=zfMUJyQaf18DNUb6IQoPN2EBeq9tMctYXXupx6T5Co8gB3t9ysmhat0ToalxZ6p2)
+
+### BFD API Details
+
+All the proposed API changes follow the FHIR specification. The first improvement is to add the `lastUpdated` to the metadata field of a resource. The current implementation returns the time the that the request was made. The proposal alters this field to be the timestamp that the ETL process wrote to the master DB. Like all instant timestamps, this timestamp must include timezone \[[4](#ref4)\]. 
+
+The second change is to support the `_lastUpdated` query parameter for resource searches \[[5](#ref5)\]. FHIR specifies a set of comparison operators to go along with this filter. Although bulk exporters will only use the `gt` or greater-than operator, a complete set will be useful for other use-cases.
```
whytheplatypus: Is there behavior that can be described here that would allow the BFD to protect itself from misbehaving consumers? Can it be made cheap for the BFD to return nothing if no resources fall within the `_lastUpdated` query?

whytheplatypus:
The overall proposal looks great to me. 

Since it covers two behaviors, the `_since / _lastUpdated` support and the ETL Feed, I think it would be more effective as two proposals that can be discussed and implemented independently.

rfcs/0000-since-parameter.md
RickHawesUSDS: Consider the case where the same record was updated in consecutive loads: after the since time passed in and again after the bulk export job was started (ie. the transactionTime). In this case, the record where would have a lastUpdate time after the transactionTime, so it would not be included in the download, when it should have been.

rfcs/0000-since-parameter.md
RickHawesUSDS: Good point. The BFD server could keep the list of updated beneficiaries itself. So the database scan can be avoided for beneficiaries without a recent update. I like this idea a lot.

RickHawesUSDS:
Closing out. This branch is merged into BFD-1506 branch.

PR #87, opened 2019-10-02
BLUEBUTTON-1288: Create BFD Pipeline EC2 instance after SGs

karlmdavis:
When I rolled out my previous PR to the `test` environment, the BFD Pipeline app failed to launch the first time, because the RDS security group it depended on wasn't in place.

This is a rather gross hack, but should fix it. It's be much simpler if Terraform just supported `depends_on` for modules, but it apparently doesn't yet.

Note: I don't think this workaround will trigger if the SG is changed; it only ensures it's _created_ first.  🤷‍♂ 

### DIFF
--- a/ops/terraform/modules/resources/bfd_pipeline/main.tf
+++ b/ops/terraform/modules/resources/bfd_pipeline/main.tf
@@ -61,5 +61,5 @@ module "ec2_instance" {
   sg_ids          = [aws_security_group.app.id]
 
   # Ensure that the DB is accessible before the BFD Pipeline is launched.
-  dependencies    = ["${aws_security_group_rule.allow_db_primary_access.id}"]
+  ec2_depends_on_1 = "aws_security_group_rule.allow_db_primary_access"
 }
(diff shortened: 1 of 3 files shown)


njdister:
I think if the intention is for the EC2 resource to be generic enough to be used by something other than the ETL pipeline, this is a good workaround.  Otherwise I don't think there isn't anything stopping you from passing the SG dependency straight to `depends_on` in the EC2 resource.

karlmdavis:
> I don't think there isn't anything stopping you from passing the SG dependency straight to `depends_on` in the EC2 resource.

@njdister: What would that `depends_on` block look like? I don't think I can pass a resource reference around a a variable -- at least, Terraform's docs don't list references as a supported variable type.

njdister:
> @njdister: What would that `depends_on` block look like? I don't think I can pass a resource reference into the module as a variable -- at least, Terraform's docs don't list references as a supported variable type.

I was able to do this:
bfd_pipeline main:
```
  sg_dep = aws_security_group_rule.allow_db_primary_access.id
```

ec2 vars:
```
variable "sg_dep" {
  type = string
}
```

ec2_pipeline main:
```
  depends_on                  = [var.sg_dep]
```

And the terraform plan didn't barf.  `depends_on` is not documented very well so I'm not sure if this might silently fail until a subsequent apply in the situation you're trying to fix.

PR #89, opened 2019-10-02
BLUEBUTTON-1288: Lock deploys to the test environment

karlmdavis:
Prevent one deploy from stomping on another. This should fix the "`Error: Error locking state: Error acquiring the state lock: ConditionalCheckFailedException: The conditional request failed`" errors that we've been getting from Terraform.


karlmdavis:
Since we're having GitHub API quota issues, Imma send this one out for review before it's built.

PR #96, opened 2019-10-05
BLUEBUTTON-1288: BFD Pipeline S3 access

karlmdavis:
Main functional difference is that it now has access to the KMS key and permissions to list the RIF bucket.

Also restructured things a bit.

### DIFF
--- a/ops/terraform/modules/resources/bfd_pipeline/main.tf
+++ b/ops/terraform/modules/resources/bfd_pipeline/main.tf
@@ -52,19 +52,19 @@ resource "aws_iam_policy" "bfd_pipeline_rif" {
   "Version": "2012-10-17",
   "Statement": [
     {
-      "Sid": "BFD Pipeline RW S3 RIF: KMS",
+      "Sid": "BFDPipelineRWS3RIFKMS",
       "Action": ["kms:Decrypt"],
       "Effect": "Allow",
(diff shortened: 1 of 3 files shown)


ops/terraform/modules/resources/bfd_pipeline/main.tf
```diff
@@ -29,6 +41,65 @@ resource "aws_security_group_rule" "allow_db_primary_access" {
...
+resource "aws_iam_role_policy_attachment" "bfd_pipeline_rif" {
+  role            = module.iam_profile_bfd_pipeline.role
+  policy_arn      = aws_iam_policy.bfd_pipeline_rif.arn
+}
+
+# Give the BFD Pipeline app read access to the Ansible Vault PW.
+#
+data "aws_iam_policy" "ansible_vault_pw_ro_s3" {
+  arn           = "arn:aws:iam::577373831711:policy/bfd-ansible-vault-pw-ro-s3"
+}
```
RickHawesUSDS: Instead of hard coding the account-id in the `arn`, I think you should get it the passed in account_id variable.

karlmdavis: Done, though a `grep` indicates it's hardcoded in several other places, which we should address later.

PR #98, opened 2019-10-05
BLUEBUTTON-1359: Enable deploys to the CCS prod environment

karlmdavis:
Also cleaning up a couple of minor things in our `Jenkinsfile`, while I'm poking at it.

https://jira.cms.gov/browse/BLUEBUTTON-1359

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -236,8 +236,7 @@ stage('Deploy to prod-stg') {
 
 if (deployEnvironment == 'ccs') {
 	stage('Build App AMIs for PROD') {
-		if (willDeployToProdEnvs && deployEnvironment != 'ccs') {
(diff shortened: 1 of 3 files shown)


njdister:
Do you need to create the new lockable resource in Jenkins first or will it be created automatically?

karlmdavis:
> Do you need to create the new lockable resource in Jenkins first or will it be created automatically?

@njdister Automagically. I _will_, however, have to go remove the old lock manually. Not that it's hurting anything, but still.

(FWIW, interesting to note that this kind of change would be a lot more dangerous if we weren't always building branches after merging in `master`.)

karlmdavis:
Had to fix a syntax error, so pushed a commit for that, rebased, and sent out for review.

karlmdavis:
Merging ahead of re-review, as this is low-risk and I'm pressed for time.

🤞

PR #99, opened 2019-10-06
BLUEBUTTON-1288: Allow EC2 instances to push to CloudWatch logs

karlmdavis:
This is apparently working for the BFD Server? No idea why; I'm not seeing the permissions that enable that anywhere in our Terraform.

Nevertheless, it definitely wasn't working for the BFD Pipeline, so I'm adding this fix, which will cover both of them.

https://jira.cms.gov/browse/BLUEBUTTON-1288

PR #103, opened 2019-10-06
Use AWS managed cloudwatch agent policy instead of custom policy

njdister:
This reverts the custom policy from #99 and #101 and instead opts to use the AWS managed cloudwatch agent policy.


karlmdavis:
Merging this for @njdister.

PR #104, opened 2019-10-07
BLUEBUTTON-1359: Fix New Relic app name in CCS prod

karlmdavis:
Was causing New Relic data for the _CCS_ `prod` env to be stored under the "BFD Server (prod, _HealthAPT_)" app. Whoops.

I have no clue how this ended up here (other than -- obviously -- I goofed it at some point). The variable isn't redefined in the same file for the other CCS envs, nor should it be: the default in the playbooks is fine.

https://jira.cms.gov/browse/BLUEBUTTON-1359


karlmdavis:
Going to merge this ahead of review, as it's low-risk and is blocking the performance tests that I'd like to run.

PR #110, opened 2019-10-08
Extend gold image defined root partition with all available free space

njdister:
@karlmdavis discovered that despite us setting the EBS volume size on deploy, the instance was not properly utilizing the space.  This fixes that.  Only issue I see is if for some reason we had to revert to older instance classes that do not support [NVMe volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances), this would break deploys without more logic.


njdister:
This has passed deployment in test and I can confirm new ETL and ASG instances have all extra available space allocated to the root partition.

PR #124, opened 2019-10-18
BLUEBUTTON-1378: startIndex out of bounds

jcgeer:
Changing our code to return a 400 (InvalidRequestException) error when a
request URLs startIndex is greater than the number of resources to be
returned.


jcgeer:
I believe the code committed here is largely complete. However, with the switch to the monorepo I was having trouble running local builds and was unable to reproduce the error mentioned in the JIRA ticket effectively, so I'm not sure if the tests are properly written. Since the Jenkins build passed before adding the "expecting error" line to the IT, I'm guessing something is missing. I've just not had the time to get my local build working to figure out what that is.

PR #125, opened 2019-10-18
BLUEBUTTON-1500: Log plan on queries slower than 6s

njdister:
**DO NOT MERGE UNTIL 10/21**

This allows us to obtain better debugging info for slow queries when they happen.  The auto_explain module can also run an analyze, but at a performance cost:

Ref: https://www.postgresql.org/docs/9.6/auto-explain.html

> Note: When this parameter is on, per-plan-node timing occurs for all statements executed, whether or not they run long enough to actually get logged. This can have an extremely negative impact on performance. Turning off auto_explain.log_timing ameliorates the performance cost, at the price of obtaining less information.

### DIFF
--- a/ops/terraform/env/prod/stateful/main.tf
+++ b/ops/terraform/env/prod/stateful/main.tf
@@ -17,6 +17,7 @@ module "stateful" {
   }
 
   db_params = [
+    {name="auto_explain.log_min_duration", value="6000", apply_on_reboot=false},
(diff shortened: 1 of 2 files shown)


njdister:
> Do you have any idea on the performance degradation that average queries would get? Should we try it on one replica, first, and measure?

The performance degradation is only when running an analyze along with an explain, and this PR (so far) does not enable auto-analyze.

PR #131, opened 2019-10-25
BLUEBUTTON-1361: Temporary debug logging

njdister:
This sets debug logging for DataSetMonitorWorker and DataSetMoveTask to help diagnose why data sets are not getting moved on successful process.  This also assists with level-setting the implementation of BLUEBUTTON-1499.


njdister:
In retrospect it may be better to apply this manually to prod for the time being via `-Dlogback.configurationFile` so we don't need to go through a whole PR + deploy just to revert.

PR #132, opened 2019-10-26
BLUEBUTTON-1508: DB Migration 21 for Since Support

RickHawesUSDS:
**Why**
This is next in a series of PRs features needed to allow BFD partners to support since. The main branch of this PR is on the BLUEBUTTON-1506 branch. This PR just migrates the db to add new columns for lastUpdated and adds tables to track database updates. 

**What Changed**
- lastUpdated columns added to the Beneficiary table and the Claim tables. Did not change the ClaimLine and Bene History tables.

**Choices Made**
- The Postgres 9.6 docs has this comment about adding columns
  > When a column is added with ADD COLUMN, all existing rows in the table are initialized with the column's default value (NULL if no DEFAULT clause is specified). If there is no DEFAULT clause, this is merely a metadata change and does not require any immediate update of the table's data; the added NULL values are supplied on readout, instead.
Based on this, there are no default values in the alter statements
- Timestamps with time zones are used to allow the DB to be restored in different time-zone. 

**Testing Done**
This migration has been applied to the `test` db. The migration took 17 milliseconds. Smoke tests done where done on the FHIR server.


whytheplatypus:
Will some indexing on these columns be necessary to make the range queries cheaper?

RickHawesUSDS:
Updated to reflect version 2 of the RFC.

PR #137, opened 2019-11-04
Add DPC dev VPC peering

RickHawesUSDS:
**Why**
DPC has new environment dpc-dev. This environment is peering with prod-sbx. 

**What**
Added the dpc-dev peering to the list of peerings for the prod-sbx. This change only affects the prod-sbx.


njdister:
Looks good but just to confirm, we don't actually create the peering connections in terraform correct?  We only apply the peering CIDR blocks to our lb security group?

RickHawesUSDS:
> Looks good but just to confirm, we don't actually create the peering connections in terraform correct? We only apply the peering CIDR blocks to our lb security group?

Yes. The VPC peering go across AWS accounts and need to be setup by GDIT. This script change is just to add the new CIDR block our LB security group.

PR #138, opened 2019-11-04
BLUEBUTTON-1499: Add delayed manifest upload integration tests for DataSetMonitor

njdister:
As part of BLUEBUTTON-1499, we want to ensure that a vacuum and analyze are only run after all data sets are processed, even if those data sets are slow to come in.  I wanted to create an integration test that would more closely simulate how data sets might get uploaded into the ETL bucket.  The bulk of this is verbatim from the `multipleDataSetsTest` function, with the manifest uploads wrapped in a `ScheduledExecutorService`.  My Java-fu is pretty weak so I'd bet this can be made better!

After thinking about this for a bit, this might be better as an enhancement to `multipleDataSetsTest` rather than its own test, let me know what you think.

### DIFF
--- a/apps/bfd-pipeline/bfd-pipeline-rif-extract/src/test/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetMonitorIT.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rif-extract/src/test/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetMonitorIT.java
@@ -7,8 +7,12 @@ import gov.cms.bfd.model.rif.RifFileType;
 import gov.cms.bfd.model.rif.samples.StaticRifResource;
 import gov.cms.bfd.pipeline.rif.extract.ExtractionOptions;
 import gov.cms.bfd.pipeline.rif.extract.s3.DataSetManifest.DataSetManifestEntry;
+import java.net.URL;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import org.awaitility.Awaitility;
 import org.awaitility.Duration;
 import org.junit.Assert;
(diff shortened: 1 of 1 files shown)


apps/bfd-pipeline/bfd-pipeline-rif-extract/src/test/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetMonitorIT.java
```diff
@@ -178,4 +183,126 @@ public void multipleDataSetsTest() throws InterruptedException {
...
+      monitor.stop();
+
+      // Verify what was handed off to the DataSetMonitorListener.
+      Assert.assertEquals(3, listener.getDataEvents().size());
+      Assert.assertEquals(0, listener.getErrorEvents().size());
+      Assert.assertEquals(manifestA.getTimestamp(), listener.getDataEvents().get(0).getTimestamp());
+      Assert.assertEquals(manifestB.getTimestamp(), listener.getDataEvents().get(1).getTimestamp());
+      Assert.assertEquals(manifestC.getTimestamp(), listener.getDataEvents().get(2).getTimestamp());
+
+      // Verify that the data sets were both renamed.
```
karlmdavis:
```suggestion
      // Verify that the data sets were all renamed.
```

apps/bfd-pipeline/bfd-pipeline-rif-extract/src/test/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetMonitorIT.java
```diff
@@ -178,4 +183,126 @@ public void multipleDataSetsTest() throws InterruptedException {
...
+          TimeUnit.SECONDS);
+
+      DataSetManifest manifestC =
+          new DataSetManifest(
+              Instant.now(), 0, new DataSetManifestEntry("carrier.rif", RifFileType.CARRIER));
+      manifestUploaderService.schedule(
+          new ManifestUploader(
+              s3Client, bucket, manifestC, StaticRifResource.SAMPLE_A_CARRIER.getResourceUrl()),
+          new Random().nextInt(10),
+          TimeUnit.SECONDS);
```
karlmdavis: Can you check the JavaDoc for this, wouldn't it schedule this as a _recurring_ task? I think that rather than a scheduled executor, you want a _delayed_ executor, which you could accomplish by just using a normal `ExecutorService` and having your upload task sleep for a bit, first.

apps/bfd-pipeline/bfd-pipeline-rif-extract/src/test/java/gov/cms/bfd/pipeline/rif/extract/s3/DataSetMonitorIT.java
```diff
@@ -178,4 +183,126 @@ public void multipleDataSetsTest() throws InterruptedException {
...
+
+      DataSetManifest manifestB =
+          new DataSetManifest(
+              manifestA.getTimestampText(),
+              1,
+              new DataSetManifestEntry("pde.rif", RifFileType.PDE));
+      manifestUploaderService.schedule(
+          new ManifestUploader(
+              s3Client, bucket, manifestB, StaticRifResource.SAMPLE_A_PDE.getResourceUrl()),
+          new Random().nextInt(10),
```
karlmdavis: If I'm reading this right, this test case could take at least 20 seconds, depending on your randoms? That's **slow**, as we tend to consider these things.

Can you think through and document: what's the minimum amount of time this test needs to be wait in order to be valid? Then, maybe remove the randoms and just always wait exactly that long.

PR #141, opened 2019-11-08
BLUEBUTTON-1584: Allow JANUS to SSH into our instances

karlmdavis:
They only _need_ access to the prod-sbx, but I don't have any simple way of restricting that on the moment. Instead, we'll rely on the fact that they're restricted to the prod-sbx VPC at the network layer.

https://jira.cms.gov/browse/BLUEBUTTON-1584


karlmdavis:
This will need to be reverted later this month, per https://jira.cms.gov/browse/BLUEBUTTON-1585.

PR #155, opened 2019-11-21
BLUEBUTTON 1506:  Bulk Export Since support

RickHawesUSDS:
**Why**
This large PR contains the work to support the Bulk Export Since epic. 

**Description**
The work in this PR is described by the RFC [0004-since-paramater](https://github.com/CMSgov/beneficiary-fhir-data/blob/BLUEBUTTON-1506-Since-Support/rfcs/0004-since-parameter.md). Details can be found in the document and the comments in the code. 

**Tickets**
- BLUEBUTTON-1507 Create and publish an RFC for bulk export since support 
- BLUEBUTTON-1508 Update schema for since support 
- BLUEBUTTON-1509 Return meta.lastUpdated
- BLUEBUTTON-1510 Search on lastUpdated for _since support
- BLUEBUTTON-1598 Optimize zero-size result-sets in since support 

**Testing Done**
In addition to the unit and integration tests added to the PR, ad-hoc testing was done on the test environment. Also, the DPC side of the _since feature was implemented and the two systems integrated and tested together. The synthetic data set was loaded, but production RIF files haven't been tried for obvious reasons. 

**Deploying this PR**
This PR contains both a schema change and changes to the pipeline and data server. Developing the PR, it was convenient to refactor and change all parts simultaneously. When approved, I'll update the schema-only PR and that will be deployed first. The deployment will be staged to prod-sbx for a time as well.

**Security Implications**
New bloom filter software package and new lastUpdated information stored in the database. 

- [x]  new software dependencies
- [ ]  altered security controls
- [x]  new data stored or transmitted
- [ ]  security checklist is completed for this change
- [ ]  requires more information or team discussion to evaluate security implications


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/QueryUtils.java
```diff
@@ -0,0 +1,163 @@
...
+    final Path lastUpdatedPath = root.get("lastUpdated");
+
+    Predicate lowerBoundPredicate = null;
+    if (lowerBound != null) {
+      switch (range.getLowerBound().getPrefix()) {
+        case GREATERTHAN:
+          lowerBoundPredicate = criteriaBuilder.greaterThan(lastUpdatedPath, lowerBound);
+          break;
+        case EQUAL:
+          lowerBoundPredicate = criteriaBuilder.greaterThanOrEqualTo(lastUpdatedPath, lowerBound);
```
shea-parkes: I know I'm coming in from way, way out of left field, and I don't have access to your requirements for this one, but this line is surprising in isolation (being under `case EQUAL`).  There are plenty of reasons this could be perfectly fine, but I felt like it was worth at least hanging a comment on it to call attention in case it was an accident.

Thanks for working on this.  I promise I'll be using it downstream.

shea-parkes: Ah, nevermind, it's the lowerbound predicate.  And you have another one on upperbound, so you're squeezing them.  Definitely fine.  Sorry to bother you.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFileFilter.java
```diff
@@ -0,0 +1,134 @@
...
+import com.google.common.hash.BloomFilter;
+import com.google.common.hash.Funnels;
+import java.util.Date;
+
+/**
+ * LoadedFile filters are used to determine if a given beneficiary was updated in particular loaded
+ * file. Beneath the covers, they use BloomFilters (see <a
+ * href="https://en.wikipedia.org/wiki/Bloom_filter">Bloom Filters</a>) which are space efficient.
+ */
+public class LoadedFileFilter {
```
whytheplatypus: What are the trade offs between the bloomfilter implementation and a optimized date range query (as I understand it they can be very cheap if the correct indexes are in place because of ordering).

whytheplatypus: answering my own question: indexing becomes too large for some of the large tables.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V21__Add_lastUpdated.sql
```diff
@@ -0,0 +1,58 @@
...
+--
+-- One row for each RIF file loaded. 
+-- The timestamps represent start and end time of processing the RIF file. 
+-- Count is the number of records in the RIF file. 
+-- The filterBytes is an array of all beneficiariesIds in the RIF file.
+-- Dev Note: The byte array format was faster (2x) than an array type
+-- The filterType allows us to upgrade the filter type in the future.
+-- 
+create table "LoadedFiles" (
+  "loadedFileId" bigint primary key,
```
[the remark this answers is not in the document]
whytheplatypus: ignore this, it's a convention within this project at the least.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java
```diff
@@ -79,6 +85,8 @@
    */
   private static final int RECORD_BATCH_SIZE = 100;
 
+  private static final int MAX_FILE_AGE = 30; // Days
```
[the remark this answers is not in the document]
RickHawesUSDS: In some past projects I would put these in configs. This project doesn't have many configs which can be a virtue because it is simpler to operate. This is especially true since we are practicing "infrastructure as code".

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/MappingSpec.java
```diff
@@ -158,6 +159,20 @@ public MappingSpec setHasLines(boolean hasLines) {
...
+   *     if not
+   */
+  public boolean getHasLastUpdated() {
+    return hasLastUpdated;
+  }
+
+  /** @param hasLastUpdated the new value for {@link #getHasLastUpdated()} */
+  public MappingSpec setHasLastUpdated(boolean hasLastUpdated) {
+    this.hasLastUpdated = hasLastUpdated;
+    return this;
```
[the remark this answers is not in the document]
whytheplatypus: hapi tends toward [fluent interfaces](https://hapifhir.io/doc_rest_client.html#_toc_the_fluentgeneric_client)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -251,17 +297,26 @@ private static int compareByClaimIdThenClaimType(IBaseResource res1, IBaseResour
   /**
    * @param claimType the {@link ClaimType} to find
    * @param patientId the {@link Beneficiary#getBeneficiaryId()} to filter by
+   * @param
```
nickrobison-usds: Missing javadoc

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java
```diff
@@ -79,6 +88,8 @@
    */
   private static final int RECORD_BATCH_SIZE = 100;
 
+  private static final int MAX_FILE_AGE = 40; // Days
```
whytheplatypus: Add `DAYS` to the name of the variable so it's clear during usage.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java
```diff
@@ -512,8 +552,7 @@ private void processAsync(
        * ensure that the failing data is captured.
        */
       try {
-        if (entityManager != null && entityManager.getTransaction().isActive())
-          entityManager.getTransaction().rollback();
+        if (txn != null && txn.isActive()) txn.rollback();
```
whytheplatypus: it looks like this `txn` isn't in the scope of this `finally`, since it's set during the `try`. I could just be counting the brackets wrong.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java
RickHawesUSDS: Good point. Will fix.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java
RickHawesUSDS: The idea behind this change is not to rely on the fact that getTransaction() will return the same transaction as the one at the start of the transaction. Instead of doing...

```
try {
  em.getTransaction().begin();
  ...
  em.getTransaction().commit();
} finally {
  em.getTransaction().rollback();
}
```

We write...
```
EntityTransaction txn = null;
try {
  txn = em.getTransaction();
  txn.begin();
  ...
  txn.commit();
} finally {
  if (txn != null) txn.rollback();
}
```
Not as elegant, but I think it is safer. What do you think? Granted, the function is a mess, as the TODO states. What do you think?

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/LoadedBatchBuilder.java
```diff
@@ -0,0 +1,48 @@
+package gov.cms.bfd.model.rif;
+
+import java.util.Date;
+import java.util.Vector;
+
+/** Class to build a LoadedBatch. Thread safe. */
+public class LoadedBatchBuilder {
+  private final Vector<String> beneficiaries;
```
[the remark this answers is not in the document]
RickHawesUSDS: Good catch. I wasn't aware it was semi deprecated, so I learned something today.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/LoadedBatch.java
```diff
@@ -0,0 +1,119 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.SequenceGenerator;
+import javax.persistence.Table;
+
+/** JPA class for the LoadedBatches table */
+@Entity
+@Table(name = "`LoadedBatches`")
```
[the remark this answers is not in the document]
RickHawesUSDS: It has to do with the convention in BFD to use mixed case table names and SQLs tendency to lower case stuff automatically if the quotes are not there.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/LoadedBatchBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+    beneficiaries.add(beneficiaryId);
+  }
+
+  /**
+   * Create a LoadedBatch from the data in the builder
+   *
+   * @return a new LoadedBatch
+   */
+  public LoadedBatch build() {
+    synchronized (this) {
```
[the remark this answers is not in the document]
RickHawesUSDS: Good point.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java
whytheplatypus: Definitely like the change to clearly use the same transaction all the way through. My question is really more about java scope in general: since `txn` is set to `em.getTransaction()` inside the `try` block does it retain that value so that `txn.rollback()` can be called in the `finally`?

whytheplatypus: answered my own question looks like it [does retain the value](https://repl.it/repls/EnchantedOpulentIrcbot) out of the `try` block.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/LoadedBatchBuilder.java
```diff
@@ -0,0 +1,48 @@
+package gov.cms.bfd.model.rif;
+
+import java.util.Date;
+import java.util.Vector;
+
+/** Class to build a LoadedBatch. Thread safe. */
+public class LoadedBatchBuilder {
```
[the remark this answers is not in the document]
RickHawesUSDS: I didn't find this annotation in the current set of packages used in the project. What package do you recommend?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitResourceProvider.java
RickHawesUSDS: Thanks

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/LoadedBatchBuilder.java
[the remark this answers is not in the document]
nickrobison-usds: I believe it's part of the javax.annotation.concurrent package, see [here](https://aalmiray.github.io/jsr-305/apidocs/javax/annotation/concurrent/NotThreadSafe.html). If not, it's probably not worth pulling in a new dependency just for an annotation.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V21__Add_lastUpdated.sql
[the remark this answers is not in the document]
karlmdavis: General reason is that it makes queries a bit easier to read in the presence of joins, without requiring aliasing. Not a cure-all (still sometimes end up needing aliases), but I've found it helps.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java
karlmdavis: Declarations are block-scoped, not assignments/values.

rfcs/0004-since-parameter.md
```diff
@@ -0,0 +1,193 @@
...
+
+Partners can use the `_lastUpdated` parameter to get a single beneficiary's recently added FHIR resources. 
+To do this without missing a resource, the partner remembers (or has it's client remember) the returned Bundle's `lastUpdated` field. 
+This date represents the last update timestamp of the service database. All future updates to the database happen after this timestamp.
+
+![Last Updated](https://www.websequencediagrams.com/files/render?link=JJsgr8NxcInDVKyKqwtBiC5bTjdEafpXz4G1vbW2QbwtiveJ8IdHKl3r35lCdKdn)
+
+### Bulk Export Implementors Details
+
+BFD partner's which are implementing bulk export can add `_since` support. Just as it was in the single beneficiary case, keeping track of timestamps is important to avoid missing data. The following sequence diagram shows how this interaction should work.
```
dkrylovsb: Let's say a new EOB is added to the BFD DB. Will that update the `_lastUpdated` timestamp on the **Patient** resource to which the EOB belongs?

dkrylovsb: The reason I am asking is to see if we could determine whether we can _completely_ avoid making a call to `/fhir/ExplanationOfBenefit?patient=...` by looking at `Patient.lastUpdated`, which would potentially make a significant positive impact on both ends.

RickHawesUSDS: The literal answer to your question, the answer is no. A Patient's `lastUpdated` timestamp reflects the update time of the underlying beneficiary record from which the resource is derived. Adding new claim records doesn't change the beneficiary record or the resultant Patient resource. 

However, there is an optimization that you have in mind that I don't understand. Can you explain it more?

dkrylovsb: Thanks for the answer, @RickHawesUSDS . Our interaction with BFD begins with obtaining a list of Patients by a given Part D contract number via an endpoint that @whytheplatypus has implemented. Then, for each Patient we search for their EOBs: a separate call per Patient. 

If we begin storing (read: caching) received EOBs on our end, then on subsequent runs _ideally_ we would like to be able to tell from a given Patient record whether or not their EOBs changed from the last time we fetched them, because that could save us from even making a network call altogether.

Does this make sense?

RickHawesUSDS: Yes. This makes sense. So my understanding is that the AB2D to service keeps a cache of the EOBs received for a particular beneficiary. Periodically, the AB2D service would like to poll the BFD to receive any new EOBs for that particular beneficiary. Do I have this right? 

If so, this is an outline on how it can be accomplished: 

**Pre-step**: In addition to caching the EOB's for a particular beneficiary, you need to store one timestamp, the `transactionTime`, with each beneficiary.

**First-Time**: Fetch all EOBs for a particular beneficiary use the current EOB search end-point. The returned bundle of EOB has a `lastUpdated` timestamp. This is the `transactionTime` to store along-side the beneficiary's EOBs.  

**Periodically**: Fetch the EOB's that have arrived after the stored `transactionTime`. The new bundles will have an updated `transactionTime` to store as well. Repeat. 

```
v1/fhir/ExplanationOfBenefit?patient=<bene_id>&_lastUpdated=gt<transactionTime>
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -218,15 +218,11 @@ public Bundle searchByLogicalId(
       }
     }
 
+    PagingLinkBuilder paging =
+        new PagingLinkBuilder(
+            requestDetails, "/Patient?", Patient.SP_RES_ID, logicalId.getValue(), lastUpdated);
```
whytheplatypus: Bringing this forward into the Providers is fantastic.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/QueryUtils.java
```diff
@@ -0,0 +1,121 @@
...
+   * Create a predicate for the lastUpdate field based on the passed _lastUpdated parameter range.
+   *
+   * @param cb to use
+   * @param root to use
+   * @param range to base the predicate on
+   * @return a predicate on the lastUpdated field
+   */
+  static Predicate createLastUpdatedPredicate(
+      CriteriaBuilder cb, Root<?> root, DateRangeParam range) {
+    final Path<Date> lastUpdatedPath = root.get("lastUpdated");
```
whytheplatypus: Could replace `root` in params and bubble up `lastUpdatedPath`.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/QueryUtils.java
```diff
@@ -0,0 +1,121 @@
...
+        return cb.or(cb.isNull(lastUpdatedPath), upperBoundPredicate);
+      } else {
+        return cb.and(lowerBoundPredicate, upperBoundPredicate);
+      }
+    } else {
+      if (lowerBoundPredicate == null) {
+        throw new IllegalArgumentException(
+            ("_lastUpdate upper and lower bound cannot both be null"));
+      } else {
+        return lowerBoundPredicate;
```
whytheplatypus: This can be outside of the if else, reducing complexity a bit.

PR #158, opened 2019-11-21
Allow local server port to be set via env var

whytheplatypus:
# Why
To allow developers to run the BFD web server on a specific port. This makes possible things like consistent port forwarding in virtualized environments, and a lower barrier to entry for local development.

# Changes
- propagate the `BFD_PORT` environment variable through the `server-start.sh` script.


whytheplatypus:
Sounds good! The `BFD_PORT` is an [existing configuration variable](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-launcher/src/main/java/gov/cms/bfd/server/launcher/AppConfiguration.java#L17), where are these typically documented?

PR #170, opened 2019-12-17
BLUEBUTTON-1679 Remove hicn-hash identifier from Patient Resource

dtisza1:
### Change Details

Summary:

This is to remove the hashed HICN identifier from the patient resource response.

- Removed the TransformerConstants.CODING_BBAPI_BENE_HICN_HASH identifier from the BeneficiaryTransformer .  The BeneficiaryTransformer is used by the PatientResourceProvider for the FHIR response. 
- Testing of the Patient resource using CURL is working as expected in the local docker setup. 
- Reviewed the code base for other areas that this change might affect. Not seeing BeneficiaryTransformer utilized anywhere besides PatientResourceProvider and in related tests.
- Updated the unit tests in the BeneficiaryTransformerTest class. This included transformSampleARecord() and transformSampleARecordWithIdentifiers() methods. In addition to validating the count/size for the number of identifiers, the values are also validated.
- Updated the IT tests. These are performed by the EndpointJsonResponseComparatorIT class. This utilizes JSON files under the apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses path. These files are used to validate related end-point responses.  The hicn-hash identifier was removed from 6 of these files.
- ALL TESTS ARE PASSING IN LOCAL DEV!
- Added a change entry to the apps/bfd-server/dev/api-changelog.md

### Acceptance Validation

- That the Patient resource responses DO NOT contain the hicn-hash identifier:  "https://bluebutton.cms.gov/resources/identifier/hicn-hash"

### External References

- [BLUEBUTTON-1679](https://jira.cms.gov/browse/BLUEBUTTON-1679)

### Security Implications

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/BeneficiaryTransformerTest.java
+++ b/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/BeneficiaryTransformerTest.java
@@ -34,8 +34,17 @@ public final class BeneficiaryTransformerTest {
     Patient patient =
         BeneficiaryTransformer.transform(
             new MetricRegistry(), beneficiary, IncludeIdentifiersMode.OMIT_HICNS_AND_MBIS);
+
     assertMatches(beneficiary, patient);
-    Assert.assertEquals(2, patient.getIdentifier().size());
+
+    // Verify patient has only one identifier.
+    Assert.assertEquals(1, patient.getIdentifier().size());
+
+    // Verify the only identifier is BENE_ID and value.
+    Assert.assertEquals(
+        patient.getIdentifier().get(0).getSystem(),
+        TransformerUtils.calculateVariableReferenceUrl(CcwCodebookVariable.BENE_ID));
+    Assert.assertEquals(patient.getIdentifier().get(0).getValue(), "567834");
   }
 
   /**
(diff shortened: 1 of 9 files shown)


apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/BeneficiaryTransformerTest.java
```diff
@@ -34,8 +35,17 @@ public void transformSampleARecord() {
...
         BeneficiaryTransformer.transform(
             new MetricRegistry(), beneficiary, IncludeIdentifiersMode.OMIT_HICNS_AND_MBIS);
+
     assertMatches(beneficiary, patient);
-    Assert.assertEquals(2, patient.getIdentifier().size());
+
+    // Verify patient has only one identifier.
+    Assert.assertEquals(1, patient.getIdentifier().size());
+
+    // Verify the only identifier is BENE_ID and value.
```
ronaldheft-usds: This is a nice assertion addition to the test.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/BeneficiaryTransformerTest.java
```diff
@@ -46,13 +56,58 @@ public void transformSampleARecord() {
...
             new MetricRegistry(), beneficiary, IncludeIdentifiersMode.INCLUDE_HICNS_AND_MBIS);
+
     assertMatches(beneficiary, patient);
-    Assert.assertEquals(7, patient.getIdentifier().size());
+
+    // Verify patient has 6 identifiers.
+    Assert.assertEquals(6, patient.getIdentifier().size());
+
+    // Verify patient identifiers and values match.
+    assertValuesInPatientIdentifiers(
```
ronaldheft-usds: If I understand this test correctly, we're asserting that when we have the INCLUDE_IDENTIFIERS flag, we're still returning the unhashed hicn, correct?

dtisza1:
> I was expecting that we would extend the IncludeIdentifiersMode enumeration to have a new INCLUDE_MBIS mode in addition to the INCLUDE_HICNS_AND_MBIS. We would still keep the current INCLUDE_HICNS_AND_MBIS mode so that we wouldn't break existing partners. It looks like this change just eliminates HICNs from the Patient resource.

@RickHawesUSDS We have a split up of related tickets. This one only includes removing the hicn-hash. Was just talking over the IncludeIdentifiersMode with @ronaldheft-usds and we are wanting to include that in the next BLUEBUTTON-1536 ticket (since it goes with that task).

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/BeneficiaryTransformerTest.java
dtisza1: That's correct.

PR #172, opened 2019-12-19
BLUEBUTTON-1516: Search patients by MBI hash

ronaldheft-usds:
### Change Details

Summary:

This PR prepares BFD to handle MBIs over HICNs for searching and the identifiers returned.

* Adds an `mbiHash` column and index to Beneficiaries and BeneficiariesHistory (props @RickHawesUSDS)
* Adds the hashing algorithm for MBIs
  * This is based on the exact same algorithm as the HICN hash. The same pepper and iterations are used, so the details will remain the same among BFD partners.
  * The exact same configuration value for the pepper and iterations was used so we did not have to update server configurations. This can be a little confusing, as the MBI hashing algorithm references the HICN pepper. If we're okay with updating server configurations, I would recommend we rename the property to something that applies to both HICN and MBI.
* Updates the RifLoader to calculate the `mbiHash` column for Beneficiaries and BeneficiariesHistory
  * This occurs by running the MBI through the hashing algorithm on load.
  * The MBI property remains optional, so if we do not have an MBI for a beneficiary, all data loading logic will continue to function.
  * The `medicareBeneficiaryId` was added to the mapping spreadsheet for BeneficiariesHistory to support testing. It appears the BeneficiariesHistory RIF loading capability is only used for testing purposes, but if this is not the case, there will be a potential data load problem with this change.
* Updates all RifLoader tests to verify the `mbiHash` is calculated correctly
  * Adds MBIs to the BeneficiariesHistory RIF files to support testing and verifying the same scenarios we were verifying for HICNs.
* Adds the `mbi-hash` identifier to Patient resources via the `mbi-hash` identifier
* Refactors the `queryDatabaseByHicnHash` method to be a generic `queryDatabaseByHash` to support reusing the same search logic for MBIs as we used for HICNs
* Adds the `mbi-hash` as a valid identifier to use for searching Patients.
  * Includes tests for all the same searching scenarios as we were testing for searching by `hicn-hash`.

Miscellaneous tasks performed:

* Adds *`.iml` files to the `.gitignore` (Using IntelliJ and wanted to ensure no IntelliJ files were checked in)

### Acceptance Validation

* Verification there are no issues adding the MBI to the BeneficiariesHistory RIF loader.
* Patients can be found via searching with the `mbi-hash` identifier.
* The `mbi-hash` identifier is returned for Patient resources.

### Feedback Requested

* General approach and validating I am conforming to BFD developer standards.

### Remaining Work

* This PR *does not* address the following related tasks. All of these tasks will be required to launch MBI search into production:
  * Calculating the MBI hash for existing beneficiaries (@RickHawesUSDS is looking into this)
  * Updating existing synthetic data to support searching by MBI ([BLUEBUTTON-1524](https://jira.cms.gov/browse/BLUEBUTTON-1524))
  * Updating the developer documentation to communicate `mbi-hash` as a validate identifier to search by

### External References

* [BLUEBUTTON-1516](https://jira.cms.gov/browse/BLUEBUTTON-1516)

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/BeneficiaryTransformer.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/BeneficiaryTransformer.java
@@ -60,6 +60,13 @@ final class BeneficiaryTransformer {
         .setSystem(TransformerConstants.CODING_BBAPI_BENE_HICN_HASH)
         .setValue(beneficiary.getHicn());
 
+    if (beneficiary.getMbiHash().isPresent()) {
+      patient
+          .addIdentifier()
+          .setSystem(TransformerConstants.CODING_BBAPI_BENE_MBI_HASH)
+          .setValue(beneficiary.getMbiHash().get());
+    }
+
     if (includeIdentifiersMode == IncludeIdentifiersMode.INCLUDE_HICNS_AND_MBIS) {
       Extension currentIdentifier =
           TransformerUtils.createIdentifierCurrencyExtension(CurrencyIdentifier.CURRENT);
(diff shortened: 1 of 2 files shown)


msnook:
@ronaldheft-usds @whytheplatypus 
Have you received confirmation on whether or not the `pepper` and/or `iteration` values will change for MBI hash?

ronaldheft-usds:
@msnook This PR keeps them the exact same:

> This is based on the exact same algorithm as the HICN hash. The same pepper and iterations are used, so the details will remain the same among BFD partners.

ronaldheft-usds:
This PR was accidentally merged. Updated PR is now here: https://github.com/CMSgov/beneficiary-fhir-data/pull/180

PR #175, opened 2019-12-22
Return synthetic resources with negative IDs

whytheplatypus:
A regex for parsing the format of EoB and Coverage resource read requests did not successfully parse synthetic Resource read requests due to their negative IDs. The result was that no synthetic EoB or Coverage resources were found on Read requests.

This updates those regex to allow for the IDs of these Resources to be negative.

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
@@ -40,7 +40,8 @@ import org.springframework.stereotype.Component;
 @Component
 public final class CoverageResourceProvider implements IResourceProvider {
   /** A {@link Pattern} that will match the {@link Coverage#getId()}s used in this application. */
-  private static final Pattern COVERAGE_ID_PATTERN = Pattern.compile("(.*)-(\\p{Alnum}+)");
+  private static final Pattern COVERAGE_ID_PATTERN =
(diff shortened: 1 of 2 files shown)


whytheplatypus:
good catch, the feature branch didn't change but the base branch did requiring this to be rebased.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
```diff
@@ -40,7 +40,8 @@
 @Component
 public final class CoverageResourceProvider implements IResourceProvider {
   /** A {@link Pattern} that will match the {@link Coverage#getId()}s used in this application. */
```
karlmdavis:
```suggestion
   /** A {@link Pattern} that will match the {@link Coverage#getId()}s used in this application, e.g. <code>part-a-1234</code> or <code>part-a--1234</code> (for negative IDs). */
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -50,7 +50,7 @@
    * A {@link Pattern} that will match the {@link ExplanationOfBenefit#getId()}s used in this
    * application.
```
karlmdavis:
```suggestion
    * application, e.g. <code>carrier-1234</code> or <code>carrier--1234</code> (for negative IDs).
```

njdister:
Closing for implementation in #258

PR #179, opened 2019-12-30
BLUEBUTTON-1516: Fixup mbiHash columns

RickHawesUSDS:
**Why**
As part of the MBI support work, fill in the mbiHash columns of the database after the mbiHash fields have been added to the database. The mbiHash columns are needed for efficient hashed MBI lookups.  

**What Changed**

A new `RifLoaderIdleTasks` class contains all the logic of this change. The new class is a helper class for the `RifLoader`. The class contains a `doIdleTask` method that is called every time a `dataNotAvailable` event occurs. Rather than running on a background thread, the idle task only runs when there are no RIF files to be processed. The intent is not to interfere with the processing of RIF files. 

This work builds on @ronaldheft-usds PR. 

**Alternatives Considered**
Discussed two alternatives:

1. A special tool that would generate RIF files that would update the beneficiary records
2. A special tool that would update columns directly

The chosen had the advantage of being fast, simple and not interfering with RIF. 

The number of threads to use for the fixups are variable. On my local laptop with 30000 fixups, I find that the fixup process is basically DB limited past 2 threads. I eventually settled on 5 threads for the fixup process. 

- 1 Thread: 33 seconds
- 2 Threads 19 seconds
- 5 Threads: 19 seconds
- 10 Threads: 19 seconds
- 20 Threads: 19 seconds

**Testing Done**
Local laptop only. Avoided deploying to test because I didn't want to waste that fixup opportunity until the code review was done.


apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,289 @@
...
+        throw new RuntimeException("Unexcpected idle task");
+    }
+  }
+
+  /**
+   * Run this task as the first idle task.
+   *
+   * @return true if done with this task.
+   */
+  public boolean doInitialTask() {
```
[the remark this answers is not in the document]
RickHawesUSDS: Good points. The code makes these assumptions. Let me think about alternatives that a little more robust.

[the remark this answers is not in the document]
RickHawesUSDS: Look at it now. Removed the counter.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,327 @@
...
+  /**
+   * Run this task after the initial task. Respect the TASK_TIME_LIMIT.
+   *
+   * @return true if done with the current task.
+   */
+  public boolean doPostStartupTask() {
+    final Instant startTime = Instant.now();
+
+    // Execute batches in parallel
+    for (int i = 0; i < THREAD_COUNT; i++) {
```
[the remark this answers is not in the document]
RickHawesUSDS: That was the intent, but it didn't work like that. I tried a few alternatives. The code that ran the fastest on my laptop, didn't have threading. 

Look at the new code again, please.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+  public Task getCurrentTask() {
+    return currentTask;
+  }
+
+  /**
+   * Run the current idle task. This method is expected to be called whenever no RIF files are
+   * present for process. It will respect the TASK_TIME_LIMIT to allow checking of RIF files and to
+   * not interfer with RIF file processing.
+   */
+  public void doIdleTask() {
```
whytheplatypus: Suggest renaming to `doCurrentTask()` or something similar that better describes the action being taken.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+  /** JPQL query for a count */
+  private static final String COUNT_UNHASHED_HISTORIES =
+      "select count(b) from BeneficiaryHistory b "
+          + "where b.mbiHash is null and b.medicareBeneficiaryId is not null and "
+          + "b.medicareBeneficiaryId is not empty";
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(RifLoaderIdleTasks.class);
+
+  /** Enum to tell what the current task is being executed. */
+  public enum Task {
```
whytheplatypus: As an alternative have `Task` represent an interface that has a function `do` or something similar. The function `getCurrentTask` would then be able to return a class the implements the `Task` interface and `doCurrentTask` can simply call the `do` function of whatever `getCurrentTask` returns.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+   *
+   * @return true if done with the current task.
+   */
+  public boolean doPostStartupTask() {
+    final Instant startTime = Instant.now();
+    LOGGER.debug("Started a PostStartup time slice");
+
+    // Execute batches in parallel
+    List<Future<Boolean>> tasks = new ArrayList<Future<Boolean>>();
+    tasks.add(executorService.submit(() -> doTransaction(startTime, this::fixupBeneficiaryBatch)));
```
whytheplatypus: Possibly change this `tasks` variable to something like `jobs`, I found it a bit confusing differententiating between these `tasks` and tasks like `doPostStartupTask`. Particularly in the `waitUntilDone` function.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+            .setFetchSize(BATCH_COUNT)
+            .scroll(ScrollMode.SCROLL_INSENSITIVE)) {
+      int count = 0;
+      while (inPeriod(startTime, TASK_TIME_LIMIT) && itemCursor.next()) {
+        final Beneficiary beneficiary = (Beneficiary) itemCursor.get(0);
+        beneficiary
+            .getMedicareBeneficiaryId()
+            .ifPresent(
+                mbi -> {
+                  if (mbi.isEmpty()) return;
```
whytheplatypus: What is the case where the MBI would be empty? Is there anything that should or can be done in that case?

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+                  final String mbiHash =
+                      "xxx"; // RifLoader.computeMbiHash(options, secretKeyFactory, mbi);
+                  beneficiary.setMbiHash(Optional.of(mbiHash));
+                });
+        isDone = itemCursor.isLast();
```
whytheplatypus: it looks like this could be pulled out of the while loop

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+      while (inPeriod(startTime, TASK_TIME_LIMIT) && itemCursor.next()) {
+        final BeneficiaryHistory beneficiary = (BeneficiaryHistory) itemCursor.get(0);
+        beneficiary
+            .getMedicareBeneficiaryId()
+            .ifPresent(
+                mbi -> {
+                  if (mbi.isEmpty()) return;
+                  final String mbiHash =
+                      "xxx"; // RifLoader.computeMbiHash(options, secretKeyFactory, mbi);
+                  beneficiary.setMbiHash(Optional.of(mbiHash));
```
whytheplatypus: There may be some amount of this logic that can be shared between the methods updating the beneficiary and beneficiaryHistory records.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+
+  /**
+   * Any time left in this time slice?
+   *
+   * @param start of the period
+   * @param period duration
+   * @return true iff current period is less than the passed in period duration;
+   */
+  public static boolean inPeriod(final Instant start, final Duration period) {
+    return Duration.between(start, Instant.now()).compareTo(period) <= 0;
```
whytheplatypus: I think this function would return `true` if `start` was after `now()`

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+      int count = 0;
+      while (inPeriod(startTime, TASK_TIME_LIMIT) && itemCursor.next()) {
+        final Beneficiary beneficiary = (Beneficiary) itemCursor.get(0);
+        beneficiary
+            .getMedicareBeneficiaryId()
+            .ifPresent(
+                mbi -> {
+                  if (mbi.isEmpty()) return;
+                  final String mbiHash =
+                      "xxx"; // RifLoader.computeMbiHash(options, secretKeyFactory, mbi);
```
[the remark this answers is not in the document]
RickHawesUSDS: Yikes. I checked in an experiment.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+   * @return true if done with all fixups
+   */
+  public Boolean fixupHistoryBatch(final EntityManager em, final Instant startTime) {
+    LOGGER.debug("Start fixing up a History batch");
+    boolean isDone = true;
+    try (ScrollableResults itemCursor =
+        em.unwrap(Session.class)
+            .createQuery(SELECT_UNHASHED_HISTORIES)
+            .setFetchSize(BATCH_COUNT)
+            .scroll(ScrollMode.FORWARD_ONLY)) {
```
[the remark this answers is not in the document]
RickHawesUSDS: Agreed.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
RickHawesUSDS: I don't think so.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
```diff
@@ -0,0 +1,355 @@
...
+
+  /**
+   * Fixup a batch of Beneficiaries. Update the beneficary metrics.
+   *
+   * @param em a {@link EntityManager} setup for a transaction
+   * @return true if done with all fixups
+   */
+  public Boolean fixupBeneficiaryBatch(final EntityManager em, final Instant startTime) {
+    LOGGER.debug("Start fixing up a Beneficiary batch");
+    boolean isDone = true;
```
whytheplatypus: should this default to false so it doesn't report done incorrectly?

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
RickHawesUSDS: I understand the thinking behind this suggestion. Here's my thinking... ENUMs are good for representing and testing state. Of course, a class type can represent a state, but doing so would defeat some of the advantages of the interface abstraction. In addition, interfaces require new files in Java. Adding an interface file and the associated implementation files will take something that is fairly simple and splits it across many files. I don't think that is good for readability. If the code became more complex, the extra abstraction would pay off, of course.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
RickHawesUSDS: I'm not sure I follow. Let's talk.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
RickHawesUSDS: Good questions. The test isn't needed.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
RickHawesUSDS: Agree. Will fix the function for this case.

apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoaderIdleTasks.java
RickHawesUSDS: I don't think so.. If the query returns an empty list, then the loop won't run. In this case, I think of the task as "done".

whytheplatypus: ok that makes sense.

Could it also return as "done" if the try block fails and it doesn't get set to false?

PR #181, opened 2020-01-02
BLUEBUTTON-1573: Configure tracing and transaction names in New Relic

karlmdavis:
See the commit messages for details, but this basically makes New Relic's APM Transactions data useful. Previously, completely different transactions were all grouped together and traces did not have enough data to be very useful.

See our test environment in New Relic for an example of the new config.

https://jira.cms.gov/browse/BLUEBUTTON-1573


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/Operation.java
```diff
@@ -0,0 +1,137 @@
...
+import org.slf4j.MDC;
+
+/**
+ * Models the canonical "operations" supported by this application, such that each meaningfully
+ * distinct operation can be assigned a unique canonical name, for use in monitoring systems.
+ *
+ * <p>Also handles the publishing of those canonical names to the places they need to be published,
+ * via the {@link #publishOperationName()} method.
+ */
+public final class Operation {
```
RickHawesUSDS: I like this class. I suggest that it may be improved by defining common options like "id". To avoid inconsistency for in calls to setOption("id", ...) or setOption("Id", ...) etc. This thinking is similar to the reason for an Endpoint enumeration.

karlmdavis: Thought about doing that but decided that, if I wanted to go there, it'd be better to create types or subtypes for every operation. Decided that wasn't worth the effort right now.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponseLoggingFilter.java
```diff
@@ -60,29 +60,37 @@ private static void handleRequest(ServletRequest request) {
...
           getClientSslPrincipalDistinguishedName(servletRequest));
 
       // Record the request headers.
       Enumeration<String> headerNames = servletRequest.getHeaderNames();
       while (headerNames.hasMoreElements()) {
         String headerName = headerNames.nextElement();
         List<String> headerValues = Collections.list(servletRequest.getHeaders(headerName));
-        if (headerValues.isEmpty()) MDC.put(computeMdcKey("request.header." + headerName), "");
+        if (headerValues.isEmpty())
+          MDC.put(computeMdcRequestKey("request.header." + headerName), "");
```
njdister: Can't `request` in `request.header.` be omitted from the computeMdcRequestKey function call?

karlmdavis:
```suggestion
          MDC.put(computeMdcRequestKey("header." + headerName), "");
```

karlmdavis: Yup -- thanks! Fixed.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
```diff
@@ -83,6 +86,10 @@ public Coverage read(@IdParam IdType coverageId) {
     if (coverageIdText == null || coverageIdText.trim().isEmpty())
       throw new IllegalArgumentException();
 
+    Operation operation = new Operation(Operation.Endpoint.V1_COVERAGE);
```
whytheplatypus: To check my understanding this will log an `operation` of `V1_COVERAGE` twice. Once here and once in `RequestResponseLoggingFilter`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -105,6 +108,11 @@ public Patient read(@IdParam IdType patientId, RequestDetails requestDetails) {
 
     List<String> includeIdentifiersValues = returnIncludeIdentifiersValues(requestDetails);
 
+    Operation operation = new Operation(Operation.Endpoint.V1_PATIENT);
+    operation.setOption("by", "id");
+    operation.setOption("IncludeIdentifiers", includeIdentifiersValues.toString());
```
whytheplatypus: These look like they're custom versions of the query params and headers sent to the system. It looks like these are also tracked in `RequestResponseLoggingFilter`.

karlmdavis: Correct, but separated out to make NDJSON searches/filtering easier.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
karlmdavis: Sort of. The `RequestResponseLoggingFilter` sets the default operation name at the start of each request to `new Operation(Operation.Endpoint.OTHER)`. That gets set in the MDC, which is a single-valued `Map`.

Any handler/endpoint that wants to customize that can override it by creating a more specific `Operation` and calling `publishOperationName()` on it. That publish call overrides the default value in the MDC.

That make sense?

karlmdavis:
> I suggest that you could attach the output from NewRelic to document how this new feature is working.

Screen Shot 2020-02-13 at 10 37 38

Doesn't look like any transaction traces form back then have survived culling, unfortunately. But they're also super nice!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
whytheplatypus: it's using `matchByHttpUri` in the default, so I'd expect that to return `V1_COVERAGE` since it's a listed endpoint?

The default logic goes from `requestHttpUri -> Endpoint enum -> requestHttpUri`

karlmdavis: Ahh, yes -- had forgotten about the matching. That's correct, though it won't have the params -- just the operation name.

Regardless, it won't get logged twice -- the access log is only written to once per response, in `gov.cms.bfd.server.war.RequestResponseLoggingFilter.addToHttpAccessLog()`.

(Being pedantic: note that I said it only gets logged to the *access log* once. The other message/event log will also include whichever values the MDC happens to have at the time of any logging event. It doesn't particularly matter and there's not a lot we can do about it, given SLF4J's API -- short of replacing the MDC entirely, ourselves, which feels like a bad idea.)

PR #183, opened 2020-01-02
BLUEBUTTON-1719: Fix build

RickHawesUSDS:
**Why**
This PR fixes the master build to allow development to proceed, without researching the root cause of the problem. 

The EOB resource includes the names of drugs. The names of drugs are determined by cross-referencing a downloading a list from the FDA. With the new year, the list of FDA drugs has apparently changed. The drug in the project's RIF SAMPLE_A is no longer in the FDA list. This caused our builds to fail. 

**What Changed**
The fix was to change RIF SAMPLE_A in both the PDE and DME records to use a similar but different drug and to update the tests to expect this different drug. 

No change was made to the transform logic. 

**Testing Done**
Full tests.

**Security Checklist**
NA


dtisza1:
> Thanks for jumping on this Rick. Quick question, what's the purpose behind removing the 'Display' field in some instances, and updating it in others?

Was wondering about this also. In some of the files the 'Display' was removed. But noticed those were also ones where you left the original  '"code" : "46122005000"' the same.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerUtils.java
```diff
@@ -2882,7 +2882,6 @@ public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
             String.format("%s-%s", nationalDrugCodeManufacturer, nationalDrugCodeIngredient),
             ndcProductColumns[3] + " - " + ndcProductColumns[13]);
       }
-      ndcProductsIn.close();
```
dtisza1: Was this change to `TransformerUtils` intended?

RickHawesUSDS: Yes. I'm cleaning up an unnecessary close. The try-resource block does the close.

RickHawesUSDS:
@dtisza1 @jzulim Thanks for catching that the `display` value was missing in the comparison JSON. The same invalid drug was being used in the BCarrier and DME RIF sample files. Applied the same replacement to these files and `display` values are now present. Please re-review.

PR #187, opened 2020-01-04
BLUEBUTTON-1644: Adding Pipfiles

karlmdavis:
They're useful for running Ansible, etc. locally, and I've had them laying around locally forever. Eventually, it'd be good to try and consolidate them and also remove the older requirements.txt files that are still there, but... baby steps.

(Note that this has nothing to do with the referenced JIRA ticket, except that I was working on it when I finally decided to clean this up.)

https://jira.cms.gov/browse/BLUEBUTTON-1644


karlmdavis:
> There's some question as to if the `pipfile` idea is still active, [from their docs](https://github.com/pypa/pipfile/blob/master/README.rst) it isn't stable yet and there hasn't been work on the spec itself for a couple years.

This makes me a very sad 🐼. The user experience with `Pipfile`s has been excellent, as compared to the... less-than-excellent user experience I've had with `requirements.txt` files.

If you were going to ask the "Python community at large" what's going on with `Pipfile` vs. `requirements.txt`, what forum would you suggest for that?

karlmdavis:
Drat: see https://github.com/pypa/pipenv/issues/3742 and linked issues. Sounds like it _might_ get resurrected, but is definitely currently dead.

whytheplatypus:
In terms of usability the pip-tools suite makes managing requirements.txt files more .... manageable.

jzulim (approved):
pipfile is alive. https://github.com/pypa/pipenv/issues/3369

LGTM.

njdister:
> pipfile is alive. [pypa/pipenv#3369](https://github.com/pypa/pipenv/issues/3369)
> 
> LGTM.

Did pipfile die and get revived in the time this PR was open? ☠️

keithdadkins (approved):

PR #198, opened 2020-01-18
BLUEBUTTON 1516: Add the RFC

RickHawesUSDS:
**Why**
The MBI changes taken in total is a significant change to the BFD contract with its partners. 

**What**
An RFC is an agreed-upon process to document significant changes to the BFD.

### DIFF
--- a/rfcs/0005-mbi-search.md
+++ b/rfcs/0005-mbi-search.md
@@ -1,0 +1,65 @@
+# RFC Proposal
+[RFC Proposal]: #rfc-proposal
+
+* RFC Proposal ID: `0005-mbi-search` 
+* Start Date: 2020-1-15
+* RFC PR: [](https://github.com/rust-lang/rfcs/pull/0000)
+* JIRA Ticket(s):
+    * [BLUEBUTTON-1516](https://jira.cms.gov/browse/BLUEBUTTON-1516)
+
+The addition of a search by Medicare Beneficiary Identifier (MBI) facility to the BFD's Patient search end-point.   
+
+## Table of Contents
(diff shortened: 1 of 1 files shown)


rfcs/0005-mbi-search.md
```diff
@@ -0,0 +1,65 @@
...
+| value    | identifiers added | Comments                   |
+|----------|-------------------|----------------------------|
+| <empty>  |                   | Default value              |
+| true     | HICN and MBI      | MBI is added to the result |
+| hicn     | HICN              | New value supported        |
+| mbi      | MBI               | New value supported        |
+| hicn,mbi | HICN and MBI      | New value supported        |
+
+### Hash Algorithm
+For convenience, MBI hashing uses the existing HICN hashing algorithm and pepper.
```
whytheplatypus: Worth including that it's using the same itterations as well.

RickHawesUSDS: Roger on adding the iterations to the section. 

What do you have in mind for future work and hashing not scaling?

whytheplatypus: The future work I was thinking of has more to do with the process of on-boarding new peering partners. For a partner to use the hash, bfd has to communicate a shared secret to each partner. As more partners join operations concerning the shared secret (pepper) get more complicated (e.g. rolling it)

I'm not sure if this RFC is the right place for the discussion but it might be.

RickHawesUSDS: Ok. I understand the concern. I don't have any suggestions to improve the situation, so I'm going to leave out the issue.

PR #208, opened 2020-01-29
BLUEBUTTON-1737 Throw ResourceNotFoundException (404) for ID multiple matches 

dtisza1:
### Change Details

Summary:

This PR updates the `PatientResourceProvider.queryDatabaseByHash()` method used for HICN and MBI searches to throw a `ResourceNotFoundException` exception when multiple bene id's are found. An entry in the access.json logs (via MDC) is also made for auditing.

The ignored (using @ignore) IT tests in the `PatientResourceProviderIT` class were updated for the new behavior.

These tests are:  `searchForExistingPatientByHicnHashWithBeneDups()` and `searchForExistingPatientByMbiHashWithBeneDups()`

A new assert method `assertPatientByHashTypeMatch()` was created to be used by these tests. It combines the overlapping code from the two separate `assertPatientByHicnHashWithBeneDupsMatch()` and `assertPatientByMbiHashWithBeneDupsMatch()` methods.

### Acceptance Validation

That HICN or MBI search requests that map to multiple bene ID's will produce a `ResourceNotFoundException` exception (404 HTTP response status code). This was previously returning a `NonUniqueResultException` exception (500 HTTP response status code).

### Feedback Requested

* Does the naming of things and the text of the error/excpetion response look OK?

### External References

* [BLUEBUTTON-1737](https://jira.cms.gov/browse/BLUEBUTTON-1737)

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
@@ -26,7 +26,6 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
-import javax.persistence.NonUniqueResultException;
 import javax.persistence.PersistenceContext;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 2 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -423,7 +423,9 @@ private Patient queryDatabaseByHash(
     if (distinctBeneIds <= 0) {
       throw new NoResultException();
     } else if (distinctBeneIds > 1) {
-      throw new NonUniqueResultException();
+      MDC.put("database_query.by_hash.collision.distinct_bene_ids", Long.toString(distinctBeneIds));
+      throw new ResourceNotFoundException(
+          "By hash query found more than one distinct BENI_ID: " + Long.toString(distinctBeneIds));
```
RickHawesUSDS: Typo... BENI_ID

dtisza1: Yikes :-)  Will update that.

PR #209, opened 2020-02-03
BLUEBUTTON-1787: GitHub Actions CI for Java and Terraform development

njdister:
This PR adds the following two GitHub Actions workflows:
- On a PR that has changes inside the apps directory, run a `mvn verify` against the parent pom of the apps directory.  This will ensure all unit and integration tests pass prior to being merged.
- On a PR that has changes inside the terraform directory, run a `terraform validate` against each independent piece of terraform state.  This will ensure terraform is syntactically valid and internally consistent prior to being merged.  

The following changes were made to the mgmt stateful module:
- Adds a user, policy, and attachment that allows the GitHub Actions runner to run S3 integration tests.

Discussion:
- What else could be added/linted without increasing the security impact?  Some of the other terraform linting tools require access to the remote state so I don't think we'd want to go down that path unless we GitHub runners in our own environment.  
- Is my current assessment that creating an AWS user for GitHub Actions to run the Java S3 integration tests is acceptable from a security standpoint?  @StewGoin 
- The `setup-java` action uses the "Zulu Community distribution of OpenJDK" which I have never heard of, and could only find specific details on in the following reddit post: https://www.reddit.com/r/java/comments/8jvv8e/what_would_be_the_reasons_to_use_the_zulu_build/?sort=top  It seems safe for our use cases but worth putting it out there in case anyone has concerns.

### DIFF
--- a/.github/workflows/ci-java.yml
+++ b/.github/workflows/ci-java.yml
@@ -1,0 +1,40 @@
+name: 'CI - Java'
+on:
+  pull_request:
+    paths:
+      - apps/**
+jobs:
+  mvn-verify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Configure AWS credentials'
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.GA_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.GA_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: 'Checkout repo'
+        uses: actions/checkout@v2
+      - name: 'Setup JDK'
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: 'Generate maven toolchain config'
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-java.yml
```diff
@@ -0,0 +1,40 @@
+name: 'CI - Java'
+on:
+  pull_request:
+    paths:
+      - apps/**
+jobs:
+  mvn-verify:
```
whytheplatypus: Needs another `job` to run linting.

```
 mvn com.coveo:fmt-maven-plugin:check
```

nickrobison-usds:
Just some added color on the Zulu JDK. Azul does really good work, they have a bunch of ex-Sun folks over there. They’re committed to support Java 8 as a ton of their customers are still running it. Their builds all pass the TCK. You might also look at the generic Adopt OpenJDK builds, which Redhat and others maintain, it’s built straight from the mercurial repo.

StewGoin:
I reviewed this against the AWS recommendations and it looks like a good approach to me (or at least nothing jumped out as an OH NO!) so 👍 from me.

PR #216, opened 2020-02-11
BLUEBUTTON-1679 Remove hicn-hash identifier from Patient Resource 2

dtisza1:
### Change Details

Summary:

This is to remove the hashed HICN identifier from the patient resource response.

- Removed the `TransformerConstants.CODING_BBAPI_BENE_HICN_HASH` identifier from the `BeneficiaryTransformer.transform()` method .  This is used by the `PatientResourceProvider` for the FHIR Patient resource response. 
- Testing of the Patient resource using CURL is working as expected in the local docker setup. 
- Updated the unit tests in the `BeneficiaryTransformerTest` class. This included `transformSampleARecord()` and `transformSampleARecordWithIdentifiers()` methods. In addition to validating the count/size for the number of identifiers, the values are also validated.
- Updated the IT tests. These are performed by the `EndpointJsonResponseComparatorIT` class. This utilizes JSON files under the apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses path. These files are used to validate related end-point responses.  The `TransformerConstants.CODING_BBAPI_BENE_HICN_HASH` identifier was removed from 6 of these files.
- ALL TESTS ARE PASSING IN LOCAL DEV!
- Added a change entry to the apps/bfd-server/dev/api-changelog.md

### Acceptance Validation

- That the Patient resource responses DO NOT contain the hicn-hash identifier:  "https://bluebutton.cms.gov/resources/identifier/hicn-hash"

### External References

- [BLUEBUTTON-1679](https://jira.cms.gov/browse/BLUEBUTTON-1679)

### Security Implications


RickHawesUSDS:
Is the intention to never return the HICN and HICN hash? Even when include identifiers is set to hicn?

dtisza1:
> Is the intention to never return the HICN and HICN hash? Even when include identifiers is set to hicn?

@RickHawesUSDS I've updated per our stand up discussion for returning the hicn-hash in the response, only if the `includeIdentifiers` header has either "hicn" or "true" values. Otherwise, it is omitted in the response.

PR #218, opened 2020-02-13
Faster startup time

RickHawesUSDS:
**Why**
Our deployment code expects that startup time will be consistent. It will timeout if startup takes too long. The current _since does a full filter refresh during startup. The duration of this operation varies by the number of RIF files recently loaded. This algorithm increases the possibility of a timeout during startup.  

**Changes**
The current design sets up `transactionTime` during init time, so the server can return Bundles with a transactionTime. Instead of calling `refreshFilters` during init time, the LoadedFilterManger just fetches transactionTime directly during init time. RefreshFilters will happen periodically in the background as before. 

**Security Implications**
None

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
@@ -3,7 +3,11 @@ package gov.cms.bfd.server.war.stu3.providers;
 import ca.uhn.fhir.rest.param.DateRangeParam;
 import gov.cms.bfd.model.rif.LoadedBatch;
 import gov.cms.bfd.model.rif.LoadedFile;
-import java.util.*;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.annotation.PostConstruct;
@@ -27,6 +31,10 @@ import org.springframework.stereotype.Component;
 public class LoadedFilterManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(LoadedFilterManager.class);
 
+  // A date before the lastUpdate feature was rolled out
+  private static final Date BEFORE_LAST_UPDATE = Date.from(Instant.parse("2020-01-01T00:00:00Z"));
+
+  // The size of the beneficiaryId column
   private static final int BENE_ID_SIZE = 15;
 
   // The connection to the DB
@@ -35,11 +43,14 @@ public class LoadedFilterManager {
   // The filter set
   private List<LoadedFileFilter> filters;
 
-  // Max created date from the database's LoadedBatch table
+  // The latest transaction time from the LoadedBatch files
   private Date transactionTime;
 
-  // Min created date from the databases's LoadedBatch table
-  private Date firstBatchUpdate;
+  // The last LoadedBatch.created in the filter set
+  private Date lastBatchCreated;
+
+  // The first LoadedBatch.created in the filter set
+  private Date firstBatchCreated;
 
   /**
    * A tuple of values: LoadedFile.loadedFileid, LoadedFile.created, max(LoadedBatch.created). Used
@@ -96,11 +107,23 @@ public class LoadedFilterManager {
    *
    * @return the first batch's created timestamp
    */
-  public Date getFirstBatchUpdate() {
-    if (transactionTime == null) {
-      throw new RuntimeException("LoadedFilterManager has not been initialized.");
+  public Date getLastBatchCreated() {
+    if (lastBatchCreated == null) {
+      throw new RuntimeException("LoadedFilterManager has not been refreshed.");
     }
-    return firstBatchUpdate;
+    return lastBatchCreated;
+  }
+
+  /**
+   * The return the first batch that the filter manager knows about
+   *
+   * @return the first batch's created timestamp
+   */
+  public Date getFirstBatchCreated() {
+    if (firstBatchCreated == null) {
+      throw new RuntimeException("LoadedFilterManager has not been refreshed.");
+    }
+    return firstBatchCreated;
   }
 
   /**
@@ -115,8 +138,8 @@ public class LoadedFilterManager {
 
   /** Called to finish initialization of the manager */
   @PostConstruct
-  public void init() {
-    refreshFilters();
+  public synchronized void init() {
+    transactionTime = fetchLastLoadedBatchCreated();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -27,6 +31,10 @@
 public class LoadedFilterManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(LoadedFilterManager.class);
 
+  // A date before the lastUpdate feature was rolled out
+  private static final Date BEFORE_LAST_UPDATE = Date.from(Instant.parse("2020-01-01T00:00:00Z"));
+
+  // The size of the beneficiaryId column
```
RickHawesUSDS: This date is used in the case that the `LoadedBatches` table is empty. It just needs to be sometime in the past.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -35,11 +43,14 @@
...
 
-  // Max created date from the database's LoadedBatch table
+  // The latest transaction time from the LoadedBatch files
   private Date transactionTime;
 
-  // Min created date from the databases's LoadedBatch table
-  private Date firstBatchUpdate;
+  // The last LoadedBatch.created in the filter set
+  private Date lastBatchCreated;
+
```
RickHawesUSDS: `lastBatchCreated` is set by the `refreshFilters`.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -35,11 +43,14 @@
   // The filter set
   private List<LoadedFileFilter> filters;
 
-  // Max created date from the database's LoadedBatch table
+  // The latest transaction time from the LoadedBatch files
   private Date transactionTime;
```
RickHawesUSDS: `transactionTime` is set by `init` and `refreshFilters`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -115,8 +138,8 @@ public void setEntityManager(EntityManager entityManager) {
 
   /** Called to finish initialization of the manager */
   @PostConstruct
-  public void init() {
-    refreshFilters();
+  public synchronized void init() {
+    transactionTime = fetchLastLoadedBatchCreated();
```
RickHawesUSDS: Here's the algorithm change. `fetchLastLoadedBatchCreated` will have a consistent duration because of the index on `LoadedBatches`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -341,29 +375,30 @@ public static LoadedFileFilter buildFilter(
...
    */
-  private Date fetchLastLoadedBatchTime() {
+  private Optional<Date> fetchLastLoadedBatchCreated() {
     Date maxCreated =
         entityManager
             .createQuery("select max(b.created) from LoadedBatch b", Date.class)
             .getSingleResult();
-    return Optional.ofNullable(maxCreated).orElse(new Date());
+    return Optional.ofNullable(maxCreated);
   }
```
RickHawesUSDS: I've moved this else clause out of this function to the caller for code readability reasons. This logic shouldn't handle the case when the table is empty.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -188,25 +214,30 @@ public synchronized void refreshFilters() {
      */
     try {
       // If new batches are present, then build new filters for the affected files
-      final Date currentLastDatabaseUpdate = fetchLastLoadedBatchTime();
-      if (this.transactionTime == null || this.transactionTime.before(currentLastDatabaseUpdate)) {
+      final Date currentLastBatchCreated =
+          fetchLastLoadedBatchCreated().orElse(BEFORE_LAST_UPDATED_FEATURE);
+      if (this.lastBatchCreated == null || this.lastBatchCreated.before(currentLastBatchCreated)) {
         LOGGER.info(
```
njdister: If I'm not mistaken the parameters have changed but the general logic behind this statement has not correct?  I ask because currently in the test environment this is logging every second and causing logs to fill up quite a bit.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -188,25 +214,30 @@ public synchronized void refreshFilters() {
      */
     try {
       // If new batches are present, then build new filters for the affected files
-      final Date currentLastDatabaseUpdate = fetchLastLoadedBatchTime();
-      if (this.transactionTime == null || this.transactionTime.before(currentLastDatabaseUpdate)) {
+      final Date currentLastBatchCreated =
+          fetchLastLoadedBatchCreated().orElse(BEFORE_LAST_UPDATED_FEATURE);
+      if (this.lastBatchCreated == null || this.lastBatchCreated.before(currentLastBatchCreated)) {
```
whytheplatypus: Is this the equivalent of saying `if lastBatchCreated is null or lastBatchCreated is before transactionTime`?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -188,25 +214,30 @@ public synchronized void refreshFilters() {
...
 
         // If batches been trimmed, then remove filters which are no longer present
-        final Date currentFirstBatchUpdate = fetchFirstLoadedBatchTime();
-        if (this.firstBatchUpdate == null
-            || this.firstBatchUpdate.before(currentFirstBatchUpdate)) {
+        final Date currentFirstBatchUpdate =
+            fetchFirstLoadedBatchCreated().orElse(BEFORE_LAST_UPDATED_FEATURE);
+        if (this.firstBatchCreated == null
+            || this.firstBatchCreated.before(currentFirstBatchUpdate)) {
           LOGGER.info("Trimmed LoadedFile filters before {}", currentFirstBatchUpdate);
```
njdister: Same as above, this is currently logging every second in test.  Can this be changed to debug?

njdister:
@RickHawesUSDS Just a few comments on logging.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
RickHawesUSDS: This code shouldn't put a log entry every second, only when new RIF batch was loaded. It doesn't on my laptop. Can you send me the logs where it is?

RickHawesUSDS: Got your logs. I understand the problem and was able to reproduce it locally. The cause was using now() for the case where there was an empty LoadedBatches table.  I also confirmed that this change fixes the problem.

PR #221, opened 2020-02-19
BLUEBUTTON-1844: TrendMicro activation workaround

njdister:
Currently our deploys are timing out due to the failure of a TrendMicro activation script that is loaded during an early stage of cloud-init (earlier than our user-data runs anyway).  This does not disable that activation script, however it sends the blocking portion of the script to the background so cloud-init can continue.  

See here for approval: https://cmsgov.slack.com/archives/CH766L1GR/p1582077487059100?thread_ts=1582076464.057400&cid=CH766L1GR

Successfully deploys to test: https://builds.bfd-management.oeda.cmscloud.local/jenkins/job/test/job/BLUEBUTTON-1844-trendmicro-activation-workaround/3/

### DIFF
--- a/ops/ansible/playbooks-ccs/build_bfd-pipeline.yml
+++ b/ops/ansible/playbooks-ccs/build_bfd-pipeline.yml
@@ -56,16 +56,6 @@
       become: true
       tags:
         - pre-ami
-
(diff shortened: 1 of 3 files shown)


StewGoin:
👍

PR #224, opened 2020-02-19
BLUEBUTTON-1841 Create indexes for partdcontactnumber columns

whytheplatypus:
# Why
Searching for beneficiaries based on `partDContractNumber<month>Id` proved slow over large data sets. This is a simple solution that would naturally go away when coverage data is stored in a more robust way.

# What changed
~Indexes are created using the `hash` method on each of the `partDContractNumber<month>Id` columns. A `hash` was chosen because only `=` operations are supported in the search.~ See comments for reason behind not using the `hash` method.

### DIFF
--- a/apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
+++ b/apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
@@ -89,7 +89,6 @@ public final class DatabaseSchemaManager {
         placeholders.put("logic.index-create-concurrently", "");
         placeholders.put("logic.sequence-start", "start with");
         placeholders.put("logic.sequence-increment", "increment by");
-        placeholders.put("logic.non-hsql-only-line", "--");
       } else {
         placeholders.put("type.int4", "int4");
         placeholders.put("logic.tablespaces-escape", "--");
@@ -98,7 +97,6 @@ public final class DatabaseSchemaManager {
         placeholders.put("logic.index-create-concurrently", "concurrently");
         placeholders.put("logic.sequence-start", "start");
         placeholders.put("logic.sequence-increment", "increment");
-        placeholders.put("logic.non-hsql-only-line", "");
       }
     } catch (SQLException e) {
       throw new UncheckedSqlException(e);
--- a/apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
+++ b/apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
@@ -89,6 +89,7 @@ public final class DatabaseSchemaManager {
         placeholders.put("logic.index-create-concurrently", "");
         placeholders.put("logic.sequence-start", "start with");
         placeholders.put("logic.sequence-increment", "increment by");
+        placeholders.put("logic.using-hash", "");
       } else {
         placeholders.put("type.int4", "int4");
         placeholders.put("logic.tablespaces-escape", "--");
@@ -97,6 +98,7 @@ public final class DatabaseSchemaManager {
         placeholders.put("logic.index-create-concurrently", "concurrently");
         placeholders.put("logic.sequence-start", "start");
         placeholders.put("logic.sequence-increment", "increment");
+        placeholders.put("logic.using-hash", "using hash");
       }
     } catch (SQLException e) {
       throw new UncheckedSqlException(e);
--- a/apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
+++ b/apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
@@ -89,7 +89,7 @@ public final class DatabaseSchemaManager {
         placeholders.put("logic.index-create-concurrently", "");
         placeholders.put("logic.sequence-start", "start with");
         placeholders.put("logic.sequence-increment", "increment by");
-        placeholders.put("logic.using-hash", "");
+        placeholders.put("logic.non-hsql-only-line", "--");
       } else {
         placeholders.put("type.int4", "int4");
         placeholders.put("logic.tablespaces-escape", "--");
@@ -98,7 +98,7 @@ public final class DatabaseSchemaManager {
         placeholders.put("logic.index-create-concurrently", "concurrently");
         placeholders.put("logic.sequence-start", "start");
         placeholders.put("logic.sequence-increment", "increment");
-        placeholders.put("logic.using-hash", "using hash");
+        placeholders.put("logic.non-hsql-only-line", "");
       }
     } catch (SQLException e) {
       throw new UncheckedSqlException(e);
--- a/apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
+++ b/apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
@@ -1,36 +1,59 @@
 
 create index ${logic.index-create-concurrently} "Beneficiaries_partd_contract_number_jan_id_idx"
-    on "Beneficiaries" ${logic.using-hash} ("partDContractNumberJanId");
+    on "Beneficiaries" ("partDContractNumberJanId")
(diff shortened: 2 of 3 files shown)


apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
```diff
@@ -0,0 +1,36 @@
+
+create index ${logic.index-create-concurrently} "Beneficiaries_partd_contract_number_jan_id_idx"
+    on "Beneficiaries" ${logic.using-hash} ("partDContractNumberJanId");
```
whytheplatypus: This is missing the update to provide `${logic.using-hash}`

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
```diff
@@ -0,0 +1,36 @@
...
+
+create index ${logic.index-create-concurrently} "Beneficiaries_partd_contract_number_feb_id_idx"
+    on "Beneficiaries" ${logic.using-hash} ("partDContractNumberFebId");
+
```
RickHawesUSDS: Did you consider a partial index (ie. an index with a where clause)? Looking at the shape of these columns, the data look's very sparse, with `0`,`N`, and `C` being the most common values. Real contracts like `H0023` only make a small ~10% of the data. If possible, using a partial index will make the index much smaller and perform much better.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
```diff
@@ -0,0 +1,36 @@
...
+create index ${logic.index-create-concurrently} "Beneficiaries_partd_contract_number_mar_id_idx"
+    on "Beneficiaries" ${logic.using-hash} ("partDContractNumberMarId");
+
```
RickHawesUSDS: I like the idea of using a hash index.

whytheplatypus:
@RickHawesUSDS Does the use of [CONCURRENTLY](https://www.postgresql.org/docs/9.1/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) when creating the index sufficiently solve the problem of locking the table to writes?

RickHawesUSDS:
> @RickHawesUSDS Does the use of [CONCURRENTLY](https://www.postgresql.org/docs/9.1/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) when creating the index sufficiently solve the problem of locking the table to writes?

Yes. That should work.

RickHawesUSDS:
Here's another concern... From https://www.postgresql.org/docs/9.6/indexes-types.html. 

> Caution
> --
> Hash index operations are not presently WAL-logged, so hash indexes might need to be rebuilt with REINDEX after a database crash if there were unwritten changes. Also, changes to hash indexes are not replicated over streaming or file-based replication after the initial base backup, so they give wrong answers to queries that subsequently use them. For these reasons, hash index use is presently discouraged.

We are using streaming replication, so I think this makes hash indexes a **no go** in our setup. This caution is not present in modern versions 11 and 12 of Postgres. I think the team finished the HASH index feature. So, one way to go is to upgrade our Postgres version.

StewGoin:
> Here's another concern... From https://www.postgresql.org/docs/9.6/indexes-types.html.
> 
> > ## Caution
> > Hash index operations are not presently WAL-logged, so hash indexes might need to be rebuilt with REINDEX after a database crash if there were unwritten changes. Also, changes to hash indexes are not replicated over streaming or file-based replication after the initial base backup, so they give wrong answers to queries that subsequently use them. For these reasons, hash index use is presently discouraged.
> 
> We are using streaming replication, so I think this makes hash indexes a **no go** in our setup. This caution is not present in modern versions 11 and 12 of Postgres. I think the team finished the HASH index feature. So, one way to go is to upgrade our Postgres version.

This raises two curiosities in my mind: does the potential migration to Aurora effectively accomplish a version upgrade of the pgsql we're using, and are there any features / deprecations we need to get a handle on.

whytheplatypus:
@StewGoin it looks like it won't force an [upgrade from 9.6](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraPostgreSQL.Updates.20180305.html). That's not to say it's not a good chance to upgrade.

cc: @jzulim in case you have more context

karlmdavis:
> This raises two curiosities in my mind: does the potential migration to Aurora effectively accomplish a version upgrade of the pgsql we're using, and are there any features / deprecations we need to get a handle on.

We are 300% planning to upgrade to the latest PostgreSQL release as part of the Aurora migration! 🎉 

It's worth noting that neither RDS nor Aurora offer any mechanisms for major version upgrades, so a major effort like this will always be required. (No reason to incur all that pain twice, so we're going to do both at once.)

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
```diff
@@ -89,6 +89,7 @@ public static void dropIndexes(DataSource dataSource) {
         placeholders.put("logic.index-create-concurrently", "");
         placeholders.put("logic.sequence-start", "start with");
         placeholders.put("logic.sequence-increment", "increment by");
+        placeholders.put("logic.non-hsql-only-line", "--");
```
karlmdavis: This name makes my head hurt. Could we just call it the `postgresql-only-line`?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
```diff
@@ -0,0 +1,59 @@
...
+create index ${logic.index-create-concurrently} "Beneficiaries_partd_contract_number_jan_id_idx"
+    on "Beneficiaries" ("partDContractNumberJanId")
+    ${logic.non-hsql-only-line} where length("partDContractNumberJanId") = 5
+    ;
```
karlmdavis: Did you want to keep the partial index, even though the `CONCURRENT` likely covers the original concern?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -379,7 +379,11 @@ private CriteriaQuery queryBeneficiariesBy(
               beneMatchesRoot.fetch(f, JoinType.LEFT);
             });
     beneMatches.select(beneMatchesRoot);
-    beneMatches.where(builder.equal(beneMatchesRoot.get(field), value));
+    beneMatches.where(
+        builder.and(
+            builder.equal(beneMatchesRoot.get(field), value),
+            // Triggers indexes built with a where field = length clause
```
karlmdavis:
```suggestion
            // This additional `WHERE` clause must be present to trigger partial indexes, which were built with a where field = length clause.
```

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
whytheplatypus: sure, although that's also confusing since there's no mention of postgres anywhere else.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
whytheplatypus: I'm not seeing an appreciable relative improvement in performance with the partial index, so I'll plan on removing in the interest of simplicity.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V25__Add_partd_contract_number_index.sql
```diff
@@ -0,0 +1,36 @@
+
```
karlmdavis: Sorry, one last thing: I'd add a header comment here that explains the reason for the change and references the JIRA ticket.

I wouldn't do that everywhere, but makes sense for DB migrations to me, as I often read them start - end, and I suspect other folks do, too. Good chance to help folks understand how & why the app has evolved over time.

whytheplatypus: good call, done.

PR #225, opened 2020-02-19
BLUEBUTTON-1841 improve coverage query speed

whytheplatypus:
# Why
Queries on the Beneficiary table against the partDContractId columns are occasionally timing out. This work clarifies some common parts of query building allowing for simple optimizations to fall out without obfuscating behavior.

# What changed
A class `PageQueryBuilder` is created to encapsulate the tools used to build queries while still exposing all of their behaviors. It also adds some convenience things like `count()` which is a common operation to get the total matching count of entries against a set of criteria.

The order of operations has changed slightly in `searchByCoverageContract` allowing the system to skip some expensive edge cases when there is no matching data to return, or less data than the requested page size.

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PageQueryBuilder.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PageQueryBuilder.java
@@ -8,9 +8,9 @@ import javax.persistence.criteria.Root;
 
 public final class PageQueryBuilder<T> {
 
-  public final Root<T> root;
-  public final CriteriaBuilder builder;
-  public final CriteriaQuery criteria;
+  private final Root<T> root;
+  private final CriteriaBuilder builder;
+  private final CriteriaQuery criteria;
   private final EntityManager entityManager;
 
   public PageQueryBuilder(Class<T> returnType, EntityManager em) {
@@ -18,6 +18,18 @@ public final class PageQueryBuilder<T> {
     builder = entityManager.getCriteriaBuilder();
     criteria = builder.createQuery(returnType);
     root = criteria.from(returnType);
+  }
+
+  public Root<T> root() {
+    return this.root;
+  }
+
+  public CriteriaBuilder builder() {
+    return this.builder;
+  }
+
+  public CriteriaQuery criteria() {
+    return this.criteria;
   }
 
   private CriteriaQuery<Long> createCountCriteria() {
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PageQueryBuilder.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PageQueryBuilder.java
@@ -1,0 +1,37 @@
+package gov.cms.bfd.server.war.stu3.providers;
+
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+
+public final class PageQueryBuilder<T> {
+
+  public final Root<T> root;
+  public final CriteriaBuilder builder;
+  public final CriteriaQuery criteria;
+  private final EntityManager entityManager;
+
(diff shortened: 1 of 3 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PageQueryBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.persistence.Query;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+
+public final class PageQueryBuilder<T> {
+
+  public final Root<T> root;
+  public final CriteriaBuilder builder;
+  public final CriteriaQuery criteria;
```
dkrylovsb: Should these be `private` for encapsulation sake and -- if they need to be exposed to users of this class -- complemented by getter methods?

whytheplatypus: They certainly could be. I wasn't sure what the current idiomatic Java would be.

whytheplatypus: Most methods on these objects modify them which is what drove me to use `public` in this case. For example reading `qb.critera.where(<predicate>)` it doesn't surprise me that `qb.criteria` is now modified to include that `predicate`. Reading `qb.getCriteria().where(<predicate>)` it seems less obvious that the internal `qb.criteria` has been modified.

Do you think that holds up? Or is it just better to move to a getter pattern?

dkrylovsb: Sounds like _builder-style_ accessor methods could be a good middle ground preserving encapsulation while avoiding awkwardness of getter-style naming? `qb.criteria().where(<predicate>)`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -258,21 +257,40 @@ public Bundle searchByCoverageContract(
     CcwCodebookVariable partDContractMonth = partDCwVariableFor(contractMonth);
     SingularAttribute<Beneficiary, String> contractMonthField = partDFieldFor(partDContractMonth);
 
+    PageQueryBuilder<Beneficiary> queryBuilder =
+        new PageQueryBuilder<Beneficiary>(Beneficiary.class, entityManager);
+
+    queryBuilder
```
RickHawesUSDS: This type of statement where `queryBuilder` is using itself as input is hard for me to reason about.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PageQueryBuilder.java
```diff
@@ -0,0 +1,55 @@
+package gov.cms.bfd.server.war.stu3.providers;
+
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+
+public final class PageQueryBuilder<T> {
```
RickHawesUSDS: Not sure what abstraction this class is trying to model. Is it trying to be a replacement for `CriteriaBuilder`. If we are going to make a builder, my suggestion is that some type of fluent class:  `builder().setValueA().setValueB().build()`

whytheplatypus: The main thing it's _trying_ to hide is the handling of the three aspects of building these queries:
```
CriteriaBuilder
CriteriaQuery<T>
Root<T>
```
With an aim for readability and allowing things like `count` easily.

I like the fluent idea.

whytheplatypus:
closing in favor of any work that will be done on [BLUEBUTTON-1948](https://jira.cms.gov/browse/BLUEBUTTON-1948)

PR #227, opened 2020-02-22
Reducing the number of load threads to easy data loading

cthulhuplus:
https://confluence.cms.gov/pages/viewpage.action?pageId=271777831

Reduced the number of data_pipeline_loader_threads to ease the data load in response to Incident listed above.


njdister:
This only changes the default, and doesn't change the formula in the build and launch playbooks:
```
playbooks-ccs/build_bfd-pipeline.yml:        data_pipeline_loader_threads: "{{ data_pipeline_ec2_instance_type_vcpu * 25 }}"
playbooks-ccs/launch_bfd-pipeline.yml:        data_pipeline_loader_threads: "{{ data_pipeline_ec2_instance_type_vcpu * 25 }}"
```

Also we don't need to actually get this into code to change the threads on the existing prod instance temporarily.

PR #228, opened 2020-02-23
Fix latency issue when RIF files are being loaded. 

RickHawesUSDS:
**Why**
During a load of RIF files in production, Coverage calls were discovered to be slowing down.

**What Changed**
Replace a lock for the entire `refreshFilters` method with one just at the end when the state is set.

**Testing Done**
- Local IT
- Ad-hoc load testing of making a request while loading

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
@@ -236,7 +236,7 @@ public class LoadedFilterManager {
           newFilters = trimFilters(newFilters, loadedFiles);
         }
 
-        set(newFilters, currentFirstBatchUpdate, currentFirstBatchUpdate);
+        set(newFilters, currentFirstBatchUpdate, currentLastBatchCreated);
       }
     } catch (Exception ex) {
       LOGGER.error("Error found refreshing LoadedFile filters", ex);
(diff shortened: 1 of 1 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -204,10 +204,11 @@ public synchronized boolean isInBounds(DateRangeParam range) {
    * call.
    */
   @Scheduled(fixedDelay = 1000, initialDelay = 2000)
-  public synchronized void refreshFilters() {
+  public void refreshFilters() {
     /*
```
RickHawesUSDS: This the mistake. The design calls for this function to not interfere with requests. By making the call synchronous, it locks the manger and requests have to wait.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/LoadedFilterManager.java
```diff
@@ -232,26 +233,25 @@ public synchronized void refreshFilters() {
...
           LOGGER.info("Trimmed LoadedFile filters before {}", currentFirstBatchUpdate);
           List<LoadedFile> loadedFiles = fetchLoadedFiles();
-          this.filters = trimFilters(this.filters, loadedFiles);
-          this.firstBatchCreated = currentFirstBatchUpdate;
+          newFilters = trimFilters(newFilters, loadedFiles);
         }
 
-        // update the transaction time as well
-        this.transactionTime = currentLastBatchCreated;
+        set(newFilters, currentFirstBatchUpdate, currentFirstBatchUpdate);
```
RickHawesUSDS: Now, instead of setting the filters and other states during the method, the state is set once at the end.

njdister: Shouldn't the third parameter passed into `set` be `currentLastBatchCreated` rather than `currentFirstBatchUpdate` or am I misunderstanding the workflow?

PR #232, opened 2020-02-26
BLUEBUTTON-1644Update the flamegraph play to fix bitrot.

karlmdavis:
Wasn't working anymore, due to changes in the AWS environments, my local
system, etc.

Also moved it out of the HealthAPT Ansible directory.

BLUEBUTTON-1644

### DIFF
--- a/ops/ansible/playbooks-ccs/ansible.cfg
+++ b/ops/ansible/playbooks-ccs/ansible.cfg
@@ -42,4 +42,4 @@ pipelining = True
 #ssh_args = -o ControlMaster=no
 
 [inventory]
-enable_plugins = ini
+#enable_plugins = ini
--- a/ops/ansible/playbooks-ccs/data_server_collect_flame_graph.yml
+++ b/ops/ansible/playbooks-ccs/data_server_collect_flame_graph.yml
@@ -5,17 +5,17 @@
 #
 # Usage:
 #
-#     $ pipenv run ./ansible-playbook-wrapper data_server_collect_flame_graph.yml --extra-vars "perf_record_seconds=15 find_process_command='pgrep java' jdk_path=/etc/alternatives/java_sdk_1.8.0_openjdk" --limit bluebutton-healthapt-test-data-server-a-1
+#     $ pipenv run ./ansible-playbook-wrapper data_server_collect_flame_graph.yml --inventory='somehost.ec2.internal' --extra-vars "perf_record_seconds=15 find_process_command='pgrep -f bfd-server-launcher.*jetty' java_user='bfd' jdk_path=/etc/alternatives/java_sdk_1.8.0_openjdk"
 ##
 
 - name: Configure Data Server System
-  hosts: data_server_systems
+  hosts: all
   tasks:
 
     - name: Set Helper Variables
       set_fact:
         tools_root_remote: '/usr/local'
-        tools_root_local: '/usr/local'
+        tools_root_local: '/tmp'
         find_process_command_default: 'pgrep java'
         perf_record_seconds_default: 15
         jdk_path_default: '/etc/alternatives/java_sdk_1.8.0_openjdk'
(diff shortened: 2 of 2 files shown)


ops/ansible/playbooks-ccs/ansible.cfg
```diff
@@ -42,4 +42,4 @@ pipelining = True
 #ssh_args = -o ControlMaster=no
 
 [inventory]
-enable_plugins = ini
+#enable_plugins = ini
```
karlmdavis: Had to do this in order to get the `--inventory=foo.ec2.internal` play parameter working, as it requires the enabled-by-default-but-disabled-here `hosts_list` plugin to be enabled.

Anyone know why this config was locked down to just `ini`? If there's not a good reason, let's just deleted this line altogether as part of this PR.

ops/ansible/playbooks-ccs/data_server_collect_flame_graph.yml
```diff
@@ -5,17 +5,17 @@
 #
 # Usage:
 #
-#     $ pipenv run ./ansible-playbook-wrapper data_server_collect_flame_graph.yml --extra-vars "perf_record_seconds=15 find_process_command='pgrep java' jdk_path=/etc/alternatives/java_sdk_1.8.0_openjdk" --limit bluebutton-healthapt-test-data-server-a-1
+#     $ pipenv run ./ansible-playbook-wrapper data_server_collect_flame_graph.yml --inventory='somehost.ec2.internal' --extra-vars "perf_record_seconds=15 find_process_command='pgrep -f bfd-server-launcher.*jetty' java_user='bfd' jdk_path=/etc/alternatives/java_sdk_1.8.0_openjdk"
```
karlmdavis: Fun fact: `pgrep -f 'java.*DataServerLauncherApp'` would be a much better filter/search command here, but won't work for reasons that escape me. It looks like maybe `pgrep -f` won't search the command line past a certain character limit?

ops/ansible/playbooks-ccs/data_server_collect_flame_graph.yml
```diff
@@ -49,6 +48,7 @@
           - cmake
           - gcc
           - gcc-c++
+          - java-1.8.0-openjdk-devel
```
karlmdavis: We should consider updating our default builds to install the JDK like this, rather than the JRE. It's super useful to have a JDK around for all sorts of things.

ops/ansible/playbooks-ccs/data_server_collect_flame_graph.yml
```diff
@@ -92,6 +92,23 @@
...
       become: true
 
+      #    - name: Attach Java Symbol Agent to Java Process
+      #      shell: "java -cp attach-main.jar:$JAVA_HOME/lib/tools.jar net.virtualvoid.perf.AttachOnce {{ process_pid }}"
+      #      args:
+      #        chdir: "{{ perf_map_agent_path_remote }}/out"
+      #      environment:
+      #        JAVA_HOME: "{{ jdk_path }}"
+      #      become: true
+      #      become_user: "{{ java_user }}"
```
karlmdavis: Whoops: need to go uncomment this and make sure that things still work.

ops/ansible/playbooks-ccs/data_server_collect_flame_graph.yml
njdister: I'm pretty sure we already do: https://github.com/CMSgov/beneficiary-fhir-data/blob/a5e50a508c7df81fadb82f8736994d94adc4f6be/ops/ansible/roles/bfd-server/tasks/install.yml#L9

ops/ansible/playbooks-ccs/ansible.cfg
jzulim: No idea why it was locked to ini. Agreed we should remove.

jzulim (dismissed):
Looks good.

njdister (dismissed):

jzulim (approved):
LGTM

keithdadkins (approved):

PR #239, opened 2020-03-06
BLUEBUTTON-1900: Ensure all FHIR resources have a lastUpdated value

RickHawesUSDS:
**Why**

Based on conversations with the FHIR community, all resources should have a lastUpdated if any of them do. 

**What Changed**
- Created a fallback `lastUpdated` value for resource loaded in the database. 
- Used the fallback value for FHIR resources based on SQL entities that do have a lastUpdated value. This all handled in the setLastUpdated helper function.
- Tests associated with this logic

**What to review**
- The handling of null and optional values

**Security Impact**
N/A


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java
```diff
@@ -225,4 +227,10 @@
    */
   static final String CODING_BBAPI_BENE_HICN_HASH_OLD =
       "http://bluebutton.cms.hhs.gov/identifier#hicnHash";
+
+  /**
+   * Fallback value to use when a record does not have a lastUpdated value. These records where
+   * loaded before the lastUpdated feature was in place.
+   */
+  static final Date FALLBACK_LAST_UPDATED = Date.from(Instant.parse("2020-01-01T00:00:00Z"));
 }
```
RickHawesUSDS: This is an arbitrary date that is before any actual lastUpdated values. I've chosen the first second of the year, to look arbitrary.

PR #241, opened 2020-03-06
BLUEBUTTON-1900: Fix an intermittent failing test

RickHawesUSDS:
**Why**
Occasionally, BFDs integration tests will fail because of a small difference in lastUpdated values. Running the tests again and they will most likely pass. 

@dtisza1 Reported this example. 
>I ran into something interesting and maybe a possible intermittent issue.  I received this error when running the IT tests related to the lastUpdated time comparison in PatientResourceProvider.   In this case, the time is just off by a short amount.
However, I've run the IT test several more times and I can't reproduce the error again :thinking_face:
Just wondering if we might get intermittent cases were this test fails, when the time is slightly off?
Untitled 
Results :
Failed tests: 
  PatientResourceProviderIT.readExistingPatientIncludeIdentifiersHicnMbi:82->assertExistingPatientIncludeIdentifiersExpected:207 Expect lastUpdated to be equal expected:<2020-02-14T18:56:00.011Z> but was:<2020-02-14T18:55:59.991Z>

**Cause**
The issue is caused by our tests being run in parallel and the same DB being

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/TransformerTestUtils.java
+++ b/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/TransformerTestUtils.java
@@ -29,6 +29,8 @@ import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.math.BigDecimal;
+import java.time.Duration;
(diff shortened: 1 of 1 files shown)


RickHawesUSDS:
Found this bug occurring sometimes in our Jenkins build. https://github.com/CMSgov/beneficiary-fhir-data/runs/568875536?check_suite_focus=true is an example.

PR #245, opened 2020-03-13
BLUEBUTTON-1788: Change RifLoader log level

RickHawesUSDS:
**Why**
As part of OUTPATIENT load work, it was discovered that the pipeline load was taking a very long time in TEST (and production). 

**What Changed**
The RifLoader log level was set to debug. At the debug level, the pipeline would count every row in every table after every RIF load. Even in PROD, this query would take hours.  

To get this fix done quickly, this PR just lowers the log level to info. 

**Testing Done**
Local testing to confirm that changing the log level stops the counting step. 

**Future Action**
Filed ticket to fix the debug level problem at the root. 
https://jira.cms.gov/browse/BLUEBUTTON-1921

### DIFF
--- a/apps/bfd-pipeline/bfd-pipeline-app/src/main/resources/logback.xml
+++ b/apps/bfd-pipeline/bfd-pipeline-app/src/main/resources/logback.xml
@@ -12,8 +12,8 @@
 	<!-- At 'debug', Hibernate will log SQL statements. -->
 	<logger name="org.hibernate.SQL" level="info" />
 
-  <!-- At 'debug', will display timing information about the idle task batches -->
-	<logger name="gov.cms.bfd.pipeline.rif.load" level="debug" />
+    <!-- At 'debug', will display timing information about the idle task batches -->
+	<logger name="gov.cms.bfd.pipeline.rif.load" level="info" />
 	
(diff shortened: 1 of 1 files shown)


njdister:
@RickHawesUSDS Could you link to the code line where this behavior happens (even if not changed)?

RickHawesUSDS:
> @RickHawesUSDS Could you link to the code line where this behavior happens (even if not changed)?

@njdister Here's the function that is called at the process loop

https://github.com/CMSgov/beneficiary-fhir-data/blob/510b5a52f10b4daf89e2a358b21fa2c07f5853dc/apps/bfd-pipeline/bfd-pipeline-rif-load/src/main/java/gov/cms/bfd/pipeline/rif/load/RifLoader.java#L724-L763

RickHawesUSDS:
> @RickHawesUSDS What about moving the relevant code to a separate logger category, instead? Feels like it'd be less likely to regress in the future -- particularly if it gets added to the logger conf with a comment explaining what it does and why it's dangerous.

@karlmdavis Good suggestion. I've added a specific line for the `recordCounts` category so that it is better documented and protected from inheriting it's parent's level. Please review again.

PR #246, opened 2020-03-18
BLUEBUTTON 1788: Outpatient Data Fixes and Tests

RickHawesUSDS:
**Why**
The last attempt at loading this Outpatient data failed. This PR documents the changes made to fix this bug and more importantly the test added to fully test every record so that we confidence in the data load. 

**What Changed**

- The root cause bug was identified as a missing PTNT_DSCHRG_STUS_CD field in some records. Fortunately, a good default value exists for this field. 0-Unknown Value was used in these cases.
- Discovered a corner case typo bug in the Outpatient transformer. This was fixed.
- A new test that loads and requests all 174k synthetic outpatient records was added. Previously, only a sample of these records was tested. 
- Work was done to optimize the execution speed of the outpatient tests. Without these optimizations, the 174k record test didn't complete in a day. This should have a small benefit to normal testing every developer does. 
- For documentation purposes, added all scripts used. 

**Testing Done**
- New integration test for every record
- Manual local loading of data and requesting every record

**Deployment Plan**
- Get PR approved
- Deploy to test and request every record
- Pair with JZ and deploy to sandbox and prod. Test every record. 

**Review Guidance**
- Look over the fixup python script for the repair to the data
- Look over the test to see if it is complete

### DIFF
--- a/ops/ccs-ops-misc/synthetic-data/scripts/outpatient/convert_files.sh
+++ b/ops/ccs-ops-misc/synthetic-data/scripts/outpatient/convert_files.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/bash
+
+# Dropping the outpclm2000_encyp_s2000 and optclm2001_encyp_s2000 files because they contain duplicate claim ids. 
+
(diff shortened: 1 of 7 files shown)


ops/ccs-ops-misc/synthetic-data/scripts/outpatient/convert_files.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/bash
```
RickHawesUSDS: For documentation purposes, I decided to put these files into a new script directory.

PR #257, opened 2020-04-21
BFD-Insights initial PR

RickHawesUSDS:
**Why**
The basic idea behind the BFD-Insights project is listed here. https://confluence.cms.gov/display/BB/Business+and+Security+Analytics+for+DASG

**What**
This is the initial check-in for bfd-insights. It creates the basic terraform for the project. As an example workflow, a food truck project.

See the `doc` folder for details. 

**Choices Made**
- `insights` top-level folder
- Use terraform for deployment including table definitions
- Food Truck example project

**Future Work**
The choice of terraform leads to thinking that GitHub actions will be used for deploys.


insights/scripts/foodtruck/Pipfile
```diff
@@ -0,0 +1,13 @@
...
+verify_ssl = true
+
+[dev-packages]
+pylint = "*"
+
+[packages]
+boto3 = "*"
+
+[requires]
+python_version = "3.8"
```
njdister, line 13: There was some discussion here as to whether the pipfile standard was being maintained: https://github.com/CMSgov/beneficiary-fhir-data/pull/187

insights/terraform/groups/main.tf
```diff
@@ -0,0 +1,59 @@
...
+  member_map  = {
+    analysts  = ["HWRI", "SKKH"]
+    authors   = ["TUVE"]
+    readers   = []
+  }
+}
+
+# Add to the IAM groups
+
+data "aws_iam_group" "groups" {
```
njdister, line 12: Shouldn't this pull from `aws_iam_user` or am I misunderstanding how this works?

njdister: Nevermind, it took me slightly longer to reason that groups and group memberships were managed in separate places

insights/terraform/modules/bucket/main.tf
```diff
@@ -0,0 +1,39 @@
...
+
+resource "aws_s3_bucket" "main" {
+  bucket    = "bfd-insights-${var.sensitivity}-${data.aws_caller_identity.current.account_id}"
+  acl       = "private"
+  tags      = merge({sensitivity = var.sensitivity}, var.tags)
+
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm     = "AES256"
```
njdister, line 11: Any reason to go with AES over KMS?

insights/terraform/modules/database/main.tf
```diff
@@ -0,0 +1,37 @@
+data "aws_caller_identity" "current" {}
+
+data "aws_s3_bucket" "moderate_bucket" {
+  bucket      = "bfd-insights-moderate-${data.aws_caller_identity.current.account_id}"
```
njdister, line 4: The module takes in a sensitivity var, but moderate is hardcoded here.  Is this intended?

insights/terraform/modules/firehose/main.tf
```diff
@@ -0,0 +1,63 @@
+data "aws_caller_identity" "current" {}
+
+data "aws_s3_bucket" "moderate_bucket" {
+  bucket      = "bfd-insights-moderate-${data.aws_caller_identity.current.account_id}"
```
njdister, line 4: The module takes in a sensitivity var, but moderate is hardcoded here. Is this intended?

insights/terraform/modules/table/main.tf
```diff
@@ -0,0 +1,59 @@
+data "aws_caller_identity" "current" {}
+
+data "aws_s3_bucket" "moderate_bucket" {
+  bucket      = "bfd-insights-moderate-${data.aws_caller_identity.current.account_id}"
```
njdister, line 4: The module takes in a sensitivity var, but moderate is hardcoded here. Is this intended?

insights/scripts/foodtruck/Pipfile
RickHawesUSDS: Interesting. I've been using pipenv because it is similar to NPM. Nevertheless, I'll go to requirements.txt.

RickHawesUSDS: Added a requirements.txt file.

insights/terraform/modules/bucket/main.tf
RickHawesUSDS: AES is uses the KMS, so fundementally the security offered is the same. From a configuration point of view, it is difficult and more complex to get all the pieces to use a CMK stored in KMS. I couldn't get it to work on with Athena, for example.

insights/terraform/modules/database/main.tf
RickHawesUSDS: Yes. We may refactor when a second bucket is introduced, but that is few weeks away.

insights/terraform/modules/table/main.tf
RickHawesUSDS: See previous answer

insights/terraform/modules/bucket/main.tf
RickHawesUSDS: Look at the new changes. Switched to a CMK.

RickHawesUSDS:
@njdister I believe your concerns have been addressed.

RickHawesUSDS:
> Looking really good, just an informational question: now that there are separate project buckets, but also a data lake bucket, what are the delineation for their uses?  Project buckets for the databases, and data lake bucket for user queries/scripts?

@njdister Project buckets will contain the raw logs from the databases. They may also contain aggregations that are project-specific. The main data lake bucket will contain common aggregations (which there are none right now). From a security perspective, the project buckets are the entry-points for the project. The data-lake bucket is not accessable by the partners.

dtisza1:
@RickHawesUSDS A good way to improve package manager level security is to utilize the pip-tools hash-checking feature. I think this should be added to this project. Maybe as a backlog item. 
For example, this uses the pip-compile "--generate-hashes" option when generating the requirements.txt file.

PR #258, opened 2020-04-22
BLUEBUTTON-2002 Return synthetic resources with negative IDs

dtisza1:
### Change Details
NOTE: This is a continuation of a previous PR with the target branch moved local.  Please reference  that PR for previous review comments at:  https://github.com/CMSgov/beneficiary-fhir-data/pull/175

#### Summary:
A regex for parsing the format of EoB and Coverage resource read requests did not successfully parse synthetic Resource read requests due to their negative IDs. The result was that no synthetic EoB or Coverage resources were found on Read requests.

This updates those regex to allow for the IDs of these Resources to be negative.

#### The following was performed:
* Moved the target branch to local and created a new PR so that this works with our deployment process.
* For the Coverage and EOB `read()` ID parameter, changed the exception from `ResourceNotFoundException` to `IllegalArgumentException` when an `IdText` does not match the expected Regex pattern. I believe this failed match should be invalid rather than indicating an ID was not found.
* Updated the IT tests for Coverage and EOB `read()` methods to test that the ID parameters work as expected for negative IDs and IDs that do not pass the regex pattern for a valid format.
* Updated comments per Karl's review suggestions.

### Acceptance Validation
* That Coverage and EOB `read()` requests are able to handle negative IDs.
* That Coverage and EOB `read()` requests with an invalid format ID throw an  `IllegalArgumentException` exception.
* That IT tests are updated to test negative IDs and invalid format.

### External References

- [BLUEBUTTON-2002](https://jira.cms.gov/browse/BLUEBUTTON-2002)
- [BFD-155](https://jira.cms.gov/browse/BFD-155)

### Security Implications

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
@@ -45,7 +45,10 @@ import org.springframework.stereotype.Component;
  */
 @Component
 public final class CoverageResourceProvider implements IResourceProvider {
-  /** A {@link Pattern} that will match the {@link Coverage#getId()}s used in this application. */
+  /**
+   * A {@link Pattern} that will match the {@link Coverage#getId()}s used in this application, e.g.
+   * <code>part-a-1234</code> or <code>part-a--1234</code> (for negative IDs).
+   */
   private static final Pattern COVERAGE_ID_PATTERN =
       Pattern.compile("(\\p{Alnum}+-\\p{Alnum})-(-?\\p{Alnum}+)");
 
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
@@ -46,7 +46,8 @@ import org.springframework.stereotype.Component;
 @Component
 public final class CoverageResourceProvider implements IResourceProvider {
   /** A {@link Pattern} that will match the {@link Coverage#getId()}s used in this application. */
-  private static final Pattern COVERAGE_ID_PATTERN = Pattern.compile("(.*)-(\\p{Alnum}+)");
+  private static final Pattern COVERAGE_ID_PATTERN =
+      Pattern.compile("(\\p{Alnum}+-\\p{Alnum})-(-?\\p{Alnum}+)");
 
   private static final Logger LOGGER = LoggerFactory.getLogger(CoverageResourceProvider.class);
 
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
@@ -106,7 +106,9 @@ public final class CoverageResourceProvider implements IResourceProvider {
     operation.publishOperationName();
 
     Matcher coverageIdMatcher = COVERAGE_ID_PATTERN.matcher(coverageIdText);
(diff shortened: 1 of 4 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
```diff
@@ -85,7 +86,9 @@ public Coverage read(@IdParam IdType coverageId) {
       throw new IllegalArgumentException();
 
     Matcher coverageIdMatcher = COVERAGE_ID_PATTERN.matcher(coverageIdText);
-    if (!coverageIdMatcher.matches()) throw new ResourceNotFoundException(coverageId);
+    if (!coverageIdMatcher.matches())
+      throw new IllegalArgumentException("Unsupported ID pattern: " + coverageIdText);
```
dtisza1, line 90: Does this change make sense to do? Does it have a potential to cause any incompatibility issues?
I believe that the exception here should indicate that the ID parameter pattern is invalid rather than not found. This also enables the ability to more easily test this via the IT tests.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -102,7 +102,9 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
     if (eobIdText == null || eobIdText.trim().isEmpty()) throw new IllegalArgumentException();
 
     Matcher eobIdMatcher = EOB_ID_PATTERN.matcher(eobIdText);
-    if (!eobIdMatcher.matches()) throw new ResourceNotFoundException(eobId);
+    if (!eobIdMatcher.matches())
+      throw new IllegalArgumentException("Unsupported ID pattern: " + eobIdText);
```
dtisza1, line 106: Does this change make sense to do? Does it have a potential to cause any incompatibility issues?
I believe that the exception here should indicate that the ID parameter pattern is invalid rather than not found. This also enables the ability to more easily test this via the IT tests.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/CoverageResourceProvider.java
dtisza1: NOTE:  The `java.lang.IllegalArgumentException` results in a `ca.uhn.fhir.rest.server.exceptions.InvalidRequestException` exception here as seen in the IT tests. Was following other related code examples, but should these all be changed to `InvalidRequestException`?

njdister: How do we differentiate between a valid pattern that doesn't have an ID in the database, and an invalid pattern?

dtisza1: > How do we differentiate between a valid pattern that doesn't have an ID in the database, and an invalid pattern?
@njdister Further down in the code, a `ResourceNotFoundException` is thrown when the ID is not found. I think your question supports the change for the pattern not matching to throw an `IllegalArgumentException` here instead. Otherwise, would not be able to tell the difference and be able to easily test for this difference.

RickHawesUSDS: I agree that this change makes sense and is helpful. We shouldn't try to preserve bad exceptions for assumed compatibility.

dtisza1:
Currently working to do some final testing with BB20 and will merge in to master.

dtisza1:
I was just able to perform some final testing of this using the BFD & BB2 TEST environments.
It is working as expected! :smile_cat: 

There is some minor work needed for BB2 to handle the `IllegalArgumentException` result code. It is currently producing a 502 result. I'll add a story in Jira for resolving this.

This is ready to merge to master!  TY!

PR #260, opened 2020-04-23
BLUEBUTTON-1864: Medicare Opt Out bucket fixes, AB2D onboarding

njdister:
### Change Details

This PR contains some minor changes:
- Fixes the KMS policy
- Fixes the bucket policy by removing a non-functional deny statement
- Removes offboarded admin
- Adds AB2D roles as readers to test (not currently used) and prod-sbx

### Acceptance Validation

- AB2D can make S3 requests to prod-sbx bucket

### External References

- [BLUEBUTTON-1864](https://jira.cms.gov/browse/BLUEBUTTON-1864)

### Security Implications

- [X] altered security controls

This is worth discussing, as it was implemented that we'd have an explicit whitelist through denying actions via `NotPrincipal`, however I haven't been able to make this work in practice.  I'm unsure how much this changes to our security posture since we already have PII buckets (data from CCW) without this in place.

### DIFF
--- a/ops/terraform/env/test/stateful/main.tf
+++ b/ops/terraform/env/test/stateful/main.tf
@@ -59,7 +59,7 @@ module "stateful" {
   victor_ops_url      = var.victor_ops_url
 
   medicare_opt_out_config = {
-    read_roles        = ["arn:aws:iam::755619740999:role/dpc-dev-consent-execution-role"]
+    read_roles        = ["arn:aws:iam::755619740999:role/dpc-dev-consent-execution-role", "arn:aws:iam::349849222861:role/Ab2dInstanceRole", "arn:aws:iam::777200079629:role/Ab2dInstanceRole", "arn:aws:iam::330810004472:role/Ab2dInstanceRole"]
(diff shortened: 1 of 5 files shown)


ops/terraform/env/test/stateful/main.tf
```diff
@@ -59,8 +59,8 @@ module "stateful" {
   victor_ops_url      = var.victor_ops_url
 
   medicare_opt_out_config = {
-    read_roles        = ["arn:aws:iam::755619740999:role/dpc-dev-consent-execution-role"]
+    read_roles        = ["arn:aws:iam::755619740999:role/dpc-dev-consent-execution-role", "arn:aws:iam::349849222861:role/Ab2dInstanceRole", "arn:aws:iam::777200079629:role/Ab2dInstanceRole", "arn:aws:iam::330810004472:role/Ab2dInstanceRole"]
```
RickHawesUSDS, line 62: Does AB2D run on different accounts?

njdister: @RickHawesUSDS They are using the new cloudtamer setup which dictates one account per environment, so yes.

PR #262, opened 2020-04-23
BLUEBUTTON 2001: Slow Coverage search with paging and identifiers

RickHawesUSDS:
**Motivation**
https://jira.cms.gov/browse/BLUEBUTTON-2001 reports a bug where a coverage searches together with paging and include-identifiers timed-out. 

**Cause**
When include-identifiers is set, the search uses a `join` clause to fetch the bene-id history. When paging is enabled, the search uses a `limit` clause to limit the number of returned results. 
There is subtle behavior with Hibernate where joins and limits cannot be used together; In this case, Hibernate drops the limit. 

This post explains the problem as well.
https://stackoverflow.com/questions/53569908/jpa-eager-fetching-and-pagination-best-practices

Note: The current code returns correct or nearly correct results. Test cases passed and the current code works small history datasets like those in SBX. 

**Changes Made**
When paging and include identifiers are set, execute the query in two steps: First, fetch bene-ids using a `limit` clause. Next, fetch a full list of beneficiaries with joins. 

Also did some refactoring was done to make the code more readable. This work included switching to JPQL as the language for the query. 

**Reviewer Guidance**
Look that the existing queries did not change. Look at the queries for edge cases. 

**Security**
No impact

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
@@ -345,7 +345,8 @@ public final class PatientResourceProvider implements IResourceProvider {
 
     // Fetching with joins is not compatible with setMaxResults as explained in this post:
     // https://stackoverflow.com/questions/53569908/jpa-eager-fetching-and-pagination-best-practices
-    // So, in cases where there are joins and paging, we query in two steps.
+    // So, in cases where there are joins and paging, we query in two steps: first fetch bene-ids
+    // with paging and the fetch full benes with joins.
     boolean useTwoSteps =
         (hasHICN(includedIdentifiers) || hasMBI(includedIdentifiers)) && paging.isPagingRequested();
     if (useTwoSteps) {
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
@@ -23,7 +23,6 @@ import gov.cms.bfd.model.rif.BeneficiaryHistory_;
 import gov.cms.bfd.model.rif.Beneficiary_;
 import gov.cms.bfd.server.war.Operation;
 import java.math.BigDecimal;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -41,7 +40,6 @@ import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.JoinType;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
-import javax.persistence.metamodel.SetAttribute;
 import javax.persistence.metamodel.SingularAttribute;
 import org.hl7.fhir.dstu3.model.Bundle;
 import org.hl7.fhir.dstu3.model.IdType;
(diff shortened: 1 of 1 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -329,19 +332,38 @@ private CcwCodebookVariable partDCwVariableFor(String system) {
    */
   private List<Beneficiary> fetchBeneficiaries(
```
RickHawesUSDS, line 333: All the changes happened in this method and the methods it calls.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -353,31 +375,74 @@ private CcwCodebookVariable partDCwVariableFor(String system) {
    * @param joins to add for many-to-one relations
    * @return the criteria
    */
-  private CriteriaQuery<Beneficiary> queryBeneficiariesBy(
+  private TypedQuery<Beneficiary> queryBeneficiariesBy(
       SingularAttribute<Beneficiary, String> field,
```
RickHawesUSDS, line 379: Moving to TypedQuery return type improves the readability of the caller.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -353,31 +375,74 @@ private CcwCodebookVariable partDCwVariableFor(String system) {
...
-                  cb.equal(b.get(field), value),
-                  cb.greaterThan(b.get("beneficiaryId"), paging.getCursor())))
-          .orderBy(cb.asc(b.get("beneficiaryId")));
-    } else {
-      return criteria
-          .select(b)
-          .where(cb.equal(b.get(field), value))
-          .orderBy(cb.asc(b.get("beneficiaryId")));
-    }
+    Predicate whereClause =
```
RickHawesUSDS, line 389: Refactoring the whereClause for readability. The select statement is now the same for both paging and non-paging cases.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -353,31 +375,74 @@ private CcwCodebookVariable partDCwVariableFor(String system) {
...
+   * @param paging to use for the result set
+   * @return the criteria
+   */
+  private TypedQuery<String> queryBeneficiaryIds(
+      SingularAttribute<Beneficiary, String> field, String value, PatientLinkBuilder paging) {
+    CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+    CriteriaQuery<String> criteria = cb.createQuery(String.class);
+    Root<Beneficiary> b = criteria.from(Beneficiary.class);
+
+    Predicate whereClause =
```
RickHawesUSDS, line 414: Although this whereClause is similar to the previous whereClause in the previous method. It is not. I tried to build a common function and it didn't work.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -350,36 +353,96 @@ private CcwCodebookVariable partDCwVariableFor(String system) {
...
+          "select b from Beneficiary b "
+              + joinsClause
+              + "where b."
+              + field
+              + " = :value "
+              + "order by b.beneficiaryId asc";
+
+      return entityManager.createQuery(query, Beneficiary.class).setParameter("value", value);
+    }
+  }
```
RickHawesUSDS, line 390: Switching over to the JPQL style for these queries. Easier to understand and read to my eyes. Simpler code as well.

RickHawesUSDS:
To verify the fix, I took a server out of rotation in PROD and ran the same curl statement that was timing out after 30 seconds before. It now runs in 1s.  
```
[richard.hawes@ip-10-252-10-202 bfd-server]$ time sudo curl --insecure --header "IncludeIdentifiers: true" --cert ./bluebutton-backend-test-data-server-client-test-keypair.pem 'https://localhost:7443/v1/fhir/Patient?_has:Coverage.extension=https://bluebutton.cms.gov/resources/variables/ptdcntrct01|S5601&_count=100&_format=json' > ~/test.json
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 15.9M    0 15.9M    0     0  12.8M      0 --:--:--  0:00:01 --:--:-- 12.8M

real	0m1.311s
user	0m0.101s
sys	0m0.153s
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
dtisza1: I like this and is easier to read for me too!

PR #271, opened 2020-05-12
make random port range configurable from mvn cmd line params - make f…

JFU-GIT:
…ixed port use case easy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,25 +1,67 @@
-MyMedicare.gov BlueButton Parent POM
-====================================
+<!--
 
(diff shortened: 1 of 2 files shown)


JFU-GIT:
Hi John, it seems that the suggested changes are already in BFD trunk, so no need for this PR. here is the pom.xml excerpt I got from bfd master:

	<properties>
		<jersey.version>2.25.1</jersey.version>

		
		<its.bfdServer.jvmargs>-Xmx4g</its.bfdServer.jvmargs>

		
		<bash.exe>bash</bash.exe>
		
		<min_port>8000</min_port>
		<max_port>9999</max_port>
	</properties>

JFU-GIT:
this PR can be deleted.

PR #273, opened 2020-05-13
BLUEBUTTON-1864: Refactor medicare opt out module to support assumed writer role

njdister:
**JIRA Ticket:**
[BLUEBUTTON-1864](https://jira.cms.gov/browse/BLUEBUTTON-1864)

Refactor medicare opt out module to support assumed writer role

### What Does This PR Do?

- Refactors the `s3_pii` module so that we no longer pass in a list of external writer roles, but instead pass in a list of writer AWS accounts that will assume a writer role we create.  This solves the issue of a bucket owner not being able to apply policies to objects owned by another account.
- While refactoring this module I ran into multiple issues with JSON formatting due to terraform template string interpolation, so I have wrapped the JSON templates in `jsonencode` and moved to native terraform expressions as per recommendations: https://www.terraform.io/docs/configuration/functions/templatefile.html#generating-json-or-yaml-from-a-template
- Restructured the bucket policy a bit so it is simpler to reason about and reduces duplication
- Minor formatting changes
- Adds @RickHawesUSDS as an admin

This has successfully been applied in test, prod-sbx, and prod (and tested by 3rd party teams in prod-sbx) however it would be good to watch out for:
- Unintended side effects
- General clarity

* If the answer to any of the questions below is **Yes**, then here's a link to the associated Security Impact Assessment: N/A.
    * Does this PR add any new software dependencies? **No**.
    * Does this PR modify or invalidate any of our security controls?**No**.
    * Does this PR store or transmit data that was not stored or transmitted before? **No**.
* If the answer to any of the questions below is **Yes**, then please add @StewGoin as a reviewer, and note that this PR should not be merged unless/until he also approves it.
    * Do you think this PR requires additional review of its security implications for other reasons? **No**.

N/A

### Submitter Checklist

Before requesting final review, I've gone through and verified that this PR complies with the following requirements:

* [x] I've refactored and rebased this PR (for help, see: [this](https://dev.to/maxwell_dev/the-git-rebase-introduction-i-wish-id-had) and [this](https://raphaelfabeni.com/git-editing-commits-part-1/), if needed, so that it is as small as it can reasonably be, in order to:
    1. Ensure that any problems it causes have a small "blast radius".
    2. Ensure that it'll be easier to rollback if that becomes necessary.
    3. Ease the burden on reviewers.
* [ ] I've verified that the commits in this PR:
    1. Reasonably explain the "what" and "why" of the changes.
    2. Leverage JIRA's [smart commits](https://confluence.atlassian.com/jirasoftwarecloud/processing-issues-with-smart-commits-788960027.html) to reference the JIRA ticket that they're associated with. For example, aim for commit messages like this (note also [the 50/72 formatting](https://stackoverflow.com/q/2290016) used here):
        
        ```
        Added the whizbang to the doodad.

        The new whizbang should really simplify things for our whoosit users.
        It was a bit tricky to get it into the doodad, but we decided that the
        flipwhee pattern was the best approach for now. Might want to revisit
        that in the future if it ends up being too hard to maintain.

        SOMEPROJECT-42
        ```
        
* [x] I've verified that this PR includes all required documentation changes, including `README` updates and changelog / release notes entries.
* [ ] I've verified that all new and modified code is appropriately commented, such that the what and why of its design would be reasonably clear to engineers, preferably ones unfamiliar with the project.
* [x] I've verified that all tech debt and/or shortcomings introduced by this PR is detailed in `TODO` and/or `FIXME` comments, which include a JIRA ticket ID for any items that require urgent attention.
* [x] I've requested review from at least two other engineers on this project, at least one of whom is a senior engineer or owns the relevant component(s) here.
* [x] I've requested review from any relevant engineers on other projects (e.g. BFD, SLS, etc.).
* [x] I've verified that this PR and its changes comply with all other policies in the [DASG Engineering Standards](../policies/engineering_standards.md) and have specifically called out any/all deviations in this PR.

### DIFF
--- a/ops/terraform/env/prod-sbx/stateful/main.tf
+++ b/ops/terraform/env/prod-sbx/stateful/main.tf
@@ -52,6 +52,7 @@ module "stateful" {
   medicare_opt_out_config = {
     read_roles        = ["arn:aws:iam::755619740999:role/dpc-dev-consent-execution-role", "arn:aws:iam::349849222861:role/Ab2dInstanceRole", "arn:aws:iam::777200079629:role/Ab2dInstanceRole", "arn:aws:iam::330810004472:role/Ab2dInstanceRole"]
     write_roles       = ["arn:aws:iam::755619740999:role/bcda-dev-nfs-instance", "arn:aws:iam::755619740999:role/bcda-test-nfs-instance"]
+    write_assume_acct = "arn:aws:iam::755619740999:root"
     admin_users       = ["arn:aws:iam::577373831711:user/DS7H", "arn:aws:iam::577373831711:user/VZG9"]
   }
 }
--- a/ops/terraform/env/prod-sbx/stateful/main.tf
+++ b/ops/terraform/env/prod-sbx/stateful/main.tf
@@ -51,8 +51,7 @@ module "stateful" {
 
   medicare_opt_out_config = {
     read_roles        = ["arn:aws:iam::755619740999:role/dpc-dev-consent-execution-role", "arn:aws:iam::349849222861:role/Ab2dInstanceRole", "arn:aws:iam::777200079629:role/Ab2dInstanceRole", "arn:aws:iam::330810004472:role/Ab2dInstanceRole"]
-    write_roles       = ["arn:aws:iam::755619740999:role/bcda-dev-nfs-instance", "arn:aws:iam::755619740999:role/bcda-test-nfs-instance"]
-    write_assume_acct = "arn:aws:iam::755619740999:root"
-    admin_users       = ["arn:aws:iam::577373831711:user/DS7H", "arn:aws:iam::577373831711:user/VZG9"]
+    write_accts       = ["arn:aws:iam::755619740999:root"]
+    admin_users       = ["arn:aws:iam::577373831711:user/DS7H", "arn:aws:iam::577373831711:user/VZG9", "arn:aws:iam::755619740999:user/HWRI"]
   }
 }
--- a/ops/terraform/env/prod-sbx/stateful/main.tf
+++ b/ops/terraform/env/prod-sbx/stateful/main.tf
@@ -50,7 +50,7 @@ module "stateful" {
   victor_ops_url      = var.victor_ops_url
 
   medicare_opt_out_config = {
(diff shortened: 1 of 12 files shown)


ops/terraform/env/prod-sbx/stateful/main.tf
```diff
@@ -51,7 +51,7 @@ module "stateful" {
 
   medicare_opt_out_config = {
     read_roles        = ["arn:aws:iam::755619740999:role/dpc-dev-consent-execution-role", "arn:aws:iam::349849222861:role/Ab2dInstanceRole", "arn:aws:iam::777200079629:role/Ab2dInstanceRole", "arn:aws:iam::330810004472:role/Ab2dInstanceRole"]
-    write_roles       = ["arn:aws:iam::755619740999:role/bcda-dev-nfs-instance", "arn:aws:iam::755619740999:role/bcda-test-nfs-instance"]
```
msnook, line 54: Confirmed from BCDA perspective, these are no longer needed.

ops/terraform/env/prod/stateful/main.tf
```diff
@@ -60,7 +60,7 @@ module "stateful" {
   medicare_opt_out_config = {
     # TODO: add read roles for DPC
     read_roles        = []
-    write_roles       = ["arn:aws:iam::755619740999:role/bcda-prod-nfs-instance"]
```
msnook, line 63: Confirmed from BCDA perspective, this is no longer needed.

ops/terraform/modules/resources/s3_pii/templates/bucket-policy.json
```diff
@@ -1,62 +1,48 @@
...
-          %{ for reader in readers ~}"${reader}",%{ endfor ~}
-          %{ for writer in writers ~}"${writer}",%{ endfor ~}
-          %{ for admin in admins ~}"${admin}",%{ endfor ~}
-          "${root}"
+          for arn in concat(readers, writers) : "${arn}"
         ]
       },
       "Action": [
-        "s3:GetObject"
+        "s3:ListBucket*",
```
RickHawesUSDS, line 15: Suggest that you could add `s3:HeadBucket`, and `s3: GetBucketLocation` to the actions list.

njdister: It appears `s3:HeadBucket` is not a valid operation for a specific S3 bucket, only for `*` so not valid within the context of the S3 policy.  The IAM writer policy already has `s3:HeadBucket`.  I have added `s3:GetBucket*` to the S3 policy to be more inclusive of read operations.

njdister:
> This LGTM. Approved from this side. However, do we need eyes from Jason on this?

I'm not really sure, since technically we're not changing the permissions structure of who gets access to what, but rather changing how its done (BCDA has to assume our role rather than we accept theirs).

PR #285, opened 2020-05-21
BFD-210: Fix broken log rotation due to selinux

njdister:
**JIRA Ticket:**

[BFD-210](https://jira.cms.gov/browse/BFD-210)

Fix broken log rotation caused by selinux file context violation.  This occurs due to the fact that cron jobs run in a limited context (issue does not persist when running logrotate manually which is an uninhibited context).

### What Does This PR Do?

This adds and applies the proper file context for logrotate to be able to run on the bfd-server directory during a cron job.

Primarily that this doesn't break anything else.  I have tested successfully on a detached instance but will test with a full deployment.

* If the answer to any of the questions below is **Yes**, then here's a link to the associated Security Impact Assessment: N/A.
    * Does this PR add any new software dependencies? **No**.
    * Does this PR modify or invalidate any of our security controls? **No**.
    * Does this PR store or transmit data that was not stored or transmitted before? **No**.
* If the answer to any of the questions below is **Yes**, then please add @StewGoin as a reviewer, and note that this PR should not be merged unless/until he also approves it.
    * Do you think this PR requires additional review of its security implications for other reasons? **No**.

N/A

### Submitter Checklist

Before requesting final review, I've gone through and verified that this PR complies with the following requirements:

* [x] I've refactored and rebased this PR (for help, see: [this](https://dev.to/maxwell_dev/the-git-rebase-introduction-i-wish-id-had) and [this](https://raphaelfabeni.com/git-editing-commits-part-1/), if needed, so that it is as small as it can reasonably be, in order to:
    1. Ensure that any problems it causes have a small "blast radius".
    2. Ensure that it'll be easier to rollback if that becomes necessary.
    3. Ease the burden on reviewers.
* [x] I've verified that the commits in this PR:
    1. Reasonably explain the "what" and "why" of the changes.
    2. Leverage JIRA's [smart commits](https://confluence.atlassian.com/jirasoftwarecloud/processing-issues-with-smart-commits-788960027.html) to reference the JIRA ticket that they're associated with. For example, aim for commit messages like this (note also [the 50/72 formatting](https://stackoverflow.com/q/2290016) used here):
        
        ```
        Added the whizbang to the doodad.

        The new whizbang should really simplify things for our whoosit users.
        It was a bit tricky to get it into the doodad, but we decided that the
        flipwhee pattern was the best approach for now. Might want to revisit
        that in the future if it ends up being too hard to maintain.

        SOMEPROJECT-42
        ```
        
* [x] I've verified that this PR includes all required documentation changes, including `README` updates and changelog / release notes entries.
* [x] I've verified that all new and modified code is appropriately commented, such that the what and why of its design would be reasonably clear to engineers, preferably ones unfamiliar with the project.
* [x] I've verified that all tech debt and/or shortcomings introduced by this PR is detailed in `TODO` and/or `FIXME` comments, which include a JIRA ticket ID for any items that require urgent attention.
* [x] I've requested review from at least two other engineers on this project, at least one of whom is a senior engineer or owns the relevant component(s) here.
* [x] I've requested review from any relevant engineers on other projects (e.g. BFD, SLS, etc.).
* [x] I've verified that this PR and its changes comply with all other policies in the [DASG Engineering Standards](../policies/engineering_standards.md) and have specifically called out any/all deviations in this PR.

### DIFF
--- a/ops/ansible/roles/bfd-server/tasks/install.yml
+++ b/ops/ansible/roles/bfd-server/tasks/install.yml
@@ -86,3 +86,14 @@
     src: 'bfd-server.j2'
     dest: /etc/logrotate.d/bfd-server
   become: true
+
+- name: Add selinux file context to allow log rotation
+  sefcontext:
+    target: "{{ data_server_dir }}(/.*)?"
+    setype: var_log_t
+    state: present
+  become: true
+
+- name: Apply selinux file context
+  command: "restorecon -rv {{ data_server_dir }}"
+  become: true
--- a/ops/ansible/roles/bfd-server/tasks/install.yml
+++ b/ops/ansible/roles/bfd-server/tasks/install.yml
@@ -89,7 +89,7 @@
 
 - name: Add selinux file context to allow log rotation
   sefcontext:
-    target: "{{ data_server_dir }}(/.*)?"
+    target: "{{ data_server_dir }}(/.*json|/.*log)?"
     setype: var_log_t
     state: present
   become: true


jzulim:
Per your PR description, do we want to deploy this PR to TEST and let is 'bake' for a few days to validate additionally?

njdister:
> Per your PR description, do we want to deploy this PR to TEST and let is 'bake' for a few days to validate additionally?

Forgot to update I have tested this with a full deployment successfully.  I also generated some traffic to test my other PR regarding cloudwatch alarms so I can confirm this doesnt break cloudwatch log shipping.

njdister:
> I'm not very familiar with selinux, but wouldn't this apply whatever labels to everything in /usr/local/bfg-server? This would include log files but also startup scripts TLS certs, etc. Ideally these certificates would not live alongside logs (and logs would be written to the standard posix /var/log path), but until then, maybe set the target to `"{{ data_server_dir }}(/*.log)?"` to narrow the scope of this change.

@keithdadkins Made and tested some changes to narrow the regex scope.  We also have `.json` files that get ingested.

PR #286, opened 2020-05-21
BFD-211: Fix for broken AWS terraform provider >= 2.61.0

njdister:
Hotfix for: https://github.com/terraform-providers/terraform-provider-aws/issues/13236

This should be reverted one we validate that the most recent AWS provider version works again.

### DIFF
--- a/ops/terraform/env/prod-sbx/stateless/main.tf
+++ b/ops/terraform/env/prod-sbx/stateless/main.tf
@@ -3,7 +3,7 @@ terraform {
 }
 
 provider "aws" {
-  # Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
+  # FIXME BFD-211: Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
   # version = "~> 2.25"
   version = "<= 2.60.0"
   region = "us-east-1"
--- a/ops/terraform/env/prod-sbx/stateless/main.tf
+++ b/ops/terraform/env/prod-sbx/stateless/main.tf
@@ -3,7 +3,9 @@ terraform {
 }
 
 provider "aws" {
-  version = "~> 2.25"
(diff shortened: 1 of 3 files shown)


ops/terraform/env/prod-sbx/stateless/main.tf
```diff
@@ -3,7 +3,9 @@ terraform {
 }
 
 provider "aws" {
-  version = "~> 2.25"
+  # Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
```
karlmdavis, line 6: I always try to leave this or a `TODO` so that these things can be found. In this case, I'd probably also throw a Jira ticket for it in the backlog and reference that ticket ID here.
```suggestion
  # FIXME Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
```

ops/terraform/env/prod/stateless/main.tf
```diff
@@ -3,7 +3,9 @@ terraform {
 }
 
 provider "aws" {
-  version = "~> 2.25"
+  # Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
```
karlmdavis, line 6:
```suggestion
  # FIXME Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
```

ops/terraform/env/test/stateless/main.tf
```diff
@@ -3,7 +3,9 @@ terraform {
 }
 
 provider "aws" {
-  version = "~> 2.25"
+  # Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
```
karlmdavis, line 6:
```suggestion
  # FIXME Revert once this is fixed: https://github.com/terraform-providers/terraform-provider-aws/issues/13236
```

PR #304, opened 2020-06-22
BFD-256 - Increase min ASG size for prod and prod-sbx

keithdadkins:
### Change Details

In support of increasing performance for AB2D, we are increasing the minimum number of running instances from 3 to 6 for prod and prod-sbx environments (test should remain at 3 to save on costs).

This is a temporary measure until a better AutoScaling metric can be created.

### Acceptance Validation

- Min and Desired counts should be twice the count of availability zones (we have 3) if prod or prod-sbx
- Test should remain at the default number of 1 per AZ (3 total)

Notes:
A `terraform plan` was run manually to verify that test would remain at 3 instances and prod would be increased to 6. This is a snippet from prod (test remained at 3):

```
      ~ min_elb_capacity.   = 3 -> 6
      ~ min_size.   = 3 -> 6
```

### External References

- [BFD-256](https://jira.cms.gov/browse/BFD-256)

### Security Implications
None


keithdadkins:
Note.. The `is_prod = substr(var.env_config.env, 0, 4) == "prod"` snippet matches both `prod` and `prod-sbx`.

PR #316, opened 2020-07-24
Dshekhar/bfd 285 BFD v2 rfc

dshekhar18:
### Change Details

This RFC proposes that our team migrate from BFD v1 platform currently on FHIR STU3 to BFD v2 platform supporting the first normative FHIR release, R4 and implement CARIN BB profiles.. BFD v2 RFC is located at:

[BFD v2 RFC.md](https://github.com/CMSgov/beneficiary-fhir-data/blob/dshekhar/BFD-285-BFDv2_RFC/rfcs/BFDv2/BFDv2%20RFC.md)

### Acceptance Validation

BFD to provide a new v2 endpoint URL (future state) and will make no architectural change to the v1 endpoint (current state). The goal is for partners to not have to make any changes in their existing integration points with BFD. BFD Stakeholders and Peer Partners will have the option to move to BFD v2 to take advantage of new features, new enhancements and new data fields. v2 will be a work in progress based on business and industry standard needs. BFD v2 (FHIR-R4) will not be backward compatible with BFD v1 (FHIR-STU3).

### Feedback Requested

Make sure all design changes are clearly outlined for  BFD v1 (FHIR STU3) migration to BFDv2 (FHIR R4/CARIN).
### External References

- BFD-285 (https://jira.cms.gov/browse/BFD-285)

### Security Implications

No

 No

Testing scenarios:
NA


dshekhar18:
> I'm unclear here on what you feel needs comment ("RFC" stands for "Request for Comment"). Can you please help me understand what it is you're proposing that you want input on?
> 
> As I read it now, it's basically just saying "we're going to add v2 endpoints with new resource providers to support them" which feels uncontroversial. But perhaps I'm missing something?
> 
> Also: it'd probably help if you removed all the unrelated stuff from the PR, like the stuff that got mixed in from your "v2 Patient" RFC, etc.

@karlmdavis  Thanks for your comments. The v2 RFC is for partners/users to have some background, motivation, proposed path and changes BFD v2 will have. If there are no comments, then thats perfectly valid scenario. Also the reason I put a Patient v1 and v2 there was for sake of completeness so partners can see that BFD v2 WILL NOT be backward compatible with BFD v1 and they will not be able to use their v1 connectors/interfaces/integration points. All concerns/comments are welcome from all partners.

rfcs/BFDv2/Patient RFC.md
```diff
@@ -0,0 +1,250 @@
...
+•	2106-3			White  
+•	UNK			Unknown  
+•	ASKU			Asked but no answer  
+
+ 		
+
+### Proposed Solution: Drawbacks
+[Proposed Solution: Drawbacks]: #proposed-solution-drawbacks
+
+Supporting both v1 and v2 in parallel may cause slight performance issues.
```
adaykin, line 225: Since at least initially there won't be as many users using v2, would it be possible to do something like dedicate 1 node to v2 and 3 to v1? Just so the new experimental doesn't affect the current stable version.

dshekhar18: + @jzulim, @sarah-tully, @karlmdavis 

@adaykin This can be definitely discussed once POC phase is DONE as part of v2 go-live. Thanks.

karlmdavis:
> the reason I put a Patient v1 and v2 there was for sake of completeness so partners can see that BFD v2 WILL NOT be backward compatible with BFD v1 and they will not be able to use their v1 connectors/interfaces/integration points.

@dshekhar18 I would state this explicitly in the RFC text, then. The lack of backwards compatibility wasn't clear to me reading through things, and I doubt it will be to others, either, unless they happen to read the comments here.

dshekhar18:
> > the reason I put a Patient v1 and v2 there was for sake of completeness so partners can see that BFD v2 WILL NOT be backward compatible with BFD v1 and they will not be able to use their v1 connectors/interfaces/integration points.
> 
> @dshekhar18 I would state this explicitly in the RFC text, then. The lack of backwards compatibility wasn't clear to me reading through things, and I doubt it will be to others, either, unless they happen to read the comments here.

@karlmdavis Comment incorporated. Thanks for bringing this up.

rfcs/BFDv2/Patient RFC.md
jzulim: @adaykin With our current deploy / infra and tooling, running V2 on a single node is not in scope here. We are limiting the V2 with a feature flag, but this will apply to all nodes. We'll only be in SBX for the time being, so if we see performance issues we'll catch and work through them.

rfcs/BFDv2/BFDv2 RFC.md
```diff
@@ -0,0 +1,235 @@
...
+FHIR R4 (US-Core, Normative release):
+http://hl7.org/fhir/R4
+
+
+CARIN BB Profiles:
+http://build.fhir.org/ig/HL7/carin-bb
+
+### Detailed Design
+
+[Detailed Design]: #detailed-design
```
jzulim, line 78: Could you use this section to detail the technical 'nuts and bolts' to how we intend to implement two versions?

dshekhar18:
> I'm unclear here on what you feel needs comment ("RFC" stands for "Request for Comment"). Can you please help me understand what it is you're proposing that you want input on?
> 
> As I read it now, it's basically just saying "we're going to add v2 endpoints with new resource providers to support them" which feels uncontroversial. But perhaps I'm missing something?
> 
> Also: it'd probably help if you removed all the unrelated stuff from the PR, like the stuff that got mixed in from your "v2 Patient" RFC, etc.

@karlmdavis Thanks Karl for all your feedback and inputs. They have been incorporated in the PR.

PR #345, opened 2020-09-04
Feature/ab2d 1863 EOB Field Selection

adaykin:
### Change Details

This is just an RFC, so a proposal at this time of how to complete server side filtering of fields for objects we are interested in receiving.

### Acceptance Validation

Whether this addresses the required questions of the RFC

### Feedback Requested

Comments on the RFC about the proposed solution of whether or not is is viable.

### External References

- [BLUEBUTTON-xxx](https://jira.cms.gov/browse/BLUEBUTTON-xxx)

### Security Implications


dshekhar18 (commented):
All trade-offs and different approaches will need to be evaluated. 

Also list of all fields which will be filtered and consumed should be part of this RFC to do performance evaluation and measure any gains.  Going through every single EOB element against the filter list on the BFD server should be avoided and other approaches should also be evaluated. 

A proof of concept should follow showing how much performance gain it led on FHIR client side (AB2D) and how much of performance burden it became on FHIR Server side (BFD).

jzulim:
Should this be implemented server side (BFD), I'm also curious to flush out how we would determine whether to filter or not? Would this be another custom header since we won't be utilizing query parameters? Something like the 'IncludeIdentifiers' header?

adaykin:
> Should this be implemented server side (BFD), I'm also curious to flush out how we would determine whether to filter or not? Would this be another custom header since we won't be utilizing query parameters? Something like the 'IncludeIdentifiers' header?

Yes it would be determined whether or not to filter based on a custom header.

adaykin:
> All trade-offs and different approaches will need to be evaluated.
> 
> Also list of all fields which will be filtered and consumed should be part of this RFC to do performance evaluation and measure any gains. Going through every single EOB element against the filter list on the BFD server should be avoided and other approaches should also be evaluated.
> 
> A proof of concept should follow showing how much performance gain it led on FHIR client side (AB2D) and how much of performance burden it became on FHIR Server side (BFD).

The alternative to looping through each field would be to put in if statements to different parts of the code where the fields we want would only be added. @karlmdavis and I discussed that approach and the downside would be maintainability of that, but yes you are correct it would be a faster approach.

Fields we keep would be:
patient
provider
organization
facility
careTeam
diagnosis
procedure
billablePeriod
item

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,127 @@
...
+    * [Proposed Solution: Unresolved Questions](#proposed-solution-unresolved-questions)
+    * [Proposed Solution: Drawbacks](#proposed-solution-drawbacks)
+    * [Proposed Solution: Notable Alternatives](#proposed-solution-notable-alternatives)
+* [Future Possibilities](#future-possibilities)
+
+## Motivation
+[Motivation]: #motivation
+
+AB2D receives lots of data back from BFD when we make API calls to gather EOBs (Explanation of Benefits), 
+where we filter out much of the data we receive. This might not be terrible for a few API calls, but for
```
karlmdavis, line 30: > where we filter out much of the data we receive

Can you be explicit about what % of the fields would be filtered out? I think that might help reduce confusion.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+private static final Set<String> allowAB2DList = new HashSet<>();
+
+static {
+    allowedAB2DList.add("patient");
+    allowedAB2DList.add("provider");
+    allowedAB2DList.add("organization");
+    ...
+}
```
karlmdavis, line 77: How will this account for nested fields? For example, `item` is a very large bucket of other fields, and I imagine you need to filter out many of those, also.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+}
+``` 
+
+### Proposed Solution: Unresolved Questions
+[Proposed Solution: Unresolved Questions]: #proposed-solution-unresolved-questions
+
+How much will this save us? We will need to benchmark this solution compared to the old one to ensure it will
+be worth putting into production.
```
karlmdavis, line 96: I think you're talking about benchmarking AB2D here, right? Either way, please clarify in the wording.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,127 @@
...
+### Proposed Solution: Notable Alternatives
+[Proposed Solution: Notable Alternatives]: #proposed-solution-notable-alternatives
+
+One alternative that would give clients more flexibility would be to send the parameters themselves, but with that
+added flexibility comes a danger of being able to more easily make a mistake. By setting the parameters on the server
+side it's more difficult to change, but also less error prone. 
+
+An alternative for using a library like PropertyUtils would be JXPath, but when I benchmarked the two, PropertyUtils
+was consistently faster. Having said that, I would like to run some more realistic tests, which would not be very hard 
+to setup, since it would just be a very simple matter of changing one line of code really to compare the two.
```
karlmdavis, line 118: Given the above question about nested/multi-level filtering, you may need to revisit this.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Future Possibilities
+[Future Possibilities]: #future-possibilities
+
+There could be filtering on the database side as well. Right now there are 194 columns
+in the beneficiary table, and we certainly do not need all of them, so filtering on the 
+database side would reduce the load on the database.
```
karlmdavis, line 125: Probably not significantly.

Or, rather, it would be **very nice** if we got to the point where _that_ change moved the needle at all, performance-wise; we're a long long way from that point right now.

karlmdavis (changes requested):
Andy, love this RFC draft and the proposed approach.

I did note some questions & concerns below that should be addressed, I think.

rfcs/0007-filtering.md
karlmdavis: To be clear: doesn't have to be exact -- a rough estimate would be great.

nflearl: Rough estimate:  I counted getters in org.hl7.fhir.dstu3.model.ExplanationOfBenefit

69 top level getters
10 used by AB2D

ExplanationOfBenefit encapsulates ItemComponent which has

42 top level getters
 6 used by AB2D

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Proposed Solution: Unresolved Questions
+[Proposed Solution: Unresolved Questions]: #proposed-solution-unresolved-questions
+
+How much will this save us?
+ 
+Proposed profiling is invoking BFD with at least 10,000 EOB Requests by Patient ID (ExplanationOfBenefit?patient=) and to measure
+the aggregate time of completion of all calls.  In other words, sum up the time taken by each call over all calls.
+
+This test will be baselined on current implementation and then ran again after the solution is deployed.
```
jzulim, line 179: Perhaps we should add the current baseline performance metrics within the RFC as well as the testing method so that we can reproduce results when a proposed code change is tested in lower environments. We'll need to use synthetic data for these tests / metrics.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+
+
+### Proposed Solution: Drawbacks
+[Proposed Solution: Drawbacks]: #proposed-solution-drawbacks
+
+This solution incurs nominal amount of extra processing to traverse the EOB object and null out fields that do not
```
jzulim, line 185: Do we have data to back this up?

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+need to be returned.
+
+This solution adds AB2D specific code within the BFD source base.  Any new required fields results in changes to BFD
```
jzulim, line 188: We are adding new data fields in the coming weeks that will land in Patient and EOB resources. Please see here for details: https://jira.cms.gov/browse/BFD-368

Does AB2D need any of these new fields?

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+[Proposed Solution: Notable Alternatives]: #proposed-solution-notable-alternatives
+
+GraphQL is a notable alternative where the fields are explicitly requested in the invocation.  Since GraphQL
+supports introspection, the client can actively determine the available fields.  A hybrid approach using
+*registered queries* requires queries to be specified in advance so that they can be locked down by a production server.
+
+The downside is that this would be a significant amount of work for BFD to adopt a different API style while continuing
+to support the REST calls.
+
+An alternative for using a library like PropertyUtils would be JXPath, but initial benchmarks showed that PropertyUtils
```
jzulim, line 201: 👍

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+was consistently faster.
+
+## Future Possibilities
+[Future Possibilities]: #future-possibilities
+
+There could be dynamic projections on the database side as well. Right now there are 194 columns
+in the beneficiary table, and we certainly do not need all of them, so retrieving only the needed columns
+would reduce the load on the database.
+
+Other groups would likely want to utilize filtering as well, but that is out of scope for this RFC.
```
jzulim, line 211: I'd assume the same. The BFD would more than likely want to explore alternative methods of filtering / exclusion / allowed, etc. that could be applied more broadly as needed. But this approach / iteration seems reasonable for the time being.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+The solution adds code in the BFD server to configure which fields AB2D requires.  Non-required fields are null'd out
+before being transformed in the result.  The specific set of required fields should remain stable
+and not require modifications.
+
+The filtering will need to happen towards the end of processing. This is to avoid
+spaghetti code since transformations from the entity object to the FHIR object happen in several places throughout
+the codebase. 
+
+
+### Proposed Solution: Detailed Design
```
jzulim, line 50: I'd suggest we put this code behind a feature flag (we have prior art on this) so we can do A/B testing in PROD.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+* [Future Possibilities](#future-possibilities)
+
+## Motivation
+[Motivation]: #motivation
+
+AB2D receives more data than needed from BFD when we make API calls to gather EOBs (Explanation of Benefits).  AB2D
+currently filters out unrelevant fields.  This might not be terrible for dozens of API calls, but for 
+10 to 100 thousand requests it can impact our performance significantly.  Since we only need a projection of the data
+(10 out of 69 top level getters; 6 out of 42 fields in the ItemComponent), it makes
+sense for the server to send us only the data consumed.  This reduces bandwidth requirements for receiving the data.
```
karlmdavis, line 33: I totally buy this argument but it'd still be good to back it up with evidence. For example: "Of the total time AB2D spends on EOB requests, XX% of this time is currently spent on data filtering, which could instead be handled much faster by BFD." (And fill in the blank.)

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+    <groupId>commons-beanutils</groupId>
+    <artifactId>commons-beanutils</artifactId>
+    <version>1.9.4</version>
+</dependency>
+```
+
+A `private static final Map<String, SubfieldSelection>` of fields will need to be maintained in `TransformerUtils`. This will
+be the AB2D required fields in the FHIR object that gets sent back to the client. Just before the
+end of the method `findByPatient` in `ExplanationOfBenefitsResourceProvider` the `eobs` can be passed to a new
+method in `TransformerUtils`.
```
karlmdavis, line 66: For what it's worth, this really doesn't relate to transformation and I'd be more inclined to include it in a new class, e.g. `gov.cms.bfd.server.war.commons.ResourceFieldFilter`.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    private static void trimFieldsThrowException(Object objToTrim, Field field, SubfieldSelection selection) {
+        try {
+            trimFields(field.get(objToTrim), selection.getExplictFields());
+        } catch (IllegalAccessException iae) {
+            throw new RuntimeException(iae);
+        }
+    }
+```
```
karlmdavis, line 169: A quick read-through of the code leads me to suspect a few things here ought to be fixed during implementation, including but not limited to:
1. Using `getDeclaredFields()` is a bug as it won't catch inherited fields (and the HAPI/FHIR models use a lot of inheritance last I looked).
2. The approach only considers two levels of nesting "top" fields and "sub" fields. This is likely insufficient and so a more generalized approach should be developed that can handle arbitrary levels of nesting.

None of these problems need to block the RFC itself, though; they're implementation issues and should be considered then. This code here is fine for proof of concept purposes.

rfcs/0007-filtering.md
jzulim, line 185: We can use our A/B testing and New Relic to confirm in PROD.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+    }
+``` 
+
+### Proposed Solution: Unresolved Questions
+[Proposed Solution: Unresolved Questions]: #proposed-solution-unresolved-questions
+
+How much will this save us?
+ 
+Proposed profiling is invoking BFD with at least 10,000 EOB Requests by Patient ID (ExplanationOfBenefit?patient=) and to measure
+the aggregate time of completion of all calls.  In other words, sum up the time taken by each call over all calls.
```
karlmdavis, line 177: I'd strongly suggest a more formal approach towards testing and verification be proposed here, for several reasons:
1. Your early testing seemed to indicate that one library was significantly more performant than another, which leads me to worry in general that this will end up being performance-critical code.
2. Us engineers are notoriously terrible at benchmarking. One way to do it right is to treat it more like a science lab, with methods written up in advance.
3. This is definitely security-critical code and so needs to be approached with a certain amount of rigor.

Can you spend some thought and time expanding this section a bit? Specifically, I'd like to see you detail how total performance will be benchmarked and then compared by AB2D. I'd strongly suggest that the approach there ought to include hiding the initial implementation behind a feature flag, to allow for that testing. I'd also spend a little bit of time considering how best to create unit and integration tests that appropriately address the security concerns around this code.

cc @StewGoin

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,211 @@
...
+
+An alternative for using a library like PropertyUtils would be JXPath, but initial benchmarks showed that PropertyUtils
+was consistently faster.
+
+## Future Possibilities
+[Future Possibilities]: #future-possibilities
+
+There could be dynamic projections on the database side as well. Right now there are 194 columns
+in the beneficiary table, and we certainly do not need all of them, so retrieving only the needed columns
+would reduce the load on the database.
```
karlmdavis, line 209: I agree in general, but just as an FYI: right now, BFD's performance is **heavily** dominated by JSON serialization -- over 80% of response time is spent on that. So DB performance is a long ways down the list, relatively speaking. (I'll also be curious to see if these filtering changes improve serialization performance!)

karlmdavis (changes requested):
I love the general approach here! I do think the test plan needs some work before approval, though. Please see my specific comments for details.

jzulim (changes requested):
RFC details a reasonable approach that limits complexity (for the time being) - The BFD system should consider a longer term solution as this type of control and granularity of data will certainly become more popular over time. Thank you for picking this up and refreshing / expanding on the RFC and proposed example implementation @nflearl - Great Job!

The requesting change I have is to add baseline (current) performance metrics AB2D captures along with your testing method to this RFC so we can duplicate when confirming the code / implementation.

rfcs/0007-filtering.md
jzulim: +1

rfcs/0007-filtering.md
nflearl: Thanks for #1 - I inherited that suggestion with the PR and did not dive deeper into that aspect.

For #2, the example code is recursive and can handle arbitrary levels of nesting.  Since AB2D only needs 2 levels, the provided example initialization demonstrates just to 2 levels.  It would be straightforward to demo deeper nesting.

jzulim:
This RFC has gone a bit stale (as they tend to do) - I've reached out to the AB2D engineering team to see if they intend to proceed with this. @nflearl @lailasharshar @wnyffenegger

nflearl:
It's been on my plate every sprint to respond with "how to test".  I think that is the only thing left to settle before the PR/RFC gets  approved.  I am actually working on defining a POC that can also be specifically purposed for this question.  My apologies for not circling back earlier.

rfcs/0007-filtering.md
```diff
@@ -172,26 +176,36 @@ Fields that represent objects are recursed into, applying the same field selecti
 [Proposed Solution: Unresolved Questions]: #proposed-solution-unresolved-questions
 
 How much will this save us?
- 
-Proposed profiling is invoking BFD with at least 10,000 EOB Requests by Patient ID (ExplanationOfBenefit?patient=) and to measure
-the aggregate time of completion of all calls.  In other words, sum up the time taken by each call over all calls.
 
-This test will be baselined on current implementation and then ran again after the solution is deployed.
+As part of a larger effort to reduce job times from days to minutes, AB2D is constructing a performance test harness
```
jzulim, line 180: Fantastic!

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,225 @@
...
+    <groupId>commons-beanutils</groupId>
+    <artifactId>commons-beanutils</artifactId>
+    <version>1.9.4</version>
+</dependency>
+```
+
+A `private static final Map<String, SubfieldSelection>` of fields will need to be maintained in `TransformerUtils`. This will
+be the AB2D required fields in the FHIR object that gets sent back to the client. Just before the
+end of the method `findByPatient` in `ExplanationOfBenefitsResourceProvider` the `eobs` can be passed to a new
+method in `TransformerUtils`.
```
jzulim, line 70: I'd also think this would be best suited as a common class and @karlmdavis naming suggestion matches our previous patterns. 

`gov.cms.bfd.server.war.commons.ResourceFieldFilter`

As well, since this RFC was initially published, we now have R4ExplanationOfBenefitsResourceProvider that supports PDE currently and should also include this new class/method.

rfcs/0007-filtering.md
jzulim: @nflearl - Do you have a proposal for how we'll get this on PROD-SBX or PROD (which only have the main branch deployed to them) - For V2, we've used a feature flag (i.e. system property flag that reads env variable) to enable per environment.

jzulim (changes requested):
Thank you again for picking this up, improving the suggested approach and for detailking the testing framework. Great job!

I have just 1 suggestion (besides mirroring Karls about moving this into a new class in the commons) and to also notate that we'll have to apply the same code changes to the R4 EOB class. 

I also wanted to raise how we might intend to deploy this out to SBX but NOT production. For V2 of the API, we put it behind a 'feature flag' via env var that's passed as a system property, something to consider during implementation but might be worth noting here in the RFC. 

Thanks again, excited to get your input on the above and give this another look ASAP.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,241 @@
...
+[RFC Proposal]: #rfc-proposal
+
+* RFC Proposal ID: `0007-filtering-fields`
+* Start Date: 2020-08-21
+* RFC PR: [rfcs#0007](https://github.com/CMSgov/beneficiary-fhir-data/pull/345)
+* JIRA Ticket(s):
+    * [https://jira.cms.gov/browse/AB2D-1863](https://jira.cms.gov/browse/AB2D-1863)
+
+This proposal will suggest adding server side filtering to API calls that return EOBs to the client in order to 
+reduce on the amount of data that needs to be transferred.
```
karlmdavis, line 11: We already do a lot of filtering, so let's reframe this summary to reflect that.
```suggestion
This proposal a new approach to server side filtering of API call results. The new approach is designed to be more comprehensive, and avoid all of the branching currently used for filtering in the transformers.
```

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,241 @@
...
+sense for the server to send us only the data consumed.  This reduces bandwidth requirements for receiving the data.
+
+For example, GraphQL solves this problem by allowing clients to specify field selection (e.g. a database projection) at
+invocation time.
+
+## Proposed Solution
+[Proposed Solution]: #proposed-solution
+
+The solution adds code in the BFD server to configure which fields AB2D requires.  Non-required fields are nulled out
+before being transformed in the result.  The specific set of required fields should remain stable
```
karlmdavis, line 46:
```suggestion
before being serialized in the result.  The specific set of required fields should remain stable
```

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,241 @@
...
+
+
+### Proposed Solution: Detailed Design
+[Proposed Solution: Detailed Design]: #proposed-solution-detailed-design
+
+The BFD_FIELD_SELECTION header enables this feature.  A value of **ALL** is current behavior, will be the default
+(if header is missing) and indicates to return all EOB fields.  A value of **AB2D_VERSION1** indicates return only
+the fields specified in this proposal.  By using an enumerated type instead of a boolean, this allows for different
+results to be specified in the future should the need arise and for different consumers to have specific field selections
+encoded.
```
karlmdavis, line 61: Please stick with idiomatic HTTP header naming conventions. See <https://stackoverflow.com/a/3561399> and elsewhere.

rfcs/0007-filtering.md
karlmdavis: Ahh, I see the recursion now.

Not a giant fan of that approach here, as it makes this code harder to read. Looking at beanutils, I see that it supports a nested property selection syntax [1]. I think this code could be reworked to make use of that, or at least reworked to use a different or perhaps custom implementation for a similar declarative nested property path style. (I'd have to dig around for a while in beanutils' source to see if it can do it directly, itself. It looks like it could certainly do whitelisting, but I'm not so sure about blacklisting.)

[1] See the "2.3 Nested Property Access" section here: <http://commons.apache.org/proper/commons-beanutils/javadocs/v1.9.4/apidocs/org/apache/commons/beanutils/package-summary.html#package.description>

rfcs/0007-filtering.md
karlmdavis: Was it faster in a way that mattered? As things stand, we spend somewhere around 5 milliseconds per request on transformation, I believe, which is a hilariously small percentage of our total response time.

If the difference wasn't meaningful enough, please instead compare the two approaches based on their relative maintainability.

karlmdavis (dismissed):
An RFC is not an implementation.

As an RFC, I love this! 

I've got a number of concerns about the proposed/prototype implementation (mentioned inline), but if the RFC is reframed a bit to frame all that as more of a prototype/example and less "this is the way It SHALL be done", I'd be happy to approve it.

rfcs/0007-filtering.md
nflearl: X- prefix has been deprecated and its use is discouraged.

the official recommendation is to just name them sensibly without the "X-" prefix

There seems to be some variance in the community in that some folks still advocate using -X for application specific parameters (which this is).

Please suggest a header name that you like better.

karlmdavis: Sorry, I should have been more clear: I meant "please don't use underscores and follow the case conventions."

rfcs/0007-filtering.md
mackec1: Not sure I really understand the issue(s) here other than to glean from the RFC and various comments.As far as I can tell it's a case of who bears the brunt of data manipulation / filtering when stock content is too verbose for some consumers; as noted, it's why GraphQL has become so popular...it solves this issue nicely!

`AB2D receives more data than needed from BFD when we make API calls to gather EOBs (Explanation of Benefits).  AB2D currently filters out irrelevant fields`

What exactly does this mean?

- Is it the structure of the EOB itself? And AB2D wants an altered EOB structure?
- The EOB structure is fine, AB2D just wants lots (much) of the data nulled out?

_**Question - Who does work reducing fluff?**_

If BFD:
  - hide under feature flag
  - still amounts to BFD compute cycles to reduce payload to minimize fluff; possible solutions:
    1) add filtering mechanism to BFD code; possibles are:
       -  "if" statements in code
       - whitelist (acceptable) map entries
       - affects (gums up?) BFD codebase
    2) use post-processor - minimal (if any) impact on current BFD codebase
       - adds an 'extra' step on way out door
       - extensible
       - no current model / mechanism in place
    3) does this align with long-term goals of BFD?
       - expect other requests like this?

If AB2D:
  - do they have impact assessment numbers on cost of current payload
    - are the requested changes enough to move the needle?
    - if BFD considers a post-processor, did AB2D consider a pre-processor?

rfcs/0007-filtering.md
mackec1: `Additional goals for the harness are to assess how much can concurrency increase.  Currently, AB2D limits EOB
retrievals to 32 concurrent requests.  The plans are to run cooperative experiments to establish a maximum concurrent limit.`

Have you considered frameworks like netty? I've seen netty handle 20k concurrent med device emulator connections running SSL WebSockets. I'm sure you could blow by the 32 concurrent without even coming close to taxing the system.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,93 @@
...
+expected benefits of changes.
+
+Proposed profiling is invoking BFD with at least 10,000 EOB Requests by Patient ID (ExplanationOfBenefit?patient=) and to measure
+the aggregate time of completion of all calls.  In other words, sum up the time taken by each call over all calls.
+
+### Proposed Solution: Drawbacks
+[Proposed Solution: Drawbacks]: #proposed-solution-drawbacks
+
+This solution incurs a nominal amount of extra processing to traverse the EOB object and null out fields that do not
+need to be returned.
```
Dranith, line 74: I think this could be done without having to take this approach, but may take some re-engineering of the current EOB processing.

rfcs/0007-filtering.md
```diff
@@ -0,0 +1,93 @@
...
+[Motivation]: #motivation
+
+A single BFD EOB retrieval call (UHC) with 2 million beneficiaries took 24 hours to retrieve three months of data.
+Larger PDPs can have in excess of 6 million beneficiaries.  To have a request take over 3 days is not acceptable.
+
+AB2D receives more data than needed from BFD when we make API calls to gather EOBs (Explanation of Benefits).  AB2D
+currently filters out irrelevant fields.  This might not be terrible for dozens of API calls, but for 
+millions of requests it can impact our performance significantly.  Since we only need a projection of the data
+(10 out of 69 top-level getters; 6 out of 42 fields in the ItemComponent), it makes
+sense for the server to send us only the data consumed.  This reduces bandwidth requirements for receiving the data.
```
Dranith, line 34: I would be interest in knowing what the actual bottleneck is.  When 2 million records are retrieved, it took 24 hours .. but how did that time spent break down?  If it was mostly due to transfer times then shrinking the output object would obviously help.  It could also help if serialization is the bottleneck.  If the bottleneck is the database retrieval or actual conversion from the native BFD data format to a FHIR EOB then other improvements would need to be applied.

From looking at the code a bit, it appears that the data retrieval and processing is all currently done serially.  If we could identify that as the main driver in this slower performance it may be easy to address without any API changes.

nflearl: We have built a testing framework specifically for this purpose and finally cleared the last configuration hurdle yesterday.  We will be experimenting in order to baseline the current peformance.  Once we have a baseline, AB2D is available for assistance in identifying additional bottlenecks.

In addition to shrinking the output object (saving on serialization and transmission times), any other performance improvements are welcome.  Work on improving database performance and conversion to FHIR EOB should be done independently of the effort to reduce payload size.

The proposal does not make any API changes, all functionality is covered by the current FHIR specification.  The proposal merely provides a higher service level guarantee than the FHIR specification.

Dranith: I agree that supporting `_elements` is a good idea in general, but it is being suggested as a fix for a performance issue.  We can't be sure that this would fix a performance issue without profiling BFD and identifying the root of the problem.  That was the only point I was making.

jzulim (approved):
Thank you for putting this together AB2D (and Earl) - I'm in favor of the BFD API EOB endpoint supporting _elements and am also curious how this impacts performance. BFD and AB2D are now in a better spot to measure this. If you want to put up a POC branch that BFD and AB2D could use for testing purposes, I think that's the next logical step here.

lsmitchell (approved):

mjburling (approved):
This is an interesting RFC. Credit obviously goes to the author, but thanks for tidying this up so we can accept it as deferred for the time being @JeremyIR!

PR #361, opened 2020-09-26
Remove ssh access from inactive bfd users

keithdadkins:
### Change Details

Removes ssh access from inactive BFD accounts. Users can be re-added as needed if and when they roll back onto the project.

### Acceptance Validation
- Inactive users removed.
- ssh_users appears correct

### Security Implications
Improves security by removing inactive/unused access.


njdister (approved):

jzulim (approved):
LGTM.

PR #385, opened 2020-11-02
Remove docker from our build servers.

keithdadkins:
### Change Details

This change removes (or does not install) the unused docker installation from our build servers. Since we do not use docker in any shape or form on our build servers, this service simply eats up system resources for no reason.

Note: I deployed this change to mgmt-test and verified that docker was not present and we were able to build.

### Acceptance Validation
The 'install_docker' ansible role is not included when building or updating jenkins.

### Security Implications
None. It actually improves security by decreasing our build server's attack surface.


jzulim (approved):
Approved

cbrunefearless (approved):
lgtm

PR #399, opened 2020-11-16
Cbrune/bfd 520 null etl fields

cbrunefearless:
### Change Details
BFD-520 - Added a function to remove /0 -> nulls in the etl process when it parses the csv.

### Acceptance Validation

Added test and passes for a trailing null for county zip code as well as other string fields.

### Feedback Requested

Make sure the change is in the right place when we get the data from the etl and before it is put in the database.

### External References

- [BFD-520](https://jira.cms.gov/browse/BFD-520)

BFD-520 | State County Zip Code Null Byte Acception

### Security Implications


apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -269,4 +270,21 @@ public static Character parseCharacter(String charText) {
...
+    String newString = string;
+
+    // Null Field
+    String nullFields = "\\0";
+    if (!Strings.isNullOrEmpty(newString)) {
+      if (newString.contains(nullFields)) newString = newString.replace(nullFields, "");
+    }
+
+    return newString;
+  }
```
karlmdavis, line 289: I'd suggest a slight wording tweak to clarify your intent here:
```suggestion
  /**
   * @param string the {@link String} to parse
   * @return a copy of the specified {@link String} but with any/all null bytes it contained removed
   */
  public static String removeInvalidCharacters(String string) {
    String newString = string;

    String nullByte = "\\0";
    if (!Strings.isNullOrEmpty(newString)) {
      if (newString.contains(nullByte)) newString = newString.replace(nullByte, "");
    }

    return newString;
  }
```

apps/bfd-pipeline/bfd-pipeline-rif-extract/pom.xml
```diff
@@ -22,6 +22,13 @@
 			<artifactId>bfd-model-rif</artifactId>
 		</dependency>
 
+		<dependency>
+			<!-- Just useful. -->
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+			<version>24.1.1-jre</version>
+		</dependency>
```
karlmdavis, line 30: If we're not already using this somewhere, you'll need to file a SIA for it. Shouldn't be a big deal, though.

You also might consider adding it to the new bfd-shared-utils module, instead, as it definitely is useful, in general.

apps/bfd-pipeline/bfd-pipeline-rif-extract/src/test/java/gov/cms/bfd/pipeline/rif/extract/RifFilesProcessorTest.java
```diff
@@ -105,6 +105,70 @@ public void process1BeneRecord() {
     Assert.assertEquals(new BigDecimal("1"), beneRow.getBeneLinkKey().get());
   }
 
+  @Test
+  public void process1BeneRecordWithTrailingNulls() {
```
karlmdavis, line 109: I'd probably simplify this way down by:
* Checking any/all fields with null bytes, to ensure they were cleaned up correctly.
* Checking one field _after_ the null byte(s), to ensure they're correct.
* Getting rid of most other assertions.

As-is, it's a bit hard to figure out which field(s) were supposed to be corrected. Maybe also add a comment mentioning that?

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -78,7 +79,7 @@ public static CSVParser createCsvParser(
    *     consistency)
    */
   public static String parseString(String string) {
-    return string;
+    return removeInvalidCharacters(string);
```
karlmdavis, line 82: Love this approach!

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -88,7 +89,7 @@ public static String parseString(String string) {
    *     value
    */
   public static Optional<String> parseOptionalString(String string) {
-    return string.isEmpty() ? Optional.empty() : Optional.of(string);
+    return string.isEmpty() ? Optional.empty() : Optional.of(removeInvalidCharacters(string));
```
karlmdavis, line 92: Love this approach!

karlmdavis (commented):
Great work!

I did add a couple of minor suggestions, just to help clarify the intent of the code.

apps/bfd-pipeline/bfd-pipeline-rif-extract/src/test/java/gov/cms/bfd/pipeline/rif/extract/RifFilesProcessorTest.java
cbrunefearless: Done

apps/bfd-pipeline/bfd-pipeline-rif-extract/pom.xml
jzulim: We are already using this for BloomFilters. see: LoadedFilterTest.java - Would this still require an SIA with this being the case?

karlmdavis: > Would this still require an SIA with this being the case?
Nope!

cbrunefearless: awesome, added it to the bfd-shared-utils

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -269,4 +270,22 @@ public static Character parseCharacter(String charText) {
...
+  /**
+   * @param inputString {@link String} that possible contains invalid characters
+   * @return an {@link String} with a null representation stripped from it if the input has a null
+   *     field representation.
+   */
+  public static String removeInvalidCharacters(String inputString) {
+    String verifiedInputString = inputString;
+
+    // Null Field
+    String nullFields = "\\0";
```
karlmdavis, line 283: I've been staring at this for a while now and I _think_ it's incorrect? What you're searching for and replacing here is a two-char sequence: a ``\'` followed by a `'0'`. Is that what is needed? Are we getting backslash-zero character sequences from the CCW that we need to strip out?

Or are we really getting null bytes hiding in strings? If we _are_, then you'd want to change this to `"\0"` or perhaps `"\0x00`", instead. You'd also likely need to update the sample RIF. I'd _guess_ that you'd need to use a hex editor to add the null bytes, as I don't think most text editors would let you type them.

jzulim (approved):
Looks like a sound approach. Test sample data with null. bytes added with tests. Approved.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -269,4 +270,20 @@ public static Character parseCharacter(String charText) {
...
+   * @param inputString {@link String} that possible contains invalid characters
+   * @return an {@link String} with a null representation stripped from it if the input has a null
+   *     field representation.
+   */
+  public static String removeInvalidCharacters(String inputString) {
+    String verifiedInputString = inputString;
+
+    if (!Strings.isNullOrEmpty(verifiedInputString)) {
+      // remove all NULLs
+      verifiedInputString = verifiedInputString.replaceAll("\\x00", "");
```
karlmdavis, line 284: This really didn't look right to me, based on my reading of [Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) (the "Backslashes, escapes, and quoting" section, in particular). However, I was wrong!

Based on some local testing, both of these code samples work (i.e. print out "fixed"):

This:
```java
public static void main(String[] args) {
  String valueToFix = "Howdy!\u0000";
  String valueFixed = valueToFix.replaceAll("\u0000", "");
  if (valueToFix.contentEquals(valueFixed))
    System.out.println("not fixed");
  else
    System.out.println("fixed");
}
```

And this:
```java
public static void main(String[] args) {
  String valueToFix = "Howdy!\u0000";
  String valueFixed = valueToFix.replaceAll("\\x00", "");
  if (valueToFix.contentEquals(valueFixed))
    System.out.println("not fixed");
  else
    System.out.println("fixed");
}
```

I still don't understand _why_ the second one works, but it does. So: neat!

karlmdavis (approved):
🚀  Great work on a tricky problem!

PR #422, opened 2020-12-22
DO NOT MERGE - Cbrune/bfd 550 update enrollment api - DO NOT MERGE

cbrunefearless:
### External References

- [BLUEBUTTON-xxx](https://jira.cms.gov/browse/BLUEBUTTON-xxx)

### Security Implications


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -29,6 +29,7 @@
 import gov.cms.bfd.server.war.commons.QueryUtils;
 import gov.cms.bfd.server.war.commons.TransformerConstants;
 import java.math.BigDecimal;
+import java.time.LocalDateTime;
```
karlmdavis, line 32: (Just hanging this comment here for lack of anywhere better.)

I think we want to end up with a FHIR ["composite" search](https://www.hl7.org/fhir/search.html#composite) that looks something like this?

```
GET [base]/Coverage?ptdcntrct=eq2020-08$http://something.cms.gov/ptdcntrct|<whatever_the_contract_number_is>
```

karlmdavis (commented):
That is an astonishing amount of code -- astonishingly good job cranking it all out! 

That said, I think we can pivot this hard towards being simpler, so that it has a better chance of being done in time.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -380,34 +386,41 @@ private String partDFieldFor(CcwCodebookVariable month) {
...
   private TypedQuery<Beneficiary> queryBeneficiariesBy(
-      String field, String value, PatientLinkBuilder paging, List<String> identifiers) {
-    String joinsClause = "";
+      String contractMonthField,
+      String contractYearField,
+      String partDContractId,
+      PatientLinkBuilder paging,
+      List<String> identifiers) {
+    String yearMonth = getYearMonth(contractYearField, contractMonthField);
+    String joinsClause = "inner join fetch b.enrollments ";
```
karlmdavis, line 395: You're not assigning this joined table an alias here, but down below, you're _assuming_ that you aliased it to "`e`". Not sure what the right syntax is for it, but you'll need to fix that.

karlmdavis (commented):
Found an additional thing you'll need to fix.

PR #455, opened 2021-02-17
remove local port config patch

ericbuckley:
### Change Details

To ease the process of developing locally, I think we can remove the local port config patch and update the `server-start.sh` script to reference an environment variable.  Using the `:-` syntax for setting the bash variable `serverPortHttps`, allows us to use the `BFD_PORT` variable locally and ignore its existence in AWS when it won't be set.

### Acceptance Validation

- locally verify that the server starts up on the port configured in their `.env` file
- visually confirm that the change to `server-start.sh` won't change the desired port in any of the AWS environments

### Feedback Requested

- I added an `env.sample` file to the contributing directory to act as a good starting point for devs, thoughts?


mackec1 (approved):
Much cleaner way to configure/run the bfd server. Also, glad that the README is no longer misleading.

keithdadkins (approved):
Thanks for pitching in Eric!

Gabasol:
Thanks Eric!

PR #462, opened 2021-02-23
BFD-652: Initial RFC draft for Pipeline orchestration

karlmdavis:
### Change Details

The BFD Pipeline already has several ETL-ish tasks/jobs that it runs. Over the next year or two, it is likely to acquire several more, including, for example the DC Geo data load.

This RFC/proposal details architectural changes that can and should be made in order to accomodate these requirements (since the BFD Pipeline architecture was originally designed with only one task/job in mind).

In addition, this PR also starts on the refactorings suggested in the RFC. It doesn't get all the way there, but it's heading in the direction. The end goal is to have a block of code that's something like this as the application's driver, in the entry point (formerly `S3ToDatabaseLoadApp.main(...)`, now `PipelineApplication.main(...)`):

```
PipelineManager pipelineManager = new PipelineManager(...);
pipelineManager.addJob(new DatabaseSchemaManagerJob(), Schedule.runOnce());
pipelineManager.addJob(new CcwRifLoadJob(), Schedule.runPeriodically(...));
pipelineManager.addJob(new CcwRifBeneficiaryCleanupJob(), Schedule.runPeriodically(...));
pipelineManager.addJob(new CcwRifBeneficiaryHistoryCleanupJob(), Schedule.runPeriodically(...));

pipelineManager.start();
```

That's all very handwavy pseudo-code, but something like that, anyways. So far, all this PR does is rename things so that kind of setup is less of a stretch. (And so that the PR for those changes won't be all muddied up with a bunch of renames.)

The HTML-rendered version of the RFC itself can be found here: [rfcs/0009-pipeline-orchestration-baby-steps.md](https://github.com/CMSgov/beneficiary-fhir-data/blob/5b6e37c9156b1a4d44955ce13eb9a71b3a1fefde/rfcs/0009-pipeline-orchestration-baby-steps.md).

### Acceptance Validation

I'm particularly interested on feedback re:

* Are there any reasons _not_ to pursue the short-term steps listed in this RFC?
* Are there pressing business needs/concerns such that some of the farther-future options detailed in the RFC should be prioritized now or earlier?
* Any other informed differences of opinion that some folks have with the recommendations in here.

### Feedback Requested

Please see previous section.

### External References

- [BFD-652](https://jira.cms.gov/browse/BFD-652)

### Security Implications

This RFC has no immediate security implications, nor do any of its short-term recommendations (and the longer-term options it considers would require separate analysis).


karlmdavis:
Marking this as ready for review, even with the TODO. It's not critical and I think the proposal is valid, regardless.

karlmdavis (commented):
Notes from review meeting. No blocking items noted.

mackec1 (approved):
Karl. I don't know enough to really dig into the code per se, but since this is an RFC draft, I see no reason why some iteration work that meets the RFC is work worth undertaking.

cbrunefearless (approved):
lgtm

keithdadkins (approved):
Looks reasonable to me.

PR #463, opened 2021-02-24
Aide fixup

keithdadkins:
There is a cron entry baked into the gold image that runs `aide` and attempts to send the results over email. This is problematic because:
1. Email is not a secure method of sending sensitive data.
2. The results of the aide check is too large so the email fails.
3. The aide database needs to be updated after installing/updating packages.

This patch removes the failing cron job and updates the aide database so our ops can run it on live instances as needed.

For more information on `aide` see:
- https://aide.github.io
- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/sec-using-aide

### Acceptance Validation
- Removes the cron job in root's crontab.
- Runs `aide --update` after all other platinum roles have run.

### External References
- [BFD-602](https://jira.cms.gov/browse/BFD-602)

### DIFF
--- a/ops/ansible/playbooks-ccs/roles/platinum_post/tasks/main.yml
+++ b/ops/ansible/playbooks-ccs/roles/platinum_post/tasks/main.yml
@@ -11,4 +11,4 @@
 # update the aide database
 - name: Update aide database
   become: true
-  ansible.builtin.shell: /sbin/aide --update
+  shell: /sbin/aide --update
--- a/ops/ansible/playbooks-ccs/roles/platinum_post/tasks/main.yml
+++ b/ops/ansible/playbooks-ccs/roles/platinum_post/tasks/main.yml
@@ -1,0 +1,14 @@
+## Final tasks to be run when building the platinum image.
+
+## AIDE fixup
+# remove the failing aide cron job
+- name: 'Ensure failing aide cron job is removed.'
+  become: true
+  ansible.builtin.cron:
+    name: "run AIDE check"
+    state: absent
+
+# update the aide database
+- name: Update aide database
+  become: true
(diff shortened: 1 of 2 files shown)


jzulim (approved):
Handy. LGTM

ops/ansible/playbooks-ccs/roles/platinum_post/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+## Final tasks to be run when building the platinum image.
+
+## AIDE fixup
+# remove the failing aide cron job
+- name: 'Ensure failing aide cron job is removed.'
+  become: true
+  ansible.builtin.cron:
```
njdister, line 7: Did BFD upgrade Ansible everywhere?  I see the new full path task definitions, looks like those are only compatible with Ansible >= 2.10.  Just checking!

njdister (approved):
Good stuff, I would re-use this for BB2, except we're moving to Amazon Linux 2.  ;)

ops/ansible/playbooks-ccs/roles/platinum_post/tasks/main.yml
keithdadkins: I assumed we were at least to 2.10, but could be mistaken. I'll refactor this just in case.

keithdadkins:
> Good stuff, I would re-use this for BB2, except we're moving to Amazon Linux 2. ;)

Lead the way and I will follow!

keithdadkins (commented):
Don't use full path ansible commands.

njdister (approved):
LGTM 👍

jzulim (approved):
Nice move. We can use this role for future checks and work.

PR #471, opened 2021-03-01
Install latest security patches during deploys.

keithdadkins:
Currently, we are installing OS updates during platinum builds. But some time can pass between builds.

This patch installs the latest security updates when building/deploying our FHIR and ETL services as well.

### Change Details
Install security patches when building FHIR/PIPELINE ami's.

### Acceptance Validation
Does what it says.

### Feedback Requested
Always welcome.

### External References

- [BFD-TODO](https://jira.cms.gov/browse/BFD-xxx)

### Security Implications
Improves security by ensuring recently deployed instances are up to date.


jzulim (approved):
Seems like a great idea to me. Thanks!

njdister (approved):
👍

PR #501, opened 2021-03-22
[BFD-695] Upsize prod-sbx to 16 cores (from 4) for performance testing

keithdadkins:
### Change Details
Increases the amount of CPU cores in prod-sbx for AB2D performance testing.
- m5.xlarge (4 cores) -> m5.4xlarge(16 cores)
- adjusts number of db connections to match

Notes: 
- sticking with the same instance class so we can see if/how AB2D scales linearly with num cores
- this will also increase the amount of ram from 16GB to 64GB, however, we have not experienced issues with memory pressure that I am aware of

### External References

- [BFD-695](https://jira.cms.gov/browse/BFD-695)

### Security Implications

### DIFF
--- a/ops/ansible/playbooks-ccs/launch_bfd-server.yml
+++ b/ops/ansible/playbooks-ccs/launch_bfd-server.yml
@@ -20,7 +20,7 @@
         dir: vars/{{ env }}
       tags: [pre-ami, post-ami]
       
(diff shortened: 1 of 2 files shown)


jzulim (approved):
LGTM

mackec1 (approved):

PR #502, opened 2021-03-22
[BFD-670] Correct Diagnosis Type coding for all EOB

Dranith:
### Change Details

The ticket in JIRA has details of what the issue is and what is being fixed.  In summary, the current code was not encoding the `type` value in Diagnosis blocks correctly.  CARIN has three different value sets to use, based on whether you are dealing with an Inpatient, Outpatient or Professional/NonClinician EOB (Pharmacy doesn't have diagnosis mapped).  This ticket updates our EOBs to use the correct value set for each EOB.

* Still uses the generic `DiagnosisLabel`s, but then translates based on the EOB type
* Refactored the `tryMethod` that was added in V2 to `TransformerUtils` to a `ReflectionUtils` class
* Refactored Diagnosis based functionality out of `TransformerUtils` to a `DiagnosisUtils` class
* The translation of `DiagnosisLabel` to the specific Value Sets is done starting in the `translateLabelCode` method of `TransformerUtils`.  This is a "best match" translation of the existing values to the CARIN compliant Value Sets.
* Moved mapping of `CLM_DRG_CD` from invalid Diagnosis element to a SupportingInfo element as noted in the Mapping profile document.

The following are the Value Sets being targeted: 
* Inpatient - https://build.fhir.org/ig/HL7/carin-bb/ValueSet-C4BBClaimInpatientInstitutionalDiagnosisType.html
* Outpatient - https://build.fhir.org/ig/HL7/carin-bb/ValueSet-C4BBClaimOutpatientInstitutionalDiagnosisType.html
* Professional/Non-Clinician - https://build.fhir.org/ig/HL7/carin-bb/ValueSet-C4BBClaimProfessionalAndNonClinicianDiagnosisType.html

### Acceptance Validation

Here is an example of an Outpatient EOB before and after:

image

All `Diagnosis` elements must have a `type` according to CARIN.  The `DiagnosisLabel` didn't have a concept of diagnosis other than the `Primary`.  This left some `Diagnosis` elements without a type.  This is fixed in this PR by including the `Other` option from the Value Sets listed above:

image

Here is an example of how external cause mappings have been updated:

image

### External References

- [BFD-670](https://jira.cms.gov/browse/BFD-670)

### Security Implications


apps/bfd-server/bfd-server-launcher/src/test/java/gov/cms/bfd/server/launcher/AuthenticationIT.java
```diff
@@ -36,7 +36,7 @@ public void authenticationWorksForTrustedClient() throws IOException {
    *
    * @throws IOException (this exception indicates a test failure)
    */
-  @Test(expected = SSLHandshakeException.class)
+  @Test(expected = SSLException.class)
```
Dranith, line 39: Seems that newer versions of Java don't throw the explicit `SSLHandshakeException`.  The good news here is that `SSLHandshakeException` derives from `SSLException` so this should be true regardless of which is thrown.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DiagnosisUtilV2.java
```diff
@@ -0,0 +1,470 @@
...
+            C4BBClaimProfessionalAndNonClinicianDiagnosisType.SECONDARY.toCode(),
+            C4BBClaimProfessionalAndNonClinicianDiagnosisType.SECONDARY.getDisplay());
+    }
+  }
+
+  /**
+   * Translates from {@link DiagnosisLabel} to an EOB type specific Coding based on the (@link
+   * ClaimTypeV2}.
+   */
+  static Coding translateLabelCode(DiagnosisLabel label, ClaimTypeV2 claimType) {
```
Dranith, line 364: Here is the start of the mapping from `DiagnosisLabel` to Diagnosis type on an EOB type basis.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -2322,19 +2245,11 @@ static void addCommonEobInformationInpatientSNF(
...
-       * ICD code. Instead, stick the DRG on the claim's primary/first diagnosis.
-       * SamhsaMatcher uses this field so if this is updated you'll need to update
-       * that as well.
-       */
-      eob.addDiagnosis()
-          .setSequence(1)
-          .setPackageCode(
-              createCodeableConcept(eob, CcwCodebookVariable.CLM_DRG_CD, diagnosisRelatedGroupCd));
-    }
+    // CLM_DRG_CD => ExplanationOfBenefit.supportingInfo
```
Dranith, line 2248: This was incorrectly being mapped into a DiagnosisComponent (this was carried forward from STU3) and has been updated to map to the field dictated in the Profile mapping.

cbrunefearless (approved):
lgtm

apps/bfd-server/bfd-server-launcher/src/test/java/gov/cms/bfd/server/launcher/AuthenticationIT.java
cbrunefearless: thats true

mackec1 (approved):
LGTM. I like that we are heading in the direction where we don't cram everything into TransformUtis.

PR #512, opened 2021-03-25
[BFD-477] Mount eft efs file systems on pipeline (test)

keithdadkins:
## Change Details
This work configures our ETL/pipeline instances to mount our partner EFT EFS file systems on boot using the `amazon-efs-utils` package (for reasons, see below).

### AWS Elastic File-System
AWS EFS is Amazon's implementation of NFS that has been optimized for their environment. Using this package, instead of standard NFS utils, allows us to control access to file systems using IAM roles and policies. 

For example, in this work we are granting our ETL/Pipeline instances permissions to query and mount partner file systems via Instance Profiles.

### FIPS and RHEL 7
Our ITOPS provided gold image is based on RedHat Enterprise Linux 7 with [FIPS](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/security_guide/sect-security_guide-federal_standards_and_regulations-federal_information_processing_standard) enabled (which limits what crypto can be used on our servers).

Using `amazone-efs-utils` on FIPS enabled RHEL 7 required us to:
- Monkey-patch `amazon-efs-utils` to be FIPS compliant.
- Manually compile `stunnel` (an EFS dependency) from source (v5.59 with FIPS).

### Directory structure

Using ansible variables, we create and mount partner file systems in the `{{eft_efs_base_dir}}/{{partner_name}}` directory. These top level directories will be owned by root.

Further, our partners will be "rooted" into a sub-directory owned by a `{{partner_name}}_eft` user and group (service accounts with no ability to login or get a shell) mapped to their assigned POSIX user and group id.

See [Working with Amazon EFS Access Points - Amazon Elastic File System](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-identity-access-points) for more info.

### Mounting
Unfortunately, because our VPC uses custom DNS and not the default AWS DNS, we cannot simply mount the EFS file systems using the standard `sudo mount -t efs -o tls,iam fs-12345678:/ efs` command without first adding a host entry for `fs-12345678`. It's also important that we mount from a mount-target hosted on the same physical az-id as the instance, else we can incur cross-AZ data charges.

This is where the `ops/ansible/roles/bfd-pipeline/templates/mount-eft-efs-file-systems.sh.j2` script comes in. It:
1. Determines which physical AZ ID the instance is on
2. Adds `/etc/hosts` entries for the partners file system
3. Mounts the file systems
4. Adds appropriate entries in `/etc/fstab`

A systemd unit file is used to ensure the above script is executed on boot.

### Acceptance Validation
- [ ] No glaring issues.
- [x] A deployment to test was performed
  - [x] Test EFS file system was mounted on boot
  - [x] File systems survive reboots
- [x] POSIX users and groups are sufficiently hardened.

### External References

- [BFD-477](https://jira.cms.gov/browse/BFD-477)

### Security Implications

- [] altered security controls


cbrunefearless (approved):
lgtm

keithdadkins:
I moved the stunnel stuff into its own role to make yanking it easier later on. Also, hopefully we can get an artifcatory (or similar) setup so we can build this is a separate pipeline and grab it from cache.

njdister (approved):

jzulim (approved):
LGTM

PR #513, opened 2021-03-28
BFD-659: Remediate-Performance-Regression_Patient_Endpoint

mackec1:
### Change Details

Because the DISTINCT JPQL keyword has two meanings based on the underlying query type, it’s important
to pass it through to the SQL statement only for scalar queries where the result set requires duplicates
to be removed by the database engine.

For parent-child entity queries where the child collection is using JOIN FETCH, the DISTINCT keyword should
only be applied after the ResultSet is got from JDBC, therefore avoiding passing DISTINCT to the SQL statement
that gets executed.

Added boolean logic to optimize distinct processing.

 .setHint("hibernate.query.passDistinctThrough", passDistinctThrough); 

This PR does not address changes to V2 patientResource handling of HICN or MBI headers; those will be addressed in another ticket.

### Acceptance Validation

Minimum acceptable processing is to marginally improve JPA Patient data retrieval; it does this by disabling 'select distinct' if IncludeIdentifiers does not include 'hicn' or 'mbi'.

Ideally, running Hibernate ExplainPlan would be used to assert performance improvement.

### Feedback Requested

Verify that code changes do not change basic JPA data retrieval.

### External References

- [BFD-659](https://jira.cms.gov/browse/BFD-659)

### Security Implications

N/A


cbrunefearless (approved):
lgtm

Dranith (approved):
Looks good.

PR #516, opened 2021-03-30
[BFD-713] Fix for `lastUpdated` time being inconsistent across BFD instances

Dranith:
### Change Details

* Changed the way that the last and current update times are compared in `LoadedFilterManager`
* Added unit tests that show the issue and protects against possible behavioral changes in future versions of Java.

### Acceptance Validation

A detailed explanation of the issue along with testing notes is available in JIRA: [BFD-713](https://jira.cms.gov/browse/BFD-713)

From a high level, the issue is that the `Date` objects being compared in `LoadedFilterManager` are actually `java.sql.Timestamp` instances.  `java.sql.Timestamp` splits the time into Seconds and Nanos, instead of keeping a complete millisecond timestamp.  This causes  `Date.before` to return `false` when the times are different because only the seconds were compared and not the milliseconds.  This meant that if the last update in the database happened on the same second, but after the last time `refreshFilters()` run, the BFD instance would never have the correct date.  This explains why when discrepancies were seen they were always only milliseconds off and never seconds+.

Reference: https://docs.oracle.com/javase/8/docs/api/java/sql/Timestamp.html

### Feedback Requested

Please verify that I removed all the pieces that were associated with the `LOADED_FILTER_MANAGER` and `REFRESHLOGGER`

### External References

- [BFD-713](https://jira.cms.gov/browse/BFD-713)

### Security Implications

### DIFF
--- a/apps/bfd-server/bfd-server-launcher/src/main/resources/logback.xml
+++ b/apps/bfd-server/bfd-server-launcher/src/main/resources/logback.xml
@@ -34,37 +34,6 @@
 		<appender-ref ref="HTTP_ACCESS" />
 	</logger>
 
-
-	<appender name="LOADED_FILTER_MANAGER" class="ch.qos.logback.core.FileAppender">
-		<file>${bfdServer.logs.dir:-./target/server-work/}loadedFilterManagerLog.json</file>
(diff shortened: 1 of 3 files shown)


cbrunefearless (approved):
LGTM

mackec1 (approved):
Andrew....good catch and fix.

PR #519, opened 2021-04-04
[BFD-735] Right-size app server instance sizes.

keithdadkins:
### Change Details
This PR is an attempt to "right-size" our application server ec2 instance sizes to better handle our current and future workloads. 

This involves migrating our application servers from memory optimized instance classes to compute optimized instance classes, and ups the number of virtual CPU's from 4 to 16 (for upper environments). Which gives prod and prod-sbx 16 cores and 32 GiB of memory (up from 4 cores and 16GiB of memory).

Specifically, this PR:
- Moves prod and prod-sbx to c5.4xlarge instance class from m5.xlarge.
- Moves test to c5.xlarge from m5.xlarge (keeping test at 4 cores to save $)
- Dynamically sets the max db connections to the Oracle recommended 10 per core.

### External References

- [BFD-735](https://jira.cms.gov/browse/BFD-735)

### Security Implications

### DIFF
--- a/ops/ansible/playbooks-ccs/build_bfd-server.yml
+++ b/ops/ansible/playbooks-ccs/build_bfd-server.yml
@@ -49,7 +49,7 @@
         data_server_db_url: "{{ vault_data_db_secondary_url }}"
         data_server_db_username: "{{ vault_data_server_db_username }}"
         data_server_db_password: "{{ vault_data_server_db_password }}"
-        data_server_db_connections_max: 40
+        data_server_db_connections_max: {{ data_server_num_cores * 10 }} # Oracle recommends 1-10 connections per CPU core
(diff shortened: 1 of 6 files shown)


keithdadkins:
Upsize instances have been deploy to prod-sbx and test. AB2D ran some smoke tests (using 10k benes) and saw a dramatic improvement in responses, however, their small test run was not enough to trigger a scaling event.

I am not sure I agree with the current method of taking 80% of memory for the JVM. In the future, I'd like to profile the instance and see how much memory is needed by the application's host, and simply give the rest to the JVM. I'm guessing 2-4 GB's of memory for the host, but I have no data to back that up. Just a guess with the ITOPS gold running AV, etc.

cbrunefearless (approved):
lgtm

njdister (approved):
🌮

PR #520, opened 2021-04-05
Added instant to set lastupdated timestamps to after compare points of setlastupdated

cbrunefearless:
### Change Details
Added instant to set lastupdated timestamps to after compare points of setlastupdated

### Acceptance Validation
Make sure it is in the correct timestamp format

### Feedback Requested
Looks correct as the before checking.

### External References

- [BFD-727](https://jira.cms.gov/browse/BFD-727)

### Security Implications

n/a

n/a

n/a

n/a


Dranith (approved):
Approved.

Was able to reproduce the original issue before the fix, and can confirm that this fix addresses the bug in this case.

jzulim (approved):
This should do the trick. Nice turnaround.

PR #521, opened 2021-04-06
[BFD-707] V2 Inpatient EOB Unit Tests

Dranith:
### Change Details

Added unit tests for the V2 Inpatient EOB.  The style of these tests diverges greatly from the V1 style.  These tests take a more deliberate approach that tests the json output while trying to make no assumptions about the code they are testing.  A "gold copy" of a transformed, json formatted Inpatient Sample A record was used as the template.  The tests lean greatly on the `equalsDeep` method that all FHIR components implement.  This allows the tests to first read from the transformed EOB and then construct a FHIR element to test against.

The benefits to using this approach are:
* If the code changes in any way, even a simple dollar amount, at least one test will fail.  In the old approach the Claim object was compared directly to the EOB.  This meant that these sort of changes would not be detected by the unit tests.  Detection is preferred because it forces the implementer to decide if the change was OK or not, and will catch more accidental changes.
* This takes more of an actual unit test approach.  Instead of having a single large test that looks at the entire EOB, there are per-element tests that give finer granularity and can isolate failures more easily.
* The generated test file may look intimidating in length, but a lot of it is simple copy paste.  Additionally the unit tests for other EOB types will be able to lean heavily on the work done here.

### Acceptance Validation

Tests all pass.  In the future we should look into adding code coverage tools and develop tests to cover any lines that are missed.

### External References

- [BFD-707](https://jira.cms.gov/browse/BFD-707)

### Security Implications


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -228,7 +228,7 @@ public Bundle findByPatient(
     Set<ClaimTypeV2> claimTypes = parseTypeParam(type);
     OffsetLinkBuilder paging = new OffsetLinkBuilder(requestDetails, "/ExplanationOfBenefit?");
 
-    Operation operation = new Operation(Operation.Endpoint.V1_EOB);
+    Operation operation = new Operation(Operation.Endpoint.V2_EOB);
```
Dranith, line 231: Minor tweaks that aren't really part of this ticket, but not big enough to put somewhere else.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -266,15 +266,12 @@ public Bundle findByPatient(
                   ClaimTypeV2.CARRIER, beneficiaryId, lastUpdated, serviceDate)));
     }
 
-    /*
-    TODO: When DME and HHA are implemented
```
Dranith, line 270: This was commented out but shouldn't have been.

cbrunefearless (dismissed):
LGTM

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/InpatientClaimTransformerV2Test.java
```diff
@@ -41,75 +67,1592 @@ public InpatientClaim generateClaim() throws FHIRException {
...
+                "Discharged"));
+
+    Assert.assertTrue(compare.equalsDeep(sic));
+  }
+
+  @Test
+  public void shouldHaveAdmissionPeriodSupInfo() {
+    SupportingInformationComponent sic =
+        TransformerTestUtilsV2.findSupportingInfoByCode("admissionperiod", eob.getSupportingInfo());
+
```
mackec1, line 255: Andrew. just curious as to whether we should do an assert prior to firing a deep comparison; to wit:

Assert.assertNotNull(sic);

Now granted if it is null the deep compare would fail; I just don't know what the failure state/message might be (i.e., it may be obvious that the compare on the null _sic_.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/TransformerTestUtilsV2.java
```diff
@@ -1011,4 +1017,316 @@ static void assertPeriodEquals(
...
+            .filter(
+                cmp ->
+                    cmp.getCategory().getCoding().stream()
+                            .filter(c -> code.equals(c.getCode()))
+                            .count()
+                        > 0)
+            .findFirst();
+
+    Assert.assertTrue(si.isPresent());
+
```
mackec1, line 1141: I had a comment that I nuked in the Transformer tests r.e. Asserting we have an object prior to constructing the element for the deep compare. But I should have checked here first...

mackec1 (approved):
Outstanding; tests individual components PLUS provides a really nice framework for our subsequent tests.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/TransformerTestUtilsV2.java
Dranith: :+1:

Yeah this seemed like a more appropriate place to check that we got a value back than in every individual test.  I'm not sure I really liked this, because as you noticed it is a bit opaque .. but it does save some copy pasting.

mackec1 (approved):

cbrunefearless (approved):
lgtm

PR #524, opened 2021-04-09
BFD-717: Send Dropwizard Metrics to New Relic

rtravitz:
### Change Details
This PR addresses [BFD-717](https://jira.cms.gov/browse/BFD-717). It adds a `NewRelicReporter` that takes existing measurements from Dropwizard, transforms them into a format that New Relic will understand, and then ships them off. It also upgrades the Dropwizard Metrics major version to 4.1.2. There is an enum introduced in v4 that the New Relic integration package relies on. This also required adding a separate `metrics-jmx` artifact, which was split out in the migration from v3 to v4.

The reporter will only be configured as part of the `MetricRegistry` Spring Bean if the `NEW_RELIC_METRIC_KEY` environment variable is set. This is a new key that differs from the existing license key. 

A writeup on how to query and analyze the new Dropwizard metrics in New Relic is forthcoming.

_Special note on Codahale_: You'll see [timers in the code](https://github.com/CMSgov/beneficiary-fhir-data/blob/d5700eb2f4d2502c9356b90a4871ceab544f84b9/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/OutpatientClaimTransformerV2.java#L3-L4) that come from a project called "codahale" that's not referenced in any of the pom files. This is the old name for a project that Dropwizard took over. Those classes all come from Dropwizard.

### Acceptance Validation

- If you add the correct key from the New Relic Insights keys UI, you should start to see events appear. You can validate that they are arriving with the following NRQL: `SELECT * from Metric WHERE appName = 'BFD Server (local)'`, where `BFD Server (local)` can be substituted to whatever you have the `NEW_RELIC_APP_NAME` env var set as locally. 
- If you don't have the metric key set, it shouldn't be attempting to send the metrics over the wire. If you tail the logs after server startup, you shouldn't see failure messages every 15 seconds that it failed to send a metric batch. (If you want to see failure logs, feel free to put in a garbled key—you'll see them real quick!)

### Feedback Requested

- Is my use of `System.getenv` fair for sourcing the configuration?
- Are we happy with a 15 second batch interval? Pulled it straight from the examples.
- Do we actually need the `JmxReporter` in there for anything? If we're not using it, it would be nice to remove. It's now a separate package. I'm also not sure where it's disabled when in production, though @mackec1 indicates that if it _was_ enabled it would bring the server to its knees.

### External References

- [BFD-717](https://jira.cms.gov/browse/BFD-717)
- [Dropwizard Metrics](https://github.com/dropwizard/metrics)
- [Dropwizard New Relic Integration](https://github.com/newrelic/dropwizard-metrics-newrelic)

### Security Implications

These changes add the [New Relic Dropwizard Reporter](https://github.com/newrelic/dropwizard-metrics-newrelic). It is an official library published by New Relic to integrate with Dropwizard. The changes also increment the major version of Dropwizard to an [actively maintained version](https://github.com/dropwizard/metrics#versions).

These changes would transmit telemetry to New Relic that currently gets stored in memory on the app server and offered up via a `/metrics/metrics` endpoint. This is not PII or PHI. It is counters and timers around certain blocks of our code that will be used to better understand usage and performance.

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
@@ -1,38 +1,36 @@
 package gov.cms.bfd.server.war;
 
-import ca.uhn.fhir.rest.server.IResourceProvider;
-import com.codahale.metrics.JmxReporter;
-import com.codahale.metrics.MetricRegistry;
-import com.codahale.metrics.health.HealthCheckRegistry;
-import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
-import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
-import com.zaxxer.hikari.HikariDataSource;
-import gov.cms.bfd.model.rif.schema.DatabaseSchemaManager;
-import gov.cms.bfd.model.rif.schema.DatabaseTestHelper;
-import gov.cms.bfd.model.rif.schema.DatabaseTestHelper.DataSourceComponents;
-import gov.cms.bfd.server.war.r4.providers.R4CoverageResourceProvider;
-import gov.cms.bfd.server.war.r4.providers.R4ExplanationOfBenefitResourceProvider;
-import gov.cms.bfd.server.war.r4.providers.R4PatientResourceProvider;
-import gov.cms.bfd.server.war.stu3.providers.CoverageResourceProvider;
-import gov.cms.bfd.server.war.stu3.providers.ExplanationOfBenefitResourceProvider;
-import gov.cms.bfd.server.war.stu3.providers.PatientResourceProvider;
-import gov.cms.bfd.sharedutils.exceptions.BadCodeMonkeyException;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
 import javax.persistence.EntityManager;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.PersistenceContext;
 import javax.persistence.PersistenceUnit;
 import javax.sql.DataSource;
-import net.ttddyy.dsproxy.support.ProxyDataSource;
-import net.ttddyy.dsproxy.support.ProxyDataSourceBuilder;
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.health.HealthCheckRegistry;
+import com.codahale.metrics.jmx.JmxReporter;
+import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
+import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
+import com.codahale.metrics.newrelic.NewRelicReporter;
+import com.newrelic.telemetry.Attributes;
+import com.newrelic.telemetry.OkHttpPoster;
+import com.newrelic.telemetry.SenderConfiguration;
+import com.newrelic.telemetry.metrics.MetricBatchSender;
+import com.zaxxer.hikari.HikariDataSource;
+
 import org.hibernate.cfg.AvailableSettings;
 import org.hibernate.jpa.HibernatePersistenceProvider;
 import org.hibernate.tool.schema.Action;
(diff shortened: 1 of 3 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -363,6 +371,37 @@ public MetricRegistry metricRegistry() {
     final JmxReporter reporter = JmxReporter.forRegistry(metricRegistry).build();
     reporter.start();
 
+    String newRelicMetricKey = System.getenv("NEW_RELIC_METRIC_KEY");
```
jzulim, line 374: This seems to be the construct / method for injection other env vars. 👍

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -363,6 +371,37 @@ public MetricRegistry metricRegistry() {
...
+
+      Attributes commonAttributes =
+          new Attributes().put("host", hostname).put("appName", newRelicAppName);
+
+      NewRelicReporter newRelicReporter =
+          NewRelicReporter.build(metricRegistry, metricBatchSender)
+              .commonAttributes(commonAttributes)
+              .build();
+
+      newRelicReporter.start(15, TimeUnit.SECONDS);
```
jzulim, line 402: 15 seconds seems very reasonable to me.

jzulim (approved):
Nice work getting this going. All looks good with a thorough description to boot (super helpful). I see the `JmxReporter` but do not see it being disabled anywhere. I suspect this is in fact being started and running and unless we're going to profile with VisualVM or another tool, I'd suggest we remove it due to it's a potential performance hit. Other than that, looks good. If you decide to rip it, re-tag me and I'll stamp again. LGTM.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -363,6 +371,37 @@ public MetricRegistry metricRegistry() {
...
+
+      String hostname;
+      try {
+        hostname = InetAddress.getLocalHost().getHostName();
+      } catch (UnknownHostException e) {
+        hostname = "unknown";
+      }
+
+      SenderConfiguration configuration =
+          SenderConfiguration.builder("https://gov-metric-api.newrelic.com", "/metric/v1")
```
Dranith, line 387: Will this never change?  Should this be an Env var as well?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
Dranith: Since 15 seconds is a bit arbitrary, again it might be good to set this via an environment variable.  Then if it is too frequent / too infrequent you can change it without having to change code.

Dranith: I guess that comment really depends on the conventions BFD typically uses.  I tend to favor configurablility over hardcoded.  One thing I often do is have a sane default (15) but allow that to be overridden via an environment variable.

This approach could be used here and above for the URL.  That way you don't have to set them, but you can if you need to at some point in the future.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
rtravitz: Oh, I wouldn't say never. In particular the `v1` to `v2` seems like it has potential to change in the future. But at that point we're likely making code changes anyway.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
rtravitz: Good point on the timer. I'll add in the variable with the default staying at 15. What's the Java way to do that? In other languages, I'll often see something like `YOUR_VAR || 15`. Is it a full on if block to check if the env var is null?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
keithdadkins: Yes, I would put the host address in an env var.

keithdadkins (dismissed):
I'm curious what if any performance hit we would take, but either way (assuming it's reasonable), it will be worth it. Thanks for this!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
keithdadkins: For a few reason. One, we may need this to change depending on the environment in the future. Two, I could see someone, somewhere, considering this semi-sensitive info. We've been pinged for similar things in the past. My two cents FWIW :)

rtravitz: I'll change them to environment variables for easy of configuration. But if anybody ever comes in worried about those two values being semi-sensitive, I'd point them to the [public New Relic documentation where they are listed](https://docs.newrelic.com/docs/security/security-privacy/compliance/fedramp-compliant-endpoints/#metric-api).

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
rtravitz: Just going to catch any parsing exceptions and set the default there instead. Save myself a step.

rtravitz:
> I'd suggest we remove it due to it's a potential performance hit.

I am removing the `JmxReporter`. Let's us get rid of a dependency. And if anybody is interested in those metrics locally, they can go to `/metrics/metrics` and see them provided by the servlet much more easily.

Dranith (approved):
Looks good.

keithdadkins (approved):
LGTM

cbrunefearless (approved):
lgtm

mackec1 (approved):

PR #526, opened 2021-04-13
DCGEO-18: PipelineJob skeleton for DC Geo project

brianburton:
### Change Details

This PR creates a skeleton `PipelineJob` implementation that will serve as the foundation for our complete ETL process.  Remaining functionality will be added over the course of the next few sprints.  We have added a subproject named `bfd-pipeline-rda-grpc` containing our ETL code.  The `DcGeoRDALoadJob` class is the `PipelineJob` implementation.  It handles the basic logic of opening and closing connections to our data source (ultimately the RDA service) and sink (ultimately the BFD database) and triggering the data transfer between them.  

The empty `PreAdjudicatedClaim` class will be expanded in later PR's to hold our actual data fields.

The complete ETL processes will be created by implementing the `RDASource` and `RDASink` interfaces appropriately in later PRs.  When it runs the skeleton establishes its connections and triggers streaming of data from the source to the sink.  Various configuration settings are supported.  These are currently controlled by environment variables.  The configuration settings are:

- expected number of minutes between invocations of the ETL process
- number of objects per batch when writing to the database
- maximum duration of the stream in a single invocation
- maximim number of objects to process in a single invocation (i.e. max objects in stream)

### Acceptance Validation

No integration with PipelineManager has been done since work on the new pipeline orchestration framework is still ongoing in another branch. This code will have no runtime impact on BFD until that integration happens.

For now the acceptance criteria should be that the unit tests pass.

### Feedback Requested

Nothing specific.

### External References

- [DCGEO-18](https://jira.cms.gov/browse/DCGEO-18)

### Security Implications

As it just introduces a skeleton implementation this PR does not process any data or establish any external connections.  Also the code is not integrated into PipelineManager so it will not be invoked so there should not be any security implications.

Unit tests use Mockito to easily create test conditions for specific scenarios without having to write mock objects.  This is only a `test` dependency so it won't be affect runtime environments.

````
        <dependency>
            <!-- Used to create mock objects for unit tests. -->
            <groupId>org.mockito</groupId>
            <artifactId>mockito-core</artifactId>
            <version>3.8.0</version>
            <scope>test</scope>
        </dependency>
````

None.

None.

No.

No.

### CODE REVIEW

File: apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java [hunk 1]
```diff
@@ -0,0 +1,149 @@
...
+  public PipelineJobOutcome call() throws Exception {
+    final long startMillis = System.currentTimeMillis();
+    int processedCount = 0;
+    Exception error = null;
+    try {
+      callsMeter.mark();
+      try (RDASource<PreAdjudicatedClaim> source = sourceFactory.call();
+          RDASink<PreAdjudicatedClaim> sink = sinkFactory.call()) {
+        processedCount =
+            source.retrieveAndProcessObjects(
```
ericbuckley, line 89: if the upcoming changes to the PipelineManager supports threading and allows for us to leave the connection open indefinitely; our metrics wouldn't move until a shutdown.  No need to address now, but we might need to rethink how to manage metrics if we want to leave the connection open

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RDASink.java
```diff
@@ -0,0 +1,48 @@
...
+public interface RDASink<T> extends AutoCloseable {
+  /**
+   * Write the object to the data store and return the number of objects successfully written.
+   *
+   * @param object single object to be written to the data store
+   * @return number of objects successfully processed
+   * @throws ProcessingException if the operation fails
+   */
+  default int writeObject(T object) throws ProcessingException {
+    return writeBatch(Collections.singleton(object));
```
ericbuckley, line 21: I'm a little confused about the dependencies between the two methods here, `writeObject` calls `writeBatch` and `writeBatch` calls `writeObject`?  Am I missing something?

ericbuckley (commented):
@brianburton looking great so far!

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RDASink.java
brianburton: Both are default implementations of the methods in the interface.  An implementing class will override one of them and get the implementation of the other for free.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>gov.cms.bfd</groupId>
+            <artifactId>bfd-pipeline-shared-utils</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
```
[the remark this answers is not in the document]
brianburton: I did not try to pare down the dependencies too much.  I retained those that seemed likely to be needed when we have implemented the database updates in later sprints.  Since the extras are retained from the ccw-rif pom.xml file they do not add any extra jars to BFD's deployment requirements.  I think that we should remove extras at the end of this PI though.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/ConfigUtils.java
```diff
@@ -0,0 +1,34 @@
...
+   * Retrieve integer configuration value with the given name. Returns the specified default value
+   * when there is no configuration value matching the key.
+   *
+   * @param key key that identifies the particular configuration value.
+   * @param defaultValue default value if there is no configuration value matching the key
+   * @return either the configuration value or the default
+   */
+  public static int getInt(String key, int defaultValue) {
+    String strValue = System.getenv(key);
+    return strValue != null ? Integer.parseInt(strValue) : defaultValue;
```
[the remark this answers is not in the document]
brianburton: I prefer to fail fast in configuration rather than falling back on a default when a config file or environment variable is incorrect.  We could catch it and log some error message before rethrowing though.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
```diff
@@ -0,0 +1,149 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(DcGeoRDALoadJob.class);
+  public static final String SCAN_INTERVAL_PROPERTY = "DCGeoRDALoadIntervalSeconds";
+  public static final String RUN_TIME_PROPERTY = "DCGeoRDALoadRunSeconds";
+  public static final String MAX_RECORDS_PROPERTY = "DCGeoRDALoadMaxRecords";
+  public static final int SCAN_INTERVAL_DEFAULT = 300;
+  public static final int RUN_TIME_DEFAULT = 300;
+  public static final int MAX_RECORDS_DEFAULT = Integer.MAX_VALUE;
+  public static final String BATCH_SIZE_PROPERTY = "DCGeoBatchSize";
+  public static final int BATCH_SIZE_DEFAULT = 1;
+  public static final String CALLS_METER_NAME =
```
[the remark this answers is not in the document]
brianburton: Thanks.  I agree that the actual metrics we record need some careful consideration.  What I have so far are low hanging fruit but I'm sure we will need others.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
```diff
@@ -0,0 +1,149 @@
...
+  private final Callable<RDASink<PreAdjudicatedClaim>> sinkFactory;
+  private final Meter callsMeter;
+  private final Meter failuresMeter;
+  private final Meter successesMeter;
+  private final Meter processedMeter;
+
+  public DcGeoRDALoadJob(
+      Config config,
+      Callable<RDASource<PreAdjudicatedClaim>> sourceFactory,
+      Callable<RDASink<PreAdjudicatedClaim>> sinkFactory,
```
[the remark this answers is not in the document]
brianburton: It's best to consider the PreAdjudicatedClaim object as a root level object containing whatever objects our schema requires.  It might actually contain a set of objects of different types or something (we'll know that when we get to the sprint that defines our schema).  RDASource is a higher level abstraction than just gRPC/RDA.  In the gRPC client case there will be an internal mapping between whatever RDA returns to us and whatever our database schema requires.

I originally left PreAdjudicatedClaim as a generic type too but thought it best to make a concrete class here to avoid confusion and prevent the PipelineManager having to deal with a generic type.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
brianburton: That's an excellent point.  There are a few other points to consider though:

- the RDASource/RDASink implementation objects will define their own internal metrics
- the processedMeter increments in tandem with the callsMeter which has the same behavior you mentioned.
- the limits on run time, idle time, and number of objects streamed in our calls to RDA will tend to cause these calls to last a finite amount of time (assuming we configure them)

The actual metrics we record definitely needs some more consideration/discussion going forward.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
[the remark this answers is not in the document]
ericbuckley: Yeah as long as we revisit this in the future.  If someone refactors the CCW-RIF and removes a dependency it would suck to still have it referenced here if we didn't need it.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/ConfigUtils.java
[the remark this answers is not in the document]
ericbuckley: 💯 on failing fast; I was thinking the latter and having a useful message in the logs that our env variable was misconfigured.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
ericbuckley: Yeah lets plan to talk about KPIs in Sprint 2.2, no sense in overthinking this before we have a clear picture of what we want to track.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RDASink.java
ericbuckley: neat; I'm guessing we'll want to override `writeBatch`, but we'll see how it plays out when we get the proto file from the RDA API team

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/SkeletonRDASink.java
```diff
@@ -0,0 +1,31 @@
...
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Skeleton RDASink implementation that just updates a Meter to count the number of times it has
+ * been called. This implementation satisfies the requirements of DCGEO-18.
+ */
+public class SkeletonRDASink implements RDASink<PreAdjudicatedClaim> {
```
ericbuckley, line 12: Come implementation time do you think you'll remove the prefix `Skeleton`?

brianburton: Yes this class would be replaced with the real sink.  That sink will have an appropriate name like PostgresRDASink.

ericbuckley (dismissed):
LGTM

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/ConfigUtils.java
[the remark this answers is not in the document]
brianburton: I have pushed a commit with this change.

ericbuckley (approved):

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RDASink.java
rtravitz: FWIW, this was also confusing to me until I opened up this resolved thread. Might be worth a comment about the intended use. I guess it sort of is with `at least one of the methods must be implemented`. But maybe making this quoted expectation clear:
> The idea is that a class that implements the interface will override just one of the two methods and get the other method implementation for free.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RDASource.java
```diff
@@ -0,0 +1,25 @@
...
+ * @param <T> the type of objects processed
+ */
+public interface RDASource<T> extends AutoCloseable {
+  /**
+   * Retrieve some number of objects from the source and pass them to the sink for processing.
+   *
+   * @param maxToProcess maximum number of objects to retrieve from the source
+   * @param maxPerBatch maximum number of objects to collect into a batch before calling the sink
+   * @param maxRunTime maximum amount of time to run before returning
+   * @param sink sink to receive batches of objects
```
rtravitz, line 19:
```suggestion
   * @param sink to receive batches of objects
```

rtravitz (approved):
Not a codeowner yet so my approval doesn't get you any further than Eric's. But one of the people who has been here longer should look before merging anyway. My two cents: I don't see anything here that should hold this back from going in as the initial skeleton. And I'm glad for the mocking in the unit tests. ✨

apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+            <artifactId>logback-classic</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <!-- Used to monitor the JVM's memory usage, garbage collection, etc.
+                    at runtime. -->
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-jvm</artifactId>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
brianburton: Good point.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/ConfigUtils.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This helper class abstracts the source of configuration values and handles insertion of default
+ * values when none are provided by the environment. Currently implemented to pull values from
+ * environment variables.
+ */
+class ConfigUtils {
```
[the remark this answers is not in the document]
brianburton: Yes I agree this would be very useful at a higher layer but I was reticent about making changes in the main code base at this point.   AppConfiguration is going to wind up loading these configs at some point but not sure it's the best place right now.  I was holding off until the orchestration PR has been merged.

ericbuckley (dismissed):

jzulim (dismissed):
LGTM. Thank you for the review last week. My only comments echo @rtravitz in regards to AppCOnfiguration vs ENV vars. When the time comes, seems reasonable to address this when tieing into the PipelineManager.

rtravitz (dismissed):

ericbuckley:
@jzulim neither Brian or myself have permissions to merge the PR, could you do it?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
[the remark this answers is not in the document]
karlmdavis: Definitely not needed here, especially at `test` scope.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+        <dependency>
+            <!-- Used to create mock objects for unit tests. -->
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.8.0</version>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
karlmdavis: @rtravitz Interesting. In general, I find myself personally not a fan of mockito-style tests; I tend to prefer to pull out interfaces for things that need to be mocked, and then just write mock implementations for them, myself.

That said, I'd love to see a research spike / PR that took a stab at the approach! It'd be good for the team to see a head-to-head comparison between the approaches, and pick one to go with.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
```diff
@@ -0,0 +1,149 @@
...
+   * @param appMetrics MetricRegistry used to track operational metrics
+   * @return a DcGeoRDALoadJob instance suitable for use by PipelineManager.
+   */
+  public static DcGeoRDALoadJob newDcGeoRDALoadJob(MetricRegistry appMetrics) {
+    return new DcGeoRDALoadJob(
+        new Config(),
+        () -> new SkeletonRDASource(appMetrics),
+        () -> new SkeletonRDASink(appMetrics),
+        appMetrics);
+  }
```
karlmdavis, line 73: Interesting style. Why a `static` method instead of just an overloaded-and-chained constructor?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
```diff
@@ -0,0 +1,149 @@
...
+
+  public Duration getScanInterval() {
+    return config.getScanInterval();
+  }
```
karlmdavis, line 77: Doesn't look like anything is using this?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
```diff
@@ -0,0 +1,149 @@
...
+    } catch (Exception ex) {
+      error = ex;
+    }
+    processedMeter.mark(processedCount);
+    final long stopMillis = System.currentTimeMillis();
+    LOGGER.info("processed {} objects in {} ms", processedCount, stopMillis - startMillis);
+    if (error != null) {
+      failuresMeter.mark();
+      LOGGER.error("processing aborted by an exception: message={}", error.getMessage(), error);
+      throw error;
```
karlmdavis, line 104: I always recommend strongly against rethrows like this, as it discards the stack from the actual cause, which will make debugging very challenging. Instead, use a wrap-and-rethrow, which will preserve and extend the original stack trace.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
```diff
@@ -0,0 +1,149 @@
...
+      try (RDASource<PreAdjudicatedClaim> source = sourceFactory.call();
+          RDASink<PreAdjudicatedClaim> sink = sinkFactory.call()) {
+        processedCount =
+            source.retrieveAndProcessObjects(
+                config.getMaxObjectsPerCall(), config.getBatchSize(), config.getMaxRunTime(), sink);
+      }
+    } catch (ProcessingException ex) {
+      processedCount += ex.getProcessedCount();
+      error = ex.getCause();
+    } catch (Exception ex) {
```
karlmdavis, line 95: I imagine you know this, given your review of the orchestration PR, but just a heads-up that this will need to be tweaked once that's merged, to allow for interrupts.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
```diff
@@ -0,0 +1,149 @@
...
+    final long startMillis = System.currentTimeMillis();
+    int processedCount = 0;
+    Exception error = null;
+    try {
+      callsMeter.mark();
+      try (RDASource<PreAdjudicatedClaim> source = sourceFactory.call();
+          RDASink<PreAdjudicatedClaim> sink = sinkFactory.call()) {
+        processedCount =
+            source.retrieveAndProcessObjects(
+                config.getMaxObjectsPerCall(), config.getBatchSize(), config.getMaxRunTime(), sink);
```
karlmdavis, line 90: Thinking about it more, I probably wouldn't have a max run time or max objects, and instead let the job run until it's interrupted.

It's less code that way, which is almost always better, and if we ever move to running jobs across multiple nodes, it's better to not have a job unnecessarily jumping around from node to node.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/PreAdjudicatedClaim.java
```diff
@@ -0,0 +1,4 @@
+package gov.cms.bfd.pipeline.rda.grpc;
+
+/** Java bean representing preadjudicated claims received from RDA API. */
+public class PreAdjudicatedClaim {}
```
karlmdavis, line 4: I'm a little iffy on this choice, similar to the conversation you & Eric were having elsewhere. I think it's fine for now, but I'll be curious to see how it plays out once you actually start catching data.

For one thing, I prolly wouldn't assume this early that you'll be _able_ to have a single base class for all the various data types that you might end up having to catch & process.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RDASink.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Collections;
+
+/**
+ * Interface for objects that process incoming objects. At least one of the methods must be
+ * implemented. All implementations are AutoCloseable since they will generally hold a database
+ * connection.
+ *
+ * @param <T> the type of objects processed
+ */
+public interface RDASink<T> extends AutoCloseable {
```
karlmdavis, line 12: Minor naming convention thing:

```suggestion
/**
 * Interface for objects that process incoming objects. At least one of the methods must be
 * implemented. All implementations are AutoCloseable since they will generally hold a database
 * connection.
 *
 * @param <T> the type of objects processed
 */
public interface RdaSink<T> extends AutoCloseable {
```

karlmdavis, line 12: I would probably also add a constraint here and/or in the methods that's something like "all implementations of this interface SHALL be idempotent." (And explain what that means, as it'll be a new term for many folks.)

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RDASource.java
```diff
@@ -0,0 +1,25 @@
...
+import java.time.Duration;
+
+/**
+ * Interface for objects that retrieve objects from some source and pass them to a sink for
+ * processing. All implementations are AutoCloseable since they will generally hold a network or
+ * database connection.
+ *
+ * @param <T> the type of objects processed
+ */
+public interface RDASource<T> extends AutoCloseable {
```
karlmdavis, line 12: Minor naming convention:

```suggestion
/**
 * Interface for objects that retrieve objects from some source and pass them to a sink for
 * processing. All implementations are AutoCloseable since they will generally hold a network or
 * database connection.
 *
 * @param <T> the type of objects processed
 */
public interface RdaSource<T> extends AutoCloseable {
```

karlmdavis, line 12: I would probably also add a constraint here and/or in the methods that's something like "all implementations of this interface SHALL be idempotent." (And explain what that means, as it'll be a new term for many folks.)

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/SkeletonRDASink.java
karlmdavis, line 12:
```suggestion
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

/**
 * Skeleton RDASink implementation that just updates a Meter to count the number of times it has
 * been called. This implementation satisfies the requirements of DCGEO-18.
 */
public class SkeletonRdaSink implements RDASink<PreAdjudicatedClaim> {
```

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/SkeletonRDASource.java
```diff
@@ -0,0 +1,35 @@
...
+import com.codahale.metrics.MetricRegistry;
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Skeleton RDASource implementation that just updates a Meter to count the number of times it has
+ * been called. This implementation satisfies the requirements of DCGEO-18.
+ */
+public class SkeletonRDASource implements RDASource<PreAdjudicatedClaim> {
```
karlmdavis, line 13:
```suggestion
public class SkeletonRdaSource implements RDASource<PreAdjudicatedClaim> {
```

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJobTest.java
```diff
@@ -0,0 +1,128 @@
...
+import java.io.IOException;
+import java.time.Duration;
+import java.util.concurrent.Callable;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DcGeoRDALoadJobTest {
```
karlmdavis, line 23:
```suggestion
public class DcGeoRdaLoadJobTest {
```

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/RDASinkTest.java
```diff
@@ -0,0 +1,52 @@
+package gov.cms.bfd.pipeline.rda.grpc;
+
+import java.io.IOException;
+import java.util.Arrays;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class RDASinkTest {
```
karlmdavis, line 10: Naming convention thing:

```suggestion
public class RdaSinkTest {
```

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,10 @@
+<configuration>
+  <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
+      <pattern>%msg%n</pattern>
+    </encoder>
+  </appender>
+  <root level="OFF">
```
karlmdavis, line 7: Is that really what you want for your tests? If you're worried about build output spam, I might instead set Surefire/Failsafe to redirect output to a file.

karlmdavis (changes requested):
All in all, I think this is 🥇-worthy stuff!

I've noted some minor things that you should address, but other than the exception rethrow item, it's no big deal.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/resources/logback-test.xml
brianburton: The tests explicitly check for exceptions so I'm not worried about losing the logging output when they run.  ERRORs in the logs during tests add a huge amount of noise and make it look like tests are failing even though the exceptions are expected.  I could switch to redirecting the logs during tests if that's the preferred convention.  I'd rather make that change in the next PR.

karlmdavis: I don't have a strong opinion either way. Though, from experience, having the log output is often the only way to diagnose why a test failed on a CI system. 🤷

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
brianburton: The stack trace is established when an exception is created, not when it's thrown or rethrown.  That can be verified with a unit test.  Rethrowing preserves the original stack trace.

If you'd prefer the exception be wrapped I can certainly add that.  I just find nested exceptions can add excessive bulk to the logs when they aren't actuall required.

karlmdavis: That is WILD: I've had that wrong in my head for, like, a damn decade. Thanks for the correction!

I do still think it's important to always wrap-and-rethrow, though, because otherwise your stack trace is painting a somewhat misleading picture about the control flow.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/resources/logback-test.xml
brianburton: Good point.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
brianburton: Good point.  When I first designed this I thought we were only going to have a time slice on a shared thread.  With our own thread it makes sense to not worry about the limits.  Though they can also be useful in a testing situation.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
brianburton: Just a style thing. I think it helps maintainability a bit. When the object will be constructed by a call from somewhere outside the module I like a static method since it decouples the constructor to use. It also avoids having too many constructors.. But it's not a major thing. Could go the overloaded constructor route too.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
brianburton: It's the basis for the equivalent methods in the new orchestration code.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
karlmdavis: To be clear, I don't have a strong opinion -- was just curious.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/DcGeoRDALoadJob.java
karlmdavis: Figured. Do you see it actually needing to be configurable, as opposed to hardcoded (at least for non-test-usage)?

ericbuckley (approved):
Thanks for making those changes

karlmdavis (approved):
🚀

jzulim (approved):

PR #527, opened 2021-04-13
DCGEO-19: gRPC Client Skeleton

brianburton:
### Change Details

***The code from this PR has been subsumed into the work for DCGEO-23 so this PR is no longer active.  Feedback from this PR will be incorporated into the PR for for DCGEO-23.***

This PR adds an RDASource implementation class called `GrpcRDASource` that can stream data from an arbitrary gRPC service using an adaptor object.  The adaptor interface is called `GrpcStreamCaller` which handles the details of any specific gRPC service and any transformation of incoming objects from the stream into our ETL's ORM base class `PreAdjudicatedClaim`.

Since the RDA 0.1 MVP release was not available during this spring this PR simply adds a sample implemenation class `HealthCheckStreamCaller`.  A subsequent PR in a later sprint will add a caller for the actual RDA pre-adjudicated claims service once that is available.

Configuration is via environment variables.  New configuration options in this PR include:

- service host and port
- max idle time for stream (stream auto-closes is no new objects arrive within this time of previous object)

### Acceptance Validation

No integration with PipelineManager has been done since work on the new pipeline orchestration framework is still ongoing in another branch. This code will have no runtime impact on BFD until that integration happens.

For now the acceptance criteria should be that the unit tests pass.

### Feedback Requested

Nothing specific.

### External References

- [DCGEO-19](https://jira.cms.gov/projects/DCGEO/issues/DCGEO-19)

### Security Implications

There should not be any security implications.

- The gRPC service targeted by this PR does not contain any sensitive data.
- The data from the service is discarded without being saved. 
- The code is not integrated into PipelineManager so it will not be invoked. 

Adds dependencies required for a gRPC client.  These are the standard gRPC related java libraries.

````
        <dependency>
          <groupId>io.grpc</groupId>
          <artifactId>grpc-netty-shaded</artifactId>
          <version>1.36.0</version>
        </dependency>
        <dependency>
          <groupId>io.grpc</groupId>
          <artifactId>grpc-protobuf</artifactId>
          <version>1.36.0</version>
        </dependency>
        <dependency>
          <groupId>io.grpc</groupId>
          <artifactId>grpc-stub</artifactId>
          <version>1.36.0</version>
        </dependency>
````

Adds a maven plugin to generate gRPC service stubs and protocol buffer data object classes.

````
            <plugin>
              <groupId>org.xolstice.maven.plugins</groupId>
              <artifactId>protobuf-maven-plugin</artifactId>
              <version>0.6.1</version>
              <configuration>
                <protocArtifact>com.google.protobuf:protoc:3.12.0:exe:${os.detected.classifier}</protocArtifact>
                <pluginId>grpc-java</pluginId>
                <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.36.0:exe:${os.detected.classifier}</pluginArtifact>
              </configuration>
              <executions>
                <execution>
                  <goals>
                    <goal>compile</goal>
                    <goal>compile-custom</goal>
                  </goals>
                </execution>
              </executions>
            </plugin>
````

Adds a maven extension to detect the local os for use by the protobuf-maven-plugin.

````
        <extensions>
          <extension>
            <groupId>kr.motd.maven</groupId>
            <artifactId>os-maven-plugin</artifactId>
            <version>1.6.2</version>
          </extension>
        </extensions>
````

None

See above.

No.

No.

### CODE REVIEW

File: apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+      LOGGER.info("exiting loop after reaching max runtime");
+      return false;
+    }
+    return true;
+  }
+
+  private int submitBatchToSink(RDASink<PreAdjudicatedClaim> sink, List<PreAdjudicatedClaim> batch)
+      throws ProcessingException {
+    LOGGER.info("submitting batch to sink: size={}", batch.size());
+    int processed = sink.writeBatch(batch);
```
ericbuckley, line 133: would it be useful to measure inserted vs updated at this level or do you think we should log those details in the sink?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/HealthCheckStreamCaller.java
```diff
@@ -0,0 +1,36 @@
...
+import com.nava.health.v1.HealthCheckResponse;
+import com.nava.health.v1.HealthGrpc;
+import gov.cms.bfd.pipeline.rda.grpc.PreAdjudicatedClaim;
+import io.grpc.ManagedChannel;
+import java.time.Duration;
+import java.util.Iterator;
+import java.util.concurrent.TimeUnit;
+
+/** GrpcStreamCaller implementation that calls the RDA HealthCheck service. */
+public class HealthCheckStreamCaller implements GrpcStreamCaller<HealthCheckResponse> {
```
ericbuckley, line 13: Are you thinking we can use this periodically in our pipeline job or is this just for testing out the RDA's gRPC service?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java
```diff
@@ -0,0 +1,152 @@
...
+    this.clock = clock;
+    callsMeter = appMetrics.meter(CALLS_METER);
+    recordsReceivedMeter = appMetrics.meter(RECORDS_RECEIVED_METER);
+    recordsStoredMeter = appMetrics.meter(RECORDS_STORED_METER);
+    batchesMeter = appMetrics.meter(BATCHES_METER);
+  }
+
+  @Override
+  public int retrieveAndProcessObjects(
+      int maxToProcess, int maxPerBatch, Duration maxRunTime, RDASink<PreAdjudicatedClaim> sink)
```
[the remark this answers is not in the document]
brianburton: NULL maxRuntime would cause an exception.  0 maxToProcess would exit the loop immediately.  I'd rather not build in overrides for misconfiguration at the lowest levels.  We could validate the values here but the caller should already have done so and I'd prefer not to do the same validation in multiple places.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java
brianburton: I would prefer to do that in the sink since that is where that particular determination would be made.  Another possibility in the sink would be records rejected or ignored for some reason.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/HealthCheckStreamCaller.java
brianburton: This was to have a concrete gRPC implementation to code against to validate the framework in this sprint.  We certainly could use this code in the later sprints as well.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java
ericbuckley: Rejections are probably something we're going to want to complain loudly about and possibly even halt processing for; thats typically been the case for the existing RIF pipeline job, but we can cross that bridge when we start implementing the sink

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java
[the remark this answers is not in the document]
ericbuckley: Let me provide more context, if we did want to run the process indefinitely, what values would we pass and/or how would we configure the job?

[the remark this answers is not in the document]
brianburton: This would do in practice.

````java
maxToProcess = Integer.MAX_VALUE;
maxRuntime = Duration.ofMilliseconds(Long.MAX_VALUE);
````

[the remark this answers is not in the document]
Comment (karlmdavis) [reply apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java:76]: > This would do in practice.

Like I noted in the other PR, I think I'd recommend removing that functionality altogether. Operationally, it has bad characteristics if we actually use it, and if we're not using it, it's just extra code we don't need.

File: apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcStreamCaller.java [hunk 4]
```diff
@@ -0,0 +1,37 @@
...
+import java.time.Duration;
+import java.util.Iterator;
+
+/**
+ * Interface for objects that handle streaming service calls and map their results to
+ * PreAdjudicatedClaim objects.
+ *
+ * @param <T> Type of objects returned by the streaming service.
+ */
+public interface GrpcStreamCaller<T> {
```
karlmdavis, line 14: This is another one of those things I'll be watching to see how it works out. At the moment, it feels like an odd interface.

For one thing, it's not obvious to me why those two methods are so strongly related that their implementations would always be part of the same interface.

For another -- and this is just a gut feel thing -- it feels a little overly abstracted. Like, this is the kind of thing I'd regret having abstracted myself, a few months/years later. When things _can_ be concrete, they're almost always clearer and more maintainable if you head that route.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/HealthCheckStreamCaller.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Override
+  public Iterator<HealthCheckResponse> callService(Duration maxRunTime) throws Exception {
+    final HealthCheckRequest request = HealthCheckRequest.newBuilder().setService(service).build();
+    return stub.withDeadlineAfter(maxRunTime.toMillis(), TimeUnit.MILLISECONDS).watch(request);
+  }
+
+  @Override
+  public PreAdjudicatedClaim convertResultToClaim(HealthCheckResponse ignored) throws Exception {
+    return new PreAdjudicatedClaim();
```
karlmdavis, line 30: This looks pretty goofy (related to my concerns about the parent interface, in general, as I noted elsewhere).

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/HealthCheckStreamCaller.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public HealthCheckStreamCaller(String service, ManagedChannel channel) {
+    this.service = service;
+    stub = HealthGrpc.newBlockingStub(channel);
+  }
+
+  @Override
+  public Iterator<HealthCheckResponse> callService(Duration maxRunTime) throws Exception {
+    final HealthCheckRequest request = HealthCheckRequest.newBuilder().setService(service).build();
+    return stub.withDeadlineAfter(maxRunTime.toMillis(), TimeUnit.MILLISECONDS).watch(request);
```
karlmdavis, line 25: It's not clear to me that this is the right deadline. I'd probably add a comment here detailing what this call does, with a focus on how long it might block for, what events it will fire (if any), and stuff like that.

This will be many folks' first encounter with a gRPC service (it is for me!) and you should take the opportunity to educate them a bit.

karlmdavis (commented):
Looks great!

Just left some comments and design musings.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java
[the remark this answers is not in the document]
brianburton: When this was developed I saw a single thread being used in PipelineManager.  In that context maxRunTime was intended to allow the job to time slice over a single thread without hogging it.  In the past I have found a setting like maxToProcess can be useful when testing or to validate the service is working without risking full blown processing (quick pre-flight check).  They could both be useful in the context of an integration test as well.  It seems I'm the only one who values this small feature.

Once the new orchestration code (in time for this sprint?) is in place this code will need to be refactored anyway.  I could make any adjustments in the next PR which will call the actual RDA 0.1 MVP service.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRDASource.java
```diff
@@ -0,0 +1,157 @@
...
+  private final Meter recordsReceivedMeter;
+  private final Meter recordsStoredMeter;
+  private final Meter batchesMeter;
+  private ManagedChannel channel;
+
+  public GrpcRDASource(
+      Config config, GrpcStreamCaller.Factory<T> callerFactory, MetricRegistry appMetrics) {
+    this(
+        ManagedChannelBuilder.forAddress(config.host, config.port)
+            .usePlaintext()
```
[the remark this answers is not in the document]
brianburton: Yeah it was strictly a temporary thing.  Will not be present in the DCGEO-23 PR.

PR #533, opened 2021-04-16
[BFD-762] Updates Ansible versions

keithdadkins:
### Change Details
Ansible versions are defined throughout the BFD project in a variety of places. Some of which are likely unused or only used locally by single engineers. However, since these versions frequently come up in our synk or dependabot scans, this work simply updates all the versions I could find where it made sense to update.

Note: Because pip is broken on FIPS enabled RHEL7 systems, we are currently using the ansible available in the default yum packages, which is 2.9.18. This is the version used on our build servers in prod, so it makes sense to try to pin the same versions.

Also, I did not update a few "requirements.freeze.txt" files as those should be artifacts from running 'pip freeze' (ideally we would not be using this anywhere). But, if they come up in a security scan, we can either remove/add to gitignore, or update the versions.

### External References

- [BFD-762](https://jira.cms.gov/browse/BFD-762)

### Security Implications


jzulim (approved):
Looks good to me. Approved. My only question would be do you feel we might want to consider a test AMI build (i.e. deployment to TEST env) to dbl check the updates don't break any jobs?

rtravitz (approved):

PR #531, opened 2021-04-15
[BFD-759] Add New Relic Metric API key and other configuration in all ENVs

rtravitz:
### Change Details
This PR addresses [BFD-759](https://jira.cms.gov/browse/BFD-759). It adds a secret key that enables us to send telemetry to the New Relic Metric API. It also adds a few public environment variables that same outcome.

### Acceptance Validation
- The key and other environment variables are added to the running server.

### Feedback Requested
- Did I get it all wired up correctly?
- Can you see the newly added key in the vault?

### External References

- [BFD-759](https://jira.cms.gov/browse/BFD-759)

### Security Implications
- [ ] new software dependencies
- [ ] altered security controls
- [ ] new data stored or transmitted
- [ ] security checklist is completed for this change
- [ ] requires more information or team discussion to evaluate security implications

PR #536, opened 2021-04-19
[BFD-760] Send pipeline app Dropwizard metrics to New Relic

rtravitz:
### Change Details
This PR addresses [BFD-760](https://jira.cms.gov/browse/BFD-760). It sends Dropwizard metrics for the Pipeline app to New Relic. It is highly similar to the changes in #524 and #531, which sent metrics for the Server app. Portions of the writeups for those two pull requests also apply here.

### Acceptance Validation
- When this application is deployed to TEST, metrics should start to flow to New Relic with an `appName` of `BFD Pipeline (test)`. 

### Feedback Requested
- It's unclear to me whether a parent `pom.xml` at the `apps` directory level is able to list the new relic reporter dependencies for both the server and the pipeline apps. When I tried to include it there, I got build failures and so added it in the respective child poms. Should I be able to hoist that dependency up to the parent? Is that even desirable?

### External References
- [BFD-760](https://jira.cms.gov/browse/BFD-760)

### Security Implications

Metrics for the pipeline app will now be sent to New Relic. They were previously only sent to Splunk as logs. There is no PII or PHI in these metrics—they are things like timers and counters of how long operations take and how often they occur.


jzulim (approved):
LGTM. Deployed to TEST without issue and seeing Pipeline metrics in NR. Nice work Ryan!

karlmdavis:
@rtravitz Sorry I didn't see this sooner, but just noticed the code in `PipelineApplication` that's calling `System.getenv(...)` directly. Any chance we can get that moved into `AppConfiguration`, so that we keep all of that code in one place? Makes it easier to track what's actually configurable, and to mock that configuration in our tests.

PR #539, opened 2021-04-20
BFD-704: Made BFD Pipeline orchestration more flexible

karlmdavis:
### Change Details

Everything is now explicitly part of a "job", and scheduled as such. This will make it much simpler for new jobs to be added, e.g. those related to DC Geo and Synthea.

Notes:

1. To figure out what all is going on here, I'd suggest looking at `PipelineApplication`, `PipelineManager`, and `PipelineManagerIT`, in that order.
2. This design is intended to enable an easy-ish transition towards running jobs on more than one node, with state synchronized via a DB. The design choices reflect that, though I do think that they're all reasonable choices even if that enhancement never happens.
3. I've left some `TODO`s in the code on purpose. I may or may not address those prior to final approval -- haven't decided if they're important enough yet.
4. I removed `RifLoaderIdleTasks` and friends as it turns out that code was quite broken. It apparently only worked accidentally; it was not working as intended, at all. Since we no longer need those jobs, it wasn't worth trying to refactor and fix all of it.

### Acceptance Validation

1. This is concurrent orchestration code, and as such, inherently complex. However, I tried hard to minimize and manage that complexity as much as possible. Did I succeed well enough?
2. Is the code correct?
3. Is it reasonably obvious how to create a new job and ensure that it gets run?

### Feedback Requested

(see above)

### External References

### Security Implications

This PR has no identified security implications.


apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
```diff
@@ -72,101 +70,31 @@ public static void main(String[] args) {
...
     /*
-     * Create the services that will be used to handle each stage in the
-     * extract, transform, and load process.
+     * Create the PipelineManager that will be responsible for running and managing the various
+     * jobs.
      */
-    RifFilesProcessor rifProcessor = new RifFilesProcessor();
-    RifLoader rifLoader = new RifLoader(appMetrics, appConfig.getLoadOptions());
+    PipelineJobRecordStore jobRecordStore = new PipelineJobRecordStore();
+    PipelineManager pipelineManager = new PipelineManager(appMetrics, jobRecordStore);
```
mackec1, line 77: Small nit; personally I like to formalize Singleton class by making a private constructor and using a static public getInstance(). For example:

private static PipelineManager myPipelineManager = null;

public static getInstance(MetricRegistry appMetrics, PipelineJobRecordStore jobRecordStore) {
    if (null == myPipelineManager) {
      myPipelineManager = new PipelineManager(appMetrics, jobRecordStore);
   }
  return myPipelineManager;
}

Never have to worry about someone trying to do something they shouldn't!

karlmdavis: Doesn't that make things more challenging to unit test? I generally have tests that construct things in all sorts of different ways, with various mocked dependencies. Can't do that with a private constructor.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -32,209 +65,403 @@
...
+    // Submit the job to be run!
+    ListenableFuture<PipelineJobOutcome> jobFuture;
+    try {
+      jobFuture = jobExecutor.submit(jobWrapper);
+    } catch (RejectedExecutionException e) {
+      // Indicates that the executor has been shutdown.
+      return false;
     }
+    jobActiveHandles.put(jobRecord.getId(), new PipelineJobHandle<>(jobWrapper, jobFuture));
+    jobsRunningCounter.incrementAndGet();
```
Dranith, line 242: Interesting thought: Does it make sense to even have a separate `AtomicInteger` here?  Since you are using a `ConcurrentHashMap` from reading `size()` it seems to have the same guarentees as `AtomicInteger` does and might actually provide more accuracy.

For example in this case, say you perform the `put` and another thread reads `jobRunningCounter`.  That will result in `jobsRunningCounter` being less than the number running but `jobActiveHandles.size()` will be correct.  

Now you have a written contract above in documentation that says this can only be incremented or read from one thread, so this works .. but it might still be simpler to just use `.size()`

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadJob.java
```diff
@@ -124,12 +127,28 @@ public CcwRifLoadJob(
...
+  @Override
+  public Optional<PipelineJobSchedule> getSchedule() {
+    return Optional.of(new PipelineJobSchedule(1, ChronoUnit.SECONDS));
+  }
+
+  /** @see gov.cms.bfd.pipeline.sharedutils.PipelineJob#isInterruptible() */
+  @Override
+  public boolean isInterruptible() {
+    /*
+     * TODO This would be a good enhancement: making this class properly interruptible. This would
```
Dranith, line 146: Agree on the TODO, but how would you know where you stopped / are picking up from?  Are RIF file loads idempotent?

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
```diff
@@ -121,101 +119,31 @@ public static void main(String[] args) {
...
-          }
-
-          /** Called when no RIF files are available to process. */
-          @Override
-          public void noDataAvailable() {
-            rifLoader.doIdleTask();
-            DataSetMonitorListener.super.noDataAvailable();
-          }
-        };
+    pipelineManager.registerJob(new DatabaseSchemaUpdateJob(appConfig.getDatabaseOptions()));
```
Dranith, line 134: Just thinking out loud, but this is the sort of thing that the Java Service Provider Interfaces are designed for.  That would allow new jobs to be discovered and picked up by just dropping them on the classpath for the `PipelineApplication`.

The code here definitely a more explicit approach, and I can see the value of that though but it feels like a "plugin" architecture could make sense here.

Dranith (approved):
Looks good.  I just really had a couple thoughts/questions about certain parts.  Nothing major though.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
karlmdavis: I think heading down the SPI/plugin route would _also_ require a dependency injection provider, e.g. Spring Beans or CDI or Guice, right? I dunno'; I should be in bed, so perhaps I'm wrong.

If so, though, I've very intentionally avoided that here. I used to be a big Guice/Dagger stan but these days find that they add way more complexity than they remove.

Even aside from that, given the small number of jobs we have (and are likely to have for the next couple years), I think I prefer how easy it is to figure out from what's going on just by looking at `PipelineApplication`. That would definitely need to be re-evaluated if the number of jobs grown past some point, e.g. towards 10.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadJob.java
karlmdavis: Very idempotent.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
karlmdavis: That's potentially a good catch. I added `jobActiveHandles` very late in development and didn't consider whether or not it could replace `jobsRunningCounter`. I'll ponder that a bit.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -32,209 +65,403 @@
...
+   *
+   * <p>Following these rules will result in this value being eventually consistent but nevertheless
+   * ALWAYS >= the number of jobs that are actually running, from the view of the {@link
+   * SchedulerJob}'s thread. This aligns with what the {@link SchedulerJob} needs the value for, as
+   * we're using it to avoid over-subscribing on work. Under-subscribing is fine, in the context.
+   *
+   * @return the counter for how many tasks are currently running on the {@link PipelineJob}s {@link
+   *     Executor}
+   */
+  private final AtomicInteger jobsRunningCounter;
```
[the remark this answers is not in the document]
karlmdavis: Actually, it was a bit incorrect: `s/SchedulerJob/VolunteerJob`.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -24,6 +48,15 @@
...
       "Starting data set monitor: watching for data sets to process...";
 
+  /**
+   * The number of jobs that can be run at one time. Because the {@link VolunteerJob} and {@link
+   * SchedulerJob} will always be running, this number must be >=3, in order for any actual jobs to
+   * get run.
+   *
+   * @see #jobExecutor
+   */
+  public static final int JOB_EXECUTOR_THREADS = 3;
```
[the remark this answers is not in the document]
karlmdavis: I think we'll eventually hit a point where we want this to be dynamic, but right now, we only have the one "real" job (for the CCW load), plus `VolunteerJob` and `CronTriggerJob`. I'd hold off on making it dynamic until that count goes up a bit.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/DefaultDataSetMonitorListener.java
```diff
@@ -0,0 +1,119 @@
...
+    this.rifProcessor = rifProcessor;
+    this.rifLoader = rifLoader;
+  }
+
+  /**
+   * @see
+   *     gov.cms.bfd.pipeline.ccw.rif.extract.s3.DataSetMonitorListener#dataAvailable(gov.cms.bfd.model.rif.RifFilesEvent)
+   */
+  @Override
+  public void dataAvailable(RifFilesEvent rifFilesEvent) {
```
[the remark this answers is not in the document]
karlmdavis: I honestly can't recall if S3 offers that particular guarantee explicitly, but at a minimum, it's Java API must, as we haven't had any errors related to that in... 3+ years.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
karlmdavis: Yeah, I definitely ought to `s/must/should` on that one. More of an architecture/design note.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -288,18 +275,18 @@ static AppConfiguration readConfigFromEnvironmentVariables() {
...
             hicnHashPepper,
-            databaseUrl,
-            databaseUsername,
-            databasePassword.toCharArray(),
             loaderThreads,
-            idempotencyRequired.get().booleanValue(),
-            fixupsEnabled,
-            fixupThreads));
+            idempotencyRequired.get().booleanValue());
+    CcwRifLoadOptions ccwRifLoadOptions = new CcwRifLoadOptions(extractionOptions, loadOptions);
```
brianburton, line 288: Loading all of the configuration variables here creates a very tight dependency between the app package and the job package implementations.  It would be nice if we could delegate configuration loading to the individual jobs rather than centralize it here.  Perhaps by passing a BiFunction<String,String,String> or some equivalent interface of our own to their constructor to abstract config setting lookup so we could choose between environment variables or java properties later without changing the submodule configs.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -32,209 +65,403 @@
...
     this.appMetrics = appMetrics;
-    this.options = options;
-    this.scanRepeatDelay = scanRepeatDelay;
-    this.listener = listener;
+    this.jobRecordStore = jobRecordStore;
+    this.jobExecutor = createJobExecutor();
+    this.jobActiveHandles = new ConcurrentHashMap<>();
+    this.jobsRunningCounter = new AtomicInteger(0);
+    this.jobMonitorsExecutor = Executors.newCachedThreadPool();
+    this.jobRegistry = new HashMap<>();
```
brianburton, line 142: jobRegistry needs to be ConcurrentHashMap since we are using stream() in various methods.  Raw HashMap will throw ConcurrentModificationException if anything is added to or removed from the map during iteration/streaming over it.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/jobs/store/PipelineJobRecordStore.java
```diff
@@ -0,0 +1,206 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(PipelineJobRecordStore.class);
+
+  /** The number of milliseconds to wait between polling job dependencies' status. */
+  private static final int JOB_DEPENDENCY_POLL_MILLIS = 100;
+
+  private final Map<PipelineJobRecordId, PipelineJobRecord<?>> jobRecords;
+
+  /** Constructs a new {@link PipelineJobRecordStore} instance. */
+  public PipelineJobRecordStore() {
+    this.jobRecords = new HashMap<>();
```
brianburton, line 34: jobRecords needs to be ConcurrentHashMap since we are using stream() in various methods.  Raw HashMap will throw ConcurrentModificationException if anything is added to or removed from the map during iteration/streaming over it.

brianburton (commented):
Looks good.  Much more flexible than the old implementation and should work for DC-Geo's needs. I would like to see the ConcurrentModificationException issues addressed though.

mackec1 (approved):
Ran IT stuff under debugger to see/understand the underpinnings in action. Lots of good work here.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
Dranith: No Java SPI doesn't require CDI.  It is in the base spec.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
karlmdavis: Yup: it was safe to do so! And in looking through things to make that change, I caught a race condition in `PipelineManager.stop()` that I'd missed, so thanks!

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
karlmdavis: Yeah, I was thinking about that myself; it's a good insight. I opted to not address it for now, but I expect it'll be one of those things we come back to in a future PR.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/jobs/store/PipelineJobRecordStore.java
karlmdavis: My brain's a bit fried after the morning's work, so I'll need a break before I try to reason through this all the way, but at first glance, I think you're right. Thanks!

karlmdavis: Yup: you were correct. Fixed. Thanks!

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
karlmdavis: I'm going to defer this, at least for now. At the moment, I'm more in favor of less magic (heh, wordplay). It's a good suggestion, though, and one worth considering in the future, particularly if we end up with large numbers of jobs.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
brianburton: I agree that's the case now and likely to remain the case in the future.  But someone less familiar with the code and unaware of the distinction could come along later and add a call to registerJob() or make some other change without knowing the risk.

brianburton (approved):
My comments from previous review have been addressed.  Looks like a good foundation for pipeline job management going forward.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
karlmdavis: I understand the concern, but I think it's low risk, just due to how the application works: hard to see why someone would lazy-/late-register jobs, rather than doing so in `PipelineApplication`.

Regardless, though, if concurrency is a concern, then just making the collection concurrent likely wouldn't be enough. You can end up with all sorts of goofy race conditions, even with a "thread safe" collection (as evidenced by the recent commits here). I think, if we want to address it, the right fix is to ensure that the registration API isn't visble to anything other than `PipelineApplication` -- probably by moving jobs into the constructor. Which... is a good idea, actually, and something we should consider doing in the future. This whole thing just went through so many iterations that there are still some goofy bits like that hanging around.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/volunteer/VolunteerJob.java
```diff
@@ -0,0 +1,112 @@
...
+    while (true) {
+      try (Timer.Context timer =
+          appMetrics
+              .timer(MetricRegistry.name(getClass().getSimpleName(), "call", "iteration"))
+              .time()) {
+        /*
+         * We want to submit up to as many jobs for execution as we can actually work, each
+         * iteration of this loop.
+         */
+        int executorSlotsOpen = pipelineManager.getOpenExecutorSlots();
```
mackec1, line 82: Just curious how much thrashing might occur here; if <= 0 it immediately goes back to loop.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/jobs/store/PipelineJobRecord.java
```diff
@@ -0,0 +1,286 @@
...
+
+  /**
+   * "Why is this variable {@code volatile}," you might ask? Good question! The short answer is the
+   * unhelpful, "because things can crash unless it is." The <em>better</em> answer is complicated:
+   * When the VolunteerJob is bootstrapped, it's enqueued by PipelineManager's constructor, and ends
+   * up setting this field -- on the application's main thread. Then, VolunteerJob starts running,
+   * and goes looking for non-enqueued jobs. Unless this field is volatile, it will sometimes "find
+   * itself" (lol), thinking that it's not enqueued, because it's reading a stale value of it. That
+   * causes things to go boom. So: we mark this as volatile, and everything's copacetic. Isn't
+   * concurrency fun?!
```
mackec1, line 51: Love the comment!

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/volunteer/VolunteerJob.java
karlmdavis: What do you mean by thrashing? It will definitely spend a lot of time waiting for something to do, but that's on purpose. If we ever move to running jobs across multiple nodes, oversubscribing becomes bad, as you could have a busy node "squatting" on jobs that it doesn't have capacity to work (and not letting any other nodes work them, either).

karlmdavis (commented):
Saving notes from review meeting.

rtravitz (approved):
Not much additional to add at this point, beyond questions asked/addressed in the review meeting. After pulling it down and tracing through, the architecture makes sense to me, and it's probably time to get it out there and battle-tested.

JeremyIR (approved):
Echoing what Ryan said with regards to last meeting. Looks good.

rtravitz (dismissed):
Re-approving after fixed conflicts.

brianburton (approved):
Looks good!

rtravitz (approved):

PR #541, opened 2021-04-21
BFD-704: Upgraded HSQL and Flyway

karlmdavis:
### Change Details

Updated these two dependencies. Not for any particular reason, it turns out -- I was hoping this would fix a problem, but it didn't. Still a good idea to bump them, though.

### Acceptance Validation

1. Does the build still pass?

### Feedback Requested

1. What's your favorite breakfast food?

### External References

### Security Implications

This change doesn't introduce any new dependencies, but does bump our versions of some. There are no identified (negative) security implications to that change.


mackec1 (approved):
LGTM

rtravitz (approved):
@karlmdavis Pumped on this change. Flyway clean fails in Postgres 11, which means we haven't been able to run the same database version locally as in production, because the integration tests fail. Hoping this fixes that without me having to write the ticket I was about to put together.

PR #545, opened 2021-04-22
BFD-704: Try to fix intermittent bfd-server SocketException test failures

karlmdavis:
### Change Details

This one is a bit harder to debug, just because it's less frequent, but I think I may have tracked down the issue. I think it's similar to the other `SocketException` errors that were just fixed in `bfd-pipeline`: the server was binding to a different host/interface than we were trying to connect to in our tests.

### Acceptance Validation

1. Does this look correct to you?
2. Is the build passing?

### Feedback Requested

1. What's your spirit animal?

### External References

### Security Implications

This change has no known security implications.


mackec1 (approved):

apps/bfd-server/bfd-server-launcher/src/main/java/gov/cms/bfd/server/launcher/AppConfiguration.java
```diff
@@ -8,7 +8,13 @@
 
 /** Models the configuration options for the launcher. */
 public final class AppConfiguration implements Serializable {
-  private static final long serialVersionUID = -8134110339292440841L;
+  private static final long serialVersionUID = 1L;
```
rtravitz, line 11: What's the reason for this change?

rtravitz (approved):
Ran locally okay 👍

mackec1 (approved):

mackec1 (approved):

cbrunefearless (approved):
lgtm

apps/bfd-server/bfd-server-launcher/src/main/java/gov/cms/bfd/server/launcher/AppConfiguration.java
karlmdavis: Good question! Having computed serialization IDs provides a false sense of security, since nothing verifies them. (We also only use serialization for the test cases, so it doesn't matter, either way.)

PR #547, opened 2021-04-22
[BFD-619] V2 (FHIR R4) Integration Tests - DO NOT MERGE!

Dranith:
### Change Details

This work adds integration tests for the R4 transformers and V2 endpoints that mimic the coverage provided by STU3.  The coverage should be similar, but the validations rely on more deep comparisons of the transformed vs API returned data structures.  Specifically this PR has added the following integration test classes:

* R4CoverageResourceProviderIT.java 
* R4ExplanationOfBenefitResourceProviderIT.java 
* R4PatientResourceProviderIT.java
* EndpointJsonResponseComparatorV2IT.java

For `EndpointJsonResponseComparator*` the gold copy files were moved into `v1` and `v2` directories to keep the two separate.  I also ran into a lot of issues with comparing EOBs returned and serialized from the endpoint vs the transformer.  This came down to trivial differences like the endpoint based serialization returning a `Quantity` and the transformer returning a `SimpleQuantity` or a number being represented as `75.0` vs `75`.  For this reason, `equalsDeep` couldn't directly be used.  This is why the EOB comparison seems messy, but provides good coverage of the fields.  There is documentation in the code that explains this process as well.

Additionally, there were several issues with the V2 endpoints that were discovered during the course of working this ticket:

* A number of claims were adding a `Diagnosis` element without a Label.  This was defaulted to `OTHER`.
* `R4SamhsaMatcher` was only dealing with PDE, causing other types to throw errors
* SNF claims were being mapped with a type of HOSPICE
* The contained `Observation` generated for some EOB types was incorrectly adding `#` to the front of the ID in the actual resource.  This should only be part of the local reference.  See: https://build.fhir.org/references.html#contained

### Acceptance Validation

All the integration tests  (V1 and V2) pass locally when run with a full clean build.

image

### External References

- [BFD-619](https://jira.cms.gov/browse/BFD-619)

### Security Implications


apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -0,0 +1,1758 @@
...
+
+  /**
+   * Verifies that {@link
+   * gov.cms.bfd.server.war.r4.providers.ExplanationOfBenefitResourceProvider#findByPatient(ca.uhn.fhir.rest.param.ReferenceParam)}
+   * handles the {@link ExplanationOfBenefitResourceProvider#HEADER_NAME_INCLUDE_TAX_NUMBERS} header
+   * properly.
+   *
+   * @throws FHIRException (indicates test failure)
+   */
+  // TODO: Fix this test .. it isn't working. Tax number is not showing up.
```
Dranith, line 1055: Someone else will need to look at this.  I couldn't figure out why TAX Number wasn't being handled here.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -0,0 +1,1769 @@
...
+      Resource actualContained = actual.getContained().get(i);
+
+      expectedContained.getMeta().setLastUpdated(null);
+      actualContained.getMeta().setLastUpdated(null);
+
+      // TODO: HAPI seems to be inserting the `#` in the ID of the contained element.
+      // This is incorrect according to the FHIR spec:
+      // https://build.fhir.org/references.html#contained
+      // This works around that problem
+      Assert.assertEquals("#" + expectedContained.getId(), actualContained.getId());
```
Dranith, line 1547: The FHIR spec suggests that the actual contained resource should *not* have an ID that starts with `#`.  Our transformers do not add this.  It appears that HAPI is adding it when you go through the actual read endpoint.  To get around this in testing, we prefix the expected ID with `#` as well.  This is not correct as far as I can tell and may be a bug with HAPI.

Reference: https://build.fhir.org/references.html#contained

mackec1 (approved):
Whoa....that was a yeoman's effort on this one. As usual, code was organized and well written. Thanks much for setting development standards that will hopefully continue on for some time.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
cbrunefearless: done

cbrunefearless (changes requested):
In V1 it is other, should it be other operating in v2? http://hl7.org/fhir/us/carin-bb/STU1/CodeSystem-C4BBClaimCareTeamRole.html

PR #549, opened 2021-04-23
[BFD-724] Send custom request headers to New Relic

rtravitz:
### Change Details
This PR addresses [BFD-724](https://jira.cms.gov/browse/BFD-724). It configures the New Relic agent to send custom http request headers with each transaction. The list of initial custom headers was taken from the [list of documented headers](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/docs/request-audit-headers.md) that BFD consumers are asked to provide with each request.

Unfortunately, there doesn't seem to be a switch in New Relic that says "send all request headers," so we have to specify them individually. This has been possible in the Java agent either through direct calls to its API in the app code or since the 5.9 release in the `newrelic.yml` [configuration file](https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-config-file-template/). The docs for the custom headers section of the [configuration file are here](https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file/#tt-custom-request-headers).

To this point we have only configured the New Relic Agent with environment variables. While there does not appear to be a way to specify these custom headers to include with environment variables, the configuration file and the env vars can live in harmony. According to the docs, ENV vars [override anything found in the configuration file](https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file/#config-options-precedence) if an option is specified in both places, so we can continue mostly configuring the agent as we were before. That said, it may be cleaner to specify all of our options in the `newrelic.yml` at some point instead of sometimes having it in env vars.

### Acceptance Validation
- Custom headers appear in New Relic after deploying to TEST and hitting the server with a request and the appropriate headers.

### Feedback Requested
- Is this ansible going to get the file in the right place?
- It seemed like people might have thought there was a "send every custom header" switch to flip. Do you know something I couldn't find?

### External References
- [BFD-724](https://jira.cms.gov/browse/BFD-724)

### Security Implications
This change would send some headers that were previously only stored in Splunk to New Relic.


ops/ansible/roles/bfd-server/templates/newrelic.yml.j2
```diff
@@ -0,0 +1,18 @@
+# This file configures the New Relic Agent.  New Relic monitors
+# Java applications with deep visibility and low overhead.  For more details and additional
+# configuration options visit https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file.
+
+transaction_events:
```
jzulim, line 5: We'll want to add the async headers as well, namely the bulk headers:

```
* `BULK-CLIENTID`: a unique identifier of the bulk client for whom this request is for (i.e. UUID, ACO ID, NPI, Part D Contract)
* `BULK-JOBID`: a unique identifier for the job (Job ID, UUID, etc.)
```

jzulim (changes requested):
Provided a few more headers.

ops/ansible/roles/bfd-server/templates/newrelic.yml.j2
```diff
@@ -0,0 +1,18 @@
...
+    - header_name: "BlueButton-OriginalQueryCounter"
+    - header_name: "BlueButton-OriginalQueryTimestamp"
+    - header_name: "BlueButton-DeveloperId"
+    - header_name: "BlueButton-Developer"
+    - header_name: "BlueButton-ApplicationId"
+    - header_name: "BlueButton-Application"
+    - header_name: "BlueButton-UserId"
+    - header_name: "BlueButton-User"
+    - header_name: "BlueButton-BeneficiaryId"
+    - header_name: "x-forwarded-for"
```
keithdadkins, line 18: Does this get merged into another config file or something? From a quick googling this appears to be a partial config. (Sorry, not yet familiar with how our new relic agent is configured)

ops/ansible/roles/bfd-server/tasks/install.yml
```diff
@@ -71,6 +71,16 @@
...
   become: true
 
+- name: Copy New Relic Java Agent Config
+  copy:
+    src: 'newrelic.yml.j2'
+    dest: "{{ data_server_dir }}/newrelic/newrelic.yml"
+    remote_src: true
+    owner: "{{ data_server_user }}"
+    group: "{{ data_server_user }}"
+    mode: u=rw,g=r,o=r
```
keithdadkins, line 81: If this file will contain any api keys or what not, I would remove `o=r`.

keithdadkins (dismissed):

ops/ansible/roles/bfd-server/templates/newrelic.yml.j2
rtravitz: It's definitely a partial file. Which I'm  hoping is fine, because the [full template is quite long](https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-config-file-template/). A lot of those fields are optional and/or have defaults that we don't override with the current environment variables. And those environment variables override values that we _would_ put in those files.

My plan was to deploy this to test before going anywhere with it to help answer that question. If we need the whole file, then I'll fix up the whole file and stop configuring it in a split-fashion between this and env vars.

ops/ansible/roles/bfd-server/tasks/install.yml
rtravitz: Right now intending it to not include the API key, but will take this out to be safe.

ops/ansible/roles/bfd-server/templates/newrelic.yml.j2
rtravitz: Added!

keithdadkins (dismissed):

ops/ansible/roles/bfd-server/tasks/install.yml
```diff
@@ -71,6 +71,15 @@
     mode: u=rw,g=r,o=r
   become: true
 
+- name: Copy New Relic Java Agent Config
+  copy:
```
keithdadkins, line 75: change this to template:

keithdadkins (changes requested):

keithdadkins (dismissed):

rtravitz:
PR is currently blocked because these headers won't send in New Relic High Security Mode which disallows custom attributes.

rtravitz:
Custom headers are now ending up in New Relic. I deployed to TEST and made a request for that specific host. Check out the attached screenshot. 💍🌋 
new-relic-custom-headers

jzulim (approved):

keithdadkins (approved):
LGTM

PR #550, opened 2021-04-25
[BFD-725] Infra to enable custom RDS reader endpoints

keithdadkins:
### Change Details
This work adds a "beta_reader" feature flag to the (terraform) `aurora` module that, when enabled, provisions two additional [custom read only endpoints](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Overview.Endpoints.html#Aurora.Endpoints.Custom). 

The purpose is to isolate one reader node from the rest of the reader nodes so that engineers can perform tasks like testing new features, debugging queries, running load tests, ML training, etc., all without impacting normal traffic and all without incurring the costs of cloning an entire cluster/environment.

Notes:
- The feature is disabled by default.
- Before deploying to production environments, be sure you have added sufficient nodes to the cluster to account for this feature (e.g. add at least one node to the cluster.. see runbook).
- The last reader reader node in the cluster is dynamically isolated and added to the `bfd-${ var.env_config.env }-beta-reader` endpoint. 
- All other readers are assigned to the second `bfd-${ var.env_config.env }-ro` endpoint.
- This does not delete any other endpoints.

An example workflow for prod would look like:
1. Add a reader node to the cluster (see runbook).
2. Set the `beta_reader` feature flag in `ops/terraform/env/prod/stateful/main.tf` to true.
3. Apply the terraform and take note of the two endpoints.
4. Update the database config in `ops/ansible/playbooks-ccs/vars/prod/group_vars/all/vault.yml` to point to the custom `bfd-prod-ro` endpoint and deploy.
5. You can now connect to the `bfd-prod-beta-reader` endpoint from your application, postgres client, etc.

### External References

- [BFD-725](https://jira.cms.gov/browse/BFD-725)

### Security Implications


mackec1 (approved):
Keith; is there a README or such mechanism to request the custom RDS endpoint be activated? And once activated, are there policies/procedures to use, tear down,e tc?

keithdadkins:
There are no policies but I will create a run book documenting the process for how to setup/tear down the isolated reader in prod when needed.

jzulim (approved):
This should set things up nicely. Nice work!

PR #556, opened 2021-04-26
[BFD-746] Add Unit Tests to PDE EOB Profile

JeremyIR:
### Change Details

Modify PDE v2 transform test code as an expansion similar to Outpatient claim transformer tests, so that there is more confidence individual fields and values are working as intended.

### Acceptance Validation

- Confirm that the same level of and unit tests for BFD V2 are applied to V2
- Ensure that all required CARIN data fields have corresponding unit tests for V2

### Feedback Requested

Do the updated set of tests for PDE v2 Transformer account for the CARIN fields in the same way Outpatient tests do?

### External References

- [BFD-746](https://jira.cms.gov/browse/BFD-746)

### Security Implications


rtravitz (commented):
I think you can get rid of the `assertMatches` function now that you're replacing it with all of these other tests. Other than that, the tests look good to me.

rtravitz (approved):

mackec1 (approved):
These can be mind-numbing, so thanks for digging...looks good.

PR #557, opened 2021-04-27
BFD-753 - Fix Performance for AB2D Patient Endpoint for Beneficiary Monthly - Drop and add new index

cbrunefearless:
### Change Details
Fix Performance for AB2D Patient Endpoint for Beneficiary Monthly - Drop and add new index
User Story:

As AB2D, we want to fix the patient endpoint, so that we can receive patient data more quickly and with less impact on the BFD.
AC:

Construct Explain plan for before and after fixing the AB2D patient endpoint when reference year is supplied
Add index (with concurrent syntax) via ETL pipeline / data migration
Test index in lower environments 
Add many records to test data 
Note - This must be deployed alone at off-peak hours (after EOD)

### Acceptance Validation

Look at the old index and make sure the drop statement is correct.  Look at the order of the new index and make sure it looks correct

### Feedback Requested
Check the index for previous index is dropped correctly and sql syntax to create a new index

Look at the old index and make sure the drop statement is correct.  Look at the order of the new index and make sure it looks correct
### External References

- [BFD-753](https://jira.cms.gov/browse/BFD-753)

### Security Implications

N/a

N/a

N/a

N/a


mackec1 (approved):

jzulim (approved):
LGTM (as we've discussed this change previously)

PR #563, opened 2021-04-28
DCGEO-39: initial RDA API 0.1 FISS claims schema

brianburton:
### Change Details

This PR adds a single Flyway migration file for issue [DCGEO-39](https://jira.cms.gov/browse/DCGEO-39).  The migration file adds two new tables and one explicit index.  Primary key indexes would also be created of course.  The tables are designed to mirror store the fields that we receive in the [0.1 MVP version of the RDA API](https://confluence.cms.gov/pages/viewpage.action?pageId=448007957).

There are some open questions that this PR leaves to be answered.  Either during review/discussion of this PR or in a subsequent one.  These questions are:

- Do we want to store the RDA API data in a separate database?  Our preference would be to use the existing database to avoid extra complexity down the line.
- Do we want to store the RDA API data in a separate schema or just add prefixes to the names of the tables?  Adding a schema would nicely separate the RDA and RIF data sets without adding too much complexity.  For now we went with a prefix "PreAdj" and used the default schema while a determination is made.
- What tablespace(s) should be used for the tables?  Do we need new tablespaces or just use the default?  For now we use the default and have commented out the tablespace declarations.
- Where will the migration file live?  In the current module or in a new one specific to RDA?  For now we just added it to `bfd-model-rif/src/main/resources/db/migration` while a determination is made.

### Acceptance Validation

Validate that:

- SQL is correct syntactically
- SQL follows appropriate conventions

### Feedback Requested

Discussion and a plan for when/how the open questions will be resolved.  We could merge this as is and apply changes in subsequent sprints as the questions are answered.

### External References

- [DCGEO-39](https://jira.cms.gov/browse/DCGEO-39)
- [0.1 MVP version of the RDA API](https://confluence.cms.gov/pages/viewpage.action?pageId=448007957)

### Security Implications

The new table includes some PII including MBI.

No

No

Yes.  Various FISS claim fields are being stored including MBI.  See the migration file for list of fields or refer to the [RDA data dictionary](https://confluenceent.cms.gov/display/MPSM/MVP+0.1+Data+Dictionary+for+Replicated+Data+Access+API).

No


karlmdavis:
Just wanted to chime in on a couple of your questions:

> Do we want to store the RDA API data in a separate database? Our preference would be to use the existing database to avoid extra complexity down the line. ... Do we want to store the RDA API data in a separate schema or just add prefixes to the names of the tables? Adding a schema would nicely separate the RDA and RIF data sets without adding too much complexity. For now we went with a prefix "PreAdj" and used the default schema while a determination is made.

Definitely not a separate DB, for the reasons @ericbuckley outlined. I'd recommend a separate schema, myself. As for naming, I'd go with a separate `pre_adj` schema, though I hope we can get a real good [bike shed](https://en.wikipedia.org/wiki/Law_of_triviality) going on this one!

> What tablespace(s) should be used for the tables? Do we need new tablespaces or just use the default? For now we use the default and have commented out the tablespace declarations.

Stick with the default tablespace. We're using Aurora and there really isn't a good reason to use separate tablespaces with that: <https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.FeatureSupport.Tablespaces>.

> Where will the migration file live? In the current module or in a new one specific to RDA? For now we just added it to `bfd-model-rif/src/main/resources/db/migration` while a determination is made.
The DB migrations _have to_ all stay together for Flyway, so that's the correct location for now. As I think I noted in another PR, that location should eventually change, but it'll change for everything, so no big deal right now.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+${logic.tablespaces-escape} SET default_tablespace = fhirdb_ts2;
+
+create table "PreAdjFissClaims"
+(
+    "dcn"               varchar(23) not null,
+    "hicNo"             varchar(12) not null,
+    "currStatus"        char(1)     not null,
+    "currLoc1"          char(1)     not null,
+    "currLoc2"          varchar(5)  not null,
+    "medaProvId"        varchar(13),
```
karlmdavis, line 23: "meda"? What's that short for?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+
+create index "PreAdjFissClaims_mbi_hash_idx"
+    on "PreAdjFissClaims" ("mbiHash");
+
+/*
+ * Columns with leading underscore in the name are unique to our database.
+ * Other columns map directly to fields in the RDA API result message.
+ *
+ * The _priority column is a 0 based index indicating the order in which the codes
+ * appear in the RDA API result message for the claim.
```
karlmdavis, line 56: That's an interesting convention. I'm curious what other folks think of it?

Personally, aside from my deep inner desire to categorize **ALL THE THINGS**, I can't come up with a reason that it'd actually be helpful. Given that, I'd probably leave it out.

karlmdavis (commented):
Great work!

I left in a couple misc comments.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
ericbuckley: "Medicare", this is the Medicare Provider Number field

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
ericbuckley: When we return the procedure codes in the FHIR resource, I'm guessing the order in which we return them doesn't matter, so in that case it's not needed. FISS does make a distinction between the first procedure code (the primary) and all the others (2-25), so if for any reason we want to map back to the primary procedure code we'll need the `_priority` value.

karlmdavis: Oh darn it, looks like I highlighted the wrong rows.

Can we pretend I attached this note to the comment about the underscore prefix, instead?

(FWIW, I would preserve order as much as possible.)

ericbuckley: Oh I missed that comment.  @brianburton I thought the RDA API team was defining the `_priority` field?  I thought the only field we're generating are `mbiHash` and `lastUpdatedBy`.

brianburton: The RDA schema has an internal field `_priority` but it is not directly passed through the API.  I am basically reconstructing it on our end based on the order of the procedure codes in the API response data.  So it is a derived field on our side.

brianburton: @karlmdavis there is some benefit to distinguishing our internal column names from those in the RDA API in case we run into a conflict with new fields later.  Especially with a young, rapidly evolving API like the RDA.

That said I did not use the underscore for lastUpdated or mbiHash because I wanted to follow conventions used in BFD so as conventions go this PR already violates it.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
rtravitz: Is that prefix coming from the upstream? I'm surprised by that versus something like `medi`

rtravitz (commented):
Mostly just agreeing with the consensus:
* Same DB
* Separate schema
* Default tablespace
* Migration location seems fine

Might just remove that TODO at the bottom about the tablespace if the group answers that question convincingly.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
rtravitz: > That said I did not use the underscore for lastUpdated or mbiHash because I wanted to follow conventions used in BFD so as conventions go this PR already violates it.

The underscore convention gets more confusing to me because of this. Would make sense to me to remove it and be diligent about documenting which columns come directly from the RDA API.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
ericbuckley: Yeah its defined within FISS

karlmdavis: I'm glad I'm not the only one to have my silly typos preserved for perpetuity.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
brianburton: Agreed.  I'll drop it.

brianburton: Latest commit removes the underscore.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
```diff
@@ -0,0 +1,57 @@
...
+
+CREATE SCHEMA IF NOT EXISTS "pre_adj";
+
+/*
+ * Where possible column names map directly to fields in the RDA API result message.
+ */
+create table "pre_adj"."FissClaims"
+(
+    "dcn"               varchar(23) not null,
+    "hicNo"             varchar(12) not null,
```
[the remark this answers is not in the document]
karlmdavis: I'd suggest sticking with the source names, as @ericbuckley describes. It will make tracing ETL issues down simpler, and also removes the temptation to "get creative" and start coming up with field names ourselves (which only makes more of a mess).

karlmdavis (dismissed):
🎸

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V30__Create_preadjudicated_claims_tables.sql
```diff
@@ -0,0 +1,57 @@
+/*
+ * Creates the RDA FISS pre-adjudicated claims tables required by the application.
+ * https://confluenceent.cms.gov/pages/viewpage.action?spaceKey=MPSM&title=MVP+0.1+Data+Dictionary+for+Replicated+Data+Access+API
+ */
+
+CREATE SCHEMA IF NOT EXISTS "pre_adj";
```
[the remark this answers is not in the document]
keithdadkins: Permissions have been added that allows the pipeline db user to create schemas. Let me know if you run into issues.

rtravitz (approved):
✨

PR #564, opened 2021-04-29
DCGEO-42: Adding a NACL to control access to MPM VPC

ericbuckley:
### Change Details
Our lower environment, and soon to be upper environments, has a VPC Peering connection setup with an MPM environment.  This was enabled to give BFD access to the upcoming RDA API that will transmit pre-adjudicated claims data.

This commit adds a network ACL (NACL) to the bfd_pipeline module in case we ever need to shut off communication with their environment. By default, this NACL will only allow communication over port 443 and all other ports will be denied.  If for some reason we need to temporarily disable all communication we can flip the action value from "allow" to "deny" on rule 100.

### Acceptance Validation
1. Command to test communication; `curl  https://app-dev.mpsm.cms.gov:443/health`
2. After we provision the NACL to our test environment we should still be able to communicate with hosts in RDA
3. Manually change the NACL rule 100 action from "allow" to "deny"
4. At this point when we test the connection, it should fail
5. Flip the action on NACL rule 100 back to the default state to re-enable communication

### Feedback Requested

- [ ] Given our problem statement of being able to turn communication on/off; is an NACL the correct choice over a Security Group?
- [ ] Do I need the `count` variable on the NACL to keep it out of prd-sbx and prd?
- [ ] Is this the variable I used the best way to manage RDA CIDR blocks we need to firewall?

### External References

- [DCGEO-42](https://jira.cms.gov/browse/DCGEO-42)

### Security Implications

- [ ] new software dependencies; **NO**

- [ ] altered security controls; **YES**
  - By default, the NACL should not deny or allow any new communication measures within the VPC
  - However if switched to "deny", it will cutoff all communication with the hosts in the MPM VPC

- [ ] new data stored or transmitted; **NO**

- [ ] security checklist is completed for this change; **N/A**

- [ ] requires more information or team discussion to evaluate security implications; **YES**


ops/terraform/modules/resources/bfd_pipeline/main.tf
```diff
@@ -134,6 +134,35 @@ resource "aws_security_group_rule" "allow_db_primary_access" {
...
+    rule_no     = 100
+    action      = "allow"
+    protocol    = "all"
+    cidr_block  = var.mpm_rda_cidr_block
+  }
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    rule_no     = 200
```
brianburton, line 159: Should we also restrict the outbound port to 443?

brianburton (dismissed):
LGTM.

ops/terraform/modules/resources/bfd_pipeline/main.tf
ericbuckley: Thats an interesting idea, we could switch this NACL to use explicit "deny" actions, then create an additional NACL with a lower rule order that "allows" communication over 443.  I think its a good idea, but would love to hear more from the BFD engineers before making this more complex.  Maybe that level of control with another "trusted" VPC isn't necessary?

ops/terraform/modules/resources/bfd_pipeline/main.tf
```diff
@@ -134,6 +134,35 @@ resource "aws_security_group_rule" "allow_db_primary_access" {
   source_security_group_id = aws_security_group.app.id # The EC2 instance for the BFD Pipeline app.
 }
 
+# NACL to manage communication between the BFD and RDA environments
+# By default this NACL won't add any additonal security, however if for any reason
+# we need to shut off communication between the two VPCs, one could simply flip the
+# action(s) defined in these rules from "ALLOW" to "DENY"
+resource "aws_network_acl" "rda" {
+  count = var.mpm_rda_cidr_block ? 1 : 0
```
keithdadkins, line 142: I don't believe this will work because the ternary expression must eval to a bool.  You could try `count = var.mpm_rda_cidr_block != "" ? 1 : 0`

ops/terraform/env/test/stateless/main.tf
```diff
@@ -22,4 +22,6 @@ module "stateless" {
   ssh_key_name    = var.ssh_key_name
   git_branch_name = var.git_branch_name
   git_commit_id   = var.git_commit_id
+
+  mpm_rda_cidr_block = "10.234.232.0/21"
```
keithdadkins, line 26: Ideally, we would store this in our vars file to prevent hard-coding and exposing inside network info to the public. Low risk, but it's something we are trying to steer away from.

ops/terraform/modules/resources/bfd_pipeline/main.tf
keithdadkins: Ideally yes. But we can harden it later if you prefer.

keithdadkins (changes requested):
One issue and a few comments. Looks good to me otherwise.

ops/terraform/modules/resources/bfd_pipeline/main.tf
ericbuckley: makes sense, I update this

ops/terraform/env/test/stateless/main.tf
ericbuckley: @keithdadkins can you link me to the vars file that would ideally hold this value?

ops/terraform/modules/resources/bfd_pipeline/main.tf
ericbuckley: @keithdadkins I took a stab at refactoring this into 2 rules, let me know what you think

keithdadkins (approved):
LGTM

brianburton (commented):
LGTM

brianburton (approved):

ops/terraform/modules/resources/bfd_pipeline/main.tf
```diff
@@ -134,6 +134,41 @@ resource "aws_security_group_rule" "allow_db_primary_access" {
...
+# allows communication and defining a "upper" rule that defaults to denying
+resource "aws_network_acl" "rda" {
+  # only create the NACL if the CIDR block has been defined
+  count = var.mpm_rda_cidr_block != null ? 1 : 0
+
+  vpc_id      = var.env_config.vpc_id
+  tags        = merge({ Name = "bfd-${var.env_config.env}-etl-app" }, var.env_config.tags)
+}
+
+resource "aws_network_acl_rule" "rda_known" {
```
jzulim, line 149: Someone help me out here pls. If we set a NACL, I think that impacts any device / host that is assigned to that subnet. Would this then block outbound 5433 / 5432 which is required for the pipeline app to reach the DB? Perhaps I'm misunderstanding NACL here.

jzulim: SG is instant level, NACL is subnet level.

ericbuckley: Correct, this should block all communication to hosts on the given subnet except over port 443.  I choose a NACL because we can specify DENY rules with that, from my understanding we can only define ALLOW rules with a SG.

The subnet I'm planning on setting in the `mpm_rda_cidr_block` should be limited to only hosts available via the VPC peering connection to the MPSM environment.  Are you thinking there will be an overlap?

jzulim (approved):
LGTM. Follows current patterns, comments helpful, love the check to see if the subnet is set.

PR #567, opened 2021-05-03
[BFD-766] - Fix BeneficairyMonthly table with custom CCW data feed(s)

mackec1:
### Change Details

Due to the determination that the 2019 and 2020 CCW beginning of calendar year data feed(s) may have contained data that would be improperly applied to the _BeneficiaryMonthly_ table, it was deemed prudent to perform a full remediation assessment and data migration to ensure data is correct.

CCW has agreed to provide a 2019 and a 2020 custom data feed that BFD will use to update its repair and track data changes that may need to be applied. This ticket contains a series of SQL scripts as well as a _playbook_ that will be used to ensure BFD data is in sync with CCW historical data.

The file, _Playbook.txt,_ contains a detailed list of instructions that need to be followed to perform the necessary data remediation and follow-up quality assurance activities.

### Acceptance Validation

- Download the branch that contains the data.
- The _/sql_ folder contains the SQL script files as well as the playbook for this tiekct.
- Assess the quality and efficacy of the SQL scripts.
- There is also a _/sql/test_ folder that contains test data the shuld match the expected CCW data feed(s).
- Follow the steps as outlined in the _Playbook.txt_ file and verify completeness as well as ability to correctly fix any data issues.

### Feedback Requested

Suggestions and/or things that may be missing from either the SQL scripts of the Playbook.

### External References

- [BFD-766](https://jira.cms.gov/browse/BFD-766)

### Security Implications

- [ Y] new data stored or transmitted

This ticket is designed to fix data discrepancies in the BFD BeneficiaryMonthly table; of particular interest is monthly Part D data.


sql/track_bene_monthly_change_func.sql
```diff
@@ -0,0 +1,38 @@
...
+			OR	OLD."partCPlanTypeCode" 				<> NEW."partCPlanTypeCode"
+			OR	OLD."partDContractNumberId" 			<> NEW."partDContractNumberId"
+			OR	OLD."partDPbpNumberId" 					<> NEW."partDPbpNumberId"
+			OR	OLD."partDSegmentNumberId" 				<> NEW."partDSegmentNumberId"
+			OR	OLD."partDRetireeDrugSubsidyInd" 		<> NEW."partDRetireeDrugSubsidyInd"
+			OR	OLD."medicaidDualEligibilityCode" 		<> NEW."medicaidDualEligibilityCode"
+			OR	OLD."partDLowIncomeCostShareGroupCode" 	<> NEW."partDLowIncomeCostShareGroupCode"
+			)
+			THEN
+		*/
```
karlmdavis, line 28: Why's this commented out code being left in? If you think you might still need it in the future, please add a comment explaining why/when.

sql/process_ccw_table.sql
```diff
@@ -0,0 +1,264 @@
+CREATE OR REPLACE PROCEDURE load_from_ccw()
```
karlmdavis, line 1: I think this fails to account for records that should be deleted? For instance, if a bene-year record _is_ in the DB but _isn't_ in this one-time export, shouldn't we remove it?

karlmdavis (changes requested):
Excellent work in impressive time, @mackec1!

I left some notes about functionality that may be missing/incorrect. Please correct me where/if I'm wrong.

sql/bene_monthly_audit.sql
```diff
@@ -0,0 +1,26 @@
+DROP TABLE IF EXISTS public.beneficiary_monthly_audit;
+
+CREATE TABLE public.beneficiary_monthly_audit
```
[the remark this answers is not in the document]
mackec1: yes and now....open to suggestions.
YES, it would be useful in subsequent audit activities (i.e., join vs. beneMonthly when looking at new vs chg'd). Also it would prevent duplicate rcds of same bene_id + yearMonth.
NO, while it is highly unlikely, if we did hit a case where we hit a unique constraint violation, then the whole job craps out and a rollback occurs...
But I think I agree that a primary key would be beneficial here from the get-go (was thinking of indexing the table post-run, but might as well do it early).

sql/ccw_load_temp.sql
```diff
@@ -0,0 +1,163 @@
+CREATE TABLE public.ccw_load_temp
+(
+	BENE_ID                     	VARCHAR(255)  COLLATE pg_catalog."default"  NOT NULL,
```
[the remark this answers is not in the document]
mackec1: ok, UTF-8 it is. Some of the playbook is a little loose and fluid in that we don't have CCW filenames yet, etc.

sql/bene_monthly_audit.sql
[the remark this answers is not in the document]
mackec1: Yup, that audit table is gonna be valuable in a number of ways post-run.

sql/ccw_load_update_trigger.sql
```diff
@@ -0,0 +1,32 @@
...
+			OLD."partCContractNumberId",
+			OLD."partCPbpNumberId",
+			OLD."partCPlanTypeCode",
+			OLD."partDContractNumberId",
+			OLD."partDPbpNumberId",
+			OLD."partDSegmentNumberId",
+			OLD."partDRetireeDrugSubsidyInd",
+			OLD."medicaidDualEligibilityCode",
+			OLD."partDLowIncomeCostShareGroupCode")
+	IS DISTINCT FROM
```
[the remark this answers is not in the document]
mackec1: I actually have an INSERT trigger in my back pocket, but just didn't see anyways we'd actually need it...thoughts?

sql/update_bene_monthly_proc.sql
```diff
@@ -0,0 +1,41 @@
...
+    		"hmoIndicatorInd"					= hmo_ind,
+    		"partCContractNumberId"				= partc_contract_number_id,
+    		"partCPbpNumberId"					= partc_pbp_number_id,
+    		"partCPlanTypeCode"					= partc_plan_type,
+    		"partDContractNumberId"				= partd_contract_number_id,
+    		"partDPbpNumberId"					= partd_pbp_number_id,
+    		"partDSegmentNumberId"				= partd_segment_num,
+    		"partDRetireeDrugSubsidyInd"		= partd_retiree_mnthly,
+    		"partDLowIncomeCostShareGroupCode"	= partd_low_inc_cost_share,
+    		"medicaidDualEligibilityCode"		= dual_elig_code
```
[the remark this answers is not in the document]
mackec1: Yea, I'll run it through a SQL formatter.

sql/update_bene_monthly_proc.sql
```diff
@@ -0,0 +1,41 @@
...
+	partd_pbp_number_id 		"BeneficiaryMonthly"."partDPbpNumberId"%TYPE,
+	partd_segment_num 			"BeneficiaryMonthly"."partDSegmentNumberId"%TYPE,
+	partd_retiree_mnthly 		"BeneficiaryMonthly"."partDRetireeDrugSubsidyInd"%TYPE,
+	partd_low_inc_cost_share 	"BeneficiaryMonthly"."partDLowIncomeCostShareGroupCode"%TYPE,
+	dual_elig_code 				"BeneficiaryMonthly"."medicaidDualEligibilityCode"%TYPE
+)
+AS
+$$
+    BEGIN
+    	UPDATE "BeneficiaryMonthly"
```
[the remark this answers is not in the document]
mackec1: See prev commonet r.e. INSERT trigger; if there's the possiblity of ever having an INSERT then this can be an UPSERT and I'll add in the INSERT trigger.

sql/track_bene_monthly_change_func.sql
mackec1: Yea, I left it in pending a review; I'll just remove.

sql/process_ccw_table.sql
mackec1: hmmm...tell me if this use case is correct:
- we have a BeneficiaryMonthly rcd for 2020-03-01.
- in our 2020 CCW data feed we don't have a rcd for 2020-03-1 for that same beneficiary.
- ideally we'd like to nuke the 2020-03-01 rcd from BeneficiaryMonthly table.

I'm thinking the best way to handle that would be to do a post-process exercise (query) where we select all bene_id, yearMonth from BeneficiaryMonthly MINUS all bene_id, yearMonth audit table (should provide a list of rcds we'd want to nuke). So wrap the above in a delete from where bene_id, yearMonth in(...
or something like that.

If we are going to do that, then I guess I'd also like to add a delete trigger; always good to see where things!

Thoughts?

sql/bene_monthly_audit.sql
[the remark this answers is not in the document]
karlmdavis: I actually like things crapping out. Wouldn't that mean there was something funky going on with the data or the code?

sql/update_bene_monthly_proc.sql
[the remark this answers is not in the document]
karlmdavis: I'd say either add support for it now, or set things to blow up explicitly if the condition is encountered. (I _think_ it'd fail as-is, which is good, but it's been a long day.)

sql/process_ccw_table.sql
karlmdavis: > tell me if this use case is correct

Yup! I can't really work out how _likely_ it is, but better safe than sorry, I think.

> I'm thinking the best way to handle that would be to do a post-process exercise (query) where we select all bene_id, yearMonth from BeneficiaryMonthly MINUS all bene_id, yearMonth audit table (should provide a list of rcds we'd want to nuke). So wrap the above in a delete from where bene_id, yearMonth in(...
or something like that.

Agreed.

> If we are going to do that, then I guess I'd also like to add a delete trigger; always good to see where things go!

Also agreed!

cbrunefearless (approved):
LGTM.. I like the scripts.. As long as the data format is correct  I am happy to approve.

sql/test/ccw_load_test_data.csv
```diff
@@ -1,5 +1,5 @@
...
+-19990000000003,2022,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,C,C,C,C,C,C,C,C,C,C,C,C,D,D,D,D,D,D,D,D,D,D,D,D,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,Z0000,Z0001,Z0002,Z0003,Z0004,Z0005,Z0006,Z0007,Z0008,Z0009,Z0010,Z0011,C,C,C,C,C,C,C,C,C,C,C,C,003,004,005,006,007,008,009,010,011,012,013,014,C,C,C,C,C,C,C,C,C,C,C,C,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA
+-19990000000004,2023,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,C,C,C,C,C,C,C,C,C,C,C,C,D,D,D,D,D,D,D,D,D,D,D,D,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,C,Z0000,Z0001,Z0002,Z0003,Z0004,Z0005,Z0006,Z0007,Z0008,Z0009,Z0010,Z0011,C,C,C,C,C,C,C,C,C,C,C,C,003,004,005,006,007,008,009,010,011,012,013,014,C,C,C,C,C,C,C,C,C,C,C,C,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA,AA
```
karlmdavis, line 5: Did you add any test data that exercises the `INSERT` and `DELETE` triggers?

sql/update_bene_monthly_proc.sql
```diff
@@ -1,41 +1,63 @@
 CREATE OR REPLACE PROCEDURE update_bene_monthly(
```
karlmdavis, line 1: The CCW data likely includes a lot of rows where all of the data fields are `NULL`; they likely didn't filter those out. We need to: the load script will have to check for and skip year-months where all of the fields are `NULL`.

sql/Playbook.txt
```diff
@@ -22,10 +22,28 @@ Verify the function was created successfully.
 ** BENEFICIARY_MONTHLY Update Trigger                          **         
 *****************************************************************
 psql -h 127.0.0.1 -U bfd -d fihr -a -f sql/ccw_load_update_trigger.sql 
+psql -h 127.0.0.1 -U bfd -d fihr -a -f sql/ccw_load_delete_trigger.sql 
+psql -h 127.0.0.1 -U bfd -d fihr -a -f sql/ccw_load_insert_trigger.sql
```
karlmdavis, line 26: A general note about files & folders: Do all of these need to be in separate files? If not, it feels like it adds some minor complexity & risk to the whole thing (e.g. someone forgets to add a trigger); I'd combine them, where possible.

karlmdavis (changes requested):
@mackec1 this is more great work!

I noted a couple of minor things and also a couple of moderate ones (test data and `NULL` handling).

sql/Playbook.txt
mackec1: Will put into minimum number of files; originally only had support for the update trigger and the insert,delete were added per orig PR review. Not a big deal to coerce into one.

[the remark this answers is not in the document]
mackec1: Will do; was unclear how to treat something that was (hopefully) a one-off / one-time only set of scripts.

sql/update_bene_monthly_proc.sql
karlmdavis: > Do I need to check for null in each month's (13) columns?

Yes. And we also need to check for _`NULL`-like_ values, e.g. if the BFD Pipeline treats "0" in a particular column as a null, then this script should, too.

> If not, what subset of columns makes sense (i.e, just check PTC_CNTRCT_ID and PTD_CNTRCT_ID)?

To frame it a different way: check for nulls in every column except for the bene ID and the date.

rtravitz (approved):
Got a similar walkthrough some parts and ran it locally. Clear instructions 🥇

PR #571, opened 2021-05-03
[BFD-749] Add Unit Tests to HHA EOB Profile

JeremyIR:
### Change Details

Modify HHA v2 transform test code as an expansion similar to Outpatient claim transformer tests, so that there is more confidence individual fields and values are working as intended.

### Acceptance Validation

- Confirm that the same level of and unit tests for BFD V2 are applied to V2
- Ensure that all required CARIN data fields have corresponding unit tests for V2

### Feedback Requested

Do the updated set of tests for HHA v2 Transformer thoroughly account for the CARIN fields?

### External References

- [BFD-749](https://jira.cms.gov/browse/BFD-749)

### Security Implications


rtravitz (commented):
Looks reasonable. Same feedback as similar PRs about removing `assertMatches` now that it's no longer used.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/HHAClaimTransformerV2Test.java
```diff
@@ -38,22 +69,805 @@ public HHAClaim generateClaim() throws FHIRException {
...
+
+    DiagnosisComponent cmp2 =
+        TransformerTestUtilsV2.createDiagnosis(
+            // Order doesn't matter
+            diag2.getSequence(),
+            new Coding("http://hl7.org/fhir/sid/icd-10", "H8888", null),
+            new Coding(
+                "http://hl7.org/fhir/us/carin-bb/CodeSystem/C4BBClaimDiagnosisType",
+                "secondary",
+                "secondary"),
```
mackec1, line 479: One of these needs to be "Secondary" .

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/HHAClaimTransformerV2Test.java
```diff
@@ -38,22 +69,805 @@ public HHAClaim generateClaim() throws FHIRException {
...
+
+    DiagnosisComponent cmp3 =
+        TransformerTestUtilsV2.createDiagnosis(
+            // Order doesn't matter
+            diag3.getSequence(),
+            new Coding("http://hl7.org/fhir/sid/icd-10", "R2222", null),
+            new Coding(
+                "http://hl7.org/fhir/us/carin-bb/CodeSystem/C4BBClaimDiagnosisType",
+                "secondary",
+                "secondary"),
```
mackec1, line 496: Same issue where Display needs to be Capitalized.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/HHAClaimTransformerV2Test.java
```diff
@@ -38,22 +69,805 @@ public HHAClaim generateClaim() throws FHIRException {
...
+
+    DiagnosisComponent cmp4 =
+        TransformerTestUtilsV2.createDiagnosis(
+            // Order doesn't matter
+            diag4.getSequence(),
+            new Coding("http://hl7.org/fhir/sid/icd-10", "R3333", null),
+            new Coding(
+                "http://hl7.org/fhir/us/carin-bb/CodeSystem/C4BBClaimDiagnosisType",
+                "secondary",
+                "secondary"),
```
mackec1, line 513: ditto

mackec1 (commented):
Couple of minor nits; the Dispaly value should be capitalized where noted.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/HHAClaimTransformerV2Test.java
JeremyIR: I was following what I saw in the JSON. When I capitalize the display name, the test fails

mackec1: Then it may well be that we have a minor bug in the Transformer code; to wit:

http://hl7.org/fhir/us/carin-bb/STU1/CodeSystem-C4BBClaimDiagnosisType.html

So that is essentially a 'base' code system; as noted in the above reference (URL), we see the following:

This Code system is referenced in the content logical definition of the following value sets:

C4BBClaimInpatientInstitutionalDiagnosisType
C4BBClaimOutpatientInstitutionalDiagnosisType
C4BBClaimProfessionalAndNonClinicianDiagnosisType

Now while the Display attribute is not that bif a deal, the fact that we reference
http://hl7.org/fhir/us/carin-bb/CodeSystem/C4BBClaimDiagnosisType
but then don't follow the spec, is effectively an error.

JeremyIR: I see. Yes. Seeing how the maven build isnt passing, Ill push this suggested change and we can address the local error.

rtravitz (approved):
I see all the feedback was addressed. Nice work ✨

mackec1 (approved):
Nice work! Thanks for making those changes.

PR #572, opened 2021-05-04
Cbrune/fix andrews branch

cbrunefearless:
### Change Details

This work adds integration tests for the R4 transformers and V2 endpoints that mimic the coverage provided by STU3.  The coverage should be similar, but the validations rely on more deep comparisons of the transformed vs API returned data structures.  Specifically this PR has added the following integration test classes:

* R4CoverageResourceProviderIT.java 
* R4ExplanationOfBenefitResourceProviderIT.java 
* R4PatientResourceProviderIT.java
* EndpointJsonResponseComparatorV2IT.java

For `EndpointJsonResponseComparator*` the gold copy files were moved into `v1` and `v2` directories to keep the two separate.  I also ran into a lot of issues with comparing EOBs returned and serialized from the endpoint vs the transformer.  This came down to trivial differences like the endpoint based serialization returning a `Quantity` and the transformer returning a `SimpleQuantity` or a number being represented as `75.0` vs `75`.  For this reason, `equalsDeep` couldn't directly be used.  This is why the EOB comparison seems messy, but provides good coverage of the fields.  There is documentation in the code that explains this process as well.

Additionally, there were several issues with the V2 endpoints that were discovered during the course of working this ticket:

* A number of claims were adding a `Diagnosis` element without a Label.  This was defaulted to `OTHER`.
* `R4SamhsaMatcher` was only dealing with PDE, causing other types to throw errors
* SNF claims were being mapped with a type of HOSPICE
* The contained `Observation` generated for some EOB types was incorrectly adding `#` to the front of the ID in the actual resource.  This should only be part of the local reference.  See: https://build.fhir.org/references.html#contained

### Acceptance Validation

All the integration tests  (V1 and V2) pass locally when run with a full clean build.

image

### External References

- [BFD-619](https://jira.cms.gov/browse/BFD-619)
### Security Implications

N/A

N/A

N/A

N/A

- [ ] N/A


apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2Test.java
```diff
@@ -506,7 +506,10 @@ public void shouldHaveDiagnosesMembers() {
             // Order doesn't matter
             diag5.getSequence(),
             new Coding("http://hl7.org/fhir/sid/icd-10", "H12345", null),
-            new Coding(null, null, null),
+            new Coding(
+                "http://hl7.org/fhir/us/carin-bb/CodeSystem/C4BBClaimDiagnosisType",
+                "secondary",
```
mackec1, line 511: chris, hate to bug you on this, but....
if you go to http://hl7.org/fhir/us/carin-bb/CodeSystem/C4BBClaimDiagnosisType, you'll see that the code == "secondary" but the display == "Secondary"...

mackec1 (commented):
Just the nit on the display text.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2Test.java
cbrunefearless: Gotcha, good catch

mackec1 (approved):
Looks good; hate those IT tests if something changes and you have to re-generate the json!

rtravitz (approved):
I get in general what's happening with this. But detailed review of whether each line is being compared correctly is kind of impenetrable to me because of the sheer scope of the PR. That said, since both @mackec1 and @cbrunefearless contributed to this after @Dranith, I'm still prepared to help move it forward.

PR #546, opened 2021-04-22
Revert "Upgraded HSQL and Flyway. (#541)"

rtravitz:
### Change Details
This PR reverts changes made in #541. The deployed pipeline application is encountering an error with Flyway that will need to be troubleshot separately.  

### Security Implications
- [ ] new software dependencies
- [ ] altered security controls
- [ ] new data stored or transmitted
- [ ] security checklist is completed for this change
- [ ] requires more information or team discussion to evaluate security implications

PR #573, opened 2021-05-05
[BFD-778] Upgrade Flyway and HSQL

rtravitz:
### Change Details
This PR finishes [BFD-778](https://jira.cms.gov/browse/BFD-778). It upgrades Flyway and HSQL versions. This PR is another attempt following work that was done in #541 and the reversion of that work in #546. After the first attempt we received the following error:
```
org.flywaydb.core.api.FlywayException: Found non-empty schema(s) "public" but no schema history table. Use baseline() or
set baselineOnMigrate to true to initialize the schema history table.
```

This exception occurred because in Flyway v5 the default name of the schema table was changed from `schema_history` to `flyway_schema_history`. This is the  table where Flyway records which migrations it previously applied. Because we are upgrading from Flyway v4 to v7, we need to specify the old table name for backwards compatibility.

This error was not noticed during the first attempt because the tests run a Flyway Clean, which wipes out the database and creates a clean slate.

### Acceptance Validation
- When starting the BFD Server against a database with an existing `schema_history` table created with an older version of Flyway, there should be no Flyway errors in the server log.

### External References
- [BFD-778](https://jira.cms.gov/browse/BFD-778)

### Security Implications

### DIFF
--- a/apps/bfd-model/bfd-model-rif/pom.xml
+++ b/apps/bfd-model/bfd-model-rif/pom.xml
@@ -137,7 +137,7 @@
 				class. -->
 			<groupId>org.flywaydb</groupId>
 			<artifactId>flyway-core</artifactId>
-			<version>4.2.0</version>
+			<version>7.8.1</version>
 		</dependency>
 
 		<dependency>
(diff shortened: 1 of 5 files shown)


mackec1 (approved):
LGTM

JeremyIR:
LGTM

JeremyIR (approved):

PR #589, opened 2021-05-11
[BFD-723] Add single instance load test instructions

rtravitz:
### Change Details
This PR supports [BFD-723](https://jira.cms.gov/browse/BFD-723). It adds instructions for running a single instance load test with [Locust](https://locust.io/). 

### Acceptance Validation
- Instructions would allow a reader to run a load test of their own.

### Feedback Requested
- Do you have what you need to be able to run your own test with these instructions?

### External References
- [BFD-723](https://jira.cms.gov/browse/BFD-723)

### Security Implications


ops/ccs-ops-misc/load-test/locustfile.py
```diff
@@ -0,0 +1,54 @@
...
+    sys.exit(1)
+
+try:
+    with open(ids_file, 'r') as f:
+        eob_ids = f.read().split('\n')
+
+        # remove empty string at the end
+        eob_ids.pop()
+
+        patient_ids = eob_ids.copy()
```
keithdadkins, line 38: what's this for? looks unused

ops/ccs-ops-misc/load-test/locustfile.py
```diff
@@ -0,0 +1,54 @@
...
+except FileNotFoundError as err:
+    print(f'key file not found: {err}')
+    sys.exit(1)
+
+after_date = '2021-04-26'
```
keithdadkins, line 43: should probably be a param, or maybe date now - n days, else as time goes by this could impact results dramatically

keithdadkins (approved):
A few things in the python script (not sure if this is the final script or just an example), but otherwise this is great. Very thorough documentation.

ops/ccs-ops-misc/load-test/locustfile.py
rtravitz: You're right! Vestigial from when I also was making calls to the `Patient` resource. Will remove it here to not add confusion.

ops/ccs-ops-misc/load-test/locustfile.py
rtravitz: This script will have to be modified for each test, so I'm not overly worried about that. I ran a test at first that didn't have any kind of date to simulate those expensive requests. 

But I can make it default to two weeks prior and add a note in the instructions, just so people don't need to solve the date stuff themselves.

keithdadkins (approved):
Very thorough, looks great.

jzulim (approved):
Great first step, that has been a long time coming. I also looked at your confluence page / results. Nice!

PR #591, opened 2021-05-11
[DCGEO-60] ssh access for eric.buckley, brian.burton and kyle.robbins

ericbuckley:
### Change Details
Adding public SSH keys for eric.buckley, brian.burton and kyle.robbins, so the three of us can SSH into the BFD hosts in the test environment.

### External References

- [DCGEO-60](https://jira.cms.gov/browse/DCGEO-60)

### Security Implications

  - the three new users will gain SSH access to the hosts in the lower environment


keithdadkins (approved):
I verified the three users were added.

jzulim (approved):
Pulled down and decrypted vault, looks good to me. 🚀

PR #592, opened 2021-05-11
Kylerobbins/dcgeo 48 pre adj claims api skeleton

KyRobbins:
### Change Details

Created "skeleton code" for new fhir resource endpoint. Added the following classes:
- PreAdjClaimResponseTypeTransformer (Functional Interface for ClaimResponses)
- PreAdjClaimTypeTransformer (Functional Interface for Claims)
- PreAdjClaimResponseTypeV2 (Enum mapping config for ClaimResponses)
- PreAdjClaimTypeV2 (Enum Mapping config for Claims)
- PreAdjFissClaimResponseTransformerV2 (Fiss -> ClaimResponse)
- PreAdjFissClaimTransformerV2 (Fiss -> Claim)
- PreAdjMcsClaimResponseTransformerV2 (MCS -> ClaimResponse)
- PreAdjMcsClaimTransformerV2 (MCS -> Claim)
- R4ClaimResourceProvider (Serves Claim fhir resource requests)
- R4ClaimResponseResourceProvider (Serves ClaimResponse fhir resource requests)

Providers have not been "activated" (not added to spring configuration) and can not be reached.  The accompanying FissClaim and McsClaim entities have not yet been implemented, and thus the Providers have no database interaction logic nor the transformers any mapping logic.

### Acceptance Validation

The goal was to create the "Skeletal" framework for needed future work.  Entities need to be created before any mapping logic can be produced or any database interactions implemented.  Implemented code should provide a basic "template" to complete the aforementioned future work.

### Feedback Requested

Ensure the new resource endpoint classes follow expectations and provide adequate groundwork for completing future work.

### External References

- [DCGEO-48](https://jira.cms.gov/browse/DCGEO-48)

### Security Implications

No PHI/PII currently being processed, only "Skeletal" code has been introduced.

- [x] security checklist is completed for this change


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,49 @@
...
+   * @param metricRegistry the {@link MetricRegistry} to use
+   * @param claimEntity the Fiss {@link PreAdjFissClaim} to transform
+   * @return a FHIR {@link ClaimResponse} resource that represents the specified claim
+   */
+  @Trace
+  static ClaimResponse transform(MetricRegistry metricRegistry, Object claimEntity) {
+    Timer.Context timer =
+        metricRegistry
+            .timer(
+                MetricRegistry.name(
```
brianburton, line 21: The metric name should be declared in a static final constant to simplify the code at the point of use.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResourceProviderTest.java
```diff
@@ -0,0 +1,129 @@
...
+    for (Parameter p : readMethod.getParameters()) {
+      if (p.getType() == IdType.class) {
+        annotation = p.getAnnotation(IdParam.class);
+      }
+    }
+
+    assertNotNull(annotation);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
brianburton, line 53: Since all the checks throw the same exception checking the message here would verify the exception is actually the one the test is looking for.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResponseResourceProviderTest.java
```diff
@@ -0,0 +1,129 @@
...
+    for (Parameter p : readMethod.getParameters()) {
+      if (p.getType() == IdType.class) {
+        annotation = p.getAnnotation(IdParam.class);
+      }
+    }
+
+    assertNotNull(annotation);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
brianburton, line 53: Since all the checks throw the same exception checking the message here would verify the exception is actually the one the test is looking for.  Same for the remaining argument tests here.

Gabasol:
Is this PR dependent on another ticket/PR?

ericbuckley:
> Is this PR dependent on another ticket/PR?

@Gabasol nope

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/PreAdjClaimResponseTypeTransformerV2.java
```diff
@@ -0,0 +1,20 @@
+package gov.cms.bfd.server.war.commons;
+
+import com.codahale.metrics.MetricRegistry;
+import org.hl7.fhir.r4.model.ClaimResponse;
+
+/**
+ * A {@link FunctionalInterface} describing the types that handle transforming Fiss/MCS records into
```
ericbuckley, line 7: small thing, but FISS (Financial Intermediary Standard System) and MCS (Mulit-Carrier System) are both acronyms

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResourceProvider.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.stereotype.Component;
+
+/** This FHIR {@link IResourceProvider} adds support for R4 {@link Claim} resources. */
+@Component
+public final class R4ClaimResourceProvider implements IResourceProvider {
+  /**
+   * A {@link Pattern} that will match the {@link Claim#getId()}s used in this application, e.g.
+   * <code>f-1234</code> or <code>m--1234</code> (for negative IDs).
+   */
+  private static final Pattern CLAIM_ID_PATTERN = Pattern.compile("(\\p{Alpha}+)-(-?\\p{Alnum}+)");
```
ericbuckley, line 30: any reason to not just limit this to `f` and `m` prefixes?
```
Pattern.compile("([fm])-(-?\\p{Alnum}+)")
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResourceProvider.java
```diff
@@ -0,0 +1,105 @@
...
+    String claimIdText = claimId.getIdPart();
+    if (claimIdText == null || claimIdText.trim().isEmpty())
+      throw new IllegalArgumentException("Resource ID can not be null/blank");
+
+    Matcher claimIdMatcher = CLAIM_ID_PATTERN.matcher(claimIdText);
+    if (!claimIdMatcher.matches())
+      throw new IllegalArgumentException("Unsupported ID pattern: " + claimIdText);
+
+    String claimIdTypeText = claimIdMatcher.group(1);
+    Optional<PreAdjClaimTypeV2> eobIdType = PreAdjClaimTypeV2.parse(claimIdTypeText);
```
ericbuckley, line 95: s/eobIdType/claimIdType/

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResponseResourceProvider.java
```diff
@@ -0,0 +1,107 @@
...
+
+  /** @param metricRegistry the {@link MetricRegistry} to use */
+  @Inject
+  public void setMetricRegistry(MetricRegistry metricRegistry) {
+    this.metricRegistry = metricRegistry;
+  }
+
+  /** @param samhsaMatcher the {@link R4SamhsaMatcher} to use */
+  @Inject
+  public void setSamhsaFilterer(R4SamhsaMatcher samhsaMatcher) {
```
ericbuckley, line 52: I don't think we'll need a SAMHSA filter on the ClaimResponse, the procedure and diagnosis codes will only be present in the Claim

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResponseResourceProvider.java
```diff
@@ -0,0 +1,107 @@
...
+    String claimIdText = claimId.getIdPart();
+    if (claimIdText == null || claimIdText.trim().isEmpty())
+      throw new IllegalArgumentException("Resource ID can not be null/blank");
+
+    Matcher claimIdMatcher = CLAIM_ID_PATTERN.matcher(claimIdText);
+    if (!claimIdMatcher.matches())
+      throw new IllegalArgumentException("Unsupported ID pattern: " + claimIdText);
+
+    String claimIdTypeText = claimIdMatcher.group(1);
+    Optional<PreAdjClaimResponseTypeV2> eobIdType =
```
ericbuckley, line 96: s/eobIdType/claimResponseIdType/

ericbuckley (changes requested):
@KyRobbins great first PR, just a few small changes needed

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResourceProvider.java
KyRobbins: I was just copying the pattern used in EoB.  I guess you could make the argument that it would increase the required code changes if adding another claim type, or that it doesn't allow for as specific of an error when determining what the issue was, but really I was just following the logic of the existing one.

brianburton: Whether we use single character or longer prefixes it could be helpful if the regex specifically matched only our valid prefixes rather than `Alpha`.

ericbuckley (approved):

brianburton (dismissed):
Looks good!

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/AssertUtils.java
```diff
@@ -0,0 +1,52 @@
+package gov.cms.bfd.server.war.utils;
+
+import static org.junit.Assert.fail;
+
+import java.util.Objects;
+
+public class AssertUtils {
```
brianburton, line 7: I love this class.  It might be better moved into `bfd-shared-utils` though so that the pipeline tests could use it too.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/AssertUtils.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public static Exception catchExceptions(Executor executor) {
+    Exception ex = null;
+
+    try {
+      executor.execute();
+    } catch (Exception e) {
+      ex = e;
+    }
```
brianburton, line 21: When I test for an expected exception I include `fail("expected an exception to the thrown");` to fail the test if none is thrown.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/AssertUtils.java
```diff
@@ -0,0 +1,52 @@
...
+
+    return ex;
+  }
+
+  public static void assertThrowEquals(Exception expected, Exception actual) {
+    if (expected.getClass() == actual.getClass()) {
```
brianburton, line 27: A check for `actual == null` here could print `fail` with a message indicated that a required exception was not thrown.  As it is this would throw an NPE if the expected exception was not thrown.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/AssertUtils.java
KyRobbins: It's not really doing an assertion test for an exception per se, it's just catching any exception that might get thrown, the assertion is actually done in the test itself which compares the caught exception (or null if no exception was caught) to the expected one.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/AssertUtils.java
KyRobbins: Good catch, I did forget the nullcheck

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/AssertUtils.java
KyRobbins: We could do that.  It's similar to how junit5 does exception testing, for when you want something more than just "an exception of type X was indeed thrown"

brianburton (approved):
Looks good to me.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/AssertUtils.java
KyRobbins: Possible solution of adding a `bfd-shared-test-utils` module to include this and other common testing classes was posed.

ericbuckley:
The goal of this PR was to obtain feedback on our "plan" for RDA->FHIR transformers and NOT the implementation.  During a review, it became apparent that having a small portion of the implementation live in this PR and the majority of it remain in PR #612, made the review process difficult.

Now that we have some good feedback on organization of these transformers, I'm recommending that we close this PR and incorporate all of these changes in PR #612.  @KyRobbins once you've incorporated all the suggestions/feedback into PR #612, please go ahead and close this.

PR #593, opened 2021-05-12
[BFD-854] Fix-PartD-Enrollment-Smaller-Transactions

mackec1:
### Change Details

BFD-766 created the main setup and processing engine for fixing BeneficiaryMonthly Part D contract data with custom load files (one for 2019, one for 2020) provided by CCW. The last step in that process, 05_process_ccw_table.sql, failed when the OS killed the database server due to extraneous memory consumption. This was probably due to a significantly high amount of memory being reserved for the rollback/commit buffers.

This PR modifies the 05_process_ccw_table.sql procedure to perform a db COMMIT every 10,000 records.

In addition, there was (very) minor logic to attempt to minimize unnecessary date calculations which might save some (minimal) processing compute cycles...but times 130M rcds might add up.

### Acceptance Validation

Look for the COMMIT statement every 10,000 rcds.

Look at the date calc cache logic (note, the process reads the CCW_LOAD_TEMP file sequentially, which means the first half of the table will be like years (2019), and then 2020 rcds will start.

### Feedback Requested

Logic look appropriate.

### External References

- [BFD-854](https://jira.cms.gov/browse/BFD-854)

### Security Implications

N/A

### DIFF
--- a/apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-854-Fix-PartD-Enrollment-Smaller-Transactions/05_process_ccw_table.sql
+++ b/apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-854-Fix-PartD-Enrollment-Smaller-Transactions/05_process_ccw_table.sql
@@ -1,0 +1,387 @@
+--
+-- Create PROCEDURE update_bene_monthly
+--
+CREATE OR REPLACE PROCEDURE update_bene_monthly(
+    bene_id                     "BeneficiaryMonthly"."parentBeneficiary"%TYPE,
+    yr_month                    "BeneficiaryMonthly"."yearMonth"%TYPE,
+    fips_cnty_code              "BeneficiaryMonthly"."fipsStateCntyCode"%TYPE,
+    medi_status_code            "BeneficiaryMonthly"."medicareStatusCode"%TYPE,
+    buy_in_ind                  "BeneficiaryMonthly"."entitlementBuyInInd"%TYPE,
+    hmo_ind                     "BeneficiaryMonthly"."hmoIndicatorInd"%TYPE,
+    partc_contract_number_id    "BeneficiaryMonthly"."partCContractNumberId"%TYPE,
+    partc_pbp_number_id         "BeneficiaryMonthly"."partCPbpNumberId"%TYPE,
+    partc_plan_type             "BeneficiaryMonthly"."partCPlanTypeCode"%TYPE,
+    partd_contract_number_id    "BeneficiaryMonthly"."partDContractNumberId"%TYPE,
+    partd_pbp_number_id         "BeneficiaryMonthly"."partDPbpNumberId"%TYPE,
+    partd_segment_num           "BeneficiaryMonthly"."partDSegmentNumberId"%TYPE,
+    partd_retiree_mnthly        "BeneficiaryMonthly"."partDRetireeDrugSubsidyInd"%TYPE,
+    partd_low_inc_cost_share    "BeneficiaryMonthly"."partDLowIncomeCostShareGroupCode"%TYPE,
+    dual_elig_code              "BeneficiaryMonthly"."medicaidDualEligibilityCode"%TYPE
+)
+AS
+$$
+BEGIN
+    IF     fips_cnty_code            IS NOT NULL
+        OR medi_status_code          IS NOT NULL
+        OR buy_in_ind                IS NOT NULL
+        OR hmo_ind                   IS NOT NULL
+        OR partc_contract_number_id  IS NOT NULL
+        OR partc_pbp_number_id       IS NOT NULL
+        OR partc_plan_type           IS NOT NULL
+        OR partd_contract_number_id  IS NOT NULL
+        OR partd_pbp_number_id       IS NOT NULL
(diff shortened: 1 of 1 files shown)


cbrunefearless (approved):
lgtm

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-854-Fix-PartD-Enrollment-Smaller-Transactions/05_process_ccw_table.sql
```diff
@@ -0,0 +1,387 @@
...
+            rcd.CST_SHR_GRP_DEC_CD,
+            rcd.META_DUAL_ELGBL_STUS_DEC_CD);
+            
+        rcd_cnt := rcd_cnt + 1;
+        
+        if rcd_cnt % 10000 = 0
+        THEN
+            COMMIT;
+            RAISE INFO 'Record Count: % ...', rcd_cnt;
+        END IF;
```
rtravitz, line 368: This seems like this is the relevant part? Are there any other places to highlight?

rtravitz (commented):
Because we'd now be committing incrementally, are there concerns about replaying the job if something else fails? Said another way, is this still idempotent?

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-854-Fix-PartD-Enrollment-Smaller-Transactions/05_process_ccw_table.sql
mackec1: That would be the 'magic bullet' (I hope)

mackec1: As for the idempotent, it is! We could run the job 100x and the results would be the same. In addition there would not be spurious audit log entries since the trigger only fires if data has changed, so after the first run (or even the first partial run) there would be not changes detected, so no audit rcd written.

rtravitz (approved):
Looks like a reasonable next step to attempt

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-854-Fix-PartD-Enrollment-Smaller-Transactions/05_process_ccw_table.sql
karlmdavis: Looks like idempotent, though it will always run through the full set of audit data from the beginning if restarted (which is fine!).

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-854-Fix-PartD-Enrollment-Smaller-Transactions/05_process_ccw_table.sql
```diff
@@ -0,0 +1,387 @@
...
+            May1 := TO_DATE(rcd.RFRNC_YR || '-05-01', 'YYYY-MM-DD');
+            Jun1 := TO_DATE(rcd.RFRNC_YR || '-06-01', 'YYYY-MM-DD');
+            Jul1 := TO_DATE(rcd.RFRNC_YR || '-07-01', 'YYYY-MM-DD');
+            Aug1 := TO_DATE(rcd.RFRNC_YR || '-08-01', 'YYYY-MM-DD');
+            Sep1 := TO_DATE(rcd.RFRNC_YR || '-09-01', 'YYYY-MM-DD');
+            Oct1 := TO_DATE(rcd.RFRNC_YR || '-10-01', 'YYYY-MM-DD');
+            Nov1 := TO_DATE(rcd.RFRNC_YR || '-11-01', 'YYYY-MM-DD');
+            Dec1 := TO_DATE(rcd.RFRNC_YR || '-12-01', 'YYYY-MM-DD');
+            currYear := rcd.RFRNC_YR;
+        END IF;
```
karlmdavis, line 144: Fun trick. Let me know if it actually seems to help -- I'm curious!

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-854-Fix-PartD-Enrollment-Smaller-Transactions/05_process_ccw_table.sql
```diff
@@ -0,0 +1,387 @@
...
+            rcd.PTD_SGMT_DEC_ID,
+            rcd.RDS_DEC_IND,
+            rcd.CST_SHR_GRP_DEC_CD,
+            rcd.META_DUAL_ELGBL_STUS_DEC_CD);
+            
+        rcd_cnt := rcd_cnt + 1;
+        
+        if rcd_cnt % 10000 = 0
+        THEN
+            COMMIT;
```
karlmdavis, line 366: Doesn't this need to be followed by another `BEGIN`? Without it, I think you're running non-transactionally.

karlmdavis (commented):
I have a question.

PR #598, opened 2021-05-13
[BFD-855] Address memory consumption issues processing the one-time CCW load

mackec1:
### Change Details

To minimize memory consumption during the one-time CCW load process, setup outer and inner loop processing such that a psql transaction will be committed every 'n' records from the inner loop processing of records derived from a psql cursor. This is accomplished by wrapping inner loop processing within a BEGIN-END block which by definition creates a transaction; upon exiting the inner loop, an implied COMMIT fires and we log the Transaction ID.

This PR addresses a shortcoming of BFD-766 which caused an out of memory panic due to the size of the transaction buffer.

### Acceptance Validation

Ensure the processing logic makes sense and performs as expected.

### Feedback Requested

Errors in logic or suggestions to make this process run faster with a minimal memory footprint.

### External References

- [BFD-855](https://jira.cms.gov/browse/BFD-55)

### Security Implications


karlmdavis (changes requested):
Apologies: but I don't trust myself to find the differences between this version and the previous ones. Can you please rework this PR to put all of them into the same directory, so that Git can actually find the diffs for us itself?

PR #599, opened 2021-05-13
[BFD-855] Fix one-time CCW BeneMonthly load process to minimize transaction buffer size

mackec1:
### Change Details

BFD-766 detailed the steps to process two one-time CCW Beneficiary Monthly files; the las step in that process caused the postgres thread to terminate due to significant memory consumption from the postgres transaction buffer. This PR provides a fix to reduce the size of the transaction buffer.

In addition some code directory cleanup was done to remove a /sql directory that was inadvertently checked in.

### Acceptance Validation

Check logic and flow with particular attention to transaction boundaries.

### Feedback Requested

Suggestions and/or logic concerns.

### External References

- [BFD-855](https://jira.cms.gov/browse/BFD-855)

### Security Implications


apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-766-fix-beneficiary-monthly-data/05_process_ccw_table.sql
```diff
@@ -90,256 +90,324 @@ DECLARE
...
+                
+            END LOOP;   -- end <<inner>>
+
+            RAISE INFO 'implied COMMIT...Transaction ID: %', TXID_CURRENT();
+        
+            IF rcd_cnt % 20000 = 0
             THEN
-                RAISE INFO 'Record Count: % ...', rcd_cnt;
+                RAISE INFO 'Record Count: % ...%', rcd_cnt, NOW();
             END IF;
```
rtravitz, line 392: Was this too noisy to have run every time it exits the inner loop? Right now it's out of step with `MAX_INNER`.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-766-fix-beneficiary-monthly-data/05_process_ccw_table.sql
```diff
@@ -90,256 +90,324 @@ DECLARE
...
+                    Mar1 := TO_DATE(rcd.RFRNC_YR || '-03-01', 'YYYY-MM-DD');
+                    Apr1 := TO_DATE(rcd.RFRNC_YR || '-04-01', 'YYYY-MM-DD');
+                    May1 := TO_DATE(rcd.RFRNC_YR || '-05-01', 'YYYY-MM-DD');
+                    Jun1 := TO_DATE(rcd.RFRNC_YR || '-06-01', 'YYYY-MM-DD');
+                    Jul1 := TO_DATE(rcd.RFRNC_YR || '-07-01', 'YYYY-MM-DD');
+                    Aug1 := TO_DATE(rcd.RFRNC_YR || '-08-01', 'YYYY-MM-DD');
+                    Sep1 := TO_DATE(rcd.RFRNC_YR || '-09-01', 'YYYY-MM-DD');
+                    Oct1 := TO_DATE(rcd.RFRNC_YR || '-10-01', 'YYYY-MM-DD');
+                    Nov1 := TO_DATE(rcd.RFRNC_YR || '-11-01', 'YYYY-MM-DD');
+                    Dec1 := TO_DATE(rcd.RFRNC_YR || '-12-01', 'YYYY-MM-DD');
```
rtravitz, line 157: This is a nice change readability change in addition to the potential perf improvement from only solving it once.

rtravitz (approved):
Shouldn't matter because of the idempotent nature of this script that we talked about last time. But knowledge checking... the commit on the sub-transaction doesn't write in a way that's visible outside of the main transaction, right? It's just a mechanism here for us to free up memory?

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-766-fix-beneficiary-monthly-data/05_process_ccw_table.sql
mackec1: Yea...the inner loop is keeping the transaction size down, while the outer loop is kinda proving a heartbeat/progress. So with 130+ M rcds, the log would be pretty noisy! Hence the discrepancy between the 10k transaction and the 20k progress.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-766-fix-beneficiary-monthly-data/05_process_ccw_table.sql
mackec1: Well the old way was 130+ M * 12 == A VERY BIG NUMBER!

So I was looking for any/all improvements 'cuz this thing could run too long anyways!

mackec1:
> Shouldn't matter because of the idempotent nature of this script that we talked about last time. But knowledge checking... the commit on the sub-transaction doesn't write in a way that's visible outside of the main transaction, right? It's just a mechanism here for us to free up memory?

> So with a Begin-END block that contains an EXCEPTION handler, you will effectively create a 'transaction block'; so the inner loop represents that transaction block; basically committing every 10k records; the outer block defers to the innermost transaction block(s). So this should limit out our rollback buffer to 10k records.

And yes, the idempotent nature of the script means it can be run multiple times without effecting the BeneficiaryMonthly table since **the data we really want**, is contained in the one-time CCW temp table...so updating recordID# 1000 with data from CCW multiple times will only cause a 'real change' one time (the first time). The by-product of the one time 'real change' also means the we will not generate spurious audit data since the UPDATE trigger will only fire when it detects data change in a given row.

karlmdavis (approved):
Really not clear to me that it _is_ implicitly committing -- this stuff just doesn't appear to be well-covered in PostgreSQL's docs. But we talked through that and have an alternative approach to address it, if needed.

Otherwise, looks great! Let's YOLO it.

PR #602, opened 2021-05-17
BFD-867: Disable High Security Mode within New Relic Agent

jzulim:
### Change Details

This change disabled HSM (high security mode) via setting the `NEW_RELIC_HIGH_SECURITY` environment variable to `false`

### Acceptance Validation

Is this the correct location to set this?
I see this is for bfs-server, but is there another location for bod-pipeline that needs to be made?

### Feedback Requested

N/A

### External References

- [BFD-867](https://jira.cms.gov/browse/BFD-867)

### Security Implications

New Relic will now have high security mode DISABLED as outlined in the [SIA](https://confluence.cms.gov/display/BFD/2021-04-28+New+Relic-+Disable+High+Security+Mode) 


keithdadkins (approved):
yes!

rtravitz (approved):
Pipeline is only running metrics exporter and not the New Relic Java Agent, so there shouldn't be another place where this is set. Quick global search of the repo confirms.

Looks like the [default is false](https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file/#cfg-enable_high_security), so we could also delete this line and expect similar behavior, I think.

jzulim:
> Pipeline is only running metrics exporter and not the New Relic Java Agent, so there shouldn't be another place where this is set. Quick global search of the repo confirms.
> 
> Looks like the [default is false](https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file/#cfg-enable_high_security), so we could also delete this line with the overriding environment variable and expect similar behavior, I think.

Thanks for that clarity. Personally, seeing this set to false feels helpful, especially if you aren't familiar with the default behavior. Thanks!

PR #605, opened 2021-05-17
[BFD-866] Modify CCW load fix-up to use combination of shell and stored function to mitigate out-of-memory issue"

mackec1:
### Change Details

The pure SQL solution continues to fail from out-of-memory resources. This version attempts to mitigate the memory consumption / transaction issues we have seen by using a combination of a shell script that performs loop processing, calling a stored function that updates 20k records at a time.

At each iteration of 20k records, the psql script terminates entirely, hopefully returning any memory/transaction overhead back to the system.

### Acceptance Validation

There are two main files to focus on:

- 05_create_update_proc_and_function.sql : performs the looping / update of 20k CCW_LOAD_FILE records for a given year.
- 06_do_update.sh : bash shell script that continually invokes function _update_bene_monthly_with_delete_ until all CCW_LOAD_TEMP files are processed for a given year.

### Feedback Requested

Focus on correctness of / holes in the bash script.

Verify the SQL processes the update(s) correctly.

### External References

- [BFD-866](https://jira.cms.gov/browse/BFD-866)

### Security Implications


rtravitz (approved):
Ran through the new steps, and it worked as advertised for the sample set of seven. The 2020 run did the first five, and the 2019 run did the remaining two. I have entries in the audit table, and no more entries in the temp load table.

No test of the looping functionality from that sample run, but it looks right when reading it.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-766-fix-beneficiary-monthly-data/Playbook.txt
```diff
@@ -145,7 +138,7 @@ action(s) need to be taken.
 
 Q. What happens if Step 6 fails?
 
-   Step 6 can be run miultiple times; it is effectively running an idempotent UPDATE so
+   Step 6 can be run multiple times; it is effectively running an idempotent UPDATE so
    re-running would simply update data with the same data as a previous run.
```
karlmdavis, line 142: It's technically not idempotent anymore.
```suggestion
   Step 6 can be run multiple times; it will pick back up where it left off.
```

karlmdavis (approved):
Great work @mackec1!

I'm okay with this if you commit the two GitHub suggestions I added (or if you address those concerns in some other way.)

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-766-fix-beneficiary-monthly-data/Playbook.txt
mackec1: Good point...we'd have to reload the CCW temp table with the csv file(s)..

cbrunefearless (dismissed):
lgtm

karlmdavis (changes requested):
You mentioned on Slack that `set -e` was breaking things, which sounds pretty sketchy to me. Can you figure out why that is and get it (or some similar error handling) added back in?

Please also run the thing through `shellcheck` (and sorry, I should have suggested that earlier).

cbrunefearless (approved):
lgtm

mackec1 (commented):
set -e back in; must have been some other issue when I removed it.

karlmdavis (approved):
🤘

PR #606, opened 2021-05-18
Changed Beneficiary.beneficiaryMonthly from Eager to Lazy

sharshar-sb:
### Change Details

In almost every patient coverage search, each Beneficiary was also retrieving information from the BeneficiaryMonthly table. This appeared to be because Beneficiary.beneficiaryMonthly was set to Eager. I changed this to Lazy load to prevent every Beneficiary search result from also loading the BeneficiaryMonthly data.

### Acceptance Validation

The Patient Coverage endpoint should work the same, just be faster

### Feedback Requested

Can you see this interfering with any other Beneficiary searches?

### External References

- [BFD-826](https://jira.cms.gov/browse/BFD-826)

### Security Implications

None, this actually reduces the amount of data being loaded.


cbrunefearless (approved):
lgtm

jzulim:
@sharshar-sb thanks for putting this up, excited to see how this might improve bene fetching for your group. Just noticed some failing tests that need some 👀

mackec1 (approved):
Looks good; thx Laila.

mackec1 (approved):
Nice! Thanks for digging into this.

cbrunefearless (approved):
lgtm

karlmdavis (changes requested):
Great catch that we were loading the data when we didn't need it. :star2: 

But, as written, this PR seems off to me... aren't you still manually triggering the load in a bunch of places where the data isn't actually needed?

And for those places where it _is_ needed, we should **not** be triggering it manually for each bene, as that leads to a query count explosion. Instead, we'd want to be rejigerring Hibernate to load the data in the first/only query via a join.

PR #607, opened 2021-05-18
[BFD-868] Only run github workflows when needed

keithdadkins:
### Change Details
As an application developer, I should not need to wait for terraform/ops related workflows to run when pushing java code.
And as an ops engineer, I should not need to wait for java workflows when updating files in ops/ or insights/.

This PR only runs workflows when changes are made to files in specified folders.

### External References

- [BFD-868](https://jira.cms.gov/browse/BFD-868)
- [BFD-1161](https://jira.cms.gov/browse/BFD-1161)

### Security Implications


rtravitz (commented):
Yes, a thousand times yes. It looks like all relevant Java code is under `apps`, so that makes sense.

Same for the terraform linting. I could see an argument there for maybe only running when a `.tf` file changes? I can also see an argument for only running the tf validation on `ops/terraform/**` and `insights/terraform/**`. No reason to run those checks on ansible or docs changes under `ops`, right?

jzulim:
> Yes, a thousand times yes. It looks like all relevant Java code is under `apps`, so that makes sense.
> 
> Same for the terraform linting. I could see an argument there for maybe only running when a `.tf` file changes? I can also see an argument for only running the tf validation on `ops/terraform/**` and `insights/terraform/**`. No reason to run those checks on ansible or docs changes under `ops`, right?

^^ I'd second this consideration, why not eh?

cbrunefearless (dismissed):
lgtm

keithdadkins:
> Same for the terraform linting. I could see an argument there for maybe only running when a `.tf` file changes? I can also see an argument for only running the tf validation on `ops/terraform/**` and `insights/terraform/**`. No reason to run those checks on ansible or docs changes under `ops`, right?

you asked for it :P

cbrunefearless (approved):
lgtm

.github/workflows/insights-code-analysis.yml
```diff
@@ -0,0 +1,188 @@
...
+  ## CHANGED FILES
+  # searches changed files and add outputs for use by other jobs
+  # e.g., ${{ needs.changed_files.outputs.tf }} would output all changed terraform (*.tf) files
+  changed_files:
+    name: checking
+    runs-on: ubuntu-latest
+    outputs:
+      all: ${{ steps.changed_files.outputs.all }}
+      sh: ${{ steps.changed_files.outputs.sh }}
+      tf: ${{ steps.changed_files.outputs.tf }}
```
karlmdavis, line 23: Is this worth the extra complexity? Our TF jobs are pretty fast, and I think this part may be overkill, compared to the `paths` qualifier above. (If that works how I think it works.)

keithdadkins: It allows us to plug in other linters as needed. And it can speed things up (shell checks can take a while). Was trying to future proof.

keithdadkins: Also, I needed a way to find all the shell scripts instead of relying on file extensions. This method allows me to find all shell scripts, `foo` and `foo.sh` as long as they have shebangs.

keithdadkins:
And fyi, these workflows are run on this PR. And, yeah, this is very much likely over engineered. This was my first foray into github actions. The complexity comes down to a few things:
1. Multiple teams running different versions of terraform.
2. Being able to find and lint shell scripts wether or not they have .sh file extensions.
3. I wanted to see the full output from terraform checks instead of stopping on the first issue found.
4. I wanted to pretty print shell check results as a PR comment, just like you would expect to see if you ran shellcheck from the command line.
5. I wanted to future proof these checks by being able to "plug in" other linters in the future.

That said, I think a better approach would be to simply have a bunch of different workflows (as Ben outlined in the recent sprint demo). 

And FIY, this was not planned sprint work and it was something I did over a weekend. So if it's deemed too complicated, we can shelve this for later.

keithdadkins:
Removed all the extra stuff.

christrotteradhoc (approved):

njdister (approved):

njdister:
I think this is running into an issue where the Java check isn't running, as expected, but it's still "required" due to the way we have branch protection set up.  I'm not sure you can have conditional required checks on branch protection.

keithdadkins:
We need to disable branch protections or have an admin (if configured to do so) merge this. I'll check with Stephen today or tomorrow to get it done.

keithdadkins:
Closing (see https://github.com/CMSgov/beneficiary-fhir-data/pull/773)

PR #609, opened 2021-05-20
BFD-734: BFD-734-synthetic-data-id-ranges

karlmdavis:
### Change Details

Just what it says on the tin.

### Acceptance Validation

Let me know if I goofed or missed something.

### Feedback Requested

(see previous)

### External References

- [BFD-734](https://jira.cms.gov/browse/BFD-734)

### Security Implications

This has no identified security concerns.


JeremyIR (commented):
Citing where the queries were derived from makes sense. In case this answer on Stack Overflow gets deleted, or Stack Overflow somehow doesn't stand the test of time, should we consider citing the link, but also adding a screenshot of the answer or rewriting it in markdown for context?

karlmdavis:
> @JeremyIR Citing where the queries were derived from makes sense. In case this answer on Stack Overflow gets deleted, or Stack Overflow somehow doesn't stand the test of time, should we consider citing the link, but also adding a screenshot of the answer or rewriting it in markdown for context?

Good question! Two thoughts:

1. Stack Overflow's content is licensed as follows (as seen at the bottom of every SO page): "user contributions licensed under cc by-sa." So if SO ever died, it's all but certain that any number of sites would end up re-hosting their content. In fact, they already do, which can be frustrating when searching the web sometimes, but nevertheless. (The [Wayback Machine](https://archive.org/web/) would probably end up with all the content, if nothing else did.)
2. That just sounds like a lot of work, and at least for me, more work than it's worth. YMMV.

rtravitz (approved):
Appreciated the thorough explanations and extra reading. This looks good to me.

mackec1 (approved):
lots of SQL here....nice work!

jzulim (approved):
Walked through the SQL, logic looks good and results are interesting. Seems we'll have some work to do for claimGroupId

PR #612, opened 2021-05-21
DCGEO-59: pre-adj claim transformers

KyRobbins:
### Change Details

Initial implementation for FISS related transformers using rda entities [https://github.com/CMSgov/beneficiary-fhir-data/tree/kylerobbins/dcgeo-62-pre-adj-claim-entities].  Transforms Fiss claims and proc codes into Claim and ClaimResponse fhir resources.

### Acceptance Validation

The transformers should adequately transform the FISS claims and proc codes into the appropriate fhir resources.

### Feedback Requested

Are solutions in line with existing solutions or otherwise include desired implementations?

### External References

- [DCGEO-59](https://jira.cms.gov/browse/DCGEO-59)

### Security Implications

PHI/PII pulled from bfd database as required by incoming requests, no immediate concerns.

Mockito - Mocking library for unit testing.


jzulim:
Is this PR ready for review? It's currently in draft so wanted to dbl check.

Gabasol:
hey @jzulim!  I will let @KyRobbins confirm but I don't believe this is ready for BFD review.

KyRobbins:
> Is this PR ready for review? It's currently in draft so wanted to dbl check.

This is still a WIP, we're still needing the MCS models to complete the transformer implementations.

apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -209,6 +214,20 @@
...
+			<scope>test</scope>
+		</dependency>
+		<!-- Hibernate brings in an older version of byte-buddy which seems
+		to conflict with mockito, which wants version 1.11.0.  Bringing it
+		in as a first class dependency to resolve the issue. -->
+		<dependency>
+			<groupId>net.bytebuddy</groupId>
+			<artifactId>byte-buddy</artifactId>
+			<version>1.11.0</version>
+		</dependency>
```
brianburton, line 230: Good catch on this dependency issue.  They can be painful to debug.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,150 @@
...
+    STATUS_TEXT = new HashMap<>();
+    STATUS_TEXT.put('a', "Active");
+    STATUS_TEXT.put('s', "Suspend");
+    STATUS_TEXT.put('p', "Paid");
+    STATUS_TEXT.put('d', "Denied");
+    STATUS_TEXT.put('i', "Inactive");
+    STATUS_TEXT.put('r', "Reject");
+    STATUS_TEXT.put('t', "RTP");
+    STATUS_TEXT.put('m', "Move");
+  }
```
brianburton, line 40: Assuming this is meant to be immutable this should be a Guava ImmutableMap rather than a HashMap.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,150 @@
...
+    STATUS_TO_OUTCOME.put(' ', ClaimResponse.RemittanceOutcome.QUEUED);
+    STATUS_TO_OUTCOME.put('a', ClaimResponse.RemittanceOutcome.QUEUED);
+    STATUS_TO_OUTCOME.put('s', ClaimResponse.RemittanceOutcome.PARTIAL);
+    STATUS_TO_OUTCOME.put('p', ClaimResponse.RemittanceOutcome.COMPLETE);
+    STATUS_TO_OUTCOME.put('d', ClaimResponse.RemittanceOutcome.ERROR);
+    STATUS_TO_OUTCOME.put('i', ClaimResponse.RemittanceOutcome.PARTIAL);
+    STATUS_TO_OUTCOME.put('r', ClaimResponse.RemittanceOutcome.ERROR);
+    STATUS_TO_OUTCOME.put('t', ClaimResponse.RemittanceOutcome.ERROR);
+    STATUS_TO_OUTCOME.put('m', ClaimResponse.RemittanceOutcome.PARTIAL);
+  }
```
brianburton, line 55: Assuming this is meant to be immutable this should be a Guava ImmutableMap rather than a HashMap.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,150 @@
...
+   * @param claimEntity the FISS {@link PreAdjFissClaim} to transform
+   * @return a FHIR {@link ClaimResponse} resource that represents the specified claim
+   */
+  @Trace
+  static ClaimResponse transform(MetricRegistry metricRegistry, Object claimEntity) {
+    Timer.Context timer = metricRegistry.timer(METRIC_NAME).time();
+
+    if (!(claimEntity instanceof PreAdjFissClaim)) {
+      throw new BadCodeMonkeyException();
+    }
```
brianburton, line 68: Maybe move the instanceof/parameter validation before the timer start?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private static List<Extension> getExtension(PreAdjFissClaim claimGroup) {
+    return Collections.singletonList(
+        new Extension()
+            .setUrl("https://dcgeo.cms.gov/resources/variables/fiss-status")
+            .setValue(
+                new Coding(
+                    "https://dcgeo.cms.gov/resources/variables/fiss-status",
+                    "" + claimGroup.getCurrStatus(),
+                    STATUS_TEXT.get(Character.toLowerCase(claimGroup.getCurrStatus())))));
```
brianburton, line 108: Just curious: what happens if the `claimGroup.getCurrStatus()` isn't a valid key so `get` returns null?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimTransformerV2.java
```diff
@@ -0,0 +1,230 @@
...
+    total.setCurrency("USD");
+
+    return total;
+  }
+
+  private static List<Resource> getContainedProvider(PreAdjFissClaim claimGroup) {
+    return Collections.singletonList(
+        new Organization()
+            .setIdentifier(
+                Arrays.asList(
```
brianburton, line 119: Style suggestion: I prefer to use`ImmutableList.of()` for these sorts of throw away lists rather than a mix of `Collections.singletonList ()` and `Arrays.asList()`.  It requires fewer imports, it's more consistent, the resulting list is clearly and always immutable (vs `Collections.singletonList()` which is but `Arrays.asList()` only sort of is), and the `ImmutableList.of()` method is overloaded with different numbers of parameters so the JVM doesn't have to create any throw away intermediate arrays for a variadic parameter.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/PreAdjClaimDaoTest.java
```diff
@@ -0,0 +1,165 @@
...
+import javax.persistence.criteria.Root;
+import org.junit.Test;
+
+public class PreAdjClaimDaoTest {
+
+  @Test
+  public void shouldInvokeGetByIdForClaimType() {
+    String claimId = "123";
+
+    MetricRegistry registry = new MetricRegistry();
```
brianburton, line 30: Maybe include a check that the timer metric was updated?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResponseResourceProviderTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+public class R4ClaimResponseResourceProviderTest {
+
+  @Test
+  public void shouldReturnCorrectIBaseResourceType() {
+    assertEquals(new R4ClaimResponseResourceProvider().getResourceType(), ClaimResponse.class);
+  }
+
+  @Test
+  public void shouldHaveReadMethodAnnotatedWithRead() throws NoSuchMethodException {
```
brianburton, line 39: I like these annotation checks as unit tests!

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResponseResourceProviderTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+  @Test
+  public void shouldThrowIllegalArgumentExceptionIfIdPartIsBlank() {
+    IdType id = new IdType(null, null, " ", null);
+
+    Exception expected = new IllegalArgumentException("Resource ID can not be null/blank");
+    Exception actual =
+        AssertUtils.catchExceptions(() -> new R4ClaimResponseResourceProvider().read(id, null));
+
+    AssertUtils.assertThrowEquals(expected, actual);
```
brianburton, line 116: These exception tests are good.  They are pretty repetitive though.  Maybe add a method in AssertUtils for the pattern?  (i.e. takes expected and Callable as params).

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
mackec1: kyle. BFD has been trying to use enums especially when they are related to Codable logic that requires a code, a display text, and a system url. We have FHIR-CARIN types like:
gov.cms.bfd.server.war.commons.carin.C4BBAdjudication.java

And we have some that don't fit into FHIR-CARIN such as:
gov.cms.bfd.server.war.commons.RaceCategory.java

ericbuckley: @mackec1 thanks Chris this is really helpful.  When possible we of course want to use existing carin types, however in this case I believe [CC4BBAdjudication](http://hl7.org/fhir/us/carin-bb/STU1/ValueSet-C4BBAdjudication.html) represents "the amount fields submitted for adjudication".   Our [outcome field](http://hl7.org/fhir/valueset-remittance-outcome.html) is meant to track the "state of the claim within the adjudication process".

Is this the field that you are using for [EOB.outcome](http://hl7.org/fhir/explanationofbenefit-definitions.html#ExplanationOfBenefit.outcome)?

mackec1:
Eric.

Yea, the choice of providing C4BBAdjudication was a poor choice here; I was
basically just trying to convey two types of enums that we implemented.

1) C4BBAdjudication was an example where your could go to the FHIR v4 spec,
and see exactly the codes, system URL and display text  that is available
form the spec as is, and created as an enum.

2) the RaceCategory was an example where we had to piece together
definitions of codes, etc. to be able to create the Coadable.

So yes, the current use of Adjudication does not meet what you'll need, but
may provide some context in design/implementation going forward.

Thanks.

Chris

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
KyRobbins: This is a good question, we'll need to define does it return a default value, or just empty.  Also, I don't quite remember, was there a question of if the current status could be null?  I know our schema currently has it as `not null`, but if it could, that would be another issue as well to address.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimTransformerV2.java
KyRobbins: It's a good suggestion.  I've just naturally been prone to Collections and Arrays because they are part of the JDK.  I'm not sure about the efficiency aspect of it, `Arrays.asList()` just directly sets the field value to the passed varargs array, where as `ImmutableList.of()` does a fair bit more than that including some validations and switching.

brianburton: ImmutableList.of() just creates an ImmutableList containing the values passed to it and returns the list.  Difference is the list is guaranteed to be immutable.  Generally immutable data is safer data and best when you don't require mutability.  Here mutability isn't required.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PreAdjFissClaimResponseTransformerV2.java
brianburton: Good point about the claim status being non-null.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/PreAdjClaimDaoTest.java
KyRobbins: We could (though not for this specific test, as the target method doesn't invoke the `MetricRegistry`).  The only downside is that the `MetricRegistry` class is not easily mockable due to use of some private inner classes (`Timer.Context`), and it's recorded using a static invocation (also difficult to mock without bringing in a hefty testing library like PowerMock).  If there was ever an issue with the dependent classes, it could break a lot of tests making the cause a bit more ambiguous. 
 Might be worth looking at that as a test case for an Integration Test or at least something "above" a unit test.

brianburton: OK if the timer isn't being updated here you can disregard my comment.  WRT testing a metric though it's trivial (at least for Meters) to query the registry for the metric and check it's value.  For example for a Meter I did this:

````
  private void assertMeterReading(long expected, String meterName) {
    long actual = appMetrics.meter(meterName).getCount();
    assertEquals("Meter " + meterName, expected, actual);
  }
````

File: apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResponseResourceProviderTest.java (hunk 9 above)
Comment (KyRobbins) [reply apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ClaimResponseResourceProviderTest.java:116]: We could, though I think these tests are pretty small already.  In the end, passing the expected and callable as parameters to another testing utility method wouldn't be any more terse then if I just modified the current test to have the `expected` and `actual` values inline in the `assertThrowEquals()` invocation.  I thought about parameterized tests, but the built in JUnit one isn't ideal, as it runs every test for every parameter if I remember correctly.  There are solutions to get around this like using an enclosed test runner, but that ends up getting rather messy looking adding extra layers of complexity to the test.  There are also some pretty solid parameterized test libraries out there that work a bit better, but I didn't think it was worth pulling in another library for this at the moment.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java [hunk 10]
```diff
@@ -0,0 +1,147 @@
...
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
+import javax.persistence.PersistenceContext;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.ClaimResponse;
+import org.hl7.fhir.r4.model.IdType;
+
+public abstract class AbstractR4ResourceProvider<T extends IBaseResource>
```
ericbuckley, line 29: Is this just for ClaimResponse resources or also for Claim resources?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,153 @@
...
+public class FissClaimResponseTransformerV2 {
+
+  private static final String METRIC_NAME =
+      MetricRegistry.name(FissClaimResponseTransformerV2.class.getSimpleName(), "transform");
+
+  private static final Map<Character, String> STATUS_TEXT;
+
+  static {
+    STATUS_TEXT =
+        ImmutableMap.<Character, String>builder()
```
ericbuckley, line 32: these values will be capitalized when we receive them from RDA

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,153 @@
...
+            .put('a', "Active")
```
ericbuckley, line 33: `a` stands for "Accepted" or "Routing"

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,153 @@
...
+            .put('s', "Suspend")
+            .put('p', "Paid")
+            .put('d', "Denied")
+            .put('i', "Inactive")
+            .put('r', "Reject")
```
ericbuckley, line 38: lets rename this to "Rejected"

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,153 @@
...
+            .put('t', "RTP")
```
ericbuckley, line 39: lets expand RTP to "Returned to Provider"

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,153 @@
...
+            .put('m', "Move")
```
ericbuckley, line 40: we may also see an empty char `' '`,  we can just show the "Unknown in that case

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimResponseTransformerV2.java
```diff
@@ -0,0 +1,153 @@
...
+  private static final Map<Character, ClaimResponse.RemittanceOutcome> STATUS_TO_OUTCOME;
+
+  static {
+    STATUS_TO_OUTCOME =
+        ImmutableMap.<Character, ClaimResponse.RemittanceOutcome>builder()
+            .put(' ', ClaimResponse.RemittanceOutcome.QUEUED)
+            .put('a', ClaimResponse.RemittanceOutcome.QUEUED)
+            .put('s', ClaimResponse.RemittanceOutcome.PARTIAL)
+            .put('p', ClaimResponse.RemittanceOutcome.COMPLETE)
+            .put('d', ClaimResponse.RemittanceOutcome.ERROR)
```
ericbuckley, line 53: I discussed this with Nicole, and for the time being we're going to remap the `d, r and t` states to "COMPLETE".  I've updated the table in the confluence page to reflect this.  Sorry for the last minute change

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimTransformerV2.java
```diff
@@ -0,0 +1,229 @@
...
+  private static List<Claim.ProcedureComponent> getProcedure(PreAdjFissClaim claimGroup) {
+    List<Claim.ProcedureComponent> procedure = new ArrayList<>();
+
+    // Sort proc codes by priority prior to building resource
+    List<PreAdjFissProcCode> procCodes = new ArrayList<>(claimGroup.getProcCodes());
+    procCodes.sort(Comparator.comparingInt(PreAdjFissProcCode::getPriority));
+
+    int i = 0;
+
+    for (PreAdjFissProcCode procCode : procCodes) {
```
ericbuckley, line 212: minor; but if we need to initialize an index for the sequence anyways, would the following be preferable?
```
for(int i=0;  i<procCodes.size(); i++)
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimTransformerV2.java
```diff
@@ -0,0 +1,43 @@
...
+      MetricRegistry.name(McsClaimTransformerV2.class.getSimpleName(), "transform");
+
+  /**
+   * @param metricRegistry the {@link MetricRegistry} to use
+   * @param claimEntity the MCS {@link PreAdjMcsClaim} to transform
+   * @return a FHIR {@link Claim} resource that represents the specified claim
+   */
+  @Trace
+  static Claim transform(MetricRegistry metricRegistry, Object claimEntity) {
+    // TODO: Update this check when entity available
```
ericbuckley, line 22: we can reference DCGEO-98 for the MCS TODOs

ericbuckley (changes requested):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
KyRobbins: It's for both, they each have a concrete implementation that extend it, specifying their specific Resource types in the parameterized value.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/FissClaimResponseTransformerV2.java
KyRobbins: The casing is converted to lowercase in logic to avoid any potential issue of unexpected casing differences.
`claim.setOutcome(STATUS_TO_OUTCOME.get(Character.toLowerCase(claimGroup.getCurrStatus())))`

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProviderTest.java
```diff
@@ -0,0 +1,314 @@
...
+    MetricRegistry mockMetricRegistry = mock(MetricRegistry.class);
+
+    ReflectionTestUtils.setField(providerSpy, "entityManager", mockEntityManager);
+    ReflectionTestUtils.setField(providerSpy, "metricRegistry", mockMetricRegistry);
+
+    doNothing().when(providerSpy).setResourceType();
+
+    providerSpy.init();
+
+    verify(providerSpy, times(1)).init();
```
brianburton, line 79: Is this verifying the right call?  Shouldn't it be verifying `setResourceType()`?

brianburton (approved):
Looks good to me.

ericbuckley (approved):
looks great, thanks Kyle!

apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -209,6 +214,21 @@
                 </exclusion>
             </exclusions>
         </dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>3.10.0</version>
+			<scope>test</scope>
+		</dependency>
+		<!-- Hibernate brings in an older version of byte-buddy which seems
```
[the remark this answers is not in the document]
cbrunefearless: Do we need a SIA?

apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -209,6 +214,21 @@
...
+		to conflict with mockito, which wants version 1.11.0.  Bringing it
+		in as a first class dependency to resolve the issue. -->
+		<dependency>
+			<groupId>net.bytebuddy</groupId>
```
cbrunefearless, line 227: Do we need a Sia for this?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
```diff
@@ -0,0 +1,154 @@
+package gov.cms.bfd.server.war.r4.providers.preadj;
```
cbrunefearless, line 1: just a question, what is preadj?

cbrunefearless: ahh preadjudicated.. sorry maybe we should spell that out

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
```diff
@@ -0,0 +1,154 @@
...
+    if (claimIdText == null || claimIdText.trim().isEmpty())
+      throw new IllegalArgumentException("Resource ID can not be null/blank");
+
+    Matcher claimIdMatcher = CLAIM_ID_PATTERN.matcher(claimIdText);
+    if (!claimIdMatcher.matches())
+      throw new IllegalArgumentException("Unsupported ID pattern: " + claimIdText);
+
+    String claimIdTypeText = claimIdMatcher.group(1);
+    Optional<ResourceTypeV2<T>> optional = parseClaimType(claimIdTypeText);
+    if (!optional.isPresent()) throw new ResourceNotFoundException(claimId);
```
cbrunefearless, line 131: just me nitpicking.. should we be consistent and move throw new to a new line?

apps/bfd-server/bfd-server-war/pom.xml
KyRobbins: ByteBuddy is already brought in by both hibernate and Mockito as a transitive dependency, this just brings it in as first class so that Mockito gets the right version of it (and only for testing scope).  That being said, I defer to others about if we should go into further analysis of it.

jzulim (approved):
Looks legit.

StewGoin:
No SIA needed for this, truly "new" libraries only impact testing harness (ByteBuddy is not really new to BBAPI).

cbrunefearless (approved):
LGTM

PR #614, opened 2021-05-24
DCGEO-26 (DRAFT): Adds FissClaimsRdaSink to write RDA FISS claims in database

brianburton:
**Note: This is a draft PR.  It requires that the PR for DCGEO-23 has been merged.**
**Note: 2021-06-10 The dependencies on the pipeline orchestration code changes have been removed from this PR. They can be added back again once the pipeline orchestration solution has been finalized. The code is ready for review (after DCGEO-23)and can be merged into master once approved.**

### Change Details

This change builds on DCGEO-23 to add an RdaSink implementation that writes FISS claims from the RDA API to the `pre_adj` schema.  Writing to the database involves conversion from the gRPC stub classes to Hibernate entity classes.  Validation and transformatino of values is done during this process as well.

**Refactoring of RifLoader:** The MBI values from the RDA claims needed to be hashed using the same algorithm and salt/pepper as the HICN.  To enable the use of the settings and algorithm used by RifLoader without adding a dependancy on `bfd-pipeline-rif-ccw` the hashing code was refactored into a new class `IdHasher` in the `bfd-pipeline-shared-utils` module.  Rather than have the code duplicated there this PR includes refactoring of `RifLoader` to use the `IdHasher` class as well.  Tests were adapted so that they could verify that the hash values are tested as before.

Key classes to review in this PR:

- `DataTransformer`: General class to validate and transform data and collect any errors detected so that they can be summarized all at once.
- `FissClaimTransformer`: Specific class to validate and transform `FissClaim` gRPC stubs class into `PreAdjFissClaim` Hibernate entity class.
- `FissClaimRdaSink`: Implementation of `RdaSink` that writes `PreAdjFissClaim` objects to the database in batches.
- `IdHasher`: Utility class that encapsulates the settings and algorithm used to hash sensitive ID strings.
- `DatabaseUtils`: Utility class to contain some common database code used by both RDA and RIF pipelines.
- `RifLoader`: This existing class from `bfd-pipeline-ccw-rif` was refactored to use `IdHasher` and `DatabaseUtils`.

### Acceptance Validation

- Are all of the claim fields transformed correctly?
- Is the `IdHasher` refactoring of `RifLoader` correct and non-breaking?
- Are the correct metrics collected?

### Feedback Requested

Nothing specific.

### External References

- [DCGEO-26](https://jira.cms.gov/browse/DCGEO-26)

### Security Implications

Code for computing hashes of various ID was refactored into a new class, `IdHasher`.

No new libraries are being used in this PR.

None

FISS claim data from the RDA API is now being written to the `pre_adj` database schema.

n/a

No


apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/IdHasher.java
```diff
@@ -0,0 +1,102 @@
...
+
+  public IdHasher(Config config) {
+    this.config = config;
+    secretKeyFactory = createSecretKeyFactory();
+  }
+
+  @VisibleForTesting
+  public static IdHasher createInstanceForTestingOnly() {
+    return new IdHasher(new Config(1000, "nottherealpepper".getBytes(StandardCharsets.UTF_8)));
+  }
```
KyRobbins, line 38: Might be better suited just having this in the test classes themselves and not in normal service code?  Could also be a testing package utility if there was a DRY concern.

brianburton: Good observation.  Yeah this one is all about DRY.  Test code isn't shared between maven modules without an explicit extra dependency in the pom, which I find to be painful.  It can be easier and harmless just to add a properly documented static method.

KyRobbins: It would be my own opinion that a single constructor call is not enough to be considered a DRY concern, I was more just speaking generally.  Personally, I think having testing code in deployed code is less desirable, and typically the DRY concept doesn't really apply to testing like it does to development code, and in fact I believe it's fine for tests to be very redundant and verbose for the sake of isolation.  Just a thought.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimTransformerTest.java
```diff
@@ -0,0 +1,201 @@
...
+import java.time.ZoneOffset;
+import java.util.Arrays;
+import org.junit.Before;
+import org.junit.Test;
+
+public class FissClaimTransformerTest {
+  // using a fixed Clock ensures our timestamp is predictable
+  private final Clock clock = Clock.fixed(Instant.ofEpochMilli(1621609413832L), ZoneOffset.UTC);
+  private final FissClaimTransformer transformer =
+      new FissClaimTransformer(clock, IdHasher.createInstanceForTestingOnly());
```
KyRobbins, line 25: Should we mock the hashing utility object instance so we're not testing code from a different class that has it's own tests?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/FissClaimRdaSink.java
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Writes the entire batch to the database in a single transaction.
+   *
+   * @param claims batch of claims to be written to the database
+   * @return the number of claims successfully written to the database
+   * @throws ProcessingException wrapped exception if an error takes place
+   */
+  @Override
+  public int writeBatch(Collection<PreAdjFissClaim> claims) throws ProcessingException {
```
KyRobbins, line 63: The logic in this class has decent complexity, do we want to create unit tests for this?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimTransformerTest.java
brianburton: The type of hashing used is directly relevant to this class so I like to include it.  If someone changed the hasher it would break this class.  Also I like to use the minimal number of mocks to get coverage.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -111,64 +105,20 @@ public RifLoader(MetricRegistry appMetrics, LoadAppOptions options) {
...
-  /**
-   * @param options the {@link LoadAppOptions} to use
-   * @param metrics the {@link MetricRegistry} to use
-   * @return a {@link HikariDataSource} for the BFD database
-   */
-  static HikariDataSource createDataSource(LoadAppOptions options, MetricRegistry metrics) {
-    HikariDataSource dataSource = new HikariDataSource();
-
     /*
-     * FIXME The pool size needs to be double the number of loader threads
```
ericbuckley, line 129: Thanks for addressing that

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/DataTransformer.java
```diff
@@ -0,0 +1,233 @@
...
+   *
+   * @param fieldName name of the field from which the value originates
+   * @param value non-null string value
+   * @param minLength minimum valid length for the string
+   * @param maxLength maximum valid length for the string
+   * @return true if the value has a valid length
+   */
+  private boolean lengthOk(String fieldName, String value, int minLength, int maxLength) {
+    final int length = value.length();
+    if (length < minLength || length > maxLength) {
```
ericbuckley, line 171: Did you mean to use `<=` and `>=`

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimTransformer.java
```diff
@@ -0,0 +1,93 @@
...
+    this.idHasher = idHasher;
+  }
+
+  public PreAdjFissClaim transformClaim(FissClaim from) {
+    final DataTransformer transformer = new DataTransformer();
+    final PreAdjFissClaim to = new PreAdjFissClaim();
+    transformer
+        .copyString("dcn", from.getDcn(), false, 1, 23, to::setDcn)
+        .copyString("hicNo", from.getHicNo(), false, 1, 12, to::setHicNo)
+        .copyCharacter("currStatus", from.getCurrStatus(), to::setCurrStatus)
```
ericbuckley, line 33: We don't have an example  yet, but the "Blank" status value might come in as "" rather than " ".

brianburton: Thanks.  Good to know.  The RDA database schema had this as a non-null char.  We might need to change our schema if the value isn't actually a char.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/DataTransformer.java
```diff
@@ -0,0 +1,233 @@
...
+   * Throws an exception if there are any transformation errors. Makes it easy to generate a
+   * standard exception following all of the transformation steps.
+   *
+   * @throws TransformationException if there are any transformation errors
+   * @return this
+   */
+  public DataTransformer throwIfErrorsPresent() throws TransformationException {
+    if (errors.size() > 0) {
+      throw new TransformationException(
+          String.format("failed with %d errors", errors.size()), getErrors());
```
ericbuckley, line 47: Would it help to log the DCN with this value, I'm sure at some point early on our assumptions about the data will be wrong and we'll immediately have to debug an invalid message.  Trying to think through how we can obtain more information about the message without logging any PII.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/IdHasher.java
```diff
@@ -0,0 +1,94 @@
...
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
+import org.apache.commons.codec.binary.Hex;
+
+/**
+ * Utility class to encapsulate the hash algorithm (PBKDF2WithHmacSHA256) and settings (iterations
+ * and salt) used to compute secure hashed versions of ID values such as HICN and MBI. Code adopted
+ * from RifLoader.
+ */
+public class IdHasher {
```
ericbuckley, line 17: Thank you for DRYing up this code

insights/terraform/projects/bcda/main.tf
```diff
@@ -121,7 +121,7 @@ resource "aws_lambda_function" "bcda_load_partitions" {
   role             = aws_iam_role.bcda_load_partitions.arn
   handler          = "index.handler"
   source_code_hash = data.archive_file.bcda_load_partitions.output_base64sha256
-  runtime          = "nodejs10.x"
+  runtime          = "nodejs12.x"
```
ericbuckley, line 124: why are we upgrading the runtime from 10 to 12?

ericbuckley (commented):
@brianburton this is looking great. I think the transformer logic works for now considering we have very little simple data to go by, but I'm guessing our validation will break down immediately once we have a larger de-identified dataset and/or real data.  Thinking through how to best debug these "invalid messages" problems is going to be valuable.

Also there are a bunch of changes to the `/ops` directory in this PR, I'm guessing the rda-api-skeleton branch needs to be merged with master?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/DataTransformer.java
brianburton: Thanks.  I have left this generic method unchanged but modified `FissClaimTransformer` to throw its own exception with the DCN included in the error message.

brianburton:
This has been supplanted by PR 659.

PR #634, opened 2021-06-09
DCGEO-28:  Adds mock RDA API server

brianburton:
### Change Details

A mock RDA API server will allow the DCGEO team to more easily test processes at various levels.  The server in this PR supports insertion of a mock API server within unit tests, integration tests, and as a stand alone process for load or end-to-end testing.  The server leverages protobuf's built in support for round trip conversion between JSON and java objects to support ingestion of sample data in NDJSON format (one object JSON per line of the input).

Most of the server implementation is in the `bfd-pipeline-rda-grpc` module.  However a new module, `bfd-pipeline-rda-grpc-server` has also been added.  The new module includes dependencies required to run the mock server as a stand-alone process.  This module is not intended to be deployed to production systems.

Key classes to review in this PR:

- `FissClaimSource`: Simple interface for sources of data for the mock server.
- `RdaService`: Simple RDA API implementation that serves FissClaims by sending all of the values returned by a `FissClaimSource`.
- `RdaServer`: Provides static factory methods to create mock servers for use in tests.
- `RandomFissClaimSource`: A `FissClaimSource` that returns completely random FissClaim objects.  The random objects aren't valid claims but are useful for load testing an ETL process or implementing certain tests.
- `JsonFissClaimSource`: A `FissClaimSource` that returns claims read from a NDJSON or a List of JSON values.
- `RdaServerApp`: Stand alone program to launch a mock RDA API server.

### Acceptance Validation

Nothing in particular.

### Feedback Requested

Nothing in particular.

### External References

- [DCGEO-28](https://jira.cms.gov/browse/DCGEO-28)

### Security Implications

As it is not meant to serve real data the mock gRPC server does not implement any authentication.

Nothing major. 

None.

JSON data can be fed through the server for testing purposes but nothing in deployment.

N/A

No


apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/DirectRdaLoadApp.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.Reader;
+import java.time.Duration;
+import java.util.Properties;
+
+/**
+ * Simple application that invokes an RDA API server and writes FISS claims to a database using
+ * RdaLoadJob. This program was written for testing purposes and is not planned to be executed in a
+ * production environment.
+ */
+public class DirectRdaLoadApp {
```
ericbuckley, line 22: So if I understand this correctly, this is just one use case to run unit tests on the server to make sure its working correctly.  We could create another LoadApp that runs the server on a process independent of the RDA client?  If for example we wanted to test that the two services can communicate over network calls?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/RandomFissClaimGenerator.java
```diff
@@ -0,0 +1,119 @@
...
+          .filter(x -> !SKIPPED_PROC_TYPE_SET.contains(x))
+          .collect(Collectors.collectingAndThen(toList(), ImmutableList::copyOf));
+
+  private final Random random;
+
+  public RandomFissClaimGenerator(long seed) {
+    this.random = new Random(seed);
+  }
+
+  public FissClaim randomClaim() {
```
ericbuckley, line 50: This may go outside of the scope of this class, but one question we're interested in knowing is at what point does the current UPSERT logic start to degrade because the number of UPDATES is much higher than the number of INSERTS. If there was a way to control the percentage of uniquely generated IDs in a run, we may be able to test that with this class

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/RdaServer.java
```diff
@@ -0,0 +1,45 @@
...
+
+import io.grpc.Server;
+import io.grpc.ServerBuilder;
+import io.grpc.inprocess.InProcessServerBuilder;
+import java.io.IOException;
+import java.util.function.Supplier;
+
+public class RdaServer {
+  /**
+   * Creates a local RDA API server on a random port for testing.
```
ericbuckley, line 11: why a random port, wouldn't we want to control this for our tests?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimStreamCallerIT.java
```diff
@@ -0,0 +1,75 @@
...
+            "test", () -> new JsonFissClaimSource(CLAIM_1 + System.lineSeparator() + CLAIM_2));
+    try {
+      final ManagedChannel channel = InProcessChannelBuilder.forName("test").build();
+      try {
+        final IdHasher hasher = new IdHasher(new IdHasher.Config(10, "justsomestring"));
+        final FissClaimTransformer transformer =
+            new FissClaimTransformer(Clock.systemUTC(), hasher);
+        final FissClaimStreamCaller caller = new FissClaimStreamCaller(transformer);
+        final GrpcResponseStream<PreAdjFissClaim> results = caller.callService(channel);
+        assertEquals(true, results.hasNext());
```
ericbuckley, line 61: pretty neat, for this test would you also be able to check the database rows created?

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/DirectRdaLoadApp.java
brianburton: Well this is a simple program that just connects to an RDA API server and dumps data to the database using the normal pipeline job.  As such it's great for testing the job without having to bring up a PipelineApp instance anywhere.  It could also be useful when we want to populate a test or sandbox database using the mock RDA API server since we could just run it locally with a connection to the database we are populating.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/RandomFissClaimGenerator.java
brianburton: That's an excellent point.  I tested that out locally by adjusting the length of the DCN generated by `RandomFissClaimGenerator` (fewer digits mean more repeats) but that's not ideal at all.  An explicit setting as you propose would be easier to use.  Maybe we could add that when extending the server to also support MCS?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/RdaServer.java
brianburton: There is an overload of this method farher down that accepts a specific port to use.  The random port option is intended for IT's rather than any sort of long term deployment of the server on an test instance.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimStreamCallerIT.java
brianburton: Yes a complete end-to-end IT could be added.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/RandomFissClaimGenerator.java
ericbuckley: yeah that sounds great.   with the current IT tests I'm less sure of its value, I'd rather see this setup with a postgres DB before we start drawing any conclusions.  so something to iterate on in the future

ericbuckley (dismissed):
nice work

KyRobbins (dismissed):
looks good

KyRobbins (approved):

KyRobbins (approved):

ericbuckley (approved):
👍

cbrunefearless (approved):
lgtm

PR #644, opened 2021-06-11
DCGEO-66: search Claim/ClaimResponse by MBI

KyRobbins:
### Change Details

Add the ability to search for Claim and ClaimResponse fhir resources by patient mbi(optional/dev use)/mbhash(default) with an optional lastUpdated date range.  Service date range not yet supported.

Dependent on: https://github.com/CMSgov/beneficiary-fhir-data/pull/612

### Acceptance Validation

Feature should effective return bundled resources for the given patient mbi/mbihash for a given lastUpdated date range.  Implementation should be consistent with existing business logic of a similar nature.

### Feedback Requested

Does the implementation solve the requested "ask"?
Is the implementation consistent with similar existing features?

### External References

- [DCGEO-66](https://jira.cms.gov/browse/DCGEO-66)

### Security Implications


brianburton (approved):
Looks good!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
```diff
@@ -151,4 +154,63 @@ public T read(@IdParam IdType claimId, RequestDetails requestDetails) {
...
+  @Search
+  @Trace
+  public Bundle findByPatient(
+      @RequiredParam(name = "mbi")
+          @Description(shortDefinition = "The patient identifier to search for")
+          ReferenceParam mbi,
+      @RequiredParam(name = "type")
+          @Description(shortDefinition = "A list of claim types to include")
+          String type,
+      @OptionalParam(name = "isHashed")
```
ericbuckley, line 167: How will the caller know if this should be set to true or false?

ericbuckley: oh nm, `isHashed` will be a query param

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
```diff
@@ -151,4 +154,63 @@ public T read(@IdParam IdType claimId, RequestDetails requestDetails) {
...
+          @Description(shortDefinition = "A list of claim types to include")
```
ericbuckley, line 168:
```suggestion
          @Description(shortDefinition = "A boolean indicating whether or not the MBI is hashed")
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -68,6 +84,75 @@ public Object getEntityById(ResourceTypeV2<?> type, String id) {
...
+          claimEntities == null || claimEntities.isEmpty() ? 0 : 1);
+    }
+
+    return claimEntities;
+  }
+
+  @VisibleForTesting
+  Predicate createDateRangePredicate(
+      Path<Instant> datePath, DateRangeParam dateRange, CriteriaBuilder builder) {
+    Instant from =
```
ericbuckley, line 133: If one of the bounds is empty, can we just skip building a predicate for it?  I think adding a where clause to these queries that makes a comparison where every row will return True will just make the query slower than it needs to be.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -68,6 +84,75 @@ public Object getEntityById(ResourceTypeV2<?> type, String id) {
...
+    CriteriaQuery<T> criteria = builder.createQuery(entityClass);
+    Root<T> root = criteria.from(entityClass);
+
+    criteria.select(root);
+    criteria.where(
+        builder.and(
+            builder.equal(root.get(attributeName), attributeValue),
+            lastUpdated == null
+                ? builder.and()
+                : createDateRangePredicate(root.get("lastUpdated"), lastUpdated, builder)));
```
ericbuckley, line 114: what do you think about adding an index to the lastUpdated field?

brianburton: Personally I'd be fine with putting the code through without the index until we can benchmark with large amounts of data to determine exactly what we'd need.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
KyRobbins: The code that invokes `createDateRangePredicate()` checks if the any value is in `lastUpdated`, so if no bounds are set it does nothing.  However, do we want to support just one bound? I.e. if upper bound is set, we're just returning everything before date X, or if only lower bound is set, we're returning everything after date X?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
KyRobbins: I'd also wait to see if any performance issues exist, as indexing the column will also have a negative hit on the the ETL process, which we're unsure how much traffic will be there atm.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
KyRobbins: Yep, and it "defaults" to true, being that if it's not explicitly "false" (either by not being set at all or being set to something else) then it's considered true in the method logic, so under normal use cases, it's completely omitted by the caller.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
ericbuckley: yeah thats what I was thinking, I don't think we need an effective start and effective end date as using those in the query will make it unnecessarily slower

brianburton (dismissed):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -68,6 +78,75 @@ public Object getEntityById(ResourceTypeV2<?> type, String id) {
...
+  @VisibleForTesting
+  Predicate createDateRangePredicate(
+      Path<Instant> datePath, DateRangeParam dateRange, CriteriaBuilder builder) {
+    List<Predicate> predicates = new ArrayList<>();
+
+    if (dateRange.getLowerBound() != null) {
+      Instant from = dateRange.getLowerBoundAsInstant().toInstant();
+
+      predicates.add(
+          ParamPrefixEnum.GREATERTHAN.equals(dateRange.getLowerBound().getPrefix())
```
ericbuckley, line 133: should this not also be `ParamPrefixEnum.GREATERTHAN_OR_EQUALS`?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -68,6 +78,75 @@ public Object getEntityById(ResourceTypeV2<?> type, String id) {
     return claimEntity;
   }
 
+  public <T> List<T> findAllByMbiHash(
+      Class<T> entityClass, String mbiHash, DateRangeParam lastUpdated) {
+    return findAllByAttribute(entityClass, "mbiHash", mbiHash, lastUpdated);
+  }
+
+  public <T> List<T> findAllByMbi(Class<T> entityClass, String mbi, DateRangeParam lastUpdated) {
+    return findAllByAttribute(entityClass, "mbi", mbi, lastUpdated);
```
ericbuckley, line 87: Will the attribute also be named "mbi" in the MCSClaim DAO?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
```diff
@@ -151,4 +157,97 @@ public T read(@IdParam IdType claimId, RequestDetails requestDetails) {
...
+
+      Bundle bundleResource;
+
+      boolean isHashed = !Boolean.FALSE.toString().equalsIgnoreCase(hashed);
+
+      Optional<ResourceTypeV2<T>> optional = parseClaimType(type);
+      if (optional.isPresent()) {
+        bundleResource =
+            createBundleFor(
+                Collections.singleton(optional.get()), mbiString, isHashed, lastUpdated);
```
ericbuckley, line 199: `options.get()` will return a list right, as it could be either `FISS`, `MCS` or `FISS,MCS`.  Would the `singleton` method work on the third option?

ericbuckley, line 199: what happens if the user specifies an invalid query parameter?  For example, `type=OUTPATIENT,INPATIENT`?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
KyRobbins: if the condition is true, it uses `greaterThan()`, if the condition is false, it uses `greaterThanOrEqualTo()`.  The idea being that the "default" logic would be `greaterThanOrEqualTo()`, since `greaterThan()` requires an exact match.  Inclusive values are typically the default behavior of date ranges for the start date, exclusive for the end date, which is why the upper and lower bound match slightly different conditions.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
KyRobbins: At the time this was created, I had no idea, so it was anticipated there may be some rework needed when the MCS implementation came around, which is now looking likely.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java
KyRobbins: I was treating it as a one or the other mistakenly, I think I do need to fix this to accept a list of enums.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
ericbuckley: we discussed this on a call and plan on using the functionality defined in [QueryUtils.createLastUpdatePredicate](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java#L20)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -123,28 +120,8 @@ public Object getEntityById(ResourceTypeV2<?> type, String id) {
 
   @VisibleForTesting
   Predicate createDateRangePredicate(
```
ericbuckley, line 122: this is now just a one line method, are you keeping it so its easy to mock in tests?

KyRobbins: exactly

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/R4ClaimResourceProvider.java
```diff
@@ -13,13 +15,17 @@
 @Component
 public class R4ClaimResourceProvider extends AbstractR4ResourceProvider<Claim> {
 
+  private static final Map<String, ResourceTypeV2<Claim>> claimTypeMap =
+      ImmutableMap.of("fiss", ClaimTypeV2.F);
```
ericbuckley, line 19: Will the type parsing accept "fiss" or "FISS" as a query parameter?

KyRobbins: it will accept both, if you look at AbstractR4ResourceProvider#parseClaimTypes()
```Set<ResourceTypeV2<T>> parseClaimTypes(@Nonnull TokenAndListParam types) {
    return types.getValuesAsQueryTokens().get(0).getValuesAsQueryTokens().stream()
        .map(TokenParam::getValue)
        .map(String::toLowerCase) // <-- Requested types are transformed to lowercase
        .map(getResourceTypeMap()::get) // <-- Then pulled from the implemented map
        .filter(Objects::nonNull)
        .collect(Collectors.toSet());
  }

**Review by ericbuckley — APPROVED** [review]
thanks @KyRobbins looks great

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/AbstractR4ResourceProvider.java [hunk 10]
```diff
@@ -144,11 +154,126 @@ public T read(@IdParam IdType claimId, RequestDetails requestDetails) {
...
+   */
+  @VisibleForTesting
+  @Nonnull
+  Set<ResourceTypeV2<T>> parseClaimTypes(@Nonnull TokenAndListParam types) {
+    return types.getValuesAsQueryTokens().get(0).getValuesAsQueryTokens().stream()
+        .map(TokenParam::getValue)
+        .map(String::toLowerCase)
+        .map(getResourceTypeMap()::get)
+        .filter(Objects::nonNull)
+        .collect(Collectors.toSet());
```
brianburton, line 179: Excellent use of a Stream here.

brianburton (approved):

Gabasol:
@KyRobbins - since this PR is technically not ready for BFD review yet, would you mind adding the WIP label?  Adding that label will help the BFD team to have a more accurate picture of which PRs are/are not ready for their attention.

cbrunefearless (approved):
lgtm

PR #645, opened 2021-06-11
MISC - fixes intermittent IT test failures due to timezone and race conditions

brianburton:
### Change Details

This PR addresses two issues that cause intermittent IT test failures:

- ~~`PipelineManagerIT` sometimes fails due to a `RejectedExecutionException`.  This was caused by the extremely short (1ms) execution interval leading to a race condition as the test shuts down the `ScheduledExcecutor`.  Increasing that to 100ms makes the race condition far less likely to happen.~~
- Various tests in `bfd-server-war` and `bfd-pipeline-ccw-rif` fail with `lastUpdated` related error messages.  These are due to the logic in the tests assuming that the timezone is set to UTC.  Adding a `TZ` environment variable to the execution environment in the pom.xml assures the timezone is set as expected by the test.

### Acceptance Validation

Unit tests should pass more reliably than before.

### Feedback Requested

Nothing specific.

### External References

None.

### Security Implications

no

no

no

no

no

### DIFF
--- a/apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineManagerIT.java
+++ b/apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineManagerIT.java
@@ -36,7 +36,7 @@ public final class PipelineManagerIT {
     // Start the pipeline against a bucket that doesn't exist.
     MockDataSetMonitorListener listener = new MockDataSetMonitorListener();
     PipelineManager pipeline =
-        new PipelineManager(new MetricRegistry(), new ExtractionOptions("foo"), 1, listener);
+        new PipelineManager(new MetricRegistry(), new ExtractionOptions("foo"), 100, listener);
     pipeline.start();
 
     // Wait for the pipeline to error out.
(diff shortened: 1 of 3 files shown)


cbrunefearless (dismissed):
lgtm

karlmdavis (dismissed):
> PipelineManagerIT sometimes fails due to a RejectedExecutionException. This was caused by the extremely short (1ms) execution interval leading to a race condition as the test shuts down the ScheduledExcecutor. Increasing that to 100ms makes the race condition far less likely to happen.

@brianburton This failure is resolved by the orchestration changes **and** if this change here gets merged, we won't be able to cleanly unrevert those changes. Can you remove that fix from this PR?

brianburton:
> > PipelineManagerIT sometimes fails due to a RejectedExecutionException. This was caused by the extremely short (1ms) execution interval leading to a race condition as the test shuts down the ScheduledExcecutor. Increasing that to 100ms makes the race condition far less likely to happen.
> 
> @brianburton This failure is resolved by the orchestration changes **and** if this change here gets merged, we won't be able to cleanly unrevert those changes. Can you remove that fix from this PR?

Thanks.  I've reverted that part of the PR.

jzulim (approved):
Thanks!

cbrunefearless (approved):
lgtm

cbrunefearless (approved):
lgtm

PR #650, opened 2021-06-15
[BFD-1000] Improve postgres logging

keithdadkins:
### Change Details

This PR improves logging queries (logs queries taking more than 1 second) and reduces the amount of tracked queries in pg_stat_statements from 5000 to 2000.

Note: Merging these stateful changes into master will not automatically trigger the terraform. This will be done manually at a time of our choosing. Also, we can apply the terraform whenever we want, but we will need to reboot the db clusters for the changes to take effect. 

### External References

- [BFD-1000](https://jira.cms.gov/browse/BFD-1000)

### Security Implications


mackec1 (approved):
Looks rgood.

keithdadkins:
Also, `pg_stat_statements` and `auto_explain` were both already configured. 

pg_stat_statements requires `CREATE EXTENSION pg_stat_statements` to be run so the tracking table is created. I will do this during maintenance. I also made a few small changes to pg_stat_statements parameters to reduce the amount of queries to track to 2000 from 5000 as postgres pre-allocates memory for these operations. If we end up with more than 2000 unique queries against or db, we can readjust later.

auto_explain will now capture queries taking longer than 1 second instead of 6 seconds.

jzulim (approved):
Approved.

cbrunefearless (approved):
LGTM

PR #665, opened 2021-06-18
BFD-897: Remove 000_cross_env_vars.yml symlink loop

karlmdavis:
BFD-897

### Change Details

This symlink just pointed directly at itself, and so contained no actual content. I have no clue why it was created in the first place, but... it couldn't have been doing anything useful, right?

I've seen all sorts of things get upset about this and fail. The latest is the `actions/upload-artifact` task for GH Actions.

### Acceptance Validation

1. Does anyone have a good reason as to why this was there and whether or not it's needed? If not, let's yank it and see what happens.

### Feedback Requested

(see above)

### External References

- [BFD-897](https://jira.cms.gov/browse/BFD-897)

### Security Implications

This PR has no known security implications.

### DIFF
--- a/ops/ansible/playbooks-ccs/host_vars/builds.bfd-mgmt.cmscloud.local/000_cross_env_vars.yml
+++ b/ops/ansible/playbooks-ccs/host_vars/builds.bfd-mgmt.cmscloud.local/000_cross_env_vars.yml
@@ -1,1 +1,0 @@
-000_cross_env_vars.yml


mackec1 (approved):
If it's an empty symlink, then let's nuke it; however, it's possible that it's a placeholder that gets turned on/off via some GH action

keithdadkins (approved):

brianburton (approved):

PR #672, opened 2021-06-22
BFD-826: Remove unneeded join for HICN history

karlmdavis:
### Change Details

At least for a couple of spot checks, this greatly improves the DB query plan, which makes sense as it's one of two large-row-count joins that the query had.

We may be able to remove other joins as well, but I want to make changes like this one step at a time, so we can keep track of what's actually impacting perforance, if anything.

### Acceptance Validation

1. Does this pass our tests? If so, I think we should merge and deploy it for testing in prod.

### Feedback Requested

(see above)

### External References

- [BFD-826](https://jira.cms.gov/browse/BFD-826)

### Security Implications

This PR should not have any security implications.


mackec1 (approved):

JeremyIR (approved):
LGTM

PR #679, opened 2021-06-29
[BFD-989] Enable DoSFilter

rtravitz:
### Change Details
This PR supports [BFD-989](https://jira.cms.gov/browse/BFD-989). It enables a [DoSFilter](http://www.eclipse.org/jetty/documentation/jetty-9/index.html#dos-filter) to rate-limit incoming requests to the FHIR API servers. The `UserAwareDoSFilter` identifies users through their client cert by overriding the `extractUserId` method.

The initial configuration of this PR for a max of 2 RPS is obviously an example. It will need to be fine-tuned before this can be merged. It's worth knowing that the rate applies across all incoming requests and is not individual per identified user. It is also not a replacement for BFD's existing auth mechanism. See the following:
> Jetty places throttled requests in a priority queue, giving priority first to authenticated users and users with an `HttpSession`, then to connections identified by their IP addresses. Connections with no way to identify them have lowest priority.

In addition to setting a per-instance rate, we need to make a decision about whether to delay or outright reject requests above the limit. By default the `DoSFilter` delays requests by 100ms. A different delay—or just rejecting the request outright with a 429—can be specified.

There is one final consideration I'd like to flag as likely important. The `DoSFilter` requires `async-supported` to be set to `true`. In Jetty 9, async report is not enabled for servlets and filters by default. When enabling it on the `UserAwareDoSFilter` as required by the documentation ([search in here](http://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/servlets/DoSFilter.html) for `async-supported`), it also then became required on the `RequestResponseLoggingFilter` entry above. It is unclear to me at this time whether this has other upstream impacts on the other two listed filters/servlets and then onto the server itself. 

So to summarize, here are the big questions:
1. What do we think for an initial per-instance rate limit?
2. Do we want to delay and then eventually reject if a request doesn't get serviced before the timeout? Or just outright reject? (I'm leaning toward the latter and forcing the retry on the Consumer).
3. Apparently Jetty 9 adheres to the Servlet API v3.1, which includes async support. This is required for the `DoSFilter` to work. Is that something we would want to enable more globally? Is there a way to determine if enabling locally for these two filters has other negative downstream effects?

### Acceptance Validation
Does it delay your request like you would expect? To see it, you can run this script: 
```bash
#! /usr/bin/env bash

set -B

for _ in {1..10}; do
  curl -Ik --cert "$BFD_PATH/apps/bfd-server/dev/ssl-stores/client-unsecured.pem" "https://localhost:$BFD_PORT/v1/fhir/Patient/-19990000000001"
done
```

The first two responses will go through without delay. The remaining eight should all have the `DoSFilter: delayed` header injected.

### Feedback Requested

Please help me with the three questions listed in the summary.

### External References

- [BFD-989](https://jira.cms.gov/browse/BFD-989)
- [DoSFilter Overview](http://www.eclipse.org/jetty/documentation/jetty-9/index.html#dos-filter)
- [DoSFilter JavaDoc](http://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/servlets/DoSFilter.html)

### Security Implications
This pulls in the Jetty servlets dependency. We already rely on the rest of Jetty, so this officially supported filter is not a big change.


apps/bfd-server/bfd-server-launcher/src/main/java/gov/cms/bfd/server/launcher/UserAwareDoSFilter.java
```diff
@@ -0,0 +1,46 @@
...
+   *     <code>null</code> if that's not available
+   */
+  private static X509Certificate getClientCertificate(ServletRequest request) {
+    X509Certificate[] certs =
+        (X509Certificate[]) request.getAttribute("javax.servlet.request.X509Certificate");
+    if (certs == null || certs.length <= 0) {
+      return null;
+    }
+    return certs[certs.length - 1];
+  }
```
rtravitz, line 45: Stolen straight from an existing implementation. We could extract if this stays how we identify users.

apps/bfd-server/bfd-server-war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -6,12 +6,29 @@
 	<filter>
 		<filter-name>RequestResponseLoggingFilter</filter-name>
 		<filter-class>gov.cms.bfd.server.war.RequestResponseLoggingFilter</filter-class>
+		<async-supported>true</async-supported>
```
rtravitz, line 9: Curious if folks have a greater understanding of how indicating that async is supported will impact the requests as they travel upstream (further from the client). This was only required on the downstream filter here and not on the filter or servlet upstream of the `UserAwareDoSFilter`.

christrotteradhoc: From the servlet spec it is permitted for an asynchronous filter/servlet to dispatch to a synchronous filter/servlet but the other way (sync to async) is not permitted. So I think that the upstream filter (the metrics filter) is fine to be left as-is without adding async-supported. Looking at that metrics filter code it looks like it does actually support async but it does not itself make a call to startAsync() so I don't think there is value unless something further along in the chain starts to take advantage of asynchronous processing.

As far as adding this to the logging filter, it looks like it will be necessary if we want to keep the logging filter as the first filter in the chain (and that seems correct). However, I think we may need to write some code in the logging filter to handle the case when the DosFilter delays a request by adding an async listener (similar to what you can find in the metrics abstract filter class). In your testing, did the logging filter perform its response actions correctly on the delayed requests?

mackec1 (commented):
Looks reasonable, but the inability to throttle based on source identifiers (i.e., IP addr, cert CN, etc.) will effectively punish the whole rather than individual source requests.

rtravitz:
> Looks reasonable, but the inability to throttle based on source identifiers (i.e., IP addr, cert CN, etc.) will effectively punish the whole rather than individual source requests.

@mackec1 I think that's a completely valid concern. When I was first researching this, it's why I was pushing so hard for proxy in front to take over rate limiting and auth together. That's still likely going to happen, but it's going to take more time. It was decided to pursue this `DoSFilter` in the meantime. It's more protection for the server to keep it alive than stopping the bad actor outright. The hard part with that idea is tuning the limits to an individual server's capacity rather than knowing the limits that an individual consumer will need.

apps/bfd-server/bfd-server-war/src/main/webapp/WEB-INF/web.xml
rtravitz: That's excellent to know—thanks for doing the work of reading the spec. Haven't had the chance yet, but I'm going to look through for my own education.

> I think we may need to write some code in the logging filter to handle the case when the DosFilter delays a request by adding an async listener

So the behavior I saw was that if `RequestResponseLoggingFilter` didn't have the `async-supported` configured, then it stopped logging anything on incoming requests. After changing the config, logging _seemed_ to resume as normal, but I need to go back and double check that all of the fields are added.

Your question is making me worry (and it would make sense) that it's just running `handleRequest` and not `handleResponse`. 

Will get back with a more definitive answer next week, unless you get bored!

christrotteradhoc: I played around with it a little bit and it appears that when the DosFilter delays a request we are getting two entries in the access.json -- one that is partially formed and one that is fully formed. I think that the partially formed one is the result of the original thread and needs to be suppressed (or perhaps just cleaned up to make it clear as to what it is) by putting in some listener logic.

mackec1 (approved):
This is about the best we can do until, as you've pointed out' that we front things with some sort of proxy listener (although I think even nginx is somewhat agnostic to source IPs, cert CN's etc)

rtravitz:
Closing as I'm easing out. You'll still be able to use this to inform future work if desired.

PR #681, opened 2021-06-30
BFD-1024: Refactor and cache test utils

karlmdavis:
### Change Details

This ended up being a rather large change, and a lot of little stuff had to happen along with it. Details:

* Refactored _most_ of the objects provided by `DatabaseTestUtils`, `PipelineTestUtils`, and `ServerTestUtils` to be cached as static state, which will be shared by across tests (though not across modules, as Maven runs each module's tests in a new JVM). This speeds up our tests **a lot**.
    * Prior to the ["Fix BFD Pipeline's init errors and job failure handling" `a9ce86a`](https://github.com/CMSgov/beneficiary-fhir-data/commit/a9ce86a668a4d9856b1901c554b4a3c0bf901393) commit, local builds took about 08:50 minutes for me.
    * After that commit, which accidentally removed caching of test DB connections, local builds took about 27:05 minutes for me.
    * With the changes, local builds now take about 06:27 minutes for me, a 27% reduction over where things _started_!
* Created a `PipelineApplicationState` to encapsulate the shared state for the BFD Pipeline application and related tests. Refactored things to use it.
* Moved the `hibernate-core` and `HikariCP` dependencies to `bfd-pipeline-shared-utils`, as everything needs them.
* Replaced most `new MetricRegistry()` calls with a reference to a cached instance, which will be shared by across tests (though not across modules, as Maven runs each module's tests in a new JVM).
    * I'm not positive that this will be useful but it at least gives us the _option_ of using the metrics to analyze our tests' performance.
* Renamed `DatabaseTestHelper` to `DatabaseTestUtils`, for consistency.
* Renamed `RifLoaderTestUtils` to `CcwRifLoadUtils`.
* Moved most of `CcwRifLoadUtils` to a new `PipelineTestUtils` in `bfd-pipeline-shared-utils`, so it can be used by more code.
* Added a `DatabaseOptions#getMaxPoolSize()` configuration value, which defaults to twice the value of `LoadAppOptions#getLoaderThreads()`.
* Removed the `ServerTestUtils#cleanDatabaseServer()` method, and replaced all usages of it with the new `PipelineTestUtils#truncateTablesInDataSource` method, which itself replaced the earlier and slower `DatabaseTestHelper#getTestDatabaseAfterClean()` method. This change shaved a minute off of our build times by itself.
* Removed all of the calls to `SessionFactoryRegistry.INSTANCE.clearRegistrations()` in our ITs, as they are no longer necessary (now that the `EntityManagerFactory` is being cached). In fact, these calls _had_ to be removed, as they'd now close the cached `EntityManagerFactory`, and thus cause our tests to fail.
* Disabled a couple of test cases in `RifLoaderIT`, which have race conditions that now reliably fail for me, given how much faster the tests execute.
* Removed the recently-added `DatabaseUtils` class, which is superceded by changes in this PR, mostly `PipelineApplicationState`.

### Acceptance Validation

* Do the tests pass locally, on GH Actions, and on Jenkins?
* Are these changes a net improvement to the code quality?

### Feedback Requested

(see previous)

### External References

- [BFD-1028](https://jira.cms.gov/browse/BFD-1028)

### Security Implications

This PR has no known security implications.


apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -338,6 +344,23 @@ static AppConfiguration readConfigFromEnvironmentVariables() {
...
+      } catch (NumberFormatException e) {
+        databaseMaxPoolSize = Optional.of(-1);
+      }
+      if (databaseMaxPoolSize.get() < 1)
+        throw new AppConfigurationException(
+            String.format(
+                "Invalid value for configuration environment variable '%s': '%s'",
+                ENV_VAR_KEY_DATABASE_MAX_POOL_SIZE, databaseMaxPoolSizeText));
+    }
+
```
christrotteradhoc, line 363: if this new block were moved below the loader threads determination then I think you could avoid the Optional and encapsulate the default case of 'loaderthreads * 2' into this section of code rather than using the .orElse() when building the databaseOptions.

karlmdavis:
The Jenkins build passed and was about two minutes faster than before the regression (down to 10 minutes, now): https://builds.bfd-management.oeda.cmscloud.local/jenkins/job/test/view/change-requests/job/PR-681/

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
karlmdavis: True, and very good catch. On reflection, though, I think I prefer the current `Optional` approach as it makes the... intent (?) of the thing a bit clearer to me. Happy to be outvoted, though.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineTestUtils.java
```diff
@@ -0,0 +1,275 @@
...
+import javax.persistence.Table;
+import javax.sql.DataSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Contains utilities that are useful when testing with or against the BFD Pipeline application.
+ *
+ * <p>This is being left in <code>src/main</code> so that it can be used from other modules' tests,
+ * without having to delve into classpath dark arts.
```
christrotteradhoc, line 54: should there be a bfd-pipeline-shared-test-utils module at some point in the future? (I'mnot suggesting it as a change for this PR)

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
christrotteradhoc: I would at least hope for consolidating the logic in one block whether an optional is used or not so that someone can tell by looking at one piece of code how the value is arrived at.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineTestUtils.java
karlmdavis: Maybe? It'd likely just have this one class, so I'm not sure the ROI is there.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
karlmdavis: Excellent suggestion, and done.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
```diff
@@ -0,0 +1,161 @@
...
+   * @return a JPA {@link EntityManagerFactory} for the application's database
+   */
+  private static EntityManagerFactory createEntityManagerFactory(DataSource dataSource) {
+    /*
+     * The number of JDBC statements that will be queued/batched within a single transaction. Most
+     * recommendations suggest this should be 5-30. Paradoxically, setting it higher seems to
+     * actually slow things down. Presumably, it's delaying work that could be done earlier in a
+     * batch, and that starts to cost more than the extra network roundtrips.
+     */
+    int jdbcBatchSize = 10;
```
christrotteradhoc, line 123: Looks like this was a static final class member earlier which seems preferable. Any particular reason for the change? I think it at least should be final.

karlmdavis: I try not to make things constants unless they're actually referenced by more than one block, as the indirection hurts readability. But sure, we can mark it `final` -- why not. Done.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -338,6 +344,29 @@ static AppConfiguration readConfigFromEnvironmentVariables() {
...
+      } catch (NumberFormatException e) {
+        databaseMaxPoolSizeOption = Optional.of(-1);
+      }
+      if (databaseMaxPoolSizeOption.get() < 1)
+        throw new AppConfigurationException(
+            String.format(
+                "Invalid value for configuration environment variable '%s': '%s'",
+                ENV_VAR_KEY_DATABASE_MAX_POOL_SIZE, databaseMaxPoolSizeText));
+    }
+
```
brianburton, line 363: This PR probably isn't the place to address this but it shouldn't take this much code to add one more environment variable.  Something like the `getIntOrDefault()` method (but throwing if the variable is missing/invalid rather than having a default) would shrink this and similar cases down to one LOC.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -338,6 +344,29 @@ static AppConfiguration readConfigFromEnvironmentVariables() {
...
+    /*
+     * Note: For CcwRifLoadJob, databaseMaxPoolSize needs to be double the number of loader threads
+     * when idempotent loads are being used. Apparently, the queries need a separate Connection?
+     */
+    Integer maxPoolSize = databaseMaxPoolSizeOption.orElse(loaderThreads * 2);
+
```
brianburton, line 369: Moving the rdaLoadOptions initialization above this would allow the default to be increased by 2 if the RDA pipeline jobs have been enabled (rdaLoadOptions != null).

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
```diff
@@ -102,11 +102,9 @@ public static void main(String[] args) throws InterruptedException {
...
     appMetricsReporter.start(1, TimeUnit.HOURS);
 
-    HikariDataSource dataSource =
-        DatabaseUtils.createDataSource(
-            appConfig.getDatabaseOptions(),
-            appMetrics,
-            2 * appConfig.getCcwRifLoadOptions().getLoadOptions().getLoaderThreads());
+    // Create the application state.
+    PipelineApplicationState appState =
+        new PipelineApplicationState(appMetrics, appConfig.getDatabaseOptions());
```
brianburton, line 107: Nice.  This is a big improvement even without thinking about test performance since it allows all of the jobs to use the same connection pool with a configurable size.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -92,38 +90,26 @@
...
-    this.dataSource =
-        DatabaseUtils.createDataSource(dataSource, appMetrics, 2 * options.getLoaderThreads());
-    this.entityManagerFactory = DatabaseUtils.createEntityManagerFactory(dataSource);
+    this.appState = appState;
 
     /*
      * We are re-using the same hash configuration for HICNs and MBIs so we only need one idHasher.
      */
     this.idHasher = new IdHasher(options.getIdHasherConfig());
+    this.secretKeyFactory = createSecretKeyFactory();
```
brianburton, line 112: This field is not used anywhere in this class.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -92,38 +90,26 @@
   private static final Logger LOGGER_RECORD_COUNTS =
       LoggerFactory.getLogger(RifLoader.class.getName() + ".recordCounts");
 
-  private final MetricRegistry appMetrics;
   private final LoadAppOptions options;
-  private final HikariDataSource dataSource;
-  private final EntityManagerFactory entityManagerFactory;
   private final IdHasher idHasher;
+  private final PipelineApplicationState appState;
+  private final SecretKeyFactory secretKeyFactory;
```
brianburton, line 96: This field is not used anywhere in this class.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
```diff
@@ -0,0 +1,161 @@
...
+      HikariDataSource pooledDataSource,
+      EntityManagerFactory entityManagerFactory) {
+    this.metrics = metrics;
+    this.pooledDataSource = pooledDataSource;
+    this.entityManagerFactory = entityManagerFactory;
+  }
+
+  /**
+   * This is marked <code>public</code> only for use in tests; non-test code that needs a {@link
+   * DataSource} should get it from the {@link PipelineApplicationState}.
```
brianburton, line 67: Except it isn't public?

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
```diff
@@ -0,0 +1,161 @@
...
+    pooledDataSource.setMaximumPoolSize(dbOptions.getMaxPoolSize());
+    pooledDataSource.setRegisterMbeans(true);
+    pooledDataSource.setMetricRegistry(metrics);
+
+    return pooledDataSource;
+  }
+
+  /**
+   * This is marked <code>public</code> only for use in tests; non-test code that needs a {@link
+   * DataSource} should get it from the {@link PipelineApplicationState}.
```
brianburton, line 89: Also private.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
```diff
@@ -0,0 +1,161 @@
...
+    pooledDataSource.setMaximumPoolSize(maxPoolSize);
+    pooledDataSource.setRegisterMbeans(true);
+    pooledDataSource.setMetricRegistry(metrics);
+
+    return pooledDataSource;
+  }
+
+  /**
+   * This is marked <code>public</code> only for use in tests; non-test code that needs a {@link
+   * EntityManagerFactory} should get it from the {@link PipelineApplicationState}.
```
brianburton, line 111: Also private.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
brianburton: Ditto.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
```diff
@@ -0,0 +1,161 @@
...
+
+  /** @see java.lang.AutoCloseable#close() */
+  @Override
+  public void close() throws Exception {
+    if (!pooledDataSource.isClosed()) {
+      pooledDataSource.close();
+    }
+    if (entityManagerFactory.isOpen()) {
+      entityManagerFactory.close();
+    }
```
brianburton, line 159: No idea if this matters, but, should the data source be closed after the entityManagerFactory just in case the latter needs it?  Maybe there is nothing to do in close() that needs it though.

brianburton (approved):
Looks good to me.  HUGE improvement in quality of life.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
brianburton: Hmm.  Git is showing shows two lines above the ocmment - meant for it just to show `secretKeyFactory`.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -338,6 +344,29 @@ static AppConfiguration readConfigFromEnvironmentVariables() {
...
+            String.format(
+                "Invalid value for configuration environment variable '%s': '%s'",
+                ENV_VAR_KEY_DATABASE_MAX_POOL_SIZE, databaseMaxPoolSizeText));
+    }
+
+    /*
+     * Note: For CcwRifLoadJob, databaseMaxPoolSize needs to be double the number of loader threads
+     * when idempotent loads are being used. Apparently, the queries need a separate Connection?
+     */
+    Integer maxPoolSize = databaseMaxPoolSizeOption.orElse(loaderThreads * 2);
```
rtravitz, line 368: Looks like there's a test failure because `loaderThreads` can't be resolved. Easiest way is probably to just move it below 386?

karlmdavis: Thanks for pointing it out -- should have checked that commit locally. Fixed now.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
karlmdavis: I can't believe I let myself get talked into this, but your wish is my command. Done.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
karlmdavis: I'd suggest making that change in a separate DC-Geo specific PR, just for tracking purposes.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
karlmdavis: Good catch -- I goofed this in the merge conflict resolution. Fixed.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
karlmdavis: Good catch -- I goofed this in the merge conflict resolution. Fixed.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
karlmdavis: Good catch: fixed.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
karlmdavis: Good catch: fixed.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/PipelineApplicationState.java
karlmdavis: Great catch: fixed.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -259,17 +265,37 @@ public String toString() {
...
+    String databaseUrl = readEnvStringRequired(ENV_VAR_KEY_DATABASE_URL);
+    String databaseUsername = readEnvStringRequired(ENV_VAR_KEY_DATABASE_USERNAME);
+    String databasePassword = readEnvStringRequired(ENV_VAR_KEY_DATABASE_PASSWORD);
+    int loaderThreads = readEnvIntPositiveRequired(ENV_VAR_KEY_LOADER_THREADS);
+    boolean idempotencyRequired = readEnvBooleanRequired(ENV_VAR_KEY_IDEMPOTENCY_REQUIRED);
+    Optional<String> newRelicMetricKey = readEnvStringOptional(ENV_VAR_NEW_RELIC_METRIC_KEY);
+    Optional<String> newRelicAppName = readEnvStringOptional(ENV_VAR_NEW_RELIC_APP_NAME);
+    Optional<String> newRelicMetricHost = readEnvStringOptional(ENV_VAR_NEW_RELIC_METRIC_HOST);
+    Optional<String> newRelicMetricPath = readEnvStringOptional(ENV_VAR_NEW_RELIC_METRIC_PATH);
+    Optional<Integer> newRelicMetricPeriod = readEnvIntOptional(ENV_VAR_NEW_RELIC_METRIC_PERIOD);
```
brianburton, line 280: Nice!  To steal Eric's pun: "Thanks for DRYing this up!"

brianburton (approved):

mackec1 (approved):
nice encapsulation of db stuff!

PR #682, opened 2021-06-30
[BFD-1010] scripts to create schema and insert scripts for loading new schema wi…

mackec1:
### Change Details

Modify schema to use postgres idiomatic table and column names that match CCW data names; change appropriate data columns to use bigint data type (bene_id, clm_id, clm_grp_id). Various load scripts to convert current schema data to new schema structure.

As this is currently intended as a draft PR until all names and data types pass peer review. This includes the new schema name, which in the scripts has been specified as 'cmac' and will be replaced at that time when the PR comes out of draft.

Questions:
========

1. Also, the new schema did not include the one-time "BeneficiaryMonthly" audit table that was part of the CCW data fixup; should we migrate this data or leave as is?
2. The current schema provides bi-directional entity mapping (i.e., Beneficiaries -> BeneficiaryMonthly, and BeneficiaryMonthly -> Beneficiaries). However, I've yet to see where the child (BeneficiaryMonthly) is used to fetch data from the parent (Beneficiaries). Do we need this child-parent type relationship(s) if we only ever use parent-child query relationships in our processing?

### Acceptance Validation

Data types are appropriate for expected content, including type constraints and sizes.

Names are appropriate and match current CCW data column names.

While the primary key constraints are provided, indexes will be forthcoming based on acceptance of table and column names.

### Feedback Requested

Any and all suggestions for data types, names, table structure that is both appropriate and meaningful for efficient data loading and querying.

### External References

- [BFD-1010](https://jira.cms.gov/browse/BFD-1010)

### Security Implications

While the db obviously contains PII data, this changes introduces no new security constraints or concerns.

- [X] new software dependencies

This schema change is the first step in converting all BFD applications to leverage the new schema structure and naming.

- [X ] new data stored or transmitted

Current schema data will be migrated to the new schema structure; however, this PR does not introduce 'new' data per se.


apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-1010-schema-changes/schema_constraints_drop_primary_keys.sql
```diff
@@ -0,0 +1,69 @@
...
+drop constraint if exists outpatient_claims_pkey;
+--
+alter table only cmac.partd_events
+drop constraint if exists partd_events_pkey;
+--
+alter table only cmac.snf_claim_lines
+drop constraint if exists snf_claim_lines_pkey;
+--
+alter table only cmac.snf_claims
+drop constraint if exists snf_claims_pkey;
```
karlmdavis, line 69: Why do we need to drop the PK indices? If it's just for speed, I'd personally rather pay the performance cost of having them, just to keep the migrations simpler and avoid oopses.

For similar reasons, I'd probably put _all_ of the indices that we will want in place up-front.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-1010-schema-changes/schema_diff.txt
```diff
@@ -0,0 +1,1588 @@
+CREATE TABLE public."Beneficiaries" (
```
karlmdavis, line 1: Apologies, but can you add a file header comment explaining what this is and why we need it?

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-1010-schema-changes/schema_modified_constraints.sql
```diff
@@ -0,0 +1,1931 @@
...
+ *
+ * 1. Clone the tables to new tables with all of the desired schema changes, copying the original data over.
+ * 2. Deploy a new version of the application that only uses those new tables.
+ * 3. Remove the original tables.
+ * 
+ * This script is for Step 1. 
+ *
+ * [1]: https://www.postgresql.org/docs/10/datatype-character.html
+ * [2]: https://www.postgresql.org/docs/10/datatype-numeric.html
+ */
```
karlmdavis, line 27: I'd add file headers to each migration script that explain **all** of the changes being made in them. That'll be very important for anyone trying to figure them out in the future.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-1010-schema-changes/schema_modified_constraints.sql
```diff
@@ -0,0 +1,1931 @@
...
+-- TODO: Right now, this is just a dump of the current BFD schema in prod, less all of the objects that I'm pretty sure we don't want to touch.
```
karlmdavis, line 28: This comment doesn't apply for this file.

karlmdavis (changes requested):
Great work, @mackec1! ⚽  🥅 

Some general notes:

* This needs to be a series of Flyway migrations, not a collection of scripts to be run manually.
    * I'd probably have one migration for the bene tables, then a separate one for each of the claim types, and another one for anything else misc (e.g. `loaded_files`).
* Ultimately, I'm not sure we want to put the tables into a separate schema. I'd leave them in `public`, myself. No big deal if I get outvoted on this one, just keep in mind that it will lead to a lot of extra typing for everyone in the future. 🤷

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-1010-schema-changes/schema_constraints_drop_primary_keys.sql
mackec1: Karl; thx for the review. I was mainly concerned with column/table names and data types. Hence the 'draft' status. A lot of the scripts were for my use (i.e., tearing down tables, playing with some constraints, etc.) so while those types of files are there, they won't be part of the real deal. 

If the table/column names are OK and the data types, then I'll start on the Flyaway script(s).

As for the schema, yea, I just did that so I would munge anything on my localhost by accident; the fact that the table names are (will be changed) allows us to put everything into public without name collisions which is fine by me.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/BFD-1010-schema-changes/schema_modified_constraints.sql
mackec1: yea, so for the most part my little node schema generator basically passed along comments, etc. from your original schema.sql file. So some of the files were just for comparison and of marginal use thereafter (i.e., schema_diff.txt). Mainly used to load into excel as needed.

mackec1:
repalce branch and create new PR since original one was somehow assigned to karl

PR #683, opened 2021-07-01
BFD-1028: Fix table order during TRUNCATE in ITs

karlmdavis:
### Change Details

Things were blowing up when building against PostgreSQL. This should fix that.

### Acceptance Validation

1. Does the GH Actions build still pass (which uses HSQL DB)?
2. Can someone build it locally against PostgreSQL?

### Feedback Requested

(see previous)

### External References

- [BFD-1028](https://jira.cms.gov/browse/BFD-1028)

### Security Implications

This PR has no security implications.


mackec1 (approved):
LGTM

christrotteradhoc:
The fix needs some rework which was discussed offline to use either a cascade option or truncate all of the tables at once on postgres only (those options are not supported in hsql).

christrotteradhoc (changes requested):
The fix needs some rework which was discussed offline to use either a cascade option or truncate all of the tables at once on postgres only (those options are not supported in hsql).

christrotteradhoc:
This was fixed under another PR

PR #685, opened 2021-07-01
BFD-1057: Add pre-flight check to Patient search by Part D

karlmdavis:
### Change Details

_May_ workaround an intermittent DB query timeout that we're seeing in production.

### Acceptance Validation

1. Does the logic here look sound?
2. Does it pass our build?
3. Are you comfortable with the performance tradeoff?
    * We'll be adding a third DB query to the first page of every search result for this operation. It _should_ only be a hundred milliseconds or so, but that's not nothing.

### Feedback Requested

(see above)

### External References

- [BFD-1057](https://jira.cms.gov/browse/BFD-1057)

### Security Implications

This PR should not have any security implications.

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
@@ -402,6 +402,22 @@ public final class PatientResourceProvider implements IResourceProvider, CommonH
     String contractCode = coverageId.getValueNotNull();
 
     /*
+     * Workaround for BFD-1057: The `ORDER BY` required on our "find the bene IDs" query (below)
+     * intermittently causes the PostgreSQL query planner to run a table scan, which takes over an
+     * hour in prod. This _seems_ to be only occurring when the query would return no results. (Yes,
+     * this is odd and we don't entirely trust it.) So, when we're on the first page of results or
+     * not paging at all here, we first pull a count of expected matches here to see if there's any
+     * reason to even run the next query.
+     */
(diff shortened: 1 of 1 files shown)


mackec1 (approved):
Cross our fingers; this should handle this edge case.

cbrunefearless (approved):
lgtm

rtravitz (approved):
Looks reasonable to me.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientResourceProvider.java
```diff
@@ -420,6 +436,48 @@ private String partDFieldByMonth(CcwCodebookVariable month) {
...
+    try {
+      matchingBeneCount =
+          Optional.of(entityManager.createQuery(beneCountCriteria).getSingleResult());
+      return matchingBeneCount.get();
+    } finally {
+      beneHistoryMatchesTimerQueryNanoSeconds = matchingBeneCountTimer.stop();
+      TransformerUtils.recordQueryInMdc(
+          "bene_count_by_year_month_part_d_contract_id",
+          beneHistoryMatchesTimerQueryNanoSeconds,
+          matchingBeneCount.isPresent() ? 1 : 0);
```
christrotteradhoc, line 477: is this what you want (0 or 1) or is is supposed to be 0 or number of records returned, i.e. the underlying Long value?

karlmdavis: The number of records returned by the query, which in this case is a `select count(distinct ...) ...`.

PR #686, opened 2021-07-01
[BFD-1046] - Add performance smoke test script with instructions

rtravitz:
### Change Details
This PR supports [BFD-1046](https://jira.cms.gov/browse/BFD-1046). It adds a small script to run calls to BFD that are specified in a configuration file. It runs each call multiple times, tags them all with a special header, averages the response times, and then prints them out. It's a speedy and repeatable way to quickly test every endpoint we care about in the application. For more information on how to run it, I added a README.

I also have a local file full of endpoints that will just need query params tweaked depending on the environment where it will be run. That can be checked in with synthetic data values if that's useful.

Hope the Go choice is fine here—this seems like a small out of the way space, it didn't require any external deps, allowed for easy concurrency to make things run faster, and is succinct. Being able to make the binary for other OS/Arch combos could also prove useful if this ever needs to run anywhere besides ✨ _this laptop_ ✨  

### Acceptance Validation
- The script provides response times for calls to configured endpoints.

### External References
- [BFD-1046](https://jira.cms.gov/browse/BFD-1046)

### Security Implications


jzulim (approved):
LGTM. Documentation is good. Go is a good choice.

cbrunefearless (approved):
lgtm

rtravitz:
Added in a sample endpoints file with synthetic data in the params for the Sample A user and a synthetic Part D contract. This makes it easy to run this script out of the box anytime you want a baseline on any environment.

rtravitz:
Also worth mentioning that the baseline results are now added as a comment in the ticket.

cbrunefearless (approved):
lgtm

christrotteradhoc (approved):
Very cool and much needed

PR #687, opened 2021-07-02
[BFD-1059] Update native instructions for README and remove synthetic data patch

christrotteradhoc:
### Change Details

Remove references to obsolete synthetic data patch (including the patch itself and create-db.sh script).
General formatting and wording changes to the Native Setup section to improve readability and copy/paste.
Minor fixes to commands.

### Acceptance Validation

View the changes in a markdown viewer and examine the Native Setup section. (no attempt was made to fix other sections)
Starting with a fresh clone and no database, execute the commands as written and make sure they do the right things.

### Feedback Requested

Do you feel strongly that either of the removed files should be preserved? The create-db.sh script, after removing the synthetic data patch became a one-liner equivalent to step 5 in the README except that the two commands had diverged so I thought it best to just have one place to maintain the suggested docker command. I could not find any additional references to this script within the repo.

### External References

- [BFD-1059](https://jira.cms.gov/browse/BFD-1059)

### Security Implications

No security implications.


contributing/create-db.sh
```diff
@@ -1,13 +0,0 @@
...
-docker run -d --name 'local_bfddb' -e 'POSTGRES_DB=bfd' -e 'POSTGRES_USER=bfd' -e 'POSTGRES_PASSWORD=InsecureLocalDev' -p '5432:5432' postgres:12
-cd ../apps/bfd-pipeline/bfd-pipeline-ccw-rif && mvn -Dits.db.url="jdbc:postgresql://localhost:5432/bfd" -Dits.db.username=bfd -Dits.db.password=InsecureLocalDev -Dit.test=RifLoaderIT#loadSampleA clean verify
-
-# Fetch Synthetic Data
-cd ../../../contributing && make synthetic-data/*.rif 
-cd ../ && git apply contributing/patches/load_local_synthetic_data.patch
-cd apps && mvn clean install -DskipITs
-
-# Run Maven test and load synthetic data
-cd bfd-pipeline/bfd-pipeline-ccw-rif && mvn -Dits.db.url="jdbc:postgresql://localhost:5432/bfd" -Dits.db.username=bfd -Dits.db.password=InsecureLocalDev -Dit.test=RifLoaderIT#loadLocalSyntheticData clean verify
```
JeremyIR, line 13: Is this file being completely removed? This is to load synthetic data in a docker container as opposed to being a substitute for having a local setup.

christrotteradhoc: It doesn't have to be removed but if we remove the synthetic data patching it no longer does much other than run a couple of steps from the README so I didn't see the value. I'm happy to keep it though if it is being relied on. Perhaps it needs to be re-written to use some other way of loading the synthetic data if we are getting away from the patching?

JeremyIR: It's more so to have an automated way to create a local dockerized PG db that reflects changes in rif files in Amazon S3, and for future plans to deploy a the db to Amazon ECR. But yes, it is redundant with steps in the README.

jzulim: I'd vote to leave it, we could use this as boiler plate for later Developer Experience (DX) automation / initiatives.

jzulim (approved):
LGTM.

contributing/create-db.sh
rtravitz: I take the opposite view. If we want to use it as an example for future things, we can certainly do that. But that example can be gleaned from the source control history. There's enough cruft in here as it is—anything that currently serves no purpose that we can remove is welcome.

JeremyIR: We can remove it. This code is checked in, so I can always re-reintroduce it.

rtravitz (approved):
Thanks for rewriting these and removing the broken patch.

PR #678, opened 2021-06-29
DCGEO-97: MCS claims RDA API caller and sink

brianburton:
**This PR is still a WIP but has been created to allow early review by DCGEO team members.**

### Change Details

This PR implements DCGEO-97 to add a new PipelineJob that calls the RDA API `getMcsClaims()` RPC and stores all claims in the resulting stream to the database.  The chane also implements the entity and protocol changes from the RDA API 0.3 MVP release.

Some of the existing code specific to FISS claim processing in master has been converted to apply to both FISS and MCS claims.  In particular:

- RDA API proto files have been updated to the 0.3 MVP release.
- A new schema migration script has been added to create the MCS tables and add new columns to the FISS tables.
- New entity classes have been added for MCS claims.  These are the same entities added for [PR 677](https://github.com/CMSgov/beneficiary-fhir-data/pull/677).
- FISS entities have been modified to include the new columns added by RDA API.
- `RdaChange` has been added to model the new RDA API `ClaimChange` wrapper object.
- `FissClaimRdaSink` has been renamed `JpaClaimRdaSink` and converted to accept the claim entity class as a type parameter.  Now this same sink is used for both FISS and MCS claims.
- The previous `PipelineJob` implementation, `RdaLoadJob` has been converted to an abstract class, `AbstractRdaLoadJob`, and trivial new subclasses have been defined for FISS and MCS (`RdaFissClaimLoadJob` and `RdaMcsClaimLoadJob`, respectively).  The subclassing was necessary to allow the `PipelineJobType` to be based on the job's class.
- A number of new integration tests have been added, including  `JpaClaimRdaSinkIT` which verifies claim entities can be written to HSQLDB correctly.
- The mock RDA API server has been extended to support `getMcsClaims` calls.

Key classes to review in this PR:

- `V31__preadj_claim_table_updates.sql` migration script.
- `McsClaimTransformer` validates and converts RDA API `McsClaim` objects into equivalent `PreAdjMcsClaim` entity objects.
- `FissClaimTransformer` validates and converts RDA API `FissClaim` objects into equivalent `PreAdjFissClaim` entity objects.  This class has been modified to add support for new fields.
- `DataTransformer` has been enhanced to handle new types in MCS claim data.
- `EnumStringExtractor` implements the special handling required for RDA API enum values.
[...]

PR #691, opened 2021-07-06
DCGEO-123: RDA API Pipeline setup for TEST environment

brianburton:
### Change Details

Changes in this PR:

- Adds RDA API pipeline environment variables to allow data ingestion from the API in the TEST environment.
- Adds additional logging to the RDA API pipeline jobs.
- Uses a different entity manager/persistence unit for RDA API pipeline jobs than the CCW Rif pipeline.
- Fixes two bugs in FISS/MCS claim transformation logic that caused transformation to fail for invalid reasons.

The persistence unit change was needed to allow Hibernate to use the entities from `bfd-model-rda`.  Hibernate only allows a persistence unit to include entities from the same jar as the `persistence.xml` file.  Since RDA has its own model module it also has its own jar.  This requires a separate persistence unit.  (Spring's entity class loader does not have this limitation so the RDA entities are useable in `bfd-server-war` module without the need for a different persistence unit.)

### Acceptance Validation

- Are the ansible changes valid.
- Do the ansible changes follow expected patterns.
- Are the code changes acceptable.

### Feedback Requested

This is my first foray into BFD ansible playbooks so there may be better ways of accomplishing the same goal.

### External References

- [DCGEO-123](https://jira.cms.gov/browse/DCGEO-123)

### Security Implications

The host and port are defined in the vault.  Other settings are not security sensitive so they are defined in `defaults/main.yml`.

No

No

RDA API is still disabled in non-TEST environments.  The `env_specific.yml` file for TEST sets `data_pipeline_rda_job_enabled` to `true` to enable the pipeline jobs there.

No

No


KyRobbins (approved):

ericbuckley (commented):
don't we also need to set `data_pipeline_rda_job_enabled` to true in the test/group_vars/all/vault.yml?

brianburton:
> don't we also need to set `data_pipeline_rda_job_enabled` to true in the test/group_vars/all/vault.yml?

Not yet because we don't want to enable it until the MVP 0.3 code is merged (PR 768).

ericbuckley:
> > don't we also need to set `data_pipeline_rda_job_enabled` to true in the test/group_vars/all/vault.yml?
> 
> Not yet because we don't want to enable it until the MVP 0.3 code is merged (PR 678).

What do you think about holding on this then till after #678 is merged, that way we don't need two PRs for this changeset?

brianburton:
> > > don't we also need to set `data_pipeline_rda_job_enabled` to true in the test/group_vars/all/vault.yml?
> > 
> > 
> > Not yet because we don't want to enable it until the MVP 0.3 code is merged (PR 678).
> 
> What do you think about holding on this then till after #678 is merged, that way we don't need two PRs for this changeset?

OK.  I've changed the flag in `ops/ansible/playbooks-ccs/vars/test/group_vars/all/env_specific.yml`.

ericbuckley (approved):

KyRobbins (dismissed):

ericbuckley (approved):

ericbuckley (approved):

KyRobbins (dismissed):

ericbuckley (approved):
👍

ops/ansible/roles/bfd-pipeline/defaults/main.yml
```diff
@@ -8,6 +8,26 @@ data_pipeline_idempotency_required: true
...
+# Set to true to enable the RDA pipeline jobs.
+data_pipeline_rda_job_enabled: false
+
+# Communication settings for RDA API server.
+# These are defined in the vault rather than plain text.
+data_pipeline_rda_grpc_host: undefined
+data_pipeline_rda_grpc_port: 443
+
+# Interval at which the PipelineManager should ensure the jobs are running.
+data_pipeline_rda_job_interval_seconds: 300
```
ericbuckley, line 23: @brianburton it looks like there are over 16k FISS claims coming in from the RDA API on DEV and probably a similar amount of MCS.  I think we should dial this back a little, maybe have it run once an hour instead of every 5 mins.  At least until we build in support for replays.  So maybe just leave this as is in this file and in make it `3600` in `test/group_vars/all/env_specific.yml`

brianburton: This is where the code I had in place originally to allow us to configure a limit on the number of records the job would write per scheduled execution would have helped.  It was removed based on reviewer comments though.

OK.  I will increase the interval as requested.  Generally speaking the value of this isn't all that important for our job since the job will run forever if the idle time is never exceeded.  If the server call fails the PipelineManager won't restart us anyway.

brianburton: The 3600 value has been added in `test/group_vars/all/env_specific.yml` now.

ericbuckley: Yeah I was assuming that we'd address the restart issues in the near future.  Maybe its good to run through a scenario real quick to see how the settings will work.

From the DB, it looks like we consumed the ~16k FISS claims in about 9 mins.  When we first connect to RDA, if we spend 9 mins consuming data, then disconnect after an additional 10 mins (the idle time).  So we'd have roughly 40 mins or so (assuming the above 3600 interval value), before we repeated the process all over again.  Does that sound right?

ericbuckley (approved):
Looks great Brian, can we just add a couple of comments to the description to outline the logging changes and the fixes to the transformer.

ops/ansible/roles/bfd-pipeline/defaults/main.yml
brianburton: Yes I think that's correct.

KyRobbins (approved):

christrotteradhoc (approved):

PR #697, opened 2021-07-13
DCGEO-117: Initial service date range implementation

KyRobbins:
### Change Details

Add ability to filter search results by a specified service date range.  For FISS claims we are filtering on the `stmtCovToDate` value and for MCS claims we are filtering on the `idrHdrToDate`.  We are trying to maintain backwards compatibility with `service-date` [filtering on ExplanationOfBenefit records](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java#L419-L461), with the exception that we are moving this logic to filter on the database rather than in Java code.

Additional gaps in unit test coverage of target PR classes also identified and filled.

### Acceptance Validation

Resulting code should return data within a specified service date range given that:
  - The claim's service start date is not null
  - The claim's service end date is not null
  - The claim's service dates overlap with the given service date range being searched on

### Feedback Requested

Does the code effectively achieve the overall ask
Does the code follow current conventions and coding practices

### External References

- [DCGEO-117](https://jira.cms.gov/browse/DCGEO-117)

### Security Implications


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/ClaimTypeV2.java
```diff
@@ -80,6 +90,14 @@ public String getEntityMbiHashAttribute() {
     return entityMbiHashAttribute;
   }
 
+  public String getEntityStartDateAttribute() {
```
ericbuckley, line 93: If we're now using lombok, can't we just use `@Getter` for this?

KyRobbins: lombok is only in the bfd-model-rda module, it has not been introduced to bfd-server-war

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -82,14 +88,17 @@ public Object getEntityById(ResourceTypeV2<?> type, String id) {
...
    * @return A list of entities of type T retrieved matching the given parameters.
    */
   public <T> List<T> findAllByAttribute(
       Class<T> entityClass,
       String attributeName,
       String attributeValue,
-      DateRangeParam lastUpdated) {
+      DateRangeParam lastUpdated,
+      DateRangeParam serviceDate,
+      String endDateAttributeName) {
```
brianburton, line 101: `endDateAttributeName` doesn't have a comment

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -124,6 +136,52 @@ Predicate createDateRangePredicate(
...
+      DateRangeParam serviceDate,
+      CriteriaBuilder builder,
+      String endDateAttributeName) {
+    Path<LocalDate> serviceDateEndPath = root.get(endDateAttributeName);
+
+    List<Predicate> predicates = new ArrayList<>();
+
+    // If the row's end service date attribute is null, don't return it
+    predicates.add(builder.isNotNull(serviceDateEndPath));
+
```
brianburton, line 151: Interesting.  Wouldn't a NULL be normally be omitted in a query result if there is any condition on that column?

brianburton (dismissed):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
ericbuckley: Maybe this was a protection in case the user DIDNT specify a `service-date` filter?  In that case I think we should remove this predicate, just in case we want to integrate with other peering APIs down the road that DONT use the `service-date` query param

ericbuckley (commented):
@KyRobbins looking good, thanks for making the changes to the `serviceDateRangePredicate` logic.  Just a couple of small changes based on Brian's comments and then I think this is ready for review

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
KyRobbins: This line checks to see if the service end date in the database is null

ericbuckley (approved):
👍

brianburton (approved):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
```diff
@@ -124,6 +137,49 @@ Predicate createDateRangePredicate(
     return QueryUtils.createLastUpdatedPredicateInstant(builder, root, dateRange);
   }
 
+  @VisibleForTesting
```
jzulim, line 140: Do you anticipate users wanting to find claims BEFORE a service-date?

jzulim: nvm, I kept reading. :)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/common/ClaimDao.java
jzulim: Bump

mackec1 (approved):
Code looks good; nice use of encapsulation and really like use of mocking in tests.

christrotteradhoc (approved):

PR #700, opened 2021-07-15
[BFD-1072] Fix unit tests on postgres

christrotteradhoc:
### Change Details

Factor out existing database vendor conditional into a new utility class (there does not appear to be such a thing already that is not test specific).
Make use of utility class to detect when we are running in Postgres and append the CASCADE option to the truncate table statements in that case.
(Added intellij files to gitignore :)

### Acceptance Validation

Should be no impact to existing production code. Tests should now work correctly when run locally on Postgres.

### Feedback Requested

Better/existing place to put these utility methods?

### External References

- [BFD-1072](https://jira.cms.gov/browse/BFD-1072)

### Security Implications

No security implications


cbrunefearless (approved):
lgtm

rtravitz (approved):
I'm getting failures that seem to be unrelated in the `bfd-server-war` integration tests where the connection is being refused. This shows up the same with the in-memory DB and Postgres, and the tests pass in CI here, so no fault of these changes.

The `DatabaseUtils` seems like a fine place to house the extra utility methods.

PR #702, opened 2021-07-16
Lsharshar/bfd 1104 serializing eobs in tests

sharshar-sb:
### Change Details

There are some difference between an EOBs created by the transformers and the one serialized due to the HAPI parser. To prevent any issues due to this, all the objects are turned into JSON and then reloaded into an EOB object before data validation is done. 

### Acceptance Validation

The tests continue to pass

### External References

- [BFD-1104](https://jira.cms.gov/browse/BFD-1104)

### Security Implications


mackec1 (approved):
Laila; as an FYI, if you look at R4ExplanationOfBenefitResourceProviderIT.java, you'll see the workarounds on some other HAPI issues (in that case it was HAPI prepending a '#' to contained Id(s) while the FHIR spec says you shouldn't have it.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/InpatientClaimTransformerV2Test.java
```diff
@@ -71,15 +72,19 @@ public InpatientClaim generateClaim() throws FHIRException {
...
+    eob = parser.parseResource(ExplanationOfBenefit.class, json);
   }
 
   private static final FhirContext fhirContext = FhirContext.forR4();
 
   /** Common top level EOB values */
   @Test
   public void shouldSetID() {
-    Assert.assertEquals("inpatient-" + claim.getClaimId(), eob.getId());
+    Assert.assertEquals("inpatient-" + claim.getClaimId(), eob.getIdElement().getIdPart());
```
christrotteradhoc, line 87: This assertion is a little bit different from the other similar ones in other classes. Is that by design?

sharshar-sb: Yes. Either should work. Just wanted to make sure one checked the actual ID object.

christrotteradhoc (approved):

PR #704, opened 2021-07-16
[BFD-1095] Allow traffic from CBJ to deployed FHIR ASGs and ETL instances

njdister:
### Change Details

This removes the unused `ci_cidrs` variable passed down to EC2/ASG modules and replaces that with a `ci_sg` variable that is the direct ID of a security group that allows traffic from CBJ.  This will be created in the stateful module to allow for sourcing the CBJ subnets from a private vars file.

### Acceptance Validation

Traffic is allowed from CBJ to instances and nothing breaks.

### Feedback Requested

I took down my old PR which attempted to reuse the `ci_cidrs` var but could not find an easy, private way to inject those subnets into the build pipeline (which is how stateless resources are created).  I think this is the right way to do this now but open to feedback.

### External References

- [BFD-914](https://jira.cms.gov/browse/BFD-914)

### Security Implications


jzulim (commented):
This indeed looks like the right approach considering the current pattern. I do have one question, are we keeping this subnet values in the vars file encrypted in Jenkins? My VPN is still jacked so I can't pop in there to confirm if I'm understanding how that piece comes together. 

Also, if these are terraform vars that are encrypted in jenkins, I believe there are duplicates in keybase that we may want to confirm are synced.

njdister:
> This indeed looks like the right approach considering the current pattern. I do have one question, are we keeping this subnet values in the vars file encrypted in Jenkins? My VPN is still jacked so I can't pop in there to confirm if I'm understanding how that piece comes together.
> 
> Also, if these are terraform vars that are encrypted in jenkins, I believe there are duplicates in keybase that we may want to confirm are synced.

The security group is now created in stateful which is applied outside of jenkins, it is only attached in stateless during jenkins deploy.  There are no terraform vars injected at deploy time in jenkins, hence why I took down my original PR (where everything was created in stateless).  So we only need vars in keybase.

jzulim (approved):
Looks clean. I finally found where you were using 'data' to grab the ci_sig. Nice worl.

keithdadkins:
Looks ok to me. We will of course need to define the `var.cloudbees_jenkins_subnets` in tfars (kb) with the appropriate values. I ran a plan against test using `10.20.30.40/32` without errors.

Screen Shot 2021-08-23 at 11 52 54 AM

keithdadkins (approved):

keithdadkins:
I've added `cloudbees_jenkins_subnets = []` stubs in each env's tfvars.

PR #707, opened 2021-07-16
adding a pre-commit hook to check for decrypted secrets

ericbuckley:
### Change Details

This PR adds a new `pre-commit` git hook to test that any "known" secret files are not being committed in plaintext.  Much of this work comes from ideas [previously introduced into the BCDA ACO API](https://github.com/CMSgov/bcda-app/pull/450).

### Feedback Requested

- [ ] is a pre-commit hook the best way to control accidental commits of decrypted data
- [ ] does it make sense to have the pre-commit hook installed with mvn (will non app engineers still benefit)
- [ ] do we want to use a `.secrets` file to manage a list of encrypted files or will it too easily become out of sync with reality 

### Security Implications

  - adds a pre-commit check to verify encrypted files aren't commited in a decrypted state


README.md
```diff
@@ -66,10 +66,10 @@ mkdir -p ~/workspaces/bfd/
 git clone git@github.com:CMSgov/beneficiary-fhir-data.git ~/workspaces/bfd/beneficiary-fhir-data.git
 ```
 
-### Native Setup
+### Initializing the Repository
 1. Install JDK 8. You'll need Java 8 to run BFD. You can install OpenJDK 8 however you prefer.
-2. Install Maven 3. Project tasks are handled by Apache Maven. Install it however you prefer.
-3. Configure your toolchain. You'll want to configure your `~/.m2/toolchains.xml` file to look like the following (change the jdkHome appropriately):
+1. Install Maven 3. Project tasks are handled by Apache Maven. Install it however you prefer.
+1. Configure your toolchain. You'll want to configure your `~/.m2/toolchains.xml` file to look like the following (change the jdkHome appropriately):
```
rtravitz, line 72: These are nice developer experience changes 🙂

rtravitz (approved):
A `.secrets` seems fine, and a git hook seems like a pragmatic thing to add here to prevent these issues going forward. I think there's work to move to a different secret management solution soon—specifically something like ParameterStore—so that this becomes a moot point.

The only other thing I'll say for this specific Git Hook plugin with maven is that it makes it exceedingly hard to build a slim Docker image. The plugin requires the `.git` directory to exist in the context in order to write the hooks, which means you have to COPY it in there. Which then means that every time _anything_ in the project changes you can't rely on the cache and have to re-copy. 

But these are my tangential struggles. This looks good.

ericbuckley:
> A `.secrets` seems fine, and a git hook seems like a pragmatic thing to add here to prevent these issues going forward. I think there's work to move to a different secret management solution soon—specifically something like ParameterStore—so that this becomes a moot point.
> 
> The only other thing I'll say for this specific Git Hook plugin with maven is that it makes it exceedingly hard to build a slim Docker image. The plugin requires the `.git` directory to exist in the context in order to write the hooks, which means you have to COPY it in there. Which then means that every time _anything_ in the project changes you can't rely on the cache and have to re-copy.
> 
> But these are my tangential struggles. This looks good.

Thanks for the feedback Ryan, a couple of thoughts

1) Yeah I figured the team *might* eventually migrate these to another system, but figured it doesn't hurt to build in an alert system if its a quick fix

2) Re the `.secrets` file, the other thought I had was to compare the current changed files to its contents in the previous commit.  However, if the migration to a different secret management system is being planned or in the works, then maybe the effort isn't worth it.

3) I'm open to removing the `githook-maven-plugin` from `apps/pom.xml` and just using a simple script to install the pre-commit hook if others think thats a sensible change.  I guess with the existing setup, any active developer is more or less forced to install the hook as soon as they run their first `mvn` command; an alternative setup could easily be skipped

.secrets
```diff
@@ -0,0 +1,17 @@
+ops/ansible/playbooks-ccs/host_vars/builds.bfd-mgmt.cmscloud.local/vault.yml
```
jzulim, line 1: @ericbuckley did you do something programmatically (scan repo in some fashion) to compile this list or was it done manually? 

@keithdadkins just flagging you in case you notice any other files that should be included.

.gitignore
```diff
@@ -4,6 +4,7 @@
 !.gitignore
 !.gitattributes
 !.github
+!.secrets
```
jzulim, line 7: 🙇

jzulim (approved):
Thank you for this, and I agree this approach get's the job done, and as Ryan mentioned we're pushing to priotize moving secrets out of our repo ASAP. I was curious how others are handling this and came across git-crypt. I wasn't familiar with it until today. Seems it will encrypt your files on commit, then unenrypt when you pull. Here's a. nice write up on the usage: https://hackernoon.com/things-you-must-know-about-git-crypt-to-successfully-protect-your-secret-data-kyi3wi6

This is far better than we currently have (nothing) and with the move to get secrets out, I agree this work belongs in our repo and can help prevents incidents. :ship:

PR #712, opened 2021-07-20
[BFD-878] Reduce Java Util Date Use W/ Bug Fix

JeremyIR:
### Change Details

There was a bug with year formatting in the TransformerUtils when creating extensions that has been addressed in this PR. Other than that there are the previously reviewed changes to remove and replace many instances of the java Date library with the Time library because of deprecation and timezone issues on different hosting environments. 
### Acceptance Validation

Use java.time classes instead of legacy java.util.Date & java.sql.Date with JDBC 4.2 or later.
Convert to/from java.time if inter-operating with code not yet updated to java.time
Keep everything in UTC timezone

### Feedback Requested

- Are there any missing opportunities to avoid using java util Date? Are there any areas where there could be inconsistencies with timezone?

### External References

- [BFD-878](https://jira.cms.gov/browse/BFD-878)

### Security Implications


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -406,18 +407,19 @@ static Extension createExtensionDate(
...
     Extension extension = null;
+    if (!dateYear.isPresent()) {
+      throw new NoSuchElementException();
+    }
     try {
-      String stringDate = dateYear.get().toString() + "-01-01";
-      Date date1 = new SimpleDateFormat("yyyy-MM-dd").parse(stringDate);
-      DateType dateYearValue = new DateType(date1, TemporalPrecisionEnum.YEAR);
+      String stringDate = String.format("%0" + 4 + "d", dateYear.get().intValue());
+      ;
```
christrotteradhoc, line 415: extraneous line

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -406,18 +407,19 @@ static Extension createExtensionDate(
...
 
     Extension extension = null;
+    if (!dateYear.isPresent()) {
+      throw new NoSuchElementException();
+    }
     try {
-      String stringDate = dateYear.get().toString() + "-01-01";
-      Date date1 = new SimpleDateFormat("yyyy-MM-dd").parse(stringDate);
-      DateType dateYearValue = new DateType(date1, TemporalPrecisionEnum.YEAR);
+      String stringDate = String.format("%0" + 4 + "d", dateYear.get().intValue());
```
christrotteradhoc, line 414: I'm not sure of the value of concatenating here. I think that format("%04d"...) is pretty typical and easier to make sense of at a glance.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerUtils.java
```diff
@@ -858,18 +859,19 @@ static Extension createExtensionDate(
...
+    if (!dateYear.isPresent()) {
+      throw new NoSuchElementException();
+    }
     try {
-      String stringDate = dateYear.get().toString() + "-01-01";
-      Date date1 = new SimpleDateFormat("yyyy-MM-dd").parse(stringDate);
-      DateType dateYearValue = new DateType(date1, TemporalPrecisionEnum.YEAR);
+      String stringDate = String.format("%0" + 4 + "d", dateYear.get().intValue());
+      ;
+      DateType dateYearValue = new DateType(stringDate);
```
christrotteradhoc, line 868: same two comment as above

christrotteradhoc (commented):
Couple of small comments inline. Did we end up with a new unit test for this case?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
JeremyIR: Good catch. I initially was thinking of creating a variable for digit length hence concatenation, but since the year is fixed at 4, there's no need.

christrotteradhoc (approved):
Thanks for adding the test. Looks good!

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/QueryUtilsTest.java
```diff
@@ -14,22 +14,32 @@ public void testInRange() {
      * Dev Note: There might be a bug in DateRangeParam and DateParam where it rounds a value that
      * is set a millisecond up or down. It makes it hard to test at the edges of a range.
      */
```
mackec1, line 16: One could always specify/allow for a 'drift' (std procedure in OCSP land)

mackec1 (approved):
lgtm

PR #714, opened 2021-07-21
[BFD-1073] Intermittent PipelineManagerIT failures

christrotteradhoc:
### Change Details

There are several similar failure modes addressed with this change but most of them come down to timing issues between one thread recording a job outcome (whether it be completion, failure, or cancellation) and the PipelineManager shutdown causing job cancellation and resulting callbacks via onFailure(). When that happens a job outcome can attempt to be recorded twice (two cancellations, failure and cancellation, success and cancellation are all possible) leading to the BadCodeMonkey exception on the second attempt to record status. This seems to be an artifact of all of the starting and stopping of the PipelineManager in the IT tests and not the sort of thing that we will see much of in the field. I suspect that recent changes to rework the tests have caused the timing of the shutdown to have been altered enough to make this more prevalent. To avoid this a change is made to only record job status one time in a thread-safe manner and write out appropriate log messages when a second attempt is made to record job status just in case it does happen in the field at some point.

WIth this fix applied a new failure mode was discovered when running the test in a loop where it is possible for the Volunteer job to attempt to enqueue the same job twice since it is building the list of pending jobs by calling isStarted() which is set by another thread which introduces a race condition. Using a new isEnqueued() method instead which is set by the same thread solves the issue.

### Acceptance Validation

Ran the suite locally over 1000 times without failure.

### External References

- [BFD-1073](https://jira.cms.gov/browse/BFD-1073)

### Security Implications

No security implications.


apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -432,28 +485,24 @@ public PipelineJobOutcome call() throws Exception {
...
+        handleJobCompletion(jobRecord.getId(), jobOutcome);
         return jobOutcome;
       } catch (InterruptedException e) {
         /*
          * This indicates that someone has successfully interrupted the job, which should only have
          * happened when we're trying to shut down. Whether or not PipelineJob.isInterruptible() for
          * this job, it's now been stopped, so we should record the cancellation.
          */
-        jobRecordStore.recordJobCancellation(jobRecord.getId());
+        handleJobCancellation(jobRecord.getId());
```
karlmdavis, line 496: I would _like_ remove this entirely, as it's redundant with `PipelineJobCallback.onFailure(Throwable)`, which _should_ always run for cancelations.

However, right now, the code doesn't appear to quite provide the guarantees necessary for that, so let's leave it for now.

apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineManagerIT.java
```diff
@@ -487,11 +487,14 @@ public void runWayTooManyJobsThenStop() throws Exception {
...
+  /**
+   * Reduce tick time on built-in jobs, to speed test execution. Setting this too low can expose a
+   * race condition in the volunteer job where it attempts to enqueue the same job twice.
+   */
   @BeforeClass
   public static void configureTimers() {
-    VolunteerJob.VOLUNTEER_TICK_MILLIS = 10;
-    SchedulerJob.SCHEDULER_TICK_MILLIS = 10;
+    VolunteerJob.VOLUNTEER_TICK_MILLIS = 500;
+    SchedulerJob.SCHEDULER_TICK_MILLIS = 500;
```
karlmdavis, line 497: Instead of this, I'd have `PipelineJobRecordStore.findPendingJobs(int)` call a new `j.isEnqueued()` method (instead of calling `j.isStarted()`, like it does currently). This should deterministically solve the problem.

karlmdavis (changes requested):
This is really astonishingly good work, @christrotteradhoc! Thanks so much for digging into this!

I am leaving a few suggestions, but NBD, overall.

christrotteradhoc:
@karlmdavis thanks for the suggestions! I've addressed those and rerun 1000 iterations of the suite successfully.

karlmdavis (approved):
Awesome work!

rtravitz (approved):
This looks great. One thousand runs is sufficiently convincing.

PR #716, opened 2021-07-22
[BFD-477] Do not automatically mount partner EFS on ETL

keithdadkins:
### Change Details
This PR stops the pipeline instances from automatically trying to mount the EFT EFS file systems on boot, but leaves the ability for us to manually mount the file system to help with testing. This is for a few reasons:
1. Originally, we had planned to do all of the EFT EFS workflow on the ETL pipeline server (ConnectDirect, EFS, etc), and that may no longer be the case. In fact, we will likely end up runing this on a dedicated EFT EFS instance.
2. I have since renamed BCDA's EFS module not thinking about the auto mounting. This is causing the boot to hang a bit, which is triggering an alert and page.
3. We will be refactoring much of the work that was done to get EFS working on FIPS enabled RHEL 7 once we migrate to Amazon Linux 2.

### External References

- [BFD-xxx](https://jira.cms.gov/browse/BFD-xxx)

### Security Implications

### DIFF
--- a/ops/ansible/roles/bfd-pipeline/tasks/eft_efs.yml
+++ b/ops/ansible/roles/bfd-pipeline/tasks/eft_efs.yml
@@ -50,8 +50,7 @@
   tags:
     - pre-ami
 
-## Create the base EFT EFS mount directory
-# e.g., /mnt/eft
+# Create the base EFT EFS mount directory (ex. /mnt/efs)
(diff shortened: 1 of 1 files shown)


jzulim (approved):
👍

rtravitz (approved):

PR #721, opened 2021-07-29
DCGEO-109: RDA pipeline integration tests

brianburton:
### Change Details

This PR primarily adds integration tests for the RDA API pipeline jobs.  In addition there are some enhancements based on experience running the jobs on TEST.

- adds the option to run PipelineApplication without the RIF/CCW job
- adds new tests in PipelineApplicationIT
- adds new integration tests for RDA pipeline jobs
- adds catch for exceptions thrown while running jobs so the jobs aren't descheduled by PipelineManager
- adds activity gauge metric for tracking pipeline job uptime
- adds latency histogram metric for tracking promptness of pipeline data ingestion from RDA API

The ability to disable CCW/RIF job was necessary to allow testing of the RDA pipeline jobs in isolation.  A new configuration environment variable named `CCW_RIF_JOB_ENABLED`.  This variable defaults to `true` if it's undefined so there is no need to modify any deployment playbooks or roles.

FISS and MCS test data has been added as NDJSON files.  The data is all randomly generated so there are no PII issues related to them.  The files live in the test jar so they will not be deployed to production.

### Acceptance Validation

- do the tests run correctly?
- are the most important tests implemented?

Here is a breakdown of what's changed by file to help in naviating the `Files changed` page during review:

|File|What's Changed|
---|---
RdaLoadJobIT.java | New end to end integration tests of the RDA API pipeline job.
PipelineApplicationIT.java | Adds end to end integration tests for RDA API pipeline job running within the PipelineApplication.
RdaServerApp.java ExceptionMessageSource.java RandomFissClaimSource.java RandomMcsClaimSource.java WrappedClaimSource.java | Changes to support the new integration test cases.
AppConfiguration.java PipelineApplication.java | Adds ability to disable CCW/RIF pipeline to allow testing of RDA pipeline.
FISS.ndjson MCS.ndjson | Randomly generated data for use in RDA API pipeline integration tests.
GrpcRdaSource.java GrpcRdaSourceTest.java | Adds new metrics to allow NewRelic tracking of `latency` (difference between RDA API timestamp and when we processed a file) and `uptime` (detecting when the job is running or not running for alerting/visual check).
RdaServerApp.java EmptyClaimSource.java JsonClaimSource.java ClaimSource.java RandomFissClaimSource.java RandomMcsClaimSource.java RdaServer.java RdaService.java WrappedClaimSource.java JsonClaimSourceTest.java RandomFissClaimSourceTest.java FissClaimStreamCallerIT.java GrpcRdaSourceIT.java | Converting various classes to directly support `RdaChange` objects to improve testability of pipeline.

### Feedback Requested

Nothing in particular.

### External References

- [DCGEO-109](https://jira.cms.gov/browse/DCGEO-109)

### Security Implications

None


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaChange.java
```diff
@@ -1,12 +1,16 @@
...
+import lombok.Getter;
+
 /**
  * RDA API wraps each claim in a `ClaimChange` object that indicates the type of change. This class
  * wraps that concept into an immutable bean used by the sink classes to optimize how they write
  * claims to the database.
  *
  * @param <T> The database entity type for the change.
  */
+@Getter
```
ericbuckley, line 13: 💯

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/JpaClaimRdaSink.java
```diff
@@ -25,10 +27,12 @@
 
   private final EntityManager entityManager;
   private final Metrics metrics;
+  private final Clock clock;
 
-  public JpaClaimRdaSink(String claimType, PipelineApplicationState appState) {
+  public JpaClaimRdaSink(String claimType, PipelineApplicationState appState, Clock clock) {
     entityManager = appState.getEntityManagerFactory().createEntityManager();
     metrics = new Metrics(appState.getMetrics(), claimType);
+    this.clock = clock;
```
ericbuckley, line 35: It seems like quite a few classes had to be updated to pass down this variable.  I didn't realize that it would cause so many code changes.  I think this will be a useful metric, but there might be other ways to capture this information (latency that is), so if you think this is adding too much complexity, we can back it out and revisit other paths for measuring latency in the future.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
```diff
@@ -159,14 +159,40 @@ public void close() throws Exception {
...
+    return metrics;
+  }
+
+  /**
+   * Indicates service is running but not actively processing a new record. Called at start of job
+   * and when a batch has been written.
+   */
+  @VisibleForTesting
+  void setUptimeToRunning() {
+    metrics.uptimeValue.set(10);
```
ericbuckley, line 172: should the values be constants?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
```diff
@@ -238,4 +265,49 @@ public int hashCode() {
...
   }
+
+  /**
+   * Metrics are tested in unit tests so they need to be easily accessible from tests. Also this
+   * class is used to write both MCS and FISS claims so the metric names need to include a claim
+   * type to distinguish them.
+   */
+  @Getter
+  @VisibleForTesting
+  static class Metrics {
```
ericbuckley, line 276: 👍 I like the encapsulation

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/sink/JpaClaimRdaSinkIT.java
```diff
@@ -116,7 +119,7 @@ public void mcsClaim() throws Exception {
     diagCode.setIdrDiagIcdType("T");
     claim.getDiagCodes().add(diagCode);
 
-    int count = sink.writeObject(new RdaChange<>(RdaChange.Type.INSERT, claim));
+    int count = sink.writeObject(new RdaChange<>(RdaChange.Type.INSERT, claim, Instant.now()));
```
ericbuckley, line 122: I had a thought during today's sync call with RDA; most of the time we should always receive the INSERT message for a claim, before any UPDATE message.  However, it's possible that the first message type we get for a claim will be an UPDATE, would this create any problems in processing?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/resources/FISS.ndjson
```diff
@@ -0,0 +1,100 @@
+{"changeType":"CHANGE_TYPE_UPDATE","fissClaim":{"dcn":"6796658","hicNo":"672783318942","totalChargeAmount":"928.08","principleDiag":"xg","fissProcCodes":[{"procCd":"xg","procDt":"2021-07-11"},{"procCd":"qvwkg","procFlag":"ptwt","procDt":"2021-06-18"},{"procCd":"kstvc"},{"procCd":"znjkq","procFlag":"bg","procDt":"2021-07-16"},{"procCd":"bxng","procDt":"2021-02-13"},{"procCd":"zd","procDt":"2021-01-31"}],"pracLocState":"cg","fissDiagCodes":[{"diagCd2":"wcwp","bitFlags":"fng","diagPoaIndUnrecognized":"r"}],"currStatusUnrecognized":"k","currLoc1Unrecognized":"b","currLoc2Unrecognized":"m"}}
```
ericbuckley, line 1: just double checking, but this is all randomly generated data right?

ericbuckley (dismissed):
Looks good @brianburton, just a few comments

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/JpaClaimRdaSink.java
brianburton: This sort of thing should be handled by a dependency manager liek Guice but the pipeline application does not use one.

brianburton: No I'm fine with the complexity.  Now that it's in place it's no big deal.  I used an explict `Clock` to facilitate testing.  It would have been possible to just call `System.currentTimeMillis()` but that would not be testable.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
brianburton: Could(maybe should) have done that.  In this case though the numbers don't really matter so much.  We just want to get a shape on a newrelic graph.  If we expand to more states I'd definitely want to make it a little more formalized.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/sink/JpaClaimRdaSinkIT.java
brianburton: Good question.  The `merge()` method is basically an upsert so if the record doesn't exist it will insert it.  It's just not as fast as persist for new records.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/resources/FISS.ndjson
brianburton: Yep it's random data generated using the mock server code.

KyRobbins (dismissed):

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
```diff
@@ -132,27 +134,36 @@ public static void main(String[] args) throws Exception {
...
+    if (appConfig.getCcwRifLoadOptions().isPresent()) {
+      // Create an application state that reuses the existing pooled data source with the ccw/rif
+      // persistence unit.
+      final PipelineApplicationState appState =
+          new PipelineApplicationState(
+              appMetrics, pooledDataSource, PipelineApplicationState.PERSISTENCE_UNIT_NAME);
+
+      pipelineManager.registerJob(
+          createCcwRifLoadJob(appConfig.getCcwRifLoadOptions().get(), appState));
+    }
```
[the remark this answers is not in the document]
christrotteradhoc: I have the same question and also wonder if we want to log something in the if block.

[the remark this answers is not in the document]
christrotteradhoc: I have the same question and also wonder if we want to log something in the if block.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaLoadOptions.java
```diff
@@ -54,17 +52,18 @@ public RdaLoadOptions(
...
-      PipelineApplicationState appState) {
+  public RdaFissClaimLoadJob createFissClaimsLoadJob(
+      PipelineApplicationState appState, Clock clock) {
     return new RdaFissClaimLoadJob(
         jobConfig,
         () ->
             new GrpcRdaSource<>(
                 grpcConfig,
                 new FissClaimStreamCaller(
                     new FissClaimTransformer(Clock.systemUTC(), new IdHasher(idHasherConfig))),
```
christrotteradhoc, line 63: Is it intentional that the clock passed in as a parameter is not being used here as well? Passing all of these clock objects around as parameters seems like it could get out of hand and lead to strange behavior that is hard to track down. Don't we always want just one clock across every piece of code (at least for business logic)?

brianburton: The clock needs to be configurable for unit testing.  Ideally we would have a dependency manager that could provide dependencies like this but we don't.   I considered adding a clock to the `PipelineApplicationState` but thought others might object.  If you'd find that preferable I could make that change.  We really should not use a singleton for the clock because those are much more error prone than just passing a dependency to a constructor.

christrotteradhoc: I was not thinking singleton either but more along the lines of dependency injection but if we don't have that available I think the PipelineApplicationState would be a reasonable compromise.

Can you enlighten me as to whether this particular line is using the wrong clock? Or is there some reason why we want to use the system clock here even though we were passed a clock as a parameter that might be different?

brianburton: LOL - yeah that's a bug.

christrotteradhoc: Got it. Same bug also exists a few lines down.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineApplication.java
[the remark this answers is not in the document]
brianburton: I have added logging for both cases.  Logging for the disabled case is a warning since in production we'd definitely need it to be disabled.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaLoadOptions.java
brianburton: Bug fixed and I've moved the Clock into PipelineApplicationState. That removed it from constructors that accept it.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -86,6 +86,12 @@
    */
   public static final String ENV_VAR_KEY_DATABASE_MAX_POOL_SIZE = "DATABASE_MAX_POOL_SIZE";
 
+  /**
+   * The name of the environment variable that should be used to indicate whether or not to
+   * configure the CCW RIF data load job. Defaults to true to run the job unless disabled.
+   */
+  public static final String ENV_VAR_KEY_CCW_RIF_JOB_ENABLED = "CCW_RIF_JOB_ENABLED";
```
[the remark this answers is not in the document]
brianburton: Good point.  If we add more jobs in the future though I think we'd want to revisit configuration more generally.  For example right now the number of threads is hard coded.  For now I think this is the safe/easy way to go.

christrotteradhoc (approved):

ericbuckley (approved):
LGTM

PR #723, opened 2021-07-30
[BFD-1127] Flyway migration to apply perms to the pre_adj schema

keithdadkins:
### Change Details
Flyway migration to apply default permissions to DCGEO's pre_adj schema and fixes perms on existing tables.

Notes:
- There are currently no sequences in this schema so no retroactive fixing was needed.
- `DO` runs an anonymous block using the default language (PL/pgSQL). Needed for IF logic.
- `$$` is to make the block readable. See https://www.postgresqltutorial.com/plpgsql-block-structure/
- PL/pgSQL's `BEGIN`/`END` are only for grouping, they do not start or end a transaction.
- The `${logic.perms}` placeholder will be replaced by `--` (e.g., commented out) if run against hsql
- So these perms will only be applied if run on postgres. See DatabaseSchemaManager.java for placeholders.

### External References

- [BFD-1127](https://jira.cms.gov/browse/BFD-1127)

### Security Implications


jzulim (approved):
Permissions and SQL syntax looks good.

ericbuckley (approved):
LGTM; nice work with finding a path for HSQLdb

keithdadkins:
I deployed this PR to test and everything looks ok. Appropriate perms were applied and it deployed to test successfully.

from etl logs
```
...
.. Successfully validated 31 migrations (execution time 00:00.047s)
... Current version of schema "public": 31
... Migrating schema "public" to version "32 - preadj schema perms"
... Successfully applied 1 migration to schema "public", now at version v32 (execution time 00:00.127s)
... Schema create/upgrade: complete.
...
```

It appears flyway reorganizes things in the schema_version table, which makes sense.
Screen Shot 2021-08-03 at 3 16 04 PM

keithdadkins:
Flyway logger says `Successfully applied 1 migration to schema "public"`, but I verified that it did actually apply it to the `pre_adj` schema.

brianburton:
> Flyway logger says `Successfully applied 1 migration to schema "public"`, but I verified that it did actually apply it to the `pre_adj` schema.

Excellent!  I think Flyway uses the name of the schema that it stores the migration status table in when reporting the outcome.

PR #725, opened 2021-08-04
DCGEO-144 (DRAFT): Adds status code checking for MCS and FISS claims.

brianburton:
### Change Details

Adds support for rejecting FissClaim and McsClaim objects from the RDA API if they contain unsupported values.

- For FISS claims all defined enum values are acceptable but "unrecognized" field values are rejected.
- For MCS claims all defined enum values EXCEPT `STATUS_CODE_NOT_USED` are acceptable but "unrecognized" field values are rejected.

In addition to adding the new feature a unit test class was added for the McsClaimTransformer.

### Acceptance Validation

- Are the correct values being rejected and other values being accepted?
- Are the unit tests sufficient to verify the change is working correctly?

### Feedback Requested

Nothing specific.

### External References

- [DCGEO-144](https://jira.cms.gov/browse/DCGEO-144)

### Security Implications

None.

No

No

No

No

No

### DIFF
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimTransformer.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimTransformer.java
@@ -1,5 +1,6 @@
 package gov.cms.bfd.pipeline.rda.grpc.source;
 
+import com.google.common.collect.ImmutableSet;
 import gov.cms.bfd.model.rda.PreAdjFissClaim;
 import gov.cms.bfd.model.rda.PreAdjFissDiagnosisCode;
 import gov.cms.bfd.model.rda.PreAdjFissProcCode;
@@ -31,21 +32,27 @@ public class FissClaimTransformer {
           FissClaim::getCurrStatusEnum,
           FissClaim::hasCurrStatusUnrecognized,
           FissClaim::getCurrStatusUnrecognized,
-          FissClaimStatus.UNRECOGNIZED);
+          FissClaimStatus.UNRECOGNIZED,
+          ImmutableSet.of(),
+          ImmutableSet.of(EnumStringExtractor.Options.RejectUnrecognized));
   private static final EnumStringExtractor<FissClaim, FissProcessingType> currLoc1 =
       new EnumStringExtractor<>(
           FissClaim::hasCurrLoc1Enum,
           FissClaim::getCurrLoc1Enum,
           FissClaim::hasCurrLoc1Unrecognized,
           FissClaim::getCurrLoc1Unrecognized,
-          FissProcessingType.UNRECOGNIZED);
+          FissProcessingType.UNRECOGNIZED,
+          ImmutableSet.of(),
+          ImmutableSet.of());
   private static final EnumStringExtractor<FissClaim, FissCurrentLocation2> currLoc2 =
       new EnumStringExtractor<>(
           FissClaim::hasCurrLoc2Enum,
           FissClaim::getCurrLoc2Enum,
           FissClaim::hasCurrLoc2Unrecognized,
           FissClaim::getCurrLoc2Unrecognized,
-          FissCurrentLocation2.UNRECOGNIZED);
+          FissCurrentLocation2.UNRECOGNIZED,
+          ImmutableSet.of(),
+          ImmutableSet.of());
   private static final EnumStringExtractor<
           FissDiagnosisCode, FissDiagnosisPresentOnAdmissionIndicator>
       diagPoaInd =
@@ -54,7 +61,9 @@ public class FissClaimTransformer {
               FissDiagnosisCode::getDiagPoaIndEnum,
               FissDiagnosisCode::hasDiagPoaIndUnrecognized,
               FissDiagnosisCode::getDiagPoaIndUnrecognized,
-              FissDiagnosisPresentOnAdmissionIndicator.UNRECOGNIZED);
+              FissDiagnosisPresentOnAdmissionIndicator.UNRECOGNIZED,
+              ImmutableSet.of(),
+              ImmutableSet.of());
 
   private final Clock clock;
   private final IdHasher idHasher;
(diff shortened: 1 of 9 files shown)


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/McsClaimTransformerTest.java
```diff
@@ -0,0 +1,484 @@
...
+    claim.setIdrClaimType("3");
+    claim.setLastUpdated(clock.instant());
+    claimBuilder
+        .setIdrClmHdIcn("123456789012345")
+        .setIdrContrId("12345")
+        .setIdrClaimTypeEnum(McsClaimType.CLAIM_TYPE_MEDICAL);
+    changeBuilder
+        .setChangeType(ClaimChange.ChangeType.CHANGE_TYPE_INSERT)
+        .setMcsClaim(claimBuilder.build());
+    assertChangeMatches(RdaChange.Type.INSERT);
```
KyRobbins, line 65: Are these testing the entire transformer class?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/TransformerTestUtils.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Set;
+import java.util.function.ToIntFunction;
+import java.util.stream.Collectors;
+
+public class TransformerTestUtils {
+  /**
+   * The McsClaimTransformerTest and FissClaimTransformerTest need to be able to compare lists of
+   * objects and there isn't a standard matcher that is compatible and handles ordering properly.
+   */
+  public static <T> void assertListContentsHaveSamePropertyValues(
```
KyRobbins, line 18: Does `assertEquals(set1, set2)` not work here?  Do the entities have `@EqualsAndHash` defined?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/McsClaimTransformerTest.java
brianburton: This is testing that a claim containing only the required fields can be transformed without errors.  So it verifies that no extra fields are required by the transformer for some reason.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/TransformerTestUtils.java
brianburton: Sadly we need to use the reflection based property matching rather than plain equality.  The entities follow the hibernate recommended practice of only including primary key fields in the equals/hashcode methods.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimTransformer.java
```diff
@@ -31,21 +32,27 @@
...
+          ImmutableSet.of(),
+          ImmutableSet.of(EnumStringExtractor.Options.RejectUnrecognized));
   private static final EnumStringExtractor<FissClaim, FissProcessingType> currLoc1 =
       new EnumStringExtractor<>(
           FissClaim::hasCurrLoc1Enum,
           FissClaim::getCurrLoc1Enum,
           FissClaim::hasCurrLoc1Unrecognized,
           FissClaim::getCurrLoc1Unrecognized,
-          FissProcessingType.UNRECOGNIZED);
+          FissProcessingType.UNRECOGNIZED,
```
ericbuckley, line 44: I missed this from before, but are BOTH currLoc1 and currLoc2 enums?

ericbuckley (dismissed):
LGTM, thanks for adding the flexibility to raises errors on ANY future enum values

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/FissClaimTransformer.java
brianburton: Yes they are.  Do we want to validate them too then?  It would be easy to do now...

ericbuckley: Not yet, I want to wait till we have a use case for them in our FHIR responses.  For now we should accept any value

KyRobbins (approved):

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/McsClaimTransformerTest.java
```diff
@@ -0,0 +1,484 @@
...
+                  "detail-0-idrKPosAddr2_1st", "invalid length: expected=[1,30] actual=33"),
+              new DataTransformer.ErrorMessage(
+                  "detail-0-idrKPosAddr2_2nd", "invalid length: expected=[1,25] actual=28"),
+              new DataTransformer.ErrorMessage(
+                  "detail-0-idrKPosCity", "invalid length: expected=[1,30] actual=33"),
+              new DataTransformer.ErrorMessage(
+                  "detail-0-idrKPosState", "invalid length: expected=[1,2] actual=5"),
+              new DataTransformer.ErrorMessage(
+                  "detail-0-idrKPosZip", "invalid length: expected=[1,15] actual=18")),
+          ex.getErrors());
```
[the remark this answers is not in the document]
brianburton: Interesting question.  There will ultimately be many more fields and I'm not thrilled by having hundreds of separate unit tests.  The individual field validations are independent of one another.  I can improve how the individual errors are reported in the test so it shows just what's different rather than the whole lists being shown.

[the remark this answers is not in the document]
christrotteradhoc: My thinking on something like this is that you are going to have a hundred of something no matter what you do. A hundred one line tests is more lines of code but they are simpler to understand and work with. Another option is a parametrized test where you don't even need a separate method for each test but you get the same effect when you run and debug them. Just some thoughts.

christrotteradhoc (approved):

ericbuckley (approved):
👍

PR #733, opened 2021-08-13
DCGEO-198: program to load rda data from ndjson files to db

brianburton:
### Change Details

This PR adds a command line program that loads NDJSON files containing RDA API claim data and uploads the data to a database.  The program is intended for use in populating local databases for testing.

A maven exec plugin configuration can be used to easily run the program from the command line.

### Acceptance Validation

Does the program work for another developer.

### Feedback Requested

Is the configuration easy to use for other developers?

### External References

- [DCGEO-198](https://jira.cms.gov/browse/DCGEO-198)

### Security Implications

None

No

No

Just uses the RDA API pipeline jobs locally to populate a database.

No

No


apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/ConfigLoader.java
```diff
@@ -0,0 +1,271 @@
...
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.function.Function;
+
+/** Simple class to abstract loading configuration values from multiple sources. */
+public class ConfigLoader {
```
ericbuckley, line 14: I could have sworn you implemented a class like this previously, was it not reusable or am I mistaken?

ericbuckley (approved):
👍

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/ConfigLoader.java
brianburton: Similar yes but less flexible than this one.  That one wound up being removed during the pipeline refactoring a while back.  I think it's important to have something like this available and I needed it here so I added it.  It's much more flexible than just reading environment variables and removes a ton of boilerplate.  It'd be nice to see it incorporated into the pipeline and/or server configuration.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/LoadRdaJsonApp.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private DatabaseOptions createDatabaseOptions() {
+      return new DatabaseOptions(dbUrl, dbUser, dbPassword, 10);
+    }
+
+    private RdaLoadOptions createRdaLoadOptions(int port) {
+      final IdHasher.Config idHasherConfig = new IdHasher.Config(hashIterations, hashPepper);
+      final AbstractRdaLoadJob.Config jobConfig =
+          new AbstractRdaLoadJob.Config(Duration.ofDays(1), batchSize);
```
KyRobbins, line 127: Any desire to have this configured (the duration)?  Could not be needed, just seems to stand out as one of the few things that is not configurable.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/DirectRdaLoadApp.java
```diff
@@ -10,13 +10,10 @@
 import gov.cms.bfd.pipeline.sharedutils.IdHasher;
 import gov.cms.bfd.pipeline.sharedutils.PipelineApplicationState;
 import gov.cms.bfd.pipeline.sharedutils.PipelineJob;
-import java.io.BufferedReader;
-import java.io.FileReader;
-import java.io.Reader;
+import java.io.File;
```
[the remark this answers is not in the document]
brianburton: Good question.  Mostly the answer is yes.  Also File is trivial to work with and can be converted to a Path on demand with toPath().

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/LoadRdaJsonApp.java
brianburton: In this case the program is reading from a local (in process) mock server so the idle timeout is just something that has to be set but won't be triggered.  So I just set it to something big.

ericbuckley (dismissed):

KyRobbins (approved):

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/LoadRdaJsonApp.java
```diff
@@ -0,0 +1,159 @@
...
+    private final String dbUser;
+    private final String dbPassword;
+    private final int batchSize;
+    private final boolean runSchemaMigration;
+    private final Optional<File> fissFile;
+    private final Optional<File> mcsFile;
+
+    private Config(ConfigLoader options) {
+      hashPepper = options.stringValue("hash.pepper", "notarealpepper");
+      hashIterations = options.intValue("hash.iterations", 2);
```
ericbuckley, line 110: any reason for a default of 2 iterations here compared to the one specified in `DirectRdaLoadApp.readRdaLoadOptionsFromProperties` (which is 100)?

brianburton: Since the default pepper isn't real I didn't think it mattered if the iterations were real either and 2 is cheaper than 100.  It could be changed though.  When loading real data we'd use a configuration that made these explicit and the default wouldn't matter.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/LoadRdaJsonApp.java
```diff
@@ -0,0 +1,157 @@
...
+      dbUrl = options.stringValue("database.url", "jdbc:hsqldb:mem:LoadRdaJsonApp");
+      dbUser = options.stringValue("database.user", "");
+      dbPassword = options.stringValue("database.password", "");
```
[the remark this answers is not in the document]
brianburton: Good point.  I didn't think it was worth it at this point but I could refactor that if/when we add another of these utility programs.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/ConfigLoader.java
```diff
@@ -0,0 +1,253 @@
...
+  private final Function<String, String> source;
```
christrotteradhoc, line 15: I was a bit thrown off by the name 'source' here and below. This does not really represent the source itself but rather the mapping function that the source gives rise to, and for this particular member it is a composite mapping that arises from potentially multiple sources, right? A javadoc on this member regardless of what we decide to call it would be appreciated.

Also, what were the factors that lead to using a Function<> instead of a Map<>? With the "new" mappings overwriting the "old" ones as sources are added I don't see anything that really requires the Function<> and it can make it more annoying to debug later if you want to print out the mapping in a debugger.

brianburton: The source of data for this loader is a function it can call to obtain a value using a key.  How that value is obtained doesn't matter to the ConfigLoader object.   So from the perspective of this class `source` actually is a source of data.  Certainly in some cases this source could be some sort of mapping that does extra stuff but generally it's more likely to be a closure of a collection and its getter method.

You are correct that the `source` is often a composite mapping but it doesn't have to be.  It could just be `map::get` or `System::getenv`.

I agree a comment is needed.  I'll add that.

Using a Function means the source can be literally anything and values can be obtained lazily.  Using a Map would require creating a new data structure up front and passing it to the ConfigLoader.   For example we don't want to create a Map of all possible environment variables and then add all system properties to that map.  Most of that work would be wasted since the ConfigLoader would only care about a few of the keys in the resulting Map.  What we need is just a way to grab only the values it needs and only when it needs them.  A lambda is great for that.

ericbuckley (approved):

KyRobbins (approved):

christrotteradhoc (approved):

PR #736, opened 2021-08-16
DCGEO-205: updating server-start to use the preadj.enabled flag

ericbuckley:
The server-start script reads in some command line arguments for determining system properties, these changes allow us to specify the value of the `bfdServer.preadj.enabled` flag via the arguments for local development.

Also, a small change was made to the timeout variable in the server-script, to allow it to be configurable as in some environments (specifically when using the docker-compose setup) it can take longer than 2 mins to start up.

### Change Details

- added an argument to the `server-start.sh` script to accept the pre-adj enabled flag
- hardcoded the above argument to `true` in `apps/pom.xml`
- updated the `SERVER_START_TIMEOUT` variable in `server-start.sh` so it can be overridden via an env variable

### External References

- [DCGEO-205](https://jira.cms.gov/browse/DCGEO-205)

### Feedback Requested
- this goes outside the scope of this PR, but I am curious as to why we manage runtime arguments in AWS with environmnent variables, but for local development many of these seem to be hard coded in pom.xml files.


apps/bfd-server/bfd-server-war/src/main/config/server-start.sh
```diff
@@ -30,7 +30,7 @@ if [[ "${cygwin}" = true ]]; then
 fi
 
 # Constants.
-serverTimeoutSeconds=120
+serverTimeoutSeconds=${SERVER_START_TIMEOUT:-120}
```
brianburton, line 33: Nice to have this configurable.

apps/bfd-server/bfd-server-war/src/main/config/server-start.sh
```diff
@@ -53,6 +53,7 @@ visualVm=""
 targetDirectory=
 dbUrl="jdbc:bfd-test:hsqldb:mem"
 v2Enabled="true"
+preadjEnabled="true"
```
brianburton, line 56: Will defaulting to true require other changes to prevent enabling it in production?

brianburton: Never mind this isn't used in prod.

KyRobbins (dismissed):

brianburton (dismissed):

jzulim (approved):
Follows previous patterns. Also, thanks for the timeout var change.

brianburton (dismissed):

christrotteradhoc:
Is there a ticket number for this?

ericbuckley:
> Is there a ticket number for this?

@christrotteradhoc indirectly yes, [creating a prototype runbook](https://jira.cms.gov/browse/DCGEO-111).  The confluence document I'm working on now is going to detail how an engineer can locally run all the BFD and BCDA services together to make preadj claim requests.  While working on this I discovered there is no way right now you can start the server locally with the `server-start.sh` script and make preadj data requests.  This PR reconciles that.

Being that it's only tangentially related, do you want me to link it to DCGEO-111 or would it be better to create a new JIRA issue?

christrotteradhoc:
I'd prefer a new JIRA if you are planning on having other commits under DCGEO-111

jzulim (dismissed):

jzulim:
Merge conflict. I can re-review along with DC Geo once that's pushed.

christrotteradhoc (approved):

KyRobbins (approved):

brianburton (approved):

PR #737, opened 2021-08-17
[BFD-1165] [SNYK] Upgrade hibernate to 5.4.24.Final 

christrotteradhoc:
### Change Details

Upgraded hibernate library as per SNYK recommendation but this resulted in two follow on issues that needed to be resolved:

- The workaround that was in place to override the byte-buddy library version due to a Mockito conflict had to be removed because it was using a scope of test and therefore not getting included in the war file. This appears not to have mattered in the older version of hibernate but the new one is blowing up because of this. The byte-buddy dependency in the new hibernate seems to be compatible with Mockito though -- all tests are passing.
- The new hibernate is not happy about an integer overflow that occurs when paging is not requested due to some code that is adding one to the page size and using that as the max results for a select query. When this code is used with the default page size of Integer.MAX_VALUE, overflow occurs and we end up passing a negative value. The proposed fix is to limit the page size to one less that MAXINT.

### Acceptance Validation

Unit and IT tests were run with no issues.

### Feedback Requested

General feedback requested.

### External References

- [BFD-1165](https://jira.cms.gov/browse/BFD-1165)

### Security Implications

Resolves a SNYK detected vulnerability in hibernate.


cbrunefearless (approved):
lgtm

mackec1 (approved):
looks good.

PR #740, opened 2021-08-18
[BFD-1095] CloudBees Core Migration (Jenkinsfile lift and shift approach)

njdister:
### Change Details

This PR takes a lift and shift approach to getting the existing Jenkinsfile and its method of execution (scripted pipeline with raw Groovy access) working in CloudBees Core.  I modified as little as possible to get it working.  This makes it easier on the review, although as I will describe in self reviewing some areas of the code, a little more risky than a pure refactor into declarative pipeline style.

Major changes:
- Create a "build container" environment using the publicly available maven docker container: https://hub.docker.com/_/maven
- Instantiate that build container environment via kubernetes pod template, specifying appropriate resource limits and service account (allows us to cross role assume in AWS)
- Wrap all appropriate steps in the build container environment
- Set persistent AWS credentials
- Remove hardcoded binary paths
- Refactor raw file object manipulation for builtin Jenkins functions since this doesn't seem to work in CloudBees Core otherwise

### Acceptance Validation

Sample full build + deploy to test env: https://jenkins-east.cloud.cms.gov/bfd/job/BFD%20-%20Multibranch%20and%20Multistage%20Pipeline/job/njdister%252FBFD-1095-cbcore-migration/49

### Feedback Requested

One potential point of contention is that we have to use a public ECR repository to host build environment container.  Private ECR is a no go due to the fact that the container checkout happens on the Kubernetes master, before the pod gets instantiated (and thus before any AWS cross role assumption can happen).  We tried without success to get private ECR working on BB2.  The only other option is to use Artifactory, which I believe we now have access to, I just have not spent the effort to learn and test it.  This would add some time if desired.  There is nothing inherently secret, private, or even specific to CMS / BFD in the build container.  It's just a bunch of pinned versions of tools we use to make our lives easier.

### External References

- [BFD-1095](https://jira.cms.gov/browse/BFD-1095)

### Security Implications


Jenkinsfile
```diff
@@ -124,161 +125,175 @@ def sendNotifications(String buildStatus = '', String stageName = '', String git
...
-	stage('Prepare') {
-		currentStage = "${env.STAGE_NAME}"
-		node {
-			// Grab the commit that triggered the build.
-			checkout scm
-
-			// Load the child Jenkinsfiles.
-			scriptForApps = load('apps/build.groovy')
-			scriptForDeploys = load('ops/deploy-ccs.groovy')
+	podTemplate(containers: [containerTemplate(name: 'bfd-cbc-build', image: 'public.ecr.aws/c2o1d8s9/bfd-cbc-build:jdk8-an29-tf12', command: 'cat', ttyEnabled: true, alwaysPullImage: true, resourceLimitCpu: '4000m', resourceLimitMemory: '8192Mi', resourceRequestCpu: '4000m', resourceRequestMemory: '8192Mi')], serviceAccount: 'bfd') {
```
njdister, line 128: Here's where we define the pod template that Jenkins will use to create our build container environment.  I'm using approximately the same amount of resources that are available on the existing EC2 Jenkins server.  I haven't noticed any significant delays in spinning up a container of this size (I was told you were allowed to request the entire resources of a node by ITOPS).

I've already mentioned the ECR vs Artifactory discussion in the PR but noting it here.

Jenkinsfile
```diff
@@ -124,161 +125,175 @@ def sendNotifications(String buildStatus = '', String stageName = '', String git
...
+					scriptForDeploys = load('ops/deploy-ccs.groovy')
+
+					// Set global AWS environment variables from role assumption
+					withCredentials([string(credentialsId: 'bfd-aws-assume-role', variable: 'awsAssumeRole')]) {
+						awsCredentials = sh(returnStdout: true, script: "aws sts assume-role --role-arn ${awsAssumeRole} --role-session-name session --output text --query Credentials").trim().split(/\s+/)
+						env.AWS_DEFAULT_REGION = 'us-east-1'
+						env.AWS_ACCESS_KEY_ID = awsCredentials[0]
+						env.AWS_SECRET_ACCESS_KEY = awsCredentials[2]
+						env.AWS_SESSION_TOKEN = awsCredentials[3]
+					}
```
njdister, line 147: Here's where we assume the role in the BFD account.  This gives us a unique set of credentials on every job run.

Jenkinsfile
```diff
@@ -124,161 +124,160 @@ def sendNotifications(String buildStatus = '', String stageName = '', String git
...
+						amiIds = scriptForDeploys.buildPlatinumAmi(amiIds)
+					}
+				} else {
+					org.jenkinsci.plugins.pipeline.modeldefinition.Utils.markStageSkippedForConditional('Build Platinum AMI')
 				}
 			}
-		} else {
-			org.jenkinsci.plugins.pipeline.modeldefinition.Utils.markStageSkippedForConditional('Deploy mgmt')
-		}
-	}
```
njdister, line 196: We remove deploy management here as it was never actually functional, and not needed anymore as we don't manage Jenkins in CB Core!

apps/bfd-server/bfd-server-war/src/main/config/server-start.sh
```diff
@@ -139,7 +139,8 @@ if [[ "${cygwin}" = true ]]; then keyStore=$(cygpath --mixed "${keyStore}"); fi
 if [[ "${cygwin}" = true ]]; then trustStore=$(cygpath --mixed "${trustStore}"); fi
 
 # Read the server port to be used from the ports file.
-serverPortHttps=${BFD_PORT:-$(grep "^server.port.https=" "${serverPortsFile}" | tr -d '\r' | cut -d'=' -f2)}
+serverPortHttps=$(grep "^server.port.https=" "${serverPortsFile}" | tr -d '\r' | cut -d'=' -f2)
```
njdister, line 142: **This** is a fun one.  I spent way too long wondering why this couldn't obtain the proper port written to file during the maven build.  As it turns out, `BFD_PORT` is an existing environment variable used by the CB Core environment so we run into an unexpected collision here when we allow it to override if present.  I would just rename it, but the BFD server relies on its presence to boot up.  For now I'm just removing the optional override and ensuring we always get the value from file.  Would be good to have @karlmdavis think about this and see if there's any possible scenario this could break local development.

apps/build.groovy
```diff
@@ -50,9 +27,8 @@ class AppBuildResults implements Serializable {
 def build(String build_env) {
 	dir ('apps') {
 
-		mvn "--update-snapshots -Dmaven.test.failure.ignore clean verify"
+		sh "mvn --update-snapshots -Dmaven.test.failure.ignore clean verify"
```
njdister, line 30: Since all maven settings and toolchain configs are embedded into our build environment container, we no longer need this and can run maven directly.  I think this is acceptable because the only time we should ever have to update maven settings or toolchains is if we bump a major version of java or maven, which would require a container rebuild anyway.

ops/deploy-ccs.groovy
```diff
@@ -86,41 +86,30 @@ def findAmis() {
...
- * Deploys/redeploys Jenkins and related systems to the management environment.
- *
- * @param amiIds an {@link AmiIds} instance detailing the IDs of the AMIs that should be used
- * @throws RuntimeException An exception will be bubbled up if the deploy tooling returns a non-zero exit code.
- */
-def deployManagement(AmiIds amiIds) {
-	echo 'Deploy to the CCS mgmt environment is not yet implemented.'
-	throw new UnsupportedOperationException('Deploy to the CCS mgmt environment is not yet implemented.')
-}
-
```
njdister, line 123: Not yet implemented, and never will be.  😉

ops/deploy-ccs.groovy
```diff
@@ -131,19 +120,16 @@ def deployManagement(AmiIds amiIds) {
...
-			varsFile.write(JsonOutput.toJson([
-				data_server_launcher: "${workspace}/${appBuildResults.dataServerLauncher}",
-				data_server_war: "${workspace}/${appBuildResults.dataServerWar}",
-				data_pipeline_jar: "${workspace}/${appBuildResults.dataPipelineUberJar}",
-			]))
+			writeFile file: "${workspace}/ops/ansible/playbooks-ccs/extra_vars.json", text: """{
+    "data_server_launcher": "${workspace}/${appBuildResults.dataServerLauncher}",
+    "data_server_war": "${workspace}/${appBuildResults.dataServerWar}",
+    "data_pipeline_jar": "${workspace}/${appBuildResults.dataPipelineUberJar}"
+}"""
```
njdister, line 129: As mentioned in the PR, all raw java file object manipulation does not work in CB Core.

ops/deploy-ccs.groovy
```diff
@@ -153,10 +139,10 @@ def buildAppAmis(String gitBranchName, String gitCommitId, AmiIds amiIds, AppBui
...
 			return new AmiIds(
 				platinumAmiId: amiIds.platinumAmiId,
-				bfdPipelineAmiId: extractAmiIdFromPackerManifest(new File(
-					"${workspace}/ops/ansible/playbooks-ccs/manifest_data-pipeline.json")),
-				bfdServerAmiId: extractAmiIdFromPackerManifest(new File(
-					"${workspace}/ops/ansible/playbooks-ccs/manifest_data-server.json")),
+				bfdPipelineAmiId: extractAmiIdFromPackerManifest(readFile(
+					file: "${workspace}/ops/ansible/playbooks-ccs/manifest_data-pipeline.json")),
+				bfdServerAmiId: extractAmiIdFromPackerManifest(readFile(
+					file: "${workspace}/ops/ansible/playbooks-ccs/manifest_data-server.json")),
```
njdister, line 145: Same as above.

ops/deploy-ccs.groovy
```diff
@@ -193,15 +179,15 @@ def deploy(String environmentId, String gitBranchName, String gitCommitId, AmiId
...
 		-no-color -input=false tfplan"
 		echo "Timestamp: ${java.time.LocalDateTime.now().toString()}"
 	}
 }
 
-def extractAmiIdFromPackerManifest(File manifest) {
+def extractAmiIdFromPackerManifest(String manifest) {
 	dir('ops/ansible/playbooks-ccs'){
-		def manifestJson = new JsonSlurper().parseText(manifest.text)
+		def manifestJson = new JsonSlurper().parseText(manifest)
```
njdister, line 190: Same as above.

ops/packer/build_bfd-all.json
```diff
@@ -69,6 +69,7 @@
     {
       "only": ["bfd-pipeline"],
       "type": "ansible",
+      "user": "ec2-user",
```
njdister, line 72: Due to differences in the default container user, I had to set this to get Ansible working.  I thought Jenkins ran as the `jenkins` user, so I"m not sure how this ever worked.  🤷

ops/packer/build_bfd-all.json
```diff
@@ -85,6 +86,7 @@
     {
       "only": ["bfd-server"],
       "type": "ansible",
+      "user": "ec2-user",
```
njdister, line 89: Same as above, default container user doesn't work with Ansible.

ops/packer/build_bfd-platinum.json
```diff
@@ -30,6 +30,7 @@
   ],
   "provisioners": [{
       "type": "ansible",
+      "user": "ec2-user",
```
njdister, line 33: Same as above, default container user doesn't work with Ansible.

ops/jenkins/Dockerfile.cbc-build
```diff
@@ -0,0 +1,22 @@
...
+
+RUN apt -y update && \
+    apt -y install python3-pip jq && \
+    pip3 install ansible==${ANSIBLE_VERSION} awscli boto && \
+    curl -L -o packer.zip https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip && \
+    unzip packer.zip -d /usr/local/bin && \
+    rm packer.zip && \
+    curl -L -o terraform.zip https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip && \
+    unzip terraform.zip -d /usr/local/bin && \
+    rm terraform.zip
```
njdister, line 22: Build a container based on the publicly available maven container, with all of the shiny tools we need.  This is a similar approach as I have taken in BB2: https://github.com/CMSgov/bluebutton-web-deployment/blob/master/Dockerfiles/Dockerfile.cbc-build

ops/jenkins/cbc-build-push.sh
```diff
@@ -0,0 +1,14 @@
...
+docker build --file Dockerfile.cbc-build \
+  --build-arg JAVA_VERSION=${2:-8} \
+  --build-arg MAVEN_VERSION=${3:-3} \
+  --build-arg ANSIBLE_VERSION=${4:-2.9.24} \
+  --build-arg PACKER_VERSION=${5:-1.6.6} \
+  --build-arg TERRAFORM_VERSION=${6:-0.12.31} \
+  --tag public.ecr.aws/c2o1d8s9/bfd-cbc-build:${DOCKER_TAG} \
+  .
+
+docker push public.ecr.aws/c2o1d8s9/bfd-cbc-build:${DOCKER_TAG}
```
njdister, line 14: Simple wrapper shell script to make building and pushing a maven environment container to ECR easier.  Similar approach I've taken in BB2: https://github.com/CMSgov/bluebutton-web-deployment/blob/master/Dockerfiles/cbc-build-push.sh

Will need updated if we go with Artifactory.

ops/jenkins/jdk8.toolchains.xml
```diff
@@ -0,0 +1,13 @@
...
+        <type>jdk</type>
+        <provides>
+            <version>1.8</version>
+            <vendor>openjdk</vendor>
+        </provides>
+        <configuration>
+            <jdkHome>/usr/local/openjdk-8</jdkHome>
+        </configuration>
+    </toolchain>
+</toolchains>
```
njdister, line 13: This is the only bit of toolchain I needed to get BFD builds workin in the maven container environment.

apps/bfd-server/bfd-server-war/src/main/config/server-start.sh
njdister: Confirmed this breaks some local development cases, refactored to only change behavior when running in Jenkins.

keithdadkins (approved):
Thanks for walking us through the code today. And great job working through all the issues!

jzulim (approved):
Great work. Thank you for this effort. 👍

PR #745, opened 2021-08-23
DCGEO-113: Support the RDA API replay feature.

brianburton:
### Change Details

The RDA API assigns a 64-bit signed sequence number to every change that it sends to its clients.  The sequence number can be used by clients to begin receiving claim updates at a specific point in the stream.  The pipeline will use this feature to allow us to roll back and re-receive claims in the event that we learn of data issues either in the data coming from the server or in our processing/storage of those claims.

Normally the pipeline job will simply pick up the stream where it left off (i.e. with the claim update immediately after the last one it has in its database) but if a special environment variable has been set the job will instead begin reading from the sequence number in the environment variable.

This PR adds support for this feature to the RDA pipeline as follows:

- Every PreAdjFissClaim and PreAdjMcsClaim record has its sequence number stored in a new column.
- Pipeline AppConfiguration loads two new optional environment variables that can be used to set a starting sequence number for each type of claim.
- GrpcRdaSource checks for either a hard coded sequence number or the maximum known value from the sink to obtain the starting sequence number to send to the RDA API.
- The RDA API callers always pass a meaningful starting sequence number to the API instead of just sending 0 as they did before.

Other changes in this PR include:
- An additional metric has been added to track the most recently written sequence number for each type of claim. This will allow a dashboard to visually represent whether or not claims are being received.
- The single JpaRdaSink has been replaced by claim specific subclasses.  These provide more meaningful log file entries as well as support using appropriate query to obtain max sequence number for each claim type.
- Several tests have been refactored to use inprocess HSQLDB and RDA API server with cleanup.  Previously these had duplicated the same cleanup code or did not have cleanup.  Now they use a static method that sets up the environment, runs a lambda, and cleans up the environment.

Many files had to be changed in this commit.  To assist reviewers in prioritizing their review the following sections list the files to be reviewed and the nature of the changes in each.  The info is broken down by module since the changes in each module had different purposes.

#### bfd-model-rda
| Files | Notes |
| --- | --- |
| PreAdjFissClaim.java PreAdjMcsClaim.java | Adds the new column to store `sequenceNumber` in claims. |
| SchemaMigrationIT.java | Updates integration test for new columns. |

#### bfd-model-rif

| Files | Notes |
| --- | --- |
| V34__preadj_rda_replay_table_updates.sql | Adds the new `sequenceNumber` column to both claim tables.  Also adds new indexes for `sequenceNumber` and `lastUpdated` to support efficient determination of what starting sequence number to use when calling RDA API. |

#### bfd-pipeline-app top level classes

Normally the pipeline jobs look in the database to get the highest known sequence number and ask the RDA APi to resume from that point.  In the event of a need to replay old data (i.e. to download corrected data from the RDA API following a bug fix) the pipeline jobs would need to be configured to start from a specific sequence number instead.

| Files | Notes |
| --- | --- |
| AppConfiguration.java | Loads optional new environment variables that can be used to specify a starting sequence number. |
| PipelineApplicationIT.java | Changed the test to add the required sequence number field in the function that creates the claim sources. |

#### bfd-pipeline-grpc-server

The server modue has programs used to mock the RDA API for various testing purposes.  Since the pipeline now supports the sequence number and the ability to replay from a specific sequence number these classes also had to be changed to add this ability.

| Files | Notes |
| --- | --- |
| RdaServerApp.java | Adds starting sequence number parameter to functions that create MessageSources. |
| ConfigLoader.java | Adds ability to read a long value from a config source. |
| DirectRdaLoadApp.java | Adds config options for setting starting sequence numbers for FISS and MCS claims. |
| LoadRdaJsonApp.java | Adds starting sequence number parameter to functions that create MessageSources. |
| ConfigLoaderTest.java | Adds unit test for long config values. |

#### bfd-pipeline-grpc

Adding support for sequence numbers and replay required a number of changes.

- The configuration objects had to accept starting sequence numbers.
- The RDASinks had to provide queries to determine appropriate starting sequence numbers.  This required splitting the single sink class into an abstract base class plus a small sub-class for each type of claim.
- The MessageSource factories needed to support starting at a particular sequence number.  This required a new functional interface that accepted a parameter rather than simply using a Supplier.

| Files | Notes |
| --- | --- |
| AbstractRdaLoadJob.java | Expanded to add support for starting sequence numbers. |
| RdaLoadOptions.java | Changed job creation methods to use proper sink subclass for each claim type. |
| RdaSink.java | Added a method to query the database for highest known sequence number.
| MessageSource.java EmptyMessageSource.java | Adds a Factory interface with a method that accepts a starting sequence number parameter. Adds a skip method for use in the tests to simulate replaying from known spots in a stream. |

**server package**

The `server` package contains classes associated with the mock RDA API server.  Changes in this package were needed to accurately simulate the RDA API's replay feature.

| Files | Notes |
| --- | --- |
| RandomFissClaimSource.java RandomMcsClaimSource.java | Add starting sequence number 0 to their MessageSources. |
| RdaServer.java | Accepts a MessageSource.Factory instead of a Supplier for creating MessageSources on demand. Adds a helper method for tests that need to run a server in process. |
| RdaService.java | Applies the `seq` value from the caller's `ClaimRequest` parameter to start the sequence at the proper point. |
| WrappedClaimSource.java | Adds support for incrementing and returning the `sequenceNumber` for each change. |

**sink package**

The `sink` package contains classes that store claims in the database.  Changes in the package were required to:

- Allow a sink to query for the highest sequence number that's been stored for a specific type of claim.
- Track the most recently used sequence number in a gauge for display in a dashboard.

Previously a single class was used as a sink for both types of claims.  Since each claim type requires its own query this was changed to have a common abstract base class plus a concrete subclass for each claim type.

| Files | Notes |
| --- | --- |
| AbstractClaimRdaSink.java | implements all the common functionality for sinks.  The metric names were adjusted to reflect the different subclasses.  The new gauge is updated as claims are written to the database. A protected method is provided for use in the subclass queries to get max sequence number. |
| FissClaimRdaSink.java McsClaimRdaSink.java | Claim type specific subclass that implements the query to get max sequence number. |

**source package**

The source package provides code for calling the RDA API, batching the resulting claims, and sending them to the sink for storage in the database.  Changes required in this package include:

- Reading the starting sequence number from each sink at startup.
- Optionally using a starting sequence number from the configuration instead of the one provided by the sink.
- Sending the starting sequence number to the RDA API server when making the gRPC call to get claims.
- Copying the sequence number from the RDA API change object into the JPA database entity objects for storage in our database.
- Adding code to test all of the above.
- Additionally log output was made more informative to assist in tracking pipeline job progress in splunk.

| Files | Notes |
| --- | --- |
| GrpcStreamCaller.java FissClaimStreamCaller.java McsClaimStreamCaller.java | Pass the starting sequence number to the RDA API server when calling the API. |
| FissClaimTransformer.java McsClaimTransformer.java | Copy the sequence number into our JPA entity object's `sequenceNumber` field. |
| GrpcRdaSource.java | Implement the logic to pick a proper starting sequence number to use, pass that to the caller.  Improve the logging output. |
| RdaPipelineTestUtils.java | Adds static helper method to encapsulate all of the boilerplate associated with running a database test in an in-memory HSQLDB database. |
| AbstractRdaLoadJobTest.java RdaLoadJobIT.java RdaLoadOptionsTest.java JsonMessageSourceTest.java RandomFissClaimSourceTest.java RandomMcsClaimSourceTest.java WrappedClaimSourceTest.java FissClaimRdaSinkIT.java FissClaimRdaSinkTest.java McsClaimRdaSinkIT.java McsClaimRdaSinkTest.java FissClaimStreamCallerIT.java FissClaimTransformerTest.java GrpcRdaSourceIT.java GrpcRdaSourceTest.java McsClaimStreamCallerIT.java McsClaimTransformerTest.java | Adds new tests.  Refactors some tests to improve readability and reliability. |

#### bfd-server-war

The `RDATestUtils.java` file was modified to add the required `sequenceNumber` field to the claims it creates.

### Acceptance Validation

Integration and unit tests verify the new functionality works as expected.

### Feedback Requested

Nothing in particular.

### External References

- [DCGEO-113](https://jira.cms.gov/browse/DCGEO-113)

### Security Implications

None

None

None

Sequence number of each claim is now being stored in a new column.

N/A

No.


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/AbstractClaimRdaSink.java
```diff
@@ -129,6 +144,35 @@ private void mergeBatch(Iterable<RdaChange<TClaim>> changes) {
     }
   }
 
+  private long maxSequenceInBatch(Collection<RdaChange<TClaim>> claims) {
+    OptionalLong value = claims.stream().mapToLong(RdaChange::getSequenceNumber).max();
```
ericbuckley, line 148: would this not always be the last claim on the stream?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/FissClaimRdaSink.java
```diff
@@ -0,0 +1,19 @@
...
+public class FissClaimRdaSink extends AbstractClaimRdaSink<PreAdjFissClaim> {
+  public FissClaimRdaSink(PipelineApplicationState appState) {
+    super(appState);
+  }
+
+  @Override
+  public Optional<Long> readMaxExistingSequenceNumber() throws ProcessingException {
+    return readMaxExistingSequenceNumber(
+        String.format(
+            "select max(c.%s) from PreAdjFissClaim c", PreAdjFissClaim.Fields.sequenceNumber));
```
ericbuckley, line 17: isn't the table name "pre_adj"."FissClaims"

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
```diff
@@ -135,12 +151,28 @@ public int retrieveAndProcessObjects(int maxPerBatch, RdaSink<TResponse> sink)
...
+   * start at the beginning.
+   *
+   * @param sink used to obtain the maximum known sequence number
+   * @return a valid RDA change sequence number
+   */
+  private long getStartingSequenceNumber(RdaSink<TResponse> sink) throws ProcessingException {
+    if (startingSequenceNumber.isPresent()) {
+      return startingSequenceNumber.get();
+    } else {
+      return sink.readMaxExistingSequenceNumber().map(seqNo -> seqNo + 1).orElse(MIN_SEQUENCE_NUM);
```
ericbuckley, line 172: Why `seqNo + 1`?

ericbuckley (dismissed):

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V34__preadj_rda_replay_table_updates.sql
```diff
@@ -0,0 +1,28 @@
...
+ * Every claim change from the RDA API contains a 64-bit sequence number that can be used
+ * to request a replay of past updates.  Each claim in the database will have the most recent
+ * sequence number that affected that claim stored in a column.
+ */
+
+/*
+ * Add the sequence number column used to track which update from RDA API was
+ * used to create or update the claim.
+ */
+alter table "pre_adj"."FissClaims"
```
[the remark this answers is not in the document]
brianburton: That's a cool observation and not one that I was aware of.  It may not be necessary since postgresql 11 though  according to this blog:  https://www.depesz.com/2018/04/04/waiting-for-postgresql-11-fast-alter-table-add-column-with-a-non-null-default/

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V34__preadj_rda_replay_table_updates.sql
```diff
@@ -0,0 +1,28 @@
...
+    ALTER COLUMN "sequenceNumber" DROP DEFAULT;
+ALTER TABLE "pre_adj"."McsClaims"
+    ALTER COLUMN "sequenceNumber" DROP DEFAULT;
+
+/* Index to allow quick determination of where to resume stream in calls to RDA API. */
+create index "FissClaims_sequenceNumber_idx" on "pre_adj"."FissClaims" ("sequenceNumber");
+create index "McsClaims_sequenceNumber_idx" on "pre_adj"."McsClaims" ("sequenceNumber");
+
+/* Index to allow fast query by update time by BFD API. */
+create index "FissClaims_lastUpdated_idx" on "pre_adj"."FissClaims" ("lastUpdated");
```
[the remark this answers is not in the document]
brianburton: The sequenceNumber index is for finding the max sequence number when the job starts.  The lastUpdated index is for helping us to find a sequence number manually to configure a replay if we know roughly what date/time bad records started to arrive from RDA API.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/AbstractClaimRdaSink.java
brianburton: Yes it should be.  This was a bit of defensive coding.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/FissClaimRdaSink.java
brianburton: Hibernate query language uses the entity name rather than the table name in the query.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
brianburton: If the last one we wrote to the database has sequence 100 then we don't need to receive that record again.  So to avoid the repeat we ask for 101.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V34__preadj_rda_replay_table_updates.sql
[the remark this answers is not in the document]
brianburton: Just to confirm the post is correct I tried out the SQL in the blog and the add column was instantaneous with 10,000,000 records.

KyRobbins (dismissed):

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
ericbuckley: ah of course, I'm not sure what I was thinking

ericbuckley (dismissed):

KyRobbins (dismissed):

ericbuckley (approved):

KyRobbins (approved):

christrotteradhoc (approved):

christrotteradhoc:
Looks good. Thanks for adding the extra documentation on the changes.

PR #763, opened 2021-09-08
DCGEO-171: Add S3 bucket as input to mock server RDA API

brianburton:
**JIRA Ticket:**
[DCGEO-171](https://jira.cms.gov/browse/DCGEO-171)

As a DCGEO developer I need to be able to configure the mock RDA API server to serve data in NDJSON format from an S3 bucket for testing purposes.

---

### What Does This PR Do?

This PR adds a feature to the RDA API mock server to allow it to read claims data from NDJSON files stored in an Amazon S3 bucket.  Accomplishing this led to the following changes:

- Added a new `MessageSource` implementation (`S3JsonMessageSource`) that reads data from a NDJSON file stored in an S3 bucket.
- Added new command line arguments to `RdaServerApp` to configure it to use `S3JsonMessageSources`.
- Moved some of the `S3Utilities` methods into a new class in the bfd-pipeline-shared module (`SharedS3Utilities`) so that it could be re-used in the RDA API pipeline.
- Added command line arguments as an option to `ConfigLoader` and modified `RdaServerApp` to use it for consistency with other apps in the `bfd-pipeline-rda-grpc-server` module.
- Added a `FilteredMessageSource` that allows messages to be filtered to implement replay logic when reading claims from an NDJSON file.
- Made various small changes to files to import and use the classes that were moved during refactoring.

* Does this make any backwards-incompatible changes that might break end user clients?
* Can you find any bugs if you run the code locally and test it manually?

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* [DCGEO-113 PR 745](https://github.com/CMSgov/beneficiary-fhir-data/pull/745)

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSource.java
```diff
@@ -0,0 +1,59 @@
...
+      return true;
+    }
+
+    while (source.hasNext()) {
+      nextValue = source.next();
+      if (filter.test(nextValue)) {
+        return true;
+      }
+    }
+    return false;
```
KyRobbins, line 42: I was going to comment on sequence number order here, as it could affect the way the filter is being used here, but it's probably reasonable to assume sequence numbers will be in order.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSourceTest.java
```diff
@@ -0,0 +1,62 @@
...
+  @Test
+  public void closePassesThrough() throws Exception {
+    MessageSource<Integer> source = createSource();
+    MessageSource<Integer> filtered = new FilteredMessageSource<>(source, i -> i != 2);
+    filtered.close();
+    verify(source).close();
+  }
+
+  @Test
+  public void multipleHashNextOk() throws Exception {
```
KyRobbins, line 30: small typo, `multipleHasNextOk()`

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSourceTest.java
```diff
@@ -0,0 +1,62 @@
...
+    assertTrue(filtered.hasNext());
+    assertTrue(filtered.hasNext());
+    verify(source, times(2)).hasNext();
+  }
+
+  @Test(expected = NoSuchElementException.class)
+  public void nextWithoutHasNextFails() throws Exception {
+    MessageSource<Integer> source = createSource();
+    MessageSource<Integer> filtered = new FilteredMessageSource<>(source, i -> i < 10);
+    filtered.next();
```
KyRobbins, line 48: Should this happen?  I think the normal behavior of iterator type APIs is that `next()` will always handle advancing the pointer as well as grabbing the next value, while `hasNext()` simply checks if the cursor is currently pointing to `null`, acting as a simple safeguard with no side effects.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSource.java
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Wrap the specified source and filter its objects using the supplied predicate. The predicate
+   * must return true for objects that are valid and false for objects that are invalid.
+   *
+   * @param source real source of objects
+   * @param filter function returning true for objects to keep or false for objects to skip
+   */
+  public FilteredMessageSource(MessageSource<T> source, Predicate<T> filter) {
+    this.source = source;
+    this.filter = filter;
```
KyRobbins, line 26: Should we advance the pointer here so that `next()` can be called?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSourceTest.java
```diff
@@ -0,0 +1,62 @@
...
+    assertEquals(Integer.valueOf(1), filtered.next());
+
+    assertTrue(filtered.hasNext());
+    assertTrue(filtered.hasNext());
+    assertTrue(filtered.hasNext());
+    verify(source, times(2)).hasNext();
+  }
+
+  @Test(expected = NoSuchElementException.class)
+  public void nextWithoutHasNextFails() throws Exception {
```
KyRobbins, line 45: Should we add a test to check if an exception is thrown when we are at the end of the iterator?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSource.java
brianburton: I agree with your logic and the thought that led to the question. This class is independent of sequence numbers so the order doesn't come into play here.  Generally the message sources needs to be provided with data in proper order.  For example JsonMessageSource doesn't sort the file because that could be very slow.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSourceTest.java
brianburton: The protocol here is that `hasNext()` has to be called before `next()`.  This isn't an `Iterator` so I feel free to deviate from that interface's protocol.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSource.java
brianburton: No it's part of the design that the constructor doesn't do any IO.  That happens in `hasNext()`.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/S3Utilities.java
```diff
@@ -1,70 +1,20 @@
...
-
-    return AmazonS3ClientBuilder.standard()
-        .withEndpointConfiguration(
-            new AwsClientBuilder.EndpointConfiguration(
-                minioConfig.minioEndpointAddress, awsS3Region.name()))
-        .withPathStyleAccessEnabled(true)
-        .withClientConfiguration(clientConfiguration)
-        .withCredentials(new AWSStaticCredentialsProvider(credentials))
-        .build();
+    return SharedS3Utilities.createS3Client(options.getS3Region());
```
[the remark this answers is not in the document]
brianburton: I retained this method because it accepts a class local to this module as an argument and it's used in 10 different places.  It's nice not to expand out the options.getS3Region() call in all of those places since it was previously an encapsulated detail.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/RdaServerApp.java
```diff
@@ -46,35 +50,27 @@ public static void main(String[] args) throws Exception {
...
-          throw new IOException("invalid argument: " + arg);
-        }
+      final ConfigLoader config =
+          ConfigLoader.builder().addKeyValueCommandLineArguments(args).build();
+      port = config.intValue("port", 5003);
+      seed = config.longOption("seed").orElseGet(System::currentTimeMillis);
+      maxToSend = config.intValue("maxToSend", 5_000);
+      fissClaimFile = config.readableFileOption("fissFile").orElse(null);
+      mcsClaimFile = config.readableFileOption("mcsFile").orElse(null);
+      final Optional<String> s3Bucket = config.stringOption("s3Bucket");
```
ericbuckley, line 63: thoughts about using a path style option here rather than just the bucket name.  I'm thinking that with the work the BFD team did last sprint in adding minio support, we could test this out locally using a minio path running an a local docker network?

brianburton: Yes I have tested this using Minio locally.  A bucket maps to a directory inside the minio data directory.  It doesn't allow path characters in the bucket name.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/server/FilteredMessageSourceTest.java
brianburton: Thanks. I have added that test.

ericbuckley (approved):
🎉

KyRobbins (approved):

christrotteradhoc (approved):

PR #784, opened 2021-09-30
DCEGO-170 (Draft): Add token generator (ICN/DCN)

KyRobbins:
**JIRA Ticket:**
[DCGEO-170](https://jira.cms.gov/browse/DCGEO-170)

Make it possible to generate unique ICN/DCN values of a specified pattern that directly correspond to a given input value (such as a unique CLAIM_ID)

---

### What Does This PR Do?

Th added feature is capable of generating structured token values of any length and of a specified pattern, backed by a numeric value allowing for deterministic generation of sequential values.  One immediate use of this feature will be to help generate ICN/DCN values for synthetic RDA claim data where the ICN/DCN are directly related to the RIF CLAIM_ID, yet uniquely distinct to avoid inaccurate association between adjudicated and pre-adjudicated synthetic data.

The feature was made generic enough that it can be used to generate sequential or random tokens of any defined pattern or length (limited only by internal BigInteger numeric backing value).

* Does the feature solve the proposed problem without becoming too bloated?
* Is it an appropriate tool to add to the shared utilities project?

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-shared-utils/pom.xml
```diff
@@ -20,18 +20,19 @@
...
 
-	<build>
-		<pluginManagement>
-			<plugins>
-			</plugins>
-		</pluginManagement>
-		<plugins>
-		</plugins>
-	</build>
-
```
brianburton, line 36: +1 for dead code removal

apps/bfd-shared-utils/src/main/java/gov/cms/bfd/sharedutils/generators/token/parser/AbstractTokenGroupParser.java
```diff
@@ -0,0 +1,78 @@
...
+          pattern.init();
+          patterns.add(pattern);
+          previousPattern = null;
+        } else {
+          throw new ParsingException("Repeat definition has no associated preceding token");
+        }
+      } else {
+        if (previousPattern != null) {
+          patterns.add(previousPattern);
+        }
```
brianburton, line 40: Seeing a downcast here is disconcerting.  How about adding previousPattern as an argument to parse() so the abstract base class doesn't have to do any casts.  The TokenRepeat could embed the previousPattern directly.  The parse() method could return a little bean with the TokenPattern it created and a new value (possibly null) for previousPattern or a flag indicating whether it consumed the previousPattern.

KyRobbins: Yep, I wasn't happy with my use of `instanceof` in this situation, it was a quick solution that I intended to look back at again but forgot to add a TODO.  I'll look at that again, and your suggestion for a refactor 👍

apps/bfd-shared-utils/src/main/java/gov/cms/bfd/sharedutils/generators/token/parser/TokenRepeatParser.java
```diff
@@ -0,0 +1,50 @@
+package gov.cms.bfd.sharedutils.generators.token.parser;
+
+import com.google.common.base.Objects;
+import gov.cms.bfd.sharedutils.generators.exceptions.ParsingException;
+import gov.cms.bfd.sharedutils.generators.token.pattern.TokenPattern;
+import gov.cms.bfd.sharedutils.generators.token.pattern.TokenRepeat;
+import java.util.Queue;
+
+/** This class parses repeated {@link TokenPattern}s, carrying overflow to the next place value. */
+public class TokenRepeatParser implements TokenParser {
```
brianburton, line 10: Adding an example of the pattern syntax to the class comment would make the intent of the class much clearer.

apps/bfd-shared-utils/src/main/java/gov/cms/bfd/sharedutils/generators/token/pattern/TokenAllOf.java
```diff
@@ -0,0 +1,115 @@
...
+
+    for (int i = 0; isValid && i < patterns.size(); ++i) {
+      TokenPattern tokenPattern = patterns.get(i);
+
+      String substring = value.substring(startIndex, startIndex + tokenPattern.tokenLength());
+      isValid = tokenPattern.isValidPattern(substring);
+      startIndex += tokenPattern.tokenLength();
+    }
+
+    return isValid;
```
brianburton, line 38: I think a `for (TokenPattern tokenPattern:patterns)` with a `return false` inside would be cleaner.

KyRobbins: It may work just fine here, I have a tendency to avoid fast fails because they are quite prone to refactoring bugs in my experience.

Gabasol:
@KyRobbins - I have a note that this PR is slated to be closed out/deleted. Is this still true? If so, when will it be closed out/deleted?

PR #794, opened 2021-10-18
DCGEO-219 (DRAFT): RFC for storing RDA claims using JSONB

brianburton:
**JIRA Ticket:**
[DCGEO-219](https://jira.cms.gov/browse/DCGEO-219)

Create an RFC proposal to store FISS and MCS claims using a JSONB column instead of existing BNF schema.

### What Does This PR Do?

The RFC says it all.  [Rendered RFC](https://github.com/CMSgov/beneficiary-fhir-data/blob/brianburton/dcgeo-219-jsonb-rfc/rfcs/0012-rda-claims-jsonb-DEFERRED.md)

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


ericbuckley (dismissed):

rfcs/0000-rda-claims-jsonb.md
```diff
@@ -0,0 +1,341 @@
...
+    "claim"          ${type.jsonb} not null,
+    constraint "FissClaimsJson_pkey" primary key ("dcn")
+);
+````
+The schema also contains one index for each of the queryable columns.
+
+Note the template macro for the `claim` column type.
+This macro is used since the type of the column will be different in postgresql vs HSQLDB:
+- In postgresql the `JSONB` type is used to allow use of JSON path querying for ad-hoc queries.
+- In HSQLDB `varchar(max)` is used since `JSONB` is not supported for that database.
```
mackec1, line 145: Will we ever exceed varchar(max) size in the jsonb payload...I think hsql maxes at 8k.

rfcs/0000-rda-claims-jsonb.md
```diff
@@ -0,0 +1,341 @@
...
+
+Since this proposal could be boiled down to using a NoSQL approach for FISS and MCS claims why not go all the way?
+A case could be made to move this data into Amazon's DynamoDB since it is built for this sort of design.
+While there could be advantages to that approach the proposed design has some practical advantages:
+- Using postgresql allows us to reap the benefits of a NoSQL approach without the need to add a whole new database technology to BFD.
+- RDS is already approved for storing PII so the changes can be implemented at any time without having to wait for review and approval of DynamoDB.
+
+
+## Prior Art
+[Prior Art]: #prior-art
```
mackec1, line 325: Not entirely sure this is true; while there may not be any prior art within BFD, document/NoSQL database are pretty pervasive these days.

rfcs/0000-rda-claims-jsonb.md
```diff
@@ -0,0 +1,341 @@
...
+
+None
+
+## Future Possibilities
+[Future Possibilities]: #future-possibilities
+
+Assuming this proposal works as expected the same technique could ultimately be applied to the rest of the BFD database.
```
mackec1, line 332: As far as I can see, most BFD queries are top-down (i.e., CarrierClaim to children CarrieClaimLines) so that lends itself to this paradigm. The only gotcha (and I'd have to check) is if we do any bottom-up (i.e., ClaimLines to Claim) queries.

mackec1 (commented):
This looks like a really good idea; I'm going to run some queries to see how this could apply to BFD proper. I'll let you know...mainly trying to see how many line items we'd have in the jsonb.

rfcs/0000-rda-claims-jsonb.md
brianburton: Good catch @mackec1 .  I actually used `LONGVARCHAR` in the prototype, not `varchar(max)`.  My mistake.  `LONGVARCHAR` defaults to 16mb but if could be be configured to use a clob in which case it could be gigabytes.

brianburton: I have fixed the RFC.

ericbuckley (approved):

keithdadkins (commented):
My only concerns: _note: I am not a DB admin, I just sometimes play one on tv_
1. The increased storage and financial costs at scale.. are we looking at millions of rows? billions? How about in 5 years?
2. If the size of a table row is greater than 8 kB (I believe is the default), postgres will compress and decompress the row (ie. TOAST). Depending on the volume, this may have an impact on your queries (and possibly the cluster) at scale.
3. I'm not familiar with how postgres handles jsonb columns internally during queries. But if it uses a lot of tmp/local disk space for sorting and whatnot, it may impact the cluster.
4. Normally the query planner looks at statistics for optimizing queries, but jsonb columns are not included. But not an issue if you are primarily relying on the other indexed columns so ymmv.
5. A few folks on the internets have complained about table bloat with jsonb (maybe VACUUM does not work the same on jsonb columns?). No idea and Aurora may do things differently so 🤷 .

Again, I'm not a db admin. Just thinking through some possible gotchas :)

ericbuckley:
> My only concerns: _note: I am not a DB admin, I just sometimes play one on tv_
> 
>     1. The increased storage and financial costs at scale.. are we looking at millions of rows? billions? How about in 5 years?
> 
>     2. If the size of a table row is greater than 8 kB (I believe is the default), postgres will compress and decompress the row (ie. TOAST). Depending on the volume, this may have an impact on your queries (and possibly the cluster) at scale.
> 
>     3. I'm not familiar with how postgres handles jsonb columns internally during queries. But if it uses a lot of tmp/local disk space for sorting and whatnot, it may impact the cluster.
> 
>     4. Normally the query planner looks at statistics for optimizing queries, but jsonb columns are not included. But not an issue if you are primarily relying on the other indexed columns so ymmv.
> 
>     5. A few folks on the internets have complained about table bloat with jsonb (maybe VACUUM does not work the same on jsonb columns?). No idea and Aurora may do things differently so 🤷 .
> 
> 
> Again, I'm not a db admin. Just thinking through some possible gotchas :)

@keithdadkins I’ll do my best to address these issues one by one.

1. The quantity of rows will decrease here, significantly compared to the current implementation, by removing child tables and their rows we will reduce the overall count.  We anticipate roughly 5 million new claims arriving each business day.

2. Yes, to my knowledge TOAST is enabled on JSONB columns, I’m no expert on this but from the reading I’ve done TOASTing and deTOASTing are extremely fast operations compared to JOINS and especially compared to multiple SELECT statements that require in memory merging (which I believe is what hibernate does by default for some parent child relationships).  From our tests, reducing our INSERTs from ~5 statements to just 1 makes the loading process much faster, even if Postgres is requiring a TOASTing strategy to store the data.

3. I wouldn’t be surprised if Postgres uses more tmp/local space for filtering/sorting on JSONB keys than a typical normalized schema, but there is no plan to use ANY JSONB keys for this in our SELECT statements in the FHIR API.  The only time we would make use of this is for analysis, say in the event of a data incident and we needed to determine the scope of the claims affected.

4. Again, all of our queries would be filtering and sorting on non-JSONB columns in the table.  So this wouldn’t be a concern for the FHIR API.

5. Storage size is a con, and outlined in more detail in drawback #2.  The main problem is that the keys are duplicated in every row, where in a fully normalized schema the column names don’t need to be repeated.  There are a few small wins regarding storage structure, like no need to store FKs or indexes on child tables, but overall we do anticipate needing more storage for a NoSQL-like schema.

KyRobbins (approved):

Gabasol:
@brianburton / @ericbuckley - should this PR revert to the draft state?

ericbuckley:
> @brianburton / @ericbuckley - should this PR revert to the draft state?

yeah just did that, we'll come back to this when we have more performance metrics to share

rfcs/0012-rda-claims-jsonb.md
```diff
@@ -0,0 +1,355 @@
+# RFC Proposal
+[RFC Proposal]: #rfc-proposal
+
+* RFC Proposal ID: `0012-rda-claims-jsonb`
+* Start Date: 2021-10-18
+* RFC PR: [CMSgov/beneficiary-fhir-data#???](https://github.com/CMSgov/beneficiary-fhir-data/pull/???)
```
ericbuckley, line 6:
```suggestion
* RFC PR: [CMSgov/beneficiary-fhir-data#794](https://github.com/CMSgov/beneficiary-fhir-data/pull/794)
```

ericbuckley (commented):
@brianburton can you change the file name to `rfcs/0012-rda-claims-jsonb[DEFERRED].json

rfcs/0012-rda-claims-jsonb.md
brianburton: Good catch!

ericbuckley (approved):

christrotteradhoc (approved):

PR #801, opened 2021-10-29
[BFD-1345] - Fix for SAMHSA Filter

cbrunefearless:
**JIRA Ticket:**
[BFD-1345](https://jira.cms.gov/browse/BFD-1345)

User Story:

As a user, I need to ensure that enabling SAMHSA filtering is not unnecessarily excluding claims (particularly inpatient), so that I have all the data that I have requested.

AC:

Ensure SAMHSA filter is NOT unnecessarily excluding inpatient claims 
A general review of the SAMHSA filtering code, for V1 and V2, is conducted to ensure it has adequate code coverage
Resolve any gaps that are identified
Notes/References:

Slack conversation where issue was originally brought up: https://cmsgov.slack.com/archives/C010WDXAZFZ/p1635273297049300

---

### What Does This PR Do?

- Modifies the in patient transformer to have an array of coding systems consisting of the original system (which is inconsistent with the rest of the code and caused the defect, but has to be preserved for backwards-compatibilty purposes) as well as the correct system (the system present in the other transformers and which is assumed in the SAMHSA matcher).
- Modifies expected results accordingly with the above.
- Adds unit and integration tests for matching and non-matching scenarios that cover all SAMHSA matcher logic including the defect as well as previously untested functionality in the matcher. (both V1 and V2)

Correctness of unit and IT tests wrt matcher logic.
Corner cases in matcher logic.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/InpatientClaimTransformerV2.java
```diff
@@ -342,7 +343,7 @@ private static ExplanationOfBenefit transformClaim(InpatientClaim claimGroup) {
               c ->
```
ericbuckley, line 343: why not use the `TransformerUtilsV2.mapHcpcs()` function that the other transformers are using? https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/OutpatientClaimTransformerV2.java#L286-L291

christrotteradhoc: That is on me actually. I asked for a smaller change because at first glance it appeared to me that there were some differences that led to this being done this way. Having looked at it more closely now I think @cbrunefearless 's original fix that did what you suggest is valid and preferable so I'm ok with going back to that.

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
```diff
@@ -2730,7 +2730,7 @@
         },
         "productOrService" : {
           "coding" : [ {
-            "system" : "https://bluebutton.cms.gov/resources/variables/hcpcs_cd",
+            "system" : "https://bluebutton.cms.gov/resources/codesystem/hcpcs",
```
karlmdavis, line 2733: Isn't this a breaking change?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/InpatientClaimTransformerV2.java
```diff
@@ -337,12 +338,8 @@ private static ExplanationOfBenefit transformClaim(InpatientClaim claimGroup) {
...
 
       // HCPCS_CD => item.productOrService
-      line.getHcpcsCode()
-          .ifPresent(
-              c ->
-                  item.setProductOrService(
-                      TransformerUtilsV2.createCodeableConcept(
-                          eob, CcwCodebookVariable.HCPCS_CD, c)));
+      TransformerUtilsV2.mapHcpcs(
+          eob, item, line.getHcpcsCode(), Optional.empty(), Collections.emptyList());
```
karlmdavis, line 342: It's been a _day_. Can someone help me, a half-braindead person, understand this?

It looks to me like this is the only functional change in the PR? I'm guessing that, previously, HCPCS codes for this one field were being mapped in a way that's inconsistent with the rest of the application? And that inconsistency was _somehow_ causing claims with those codes to be filtered out when they shouldn't have been?

karlmdavis (changes requested):
Thanks so much for your work on this!

Need a couple questions answered, I think.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/InpatientClaimTransformerV2.java
ericbuckley: Yeah, that's correct.  The current version of InpatientClaimTransformerV2 uses the "https://bluebutton.cms.gov/resources/variables/hcpcs_cd" coding, this coding system is unrecognized by the filter and as a [fail safe assumes that every value could be a SAMHSA code](https://github.com/CMSgov/beneficiary-fhir-data/blob/ae0b5af919d2923c5cd7b69d1f50cf28c04f5415/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java#L532-L538)

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
karlmdavis: Confirmed: this is a breaking change. We do **not** want to do that, as things stand right now.

Instead, we'll need to _add_ a `Coding` to the `CodeableConcept` that has the correct `system`, and leave the one with the goofy `system` there for backwards-compatibility purposes. In JSON terms, you're adding another `system+code` dict to the `coding` array.

This will require changes to our current SAMHSA filtering to work correctly. Right now, I think the relevant portion of the code is this:

```java
  private boolean containsSamhsaProcedureCode(CodeableConcept procedureConcept) {
    for (Coding procedureCoding : procedureConcept.getCoding()) {
      if (TransformerConstants.CODING_SYSTEM_HCPCS.equals(procedureCoding.getSystem())) {
        if (isSamhsaCptCode(procedureCoding)) return true;
      } else {
        /*
         * Fail safe: if we don't know the procedure Coding system, assume the code is
         * SAMHSA.
         */
        return true;
      }
    }

    // No blacklisted procedure Codings found: this procedure isn't SAMHSA-related.
    return false;
  }
```

In pseudo-code terms, we'll instead need it to look more like this:

```java
  private boolean containsSamhsaProcedureCode(CodeableConcept procedureConcept) {
    // Does the CodeableConcept have a legit HCPCS Coding?
    Optional<Coding> hcpcsCoding = findHcpcsCoding(procedureConcept);

    // Check that Coding to see if it's blacklisted.
    if(hcpcsCoding.isPresent() && isSamhsaCptCode(procedureCoding)) {
      return true;
    } else if(hcpcsCoding.isPresent() && hasUnknownSystem(procedureConcept)) {
        /*
         * Fail safe: if we don't know the procedure Coding system, assume the code is
         * SAMHSA.
         */
        return true;
    } else {
      // Otherwise, this only has known & non-SAMHSA-blacklisted procedure codes.
      return false;
    }
  }

  private boolean findHcpcsCoding(CodeableConcept procedureConcept) {
    ...
  }

  private boolean hasUnknownSystem(CodeableConcept procedureConcept) {
    Set<String> codingSystems = procedureConcept.getCoding().stream().map(coding -> coding.getSystem()).collect(Collectors.toSet());

    if (codingSystem.size() == 2 && codingSystem.contains(TransformerConstants.CODING_SYSTEM_HCPCS) && codingSystem.contains(CcwCodebookVariable.HCPCS_CD.whatever())) {
      /* add a comment here explaining why this complex check is needed and referencing the JIRA ticket */
      return false;
    } else if (codingSystem.size() == 1 && codingSystem.contains(TransformerConstants.CODING_SYSTEM_HCPCS)) {
      return false;
    } else {
      return true;
    }
  }
```

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
```diff
@@ -2730,6 +2730,10 @@
         },
         "productOrService" : {
           "coding" : [ {
+            "system" : "https://bluebutton.cms.gov/resources/variables/hcpcs_cd",
+            "code" : "M55"
+          },
+          {
             "system" : "https://bluebutton.cms.gov/resources/codesystem/hcpcs",
             "code" : "M55"
           } ]
```
KyRobbins: Would this not still trigger the filtering?  The samhsa logic seems to iterate over each coding in the `productOrService`. https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java#L528-L543

christrotteradhoc: more changes to come ; converted to draft for now

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -262,6 +271,14 @@ public Bundle findByPatient(
...
     List<IBaseResource> eobs = new ArrayList<IBaseResource>();
 
+    if (claimTypes.contains(ClaimTypeV2.INPATIENT)) {
+      if (lastUpdated.getLowerBound().getValue().after(DATE_OF_DEFECT)
+          && lastUpdated.getLowerBound().getValue().before(DATE_OF_FIX)) {
+        lastUpdated.setLowerBound(
+            new DateParam().setPrefix(ParamPrefixEnum.LESSTHAN_OR_EQUALS).setValue(DATE_OF_DEFECT));
+      }
+    }
+
```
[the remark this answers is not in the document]
albertico-gov: @KyRobbins Makes sense to me. cc: @christrotteradhoc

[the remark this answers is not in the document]
christrotteradhoc: Agreed - we will need more commenting in all of the places (this particular piece of code is very much a work in progress that is still pending some discussions within the team). As for removing this after a certain date, we are going to add it to the list of things to remove in V3 for sure but I think we were leaning more towards leaving it in V2 forever since we don't know for sure a date by which all consumers would have issued a new call. If we do have such a date though we can add that in a comment.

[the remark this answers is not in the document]
KyRobbins: Just to pose the question, was there any thought in just updating the `lastUpdated` of the affected claims rather then adding a temporary code patch?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +523,89 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+      return false;
+    }
+  }
+
+  /**
+   * @param procedureConcept the procedure {@link CodeableConcept} to check
+   * @return <code>true</code> if the specified procedure {@link CodeableConcept} contains the
+   *     {@link Coding} with the HCPCS {@link System}, <code>false</code> if it does not
+   */
+  private boolean findHcpcsCoding(CodeableConcept procedureConcept) {
```
lsmitchell, line 553: Might be better called "hasHcpcsCoding" or "containsHcpcsCoding" since we're not retrieving the Coding here and we're returning a boolean.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +523,89 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
     /*
      * Note: CPT codes represent a subset of possible HCPCS codes (but are the only
      * subset that we blacklist from).
      */
-    if (!TransformerConstants.CODING_SYSTEM_HCPCS.equals(procedureCoding.getSystem()))
-      throw new IllegalArgumentException();
+    for (Coding procedureCoding : procedureConcept.getCoding()) {
+      if (!TransformerConstants.CODING_SYSTEM_HCPCS.equals(procedureCoding.getSystem())) {
+        throw new IllegalArgumentException();
+      }
```
lsmitchell, line 601: This causes an issue if any of the system values are not HPCPCS (which they will always be, because we're adding both HCPCS and HCPCS_CS). The check here should allow for both systems codes, since we are adding both.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -595,8 +595,12 @@ private boolean isSamhsaCptCode(CodeableConcept procedureConcept) {
...
      */
+    Set<String> codingSystems =
+        procedureConcept.getCoding().stream()
+            .map(coding -> coding.getSystem())
+            .collect(Collectors.toSet());
     for (Coding procedureCoding : procedureConcept.getCoding()) {
-      if (!TransformerConstants.CODING_SYSTEM_HCPCS.equals(procedureCoding.getSystem())) {
+      if (!codingSystems.contains(TransformerConstants.CODING_SYSTEM_HCPCS)) {
         throw new IllegalArgumentException();
       }
```
lsmitchell: This can be moved outside the for loop.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +523,93 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+      }
 
-    /*
-     * Note: per XXX all codes in icd10DiagnosisCodes are already normalized.
-     */
-    return cptCodes.contains(normalizeHcpcsCode(procedureCoding.getCode()));
+      /*
+       * Note: per XXX all codes in icd10DiagnosisCodes are already normalized.
+       */
+      return cptCodes.contains(normalizeHcpcsCode(procedureCoding.getCode()));
```
lsmitchell, line 610: 🍰  It would be nice to check all/both codes, so we should probably only return if true, else continue. Although both codes should theoretically be the same, I'd recommend being cautious here.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
KyRobbins: Could this ever happen?  For `isSamhsaCptCode()` to get invoked, it would have had to pass the `hasHcpcsCode` check it would seem.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +523,89 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+        && codingSystems.contains(
+            TransformerUtilsV2.calculateVariableReferenceUrl(CcwCodebookVariable.HCPCS_CD))) {
+      return false;
+    } else if (codingSystems.size() == 1
+        && codingSystems.contains(TransformerConstants.CODING_SYSTEM_HCPCS)) {
+      return false;
+    } else {
+      return true;
+    }
+  }
```
KyRobbins, line 585: Seems like maybe this could be condensed to:
```java
private boolean hasUnknownSystem(CodeableConcept concept) {
  Set<String> codingSystems =
        procedureConcept.getCoding().stream()
            .map(coding -> coding.getSystem())
            .collect(Collectors.toSet());
            
  return !codingSystems.isEmpty()
    && !codingSystems.equals(Set.of(TransformerConstants.CODING_SYSTEM_HCPCS))
    && !codingSystems.equals(Set.of(TransformerConstants.CODING_SYSTEM_HCPCS, CcwCodebookVariable.HCPCS_CD));
}```

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 8 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java]: Yeah at present this can't be triggered. For super safety you could make the argument that this method could be called from elsewhere in the future and to guard against improper use, but I've seen the argument made not to be overly defensive too.

Comment (KyRobbins) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java]: I'm also wondering if this should remain one method as well?  I like breaking things out into simple logical methods, but I'm wondering if this all falls under the same check concept?
```java
  protected boolean containsSamhsaProcedureCode(CodeableConcept procedureConcept) {
    Map<String, Coding> codings = procedureConcept.getCoding().stream()
            .collect(Collectors.toMap(Coding::getSystem, Function.identity()));

    if (codings.containsKey(TransformerConstants.CODING_SYSTEM_HCPCS)) {
      return isSamhsaCptCode(codings.get(TransformerConstants.CODING_SYSTEM_HCPCS)) // HCPCS Samhsa code
              // Or if there are extra unwanted systems
              || (!codings.keySet().equals(Set.of(TransformerConstants.CODING_SYSTEM_HCPCS))
                && !codings.keySet().equals(Set.of(TransformerConstants.CODING_SYSTEM_HCPCS, TransformerUtilsV2.calculateVariableReferenceUrl(CcwCodebookVariable.HCPCS_CD))));
    } else {
      return !codings.isEmpty();
    }
  }

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 10 above)
Comment (christrotteradhoc) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:585]: I agree that what is there now is a bit clunky. I am wondering if it might be more intuitive to write this function in positive logic and adjust the code that is calling it:

```java
private boolean hasKnownSystem(CodeableConcept concept) {
  Set<String> codingSystems =
        procedureConcept.getCoding().stream()
            .map(Coding::getSystem)
            .collect(Collectors.toSet());

  // Write a nice comment here about why this is considered to be a known system.
  Set<String> knownHcpcsSystem = Set.of(TransformerConstants.CODING_SYSTEM_HCPCS);

  // Write a nice comment here about why we need this.
  Set<String> backwardsCompatibleHcpcsSystem = Set.of(TransformerConstants.CODING_SYSTEM_HCPCS, TransformerUtilsV2.calculateVariableReferenceUrl(CcwCodebookVariable.HCPCS_CD);
  
  return codingSystems.equals(knownHcpcsSystem) || codingSystem.equals(backwardsCompatibleHcpcsSystem);
}```

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java (hunk 5 above)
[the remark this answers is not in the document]
Comment (christrotteradhoc) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java]: we are not doing this anymore but one of the main reasons to not update the DB was to allow this to be specific to v2.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 8 above)
Comment (karlmdavis) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java]: Just want to chime in on the general principle here: we should bias **heavily** towards being overly defensive in any/all of our systems. Extra assertions hurt **nothing** and often make the code much simpler for readers to reason about, which is a major win.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 10 above)
Comment (karlmdavis) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:585]: One minor comment: I'm a bit concerned that there are V1/V2 versions of this method: `TransformerUtilsV2.calculateVariableReferenceUrl(...)`. If they're actually the same code, we should get rid of the duplicate. If they're _not_ the same code, it's probably a bad idea to use them here, as the SAMHSA filtering logic should not be version-specific.

Comment (karlmdavis) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:585]: As an aside, that's delightfully clever code, @christrotteradhoc ! I haven't had enough coffee to try and verify it yet, but like I said, I'm going to mostly be watching the test coverage from here on out, anyways.

Comment (karlmdavis) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:585]: @KyRobbins Same compliment for your proposed alternative, too. Bravo!

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 8 above)
Comment (KyRobbins) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java]: As a single method, it wouldn't be an issue, I'm just wondering if that might serve better here, as the check is very specific to this one use case it seems like, and it's arguably not too over complicated as such.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 10 above)
Comment (KyRobbins) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:585]: > I agree that what is there now is a bit clunky. I am wondering if it might be more intuitive to write this function in positive logic and adjust the code that is calling it:
> 
> ```java
> private boolean hasKnownSystem(CodeableConcept concept) {
>   Set<String> codingSystems =
>         procedureConcept.getCoding().stream()
>             .map(Coding::getSystem)
>             .collect(Collectors.toSet());
> 
>   // Write a nice comment here about why this is considered to be a known system.
>   Set<String> knownHcpcsSystem = Set.of(TransformerConstants.CODING_SYSTEM_HCPCS);
> 
>   // Write a nice comment here about why we need this.
>   Set<String> backwardsCompatibleHcpcsSystem = Set.of(TransformerConstants.CODING_SYSTEM_HCPCS, TransformerUtilsV2.calculateVariableReferenceUrl(CcwCodebookVariable.HCPCS_CD);
>   
>   return codingSystems.equals(knownHcpcsSystem) || codingSystem.equals(backwardsCompatibleHcpcsSystem);
> }
> ```

In that scenario, I think you'd want to name it `containsOnlyKnownSystems()`, as `hasKnownSystem()` infers that there is at least one known one.  Also not sure if you need to account for an empty list scenario.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java [hunk 11]
```diff
@@ -522,43 +523,93 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
-  private boolean isSamhsaCptCode(Coding procedureCoding) {
+  private boolean containsOnlyKnownSystems(CodeableConcept procedureConcept) {
+    Set<String> codingSystems =
+        procedureConcept.getCoding().stream().map(Coding::getSystem).collect(Collectors.toSet());
+
+    String hcpcsCdSystem =
+        String.format(
+            "%s/%s",
+            TransformerConstants.BASE_URL_CCW_VARIABLES,
+            CcwCodebookVariable.HCPCS_CD.getVariable().getId().toLowerCase());
```
christrotteradhoc, line 575: makes more sense to me to use the method like you were doing before (the one that gets called from the transformer) but just move that method into a class that is not v2 specific.

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
christrotteradhoc: @karlmdavis We ended up needing to deviate from this some so would appreciate the extra scrutiny on what was implemented vs what is here.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
christrotteradhoc: agreed and we ended up having to account for the empty systems in the higher level method to match v1 behavior.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +524,95 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+   * @return <code>true</code> if the specified procedure {@link CodeableConcept} contains at least
+   *     one {@link Coding} and only contains {@link Coding}s that have known coding systems <code>
+   *     false</code> otherwise
+   */
+  private boolean containsOnlyKnownSystems(CodeableConcept procedureConcept) {
+    Set<String> codingSystems =
+        procedureConcept.getCoding().stream().map(Coding::getSystem).collect(Collectors.toSet());
+
+    String hcpcsCdSystem = CCWUtils.calculateVariableReferenceUrl(CcwCodebookVariable.HCPCS_CD);
+    String hcpcsSystem = TransformerConstants.CODING_SYSTEM_HCPCS;
```
mackec1, line 579: sort of OK, I guess; just not crazy about creating a variable from a static constant variable.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -1442,7 +1632,9 @@ private Beneficiary loadSampleA() {
    * Find the first Beneficiary from a record list returned by {@link ServerTestUtils#loadData(List}
    *
    * @param loadedRecords to use
-   * @return the first Beneficiary
+   * @param loadedRecords the loaded records
+   * @return the first Beneficiary*
+   * @return the beneficiary
```
mackec1, line 1637: not sure about the 2 'return' entries for JavaDoc.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +524,95 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+      return false;
+    }
+
+    // Does the CodeableConcept have a legit HCPCS Coding?
+    boolean hasHcpcsCoding = hasHcpcsCoding(procedureConcept);
+
+    // Check that Coding to see if it's blacklisted.
+    if (hasHcpcsCoding && isSamhsaCptCode(procedureConcept)) {
+      return true;
+    } else if (!containsOnlyKnownSystems(procedureConcept)) {
```
KyRobbins, line 542: The only thing I'm not thrilled about with this is that it technically iterates over the same codings 3 separate times (assuming you made it through the `else if` block)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +524,95 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
       throw new IllegalArgumentException();
+    }
 
     /*
      * Note: per XXX all codes in icd10DiagnosisCodes are already normalized.
      */
-    return cptCodes.contains(normalizeHcpcsCode(procedureCoding.getCode()));
+    return procedureConcept.getCoding().stream()
+        .anyMatch(
+            procedureCoding -> cptCodes.contains(normalizeHcpcsCode(procedureCoding.getCode())));
```
KyRobbins, line 615: This seems a little disconnected to me, as it checks if there is a HCPCS systems coding, but then checks if _any_ of the codings has a HCPCS code, which theoretically might not be the HCPCS system coding.  Would seem like maybe just checking if it's both a HCPCS system & code would serve the purpose?
If it's not a HCPCS system, this would never have been called in the current implementation.  It's also explicitly checking if a `CodeableConcept` has a HCPCS samhsa code, which would **not** seem to infer a throw if it didn't have a hcpcs system imo, my intuition would expect it to just return false if there is no hcpcs system coding, since it does not contain a hcpcs system code.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
christrotteradhoc: good catch, fixed it.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
KyRobbins: I kind of agree here, seems a bit redundant, particularly when they are only used to build a couple of sets, but not a big deal either way.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +524,95 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+
+    // Valid system url for productOrService coding
+    Set<String> knownHcpcsSystem = Set.of(hcpcsSystem);
+
+    // Additional valid coding system URL for backwards-compatibility
+    // See: https://jira.cms.gov/browse/BFD-1345
+    Set<String> backwardsCompatibleHcpcsSystem = Set.of(hcpcsSystem, hcpcsCdSystem);
+
+    return codingSystems.equals(knownHcpcsSystem)
+        || codingSystems.equals(backwardsCompatibleHcpcsSystem);
```
KyRobbins: It was mentioned I believe that empty collections are handled at a "higher" level, but related to the "defensive coding" ideas mentioned, I think checking for an empty set here could be appropriate.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -900,126 +901,229 @@ public void searchForEobsByMissingPatient() {
...
+      entityManager = entityManagerFactory.createEntityManager();
+
+      // Tweak the SAMPLE_A claims such that they are SAMHSA-related.
+      adjustCarrierClaimForSamhsaDiagnosis(loadedRecords, entityManager);
+      adjustInpatientRecordForSamhsaDiagnosis(loadedRecords, entityManager);
+      adjustOutpatientRecordForSamhsaDiagnosis(loadedRecords, entityManager);
+      adjustHhaRecordForSamhsaDiagnosis(loadedRecords, entityManager);
+      adjustSnfRecordForSamhsaDiagnosis(loadedRecords, entityManager);
+      adjustHospiceRecordForSamhsaDiagnosis(loadedRecords, entityManager);
+      adjustDmeRecordForSamhsaDiagnosis(loadedRecords, entityManager);
```
KyRobbins, line 929: Are these only setting up tests for one possible samhsa code?  Would we want to check each place a samhsa code could be present? (ProcedureComp, DiagnosisComp, Line item, both for recognized SAMHSA codes and unknown systems for each)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
christrotteradhoc: I don't really like it either, will change it

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -900,126 +901,229 @@ public void searchForEobsByMissingPatient() {
...
-   * with <code>excludeSAMHSA=true</code> properly filters out SAMHSA-related claims.
+   * Load the SAMPLE_A resources and then tweak each of the claim types that support SAMHSA (all
+   * except PDE) to have a SAMHSA diagnosis code.
    *
-   * @throws FHIRException (indicates test failure)
+   * @return the beneficary record loaded by Sample A
    */
-  @Test
-  public void searchForEobsWithSamhsaFiltering() throws FHIRException {
+  private Beneficiary loadSampleAWithSamhsa() {
```
KyRobbins, line 909: It's easier to just pull data from the db and modify it slightly for the needs of the test, but I'm wondering if it would be worth just building claims as needed for the test in code, as it really only needs the values related to the SamhsaMatcher logic.  This would speed up these tests quite a bit not requiring a DB be set up for them.
For testing the endpoint, it would seem just a single (or two to test not set) tests would suffice, since testing if the `excludeSamhsa` parameter is set should be the same for all I think?

KyRobbins:
Woops, meant to just delete my reply to Chris's comment because it was already resolved, sorry about that

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
```diff
@@ -0,0 +1,751 @@
...
+          {inpatientClaim, inpatientEob},
+          {outpatientClaim, outpatientEob},
+          {dmeClaim, dmeEob},
+          {hhaClaim, hhaEob},
+          {hospiceClaim, hospiceEob},
+          {snfClaim, snfEob},
+          {carrierClaim, carrierEob},
+          {pdeClaim, pdeEob}
+        });
+  }
```
KyRobbins, line 114: Ah, this is where all the different types of samhsa codes are checked.  So my question here is just do we need to start with a real claim?  Seems like we only need to worry about the samhsa matcher specific data here?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
```diff
@@ -0,0 +1,751 @@
...
+    RifRecordBase carrierClaim = getClaim(CarrierClaim.class);
+    ExplanationOfBenefit carrierEob =
+        CarrierClaimTransformerV2.transform(new MetricRegistry(), carrierClaim, Optional.empty());
+
+    RifRecordBase pdeClaim = getClaim(PartDEvent.class);
+    ExplanationOfBenefit pdeEob =
+        PartDEventTransformerV2.transform(new MetricRegistry(), pdeClaim, Optional.empty());
+
+    // Load the claim types into the test data that will be run against each test
+    return Arrays.asList(
```
KyRobbins, line 103: Just a point of note, `List.of()` is a little bit of an "upgrade" now that BFD is in jdk11, it produces an immutable list rather then a fixed length array list.  It's also just a tiny bit terser.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/SamhsaMatcherFromClaimTransformerTest.java
```diff
@@ -0,0 +1,649 @@
...
+        diagnosisComponent.setPackageCode(null);
+      }
+    }
+
+    // Set procedure to empty so we dont check it for matches
+    for (ExplanationOfBenefit.ProcedureComponent diagnosisComponent : modifiedEob.getProcedure()) {
+      CodeableConcept codeableConcept = diagnosisComponent.getProcedureCodeableConcept();
+      ArrayList<Coding> codingList = new ArrayList<>();
+      codeableConcept.setCoding(codingList);
+    }
```
KyRobbins, line 572: Small thing, but this seems a bit verbose, I think it could just be
```java
modifiedEob.getProcedure().forEach(c -> c.getProcedureCodeableConcept().setCoding(new ArrayList<>()));

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 16 above)
Comment (christrotteradhoc) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java]: Adding the empty set condition here would make this more permissive and less defensive -- there would be more things that qualify as only having known systems and the failsafe condition is based on whether we have unknowns.

I've gone back and forth a few times on this aspect and ultimately decided to document the more defensive behavior in the method comment here and have an explicit check with comment in the caller.

Comment (KyRobbins) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java]: Right, I wasn't meaning in order to make it more defensive, just in the aspect of being thorough.  As you said, the "failsafe" is expected to fire is based on whether we have unknowns, firing "if we don't know the procedure coding system", which doesn't seem as intuitive imo if there is no coding system.
Handling it in the invoking method does of course work, just means you'd have to replicate that step for every method that would want to invoke this one, as it doesn't seem like there'd be a use case, as far as I'm aware of, where you'd want an empty list of codings to actually trigger a samhsa filter.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 14 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:542]: Good point, this can be simplified at least by merging the first two iterations into one and making a single method for the first condition.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 15 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:615]: Indeed, it was a little repetitive. Merged the two checks here and just returned false if we didn't have a HCPCS Coding, which should reduce the redundancy and confusion.

File: apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java (hunk 17 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java:929]: Yes these set up the ITs for Diagnosis checks only. The unit tests more thoroughly go over each of these SAMHSA trigger areas for each of the claim types where they apply.

File: apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java (hunk 19 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java:114]: The test is also checking that the output of the ClaimTransformer can be appropriately caught in the matcher, thats why we're ingesting the data from a "real" (test data) claim. The issue we had was related to a mismatch in the code set by the transformer and what we filter on, so this seemed like a good way to enforce that contract.

File: apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java (hunk 20 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java:103]: Good call! I'll replace it, since I do like that it reads better.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 8 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java]: Resolved by making it one method :)

File: apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/SamhsaMatcherFromClaimTransformerTest.java (hunk 21 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/SamhsaMatcherFromClaimTransformerTest.java]: Adjusted!

File: apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java (hunk 18 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java:909]: What you described is basically the testing we did in the unit tests, where we added a lot more coverage and pulled from static files. For the integration tests we're going to err on the side of slightly more coverage and leave the four tests, since they each explicitly setup and test a scenario.

**Comment by ericbuckley** [comment]
@christrotteradhoc @karlmdavis I understand the need for making the coding system change backwards compatible in the short-term.  However, long-term, will the `hcpcs_cd` coding system ever be removed or will it stay here for the life of V2?

Also I just noticed that the [hcpc_cd](https://bluebutton.cms.gov/resources/variables/hcpcs_cd/) URL resolves, but [hcpcs](https://bluebutton.cms.gov/resources/variables/hcpcs/) does not, so we need a separate ticket in the static website to address that?

**Review by KyRobbins — COMMENTED** [review]
Only one last question, the rest looks good 👍

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java [hunk 22]
```diff
@@ -522,43 +524,79 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+    } else if (!containsOnlyKnownSystems(procedureConcept)) {
+      /*
+       * Fail safe: if we don't know the procedure Coding system, assume the code is
+       * SAMHSA.
+       */
+      return true;
+    } else {
+      // Otherwise, this only has known & non-SAMHSA-blacklisted procedure codes.
+      return false;
+    }
```
KyRobbins, line 548: Actually I just noticed, I think this could also be simplified to:
```java
if (hasHspcAndSamhsaCptCode(procedureConcept)) {
  return true;
} else {
  return !containsOnlyKnownSystems(procedureConcept));
}

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java [hunk 23]
```diff
@@ -522,43 +524,79 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
 
     /*
      * Note: per XXX all codes in icd10DiagnosisCodes are already normalized.
      */
-    return cptCodes.contains(normalizeHcpcsCode(procedureCoding.getCode()));
+    return procedureConcept.getCoding().stream()
+        .anyMatch(
+            procedureCoding -> cptCodes.contains(normalizeHcpcsCode(procedureCoding.getCode())));
```
KyRobbins, line 599: I was noticing this seems different from the existing checks.  `isSamhsaCptCode` and `isSamhsaIcd9[10]procedure[diagnosis]` all throw if it's the wrong system, where this (`hasHspcAndSamhsaCptCode`) just returns false.  Just wondering if it would be good to make all the API methods a little more consistent in behavior?

lsmitchell: Could use some domain knowledge input here, do we expect HCPCS to _always_ exist? Theres a bit of a difference between gracefully bowing out of the filter check with false, and blowing up with an exception. (Maybe the other methods could/should be changed to be more graceful, or maybe this one should be more restrictive, depending on the answer to that).

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +524,79 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
   }
```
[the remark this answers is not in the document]
lsmitchell: See below comment on if we should actually keep this false (in which case we could do this) or if it should be an exception like the other methods.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
KyRobbins: Perhaps that could be done as a documentation comment of the method just to keep the code a little more terse?  Just a thought.
```java
/**
 * This method will return true (contains SAMHSA) if the concept contains a HCPCS samhsa coding
 * or a coding with an unknown system.
 */

Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:548]: Actually this made me notice this method was missing a description entirely, I'll simplify and move the explanation to the method description. Good call!

**Review by karlmdavis — CHANGES_REQUESTED** [review]
Bravo!

There's a couple of moderate issues here that I'd like to see fixed before this is merged, but they're both super simple.

Also left a bunch of minor suggestions that shouldn't block approval.

File: apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java (hunk 23 above)
Comment (lsmitchell) [reply apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java:591]: Thanks for the insight Karl! I'll adjust this to be an exception (and fix the tests to expect that for a wrong system scenario).

File: apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/SamhsaMatcherTest.java [hunk 25]
```diff
@@ -688,6 +699,146 @@ public void matchSnfClaimsByDrg() throws FHIRException {
...
+        .getService()
+        .setCoding(
+            Collections.singletonList(
+                new Coding(TransformerConstants.CODING_SYSTEM_HCPCS, "H0050", "")));
+
+    // When
+    boolean isMatch = samhsaMatcher.test(explanationOfBenefit);
+
+    // Then
+    assertTrue(isMatch);
```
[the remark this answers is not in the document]
lsmitchell: Yes, copy-paste error! I'll fix this.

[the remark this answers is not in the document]
lsmitchell: Actually these tests were copies of the v2 test class for testing v1 before we just copied the whole class and modified it for v1, so these are duplicates of the tests in SamhsaMatcherFromClaimTransformerTest and can be removed. I didn't notice that the initial effort had gone into this class.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
```diff
@@ -0,0 +1,845 @@
...
+  public void setup() {
+    samhsaMatcherV2 = new R4SamhsaMatcher();
+  }
+
+  /**
+   * Data collection.
+   *
+   * @return the collection
+   */
+  @Parameterized.Parameters(name = "{0}")
```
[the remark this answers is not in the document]
lsmitchell: .class doesn't seem to work, the parameterized namestring seems pretty limited. It can only use one of the parameters' toString, so to make it prettier i'd have to format the classname and pass a separate list of params just for test display purposes. At the moment it uses the ugly full path+classname which was a good improvement over the default of just using the parameter index, and was the best thing I could get to work without doing something hacky(er) just for display purposes:

image

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
```diff
@@ -0,0 +1,845 @@
...
+
+    // PDE has no SAMHSA, so expect no match on SAMSHA filter
+    // DME, HHA, Hospice, Carrier does not look at procedure so it wont match
+    if (claim instanceof PartDEvent
+        || claim instanceof DMEClaim
+        || claim instanceof HHAClaim
+        || claim instanceof HospiceClaim
+        || claim instanceof CarrierClaim) {
+      expectMatch = false;
+    }
```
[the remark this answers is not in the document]
lsmitchell: Claim is actually dual-purpose:
1. Determine which claimtype we're using, which to your point could be done with the type field and the util
2. Have a param list with which to tie the testname to, see above comment

If I didn't need the test name as well I'd be inclined to nix it and use the type field, but since it also names the tests and we have no better option to fix the display name that I know of, I think we should keep it as-is for now.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
```diff
@@ -0,0 +1,845 @@
...
+
+    verifySamhsaMatcherForProcedureIcd(
+        IcdCode.CODING_SYSTEM_ICD_9,
+        BLACKLISTED_IC9_PROCEDURE_CODE,
+        expectMatch,
+        loadedExplanationOfBenefit);
+  }
```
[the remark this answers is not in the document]
lsmitchell: Yeah, I was iterating over solutions as I was writing these up and starting to consider the various claim types (and started realizing that all scenarios dont apply to each one, which led to these exclusions here.)

Created https://jira.cms.gov/browse/BFD-1345 for cleaning this up a bit with properties so we don't need so much exclusion code in the tests themselves.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/CCWUtils.java
```diff
@@ -0,0 +1,18 @@
+package gov.cms.bfd.server.war.commons;
+
+import gov.cms.bfd.model.codebook.model.CcwCodebookInterface;
+
+/** As set of methods to work with CCW Codebook variables. */
```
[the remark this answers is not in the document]
lsmitchell: Fixed, good eye!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +524,79 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+  }
+
+  /**
+   * @param procedureConcept the procedure {@link CodeableConcept} to check
+   * @return <code>true</code> if the specified procedure {@link CodeableConcept} contains any
+   *     {@link Coding}s that match any of the {@link #cptCodes}, <code>false</code> if they all do
+   *     not
    */
-  private boolean isSamhsaCptCode(Coding procedureCoding) {
+  private boolean hasHspcAndSamhsaCptCode(CodeableConcept procedureConcept) {
```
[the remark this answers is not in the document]
lsmitchell: Good catch, fixed.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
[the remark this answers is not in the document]
lsmitchell: Since I changed it back to an exception, I left this as-is.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
```diff
@@ -522,43 +524,79 @@ private static String normalizeIcd10Code(String icd10DiagnosisCode) {
...
+    Set<String> codingSystems =
+        procedureConcept.getCoding().stream().map(Coding::getSystem).collect(Collectors.toSet());
+
+    // Does the CodeableConcept have a legit HCPCS Coding?
+    if (!codingSystems.contains(TransformerConstants.CODING_SYSTEM_HCPCS)) {
+      return false;
+    }
 
     /*
      * Note: per XXX all codes in icd10DiagnosisCodes are already normalized.
```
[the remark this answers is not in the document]
lsmitchell: Fixed the CPT code to be normalized in the constructor, added that these codes are normalized in the javadoc, and removed these comments.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -1029,33 +1133,119 @@ public void searchForEobsWithSamhsaFiltering() throws FHIRException {
...
+    IGenericClient fhirClient = ServerTestUtils.get().createFhirClientV2();
+
+    Bundle searchResults =
+        fhirClient
+            .search()
+            .forResource(ExplanationOfBenefit.class)
+            .where(
+                ExplanationOfBenefit.PATIENT.hasId(
+                    TransformerUtilsV2.buildPatientId(beneficiary.getBeneficiaryId())))
+            .and(new StringClientParam("excludeSAMHSA").matches().value("true"))
```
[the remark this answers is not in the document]
lsmitchell: Good call, fixed all of these.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -1490,6 +1681,12 @@ private Bundle fetchWithLastUpdated(
...
   }
 
+  /**
+   * Assert each eob.
+   *
+   * @param searchResults the search results
+   * @param loadedRecords the loaded records
+   */
   /*
    * Verify that each of the expected claims (one for every claim type) is present
```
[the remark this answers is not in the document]
lsmitchell: Indeed, I think autodoc goofed it since the comment was not a javadoc. I combined them.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/SamhsaMatcherTest.java
```diff
@@ -688,6 +699,146 @@ public void matchSnfClaimsByDrg() throws FHIRException {
...
 
+  /**
+   * Verifies that when transforming a claim into an ExplanationOfBenefit which has
+   * item[n].productOrService.coding[n].system (procedure code) values which = has eob.procedure
+   * SAMSHA code, then the SAMHSA matcher's test method will identify this as a SAMSHA related
+   * ExplanationOfBenefit.
+   */
+  @Test
+  public void
+      testSamhsaMatcherWhenTransformedInpatientHasItemWithNoValidSamhsaProcedureCodeReturnsFalse() {
```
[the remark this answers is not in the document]
lsmitchell: My explictly-constructed-but-wordy test names :)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
lsmitchell: One thing to note about this though, the same test for an unknown system does NOT cause an exception in v1. Returning false actually matches v1 functionality, for consistency. Maybe that means v1 is less secure and we've "upgraded" v2, but just an FYI.

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
karlmdavis: This was addressed.

karlmdavis:
> @christrotteradhoc @karlmdavis I understand the need for making the coding system change backwards compatible in the short-term. However, long-term, will the hcpcs_cd coding system ever be removed or will it stay here for the life of V2?

The life of V2. Backwards compatibility is a long-term need for our users, not a short-term one.

> Also I just noticed that the hcpc_cd URL resolves, but hcpcs does not, so we need a separate ticket in the static website to address that?

It's a good idea.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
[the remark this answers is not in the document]
christrotteradhoc: I had a small issue with this where from the test result view, if I tried to right click and rerun a test, it was passing this reference from the previous run and the test was failing. Given that plus the conversation below about the claim not being needed, I'm leaning slightly toward the unused parameter solution so that the name of the test would be things like 'INPATIENT', 'HHA', etc.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
karlmdavis: I did not explain this particularly well.

@lsmitchell Is this method ever called for `CodeableConcept`s that don't contain the HCPCS code? Like, on purpose? If so, throwing an exception is inappropriate. If not, it is.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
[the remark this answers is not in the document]
karlmdavis: I don't think it's great to have the test case names change every time, as it could confuse things that track test case successes vs. failures over time, e.g. Jenkins.

This is still a minor point that shouldn't block approval, but I'd add a new name param as `{0}` in the params array to address this.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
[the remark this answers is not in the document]
karlmdavis: See previous comment on test case names.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/SamhsaMatcherTest.java
```diff
@@ -699,146 +689,6 @@ public void matchSnfClaimsByDrg() throws FHIRException {
...
-    if (claim != null) {
-      claim.setLastUpdated(Instant.now());
-    } else {
-      throw new IllegalStateException(
-          "Test setup issue, did not find expected InpatientClaim in sample record.");
-    }
-
-    return claim;
-  }
-
```
karlmdavis: Just curious: why'd all these test cases drop out in this revision?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
```diff
@@ -236,26 +237,34 @@ public SamhsaMatcherR4FromClaimTransformerV2Test(
...
+          "unknown/system/code",
+          NON_SAMHSA_HCPCS_CODE,
+          "unknown/system/code",
+          NON_SAMHSA_HCPCS_CODE,
+          expectMatch,
+          loadedExplanationOfBenefit);
+    } catch (IllegalArgumentException ex) {
+      // Expect an exception; dont use junit built in because some paths dont hit an exception
+      return;
+    }
```
karlmdavis, line 264: Please see my clarification on this. I'm guessing, though, that we _are_ expecting exceptions to occur now? If so, we should likely revert that change.

karlmdavis (commented):
Couple questions. None of these are blocking.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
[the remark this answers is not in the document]
lsmitchell: Wrote up an adjustment to pass the output from the util, and confirmed with the fix that the tests can be rerun individually:
image

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/SamhsaMatcherTest.java
lsmitchell: @cbrunefearless started porting tests a couple tests from v1 when we were first starting, and they got put in this class. Eventually to port the v1 tests we just copied the whole test file and made v1 modifications, so the tests in this class were duplicates.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
lsmitchell: Adjusted back to false, apologies for the misunderstanding!

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/SamhsaMatcherR4FromClaimTransformerV2Test.java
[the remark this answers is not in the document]
lsmitchell: Changed to be a string parameter, which cleans up the test names.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4SamhsaMatcher.java
lsmitchell: Since DME can have other codes, switched back to false and left a comment noting why.

christrotteradhoc (approved):
Big thank you to everyone that participated in the coding and reviewing of this!

PR #802, opened 2021-10-29
BFD-1165: upgrade hibernate to 5.6.1

christrotteradhoc:
**JIRA Ticket:**
[BFD-1165](https://jira.cms.gov/browse/BFD-1165)

---

### What Does This PR Do?

Upgraded hibernate library as per SNYK recommendation but this resulted in two follow on issues that needed to be resolved:

- The workaround that was in place to override the byte-buddy library version due to a Mockito conflict had to be removed because it was using a scope of test and therefore not getting included in the war file. This appears not to have mattered in the older version of hibernate but the new one is blowing up because of this. The byte-buddy dependency in the new hibernate seems to be compatible with Mockito though -- all tests are passing.
- The new hibernate is not happy about an integer overflow that occurs when paging is not requested due to some code that is adding one to the page size and using that as the max results for a select query. When this code is used with the default page size of Integer.MAX_VALUE, overflow occurs and we end up passing a negative value. The proposed fix is to limit the page size to one less that MAXINT.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -219,15 +219,6 @@
...
 		</dependency>
-		<!-- Hibernate brings in an older version of byte-buddy which seems
-		to conflict with mockito, which wants version 1.11.0.  Bringing it
-		in as a first class dependency to resolve the issue. -->
-		<dependency>
-			<groupId>net.bytebuddy</groupId>
-			<artifactId>byte-buddy</artifactId>
-			<version>1.11.0</version>
-			<scope>test</scope>
-		</dependency>
```
christrotteradhoc, line 230: Latest version of hibernate does not have this issue. Leaving this here with the newer version of hibernate leads to other issues so removing this now unnecessary workaround.

mackec1 (approved):
LGTM, but it feels a little wierd that we don't have properties to define a max page size that we're willing to support.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/PatientLinkBuilder.java
```diff
@@ -13,6 +13,13 @@
 
 /** A link builder for Patient resources using bene-id cursors */
 public final class PatientLinkBuilder implements LinkBuilder {
+  /**
+   * Maximum page size is one less than the maximum integer value to allow clients to request one
+   * more than the page size as a means to see if an additional page is necessary without having an
+   * integer overflow.
+   */
+  public static final int MAX_PAGE_SIZE = Integer.MAX_VALUE - 1;
```
jzulim, line 21: 👍

jzulim (approved):
Nice work!

PR #810, opened 2021-11-10
BFD-1352: Adjust BFD unit tests to use golden bene ID that matches data in environment

lsmitchell:
**JIRA Ticket:**
[BFD-1352](https://jira.cms.gov/browse/BFD-1352)

As an engineer I would like the unit and integration test data to use the golden beneficiary ID for the sample-A files so that when doing calls against the test server I know which environment data corresponds to the test data.

---

### What Does This PR Do?
Updates the BFD unit tests and sample-A data to use the golden bene ID instead of the id they were using before (567834). This does not change the test data other than which ID is used, so that the sample-A bene id in source control matches what was loaded into the actual environments.

Replace me.

* There are no unit tests left with bene ID 567834
* The IT tests do not need to change as a result of this
* All tests are passing as a result of this change

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/TransformerTestUtilsV2.java
```diff
@@ -113,6 +115,15 @@ static void assertEobCommonClaimHeaderData(
...
 
+  /**
+   * Gets the golden beneficiary id.
+   *
+   * @return the golden beneficiary id
+   */
+  static String getGoldenBeneId() {
+    return GOLDEN_BENE_ID;
+  }
+
```
lsmitchell, line 126: Is there a better place anyone can think of to put this? It seems ok here, but realistically we need it outside this package too. If there is some common util that multiple packages could read from to get this data, this might be better placed there?

christrotteradhoc: maybe in a class under gov.cms.bfd.model.rif.samples since that is where the data lives too? Do you think we need the method though or can we just go with a single constant?

christrotteradhoc:
There is a reference to this ID in the top level README as well.

mackec1 (approved):
I'll defer to your judgement; however, Trotter's notion of putting the definition outside a version-dependant file is appleaing; some candidates:

./bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/CCWUtils.java
./bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
./bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseTestUtils.java

cbrunefearless (approved):
LGTM

mackec1 (approved):
Hopefull you were able to use/apply a 'replace all' for this one. Nice work!

PR #869, opened 2021-12-14
PACA-275: RDA claim parallel writes

brianburton:
**JIRA Ticket:**
[PACA-275](https://jira.cms.gov/browse/PACA-275)

Modify the RDA API pipeline to improve claim ingestion performance.

---

### What Does This PR Do?

Thie PR implements several changes to increase claim processing throughput:

- Modifies the `RdaSink` interface to allow it to be implemented asynchronously.
- Modifies `GrpcRdaSource` to use the new `RdaSink` methods when saving data to the database.
- Adds a new table `RdaApiProgress` to store sequence numbers for each claim type.  These are used to know what starting sequence number to request when calling the RDA API.  During data ingestion the `WriterThreadPool` class uses a `SequenceNumberTracker` class to determine which sequence number should be written to this table and the `SequenceNumberWriterThread` does the actual writing.  Comments in these classes explain the logic.
- Moves `message->claim` transformation into the `RdaSink` so that it can be done in worker threads.  Previously this transformation was done in the `GrpcStreamCaller` classes which was a bottleneck when using asynchronous writes.
- Adds a new `ConcurrentRdaSink` that uses a pool of worker threads to write claims and sequence numbers to the database asynchronously.  This is a general wrapper that delegates work to entity specific single-threaded sink implementations.  During data ingestion the `ConcurrentRdaSink` uses a `WriterThreadPool` to manage a pool of multiple `ClaimWriterThread`s and a single `SequenceNumberWriterThread` to perform the database updates.
- Modifies `AppConfiguration` to allow the number of claim writer threads to be configured for each job using an environment variable.  Note that this controls the number of writers per job.  Since there are two jobs (one for FISS and another for MCS claims) the actual number of writer threads will be double this number.
- Splits the `RdaSink` implementation classes into two new packages (`gov.cms.bfd.pipeline.rda.grpc.sink.direct` and `gov.cms.bfd.pipeline.rda.grpc.sink.concurrent`) to make the code easier to follow.  **This refactoring created some noise in the `Files changed` list but was worthwhile since the code in the two packages are logically very different.**. The `direct` classes have all the knowledge about FISS/MCS claims while the `concurrent` classes have no knowledge of those but instead have knowledge of threads/queues etc.
- Removes `persist() then merge()` optimization logic from `AbstractClaimRdaSink`.  That feature added complexity and was tuned for incoming data that was almost all new claims.  Current understanding is that many or most claims will be updates to existing records so the extra complexity wasn't worthwhile.  We can revisit this once we have prod data history next year.
- Adds a `CachingIdHasher` to avoid recomputing recently computed MBI values.  For now uses an in memory cache but can easily be extended to perform database lookups in a future PR.
- Allows `RdaServer` to be configured to listen on a specific host name for added security when benchmarking.

### How do the pieces fit together?

The pipeline job defines an interface named `RdaSink` for objects that handle data transformation and storage of the objects it ingests.  Objects that implement `RdaSink` do the following:

- Read and write sequence numbers.  More on these below.
- Transform RDA API change messages to JPA entities.
- Write JPA change entities to the database.

The RDA API returns a stream of change messages.  Each change message has a sequence number and a claim.  The sequence numbers are unique to each change and increase with every change.  Callers can send a starting sequence number to resume at that point in the stream.  We store only the latest version of each claim in our database.

There are 3 `RdaSink` implementations:

- `FissClaimRdaSink` synchronous implementation that supports FISS claims.
- `McsClaimRdaSink` synchronous implementation that supports MCS claims.
- `ConcurrentRdaSink` asynchronous implementation that is a wrapper on top of one of the other two.

The `ConcurrentRdaSink`'s worker threads use the appropriate sink implementation to perform the actual transformation and database writes.  Every incoming claim is assigned to a worker thread based on a hash of its claim ID.  This is done to ensure that we can't write two changes for the same claim out of sequence.

As claims are scheduled for writing we have to track their sequence numbers so that we can maintain the correct value in the `RdaApiProgress` table.  The correct value is always the highest sequence number for which we know there are no unsaved changes with a lower sequence number.   This logic is handled using a class called `SequenceNumberTracker`.

There are two types of worker threads.
- `SequenceNumberWriterThread` updates the `RdaApiProgress` table.  There is always one of these for each pool.
- `ClaimWriterThread` updates the appropriate claim tables.  There are a variable number of these based on configuration settings.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* This PR contains a schema change that appeared to require a separate PR to be merged ahead of it to disable hibernate schema validation.
* [PR 904](https://github.com/CMSgov/beneficiary-fhir-data/pull/904)

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/WriterThreadPool.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  public Optional<Long> readMaxExistingSequenceNumber() throws ProcessingException {
+    return sink.readMaxExistingSequenceNumber();
+  }
+
+  /**
+   * Schedules the current sequence number to be written to the database by a worker thread.
+   *
+   * @throws ProcessingException
```
KyRobbins, line 140: Description here for throws

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -49,33 +70,84 @@ public void close() throws Exception {
...
+      String query =
+          String.format(
+              "select p.%s from RdaApiProgress p where p.claimType='%s'",
+              RdaApiProgress.Fields.lastSequenceNumber, claimType.name());
+      final List<Long> sequenceNumber =
+          entityManager.createQuery(query, Long.class).getResultList();
+      final Optional<Long> answer =
+          sequenceNumber.isEmpty() ? Optional.empty() : Optional.of(sequenceNumber.get(0));
+      logger.info(
+          "max sequence number result is {}", answer.map(n -> Long.toString(n)).orElse("none"));
```
KyRobbins, line 92: This seems like it could possibly be a non-trivial calculation, should we wrap this with a `logger.infoEnabled()` check?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
```diff
@@ -113,7 +114,7 @@ public GrpcRdaSource(
    * @throws ProcessingException wrapper around any Exception thrown by the service or sink
    */
   @Override
-  public int retrieveAndProcessObjects(int maxPerBatch, RdaSink<TResponse> sink)
+  public int retrieveAndProcessObjects(int maxPerBatch, RdaSink<TMessage, TClaim> sink)
```
KyRobbins, line 117: This seems like a very complex method, curious if this can be broken up a little for better readability

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaSink.java
```diff
@@ -39,30 +56,33 @@
    * @return number of objects successfully processed
    * @throws ProcessingException if the operation fails
    */
-  default int writeObject(T object) throws ProcessingException {
-    return writeBatch(Collections.singleton(object));
+  default int writeMessage(String dataVersion, TMessage object) throws ProcessingException {
```
KyRobbins, line 59: Update javadocs for params

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/DirectRdaLoadApp.java
```diff
@@ -54,9 +54,12 @@ public static void main(String[] args) throws Exception {
     reporter.start(5, TimeUnit.SECONDS);
 
     final RdaLoadOptions jobConfig = readRdaLoadOptionsFromProperties(options);
-    final DatabaseOptions databaseConfig = readDatabaseOptions(options);
+    final DatabaseOptions databaseConfig =
+        readDatabaseOptions(options, jobConfig.getJobConfig().getWriteThreads());
     HikariDataSource pooledDataSource =
         PipelineApplicationState.createPooledDataSource(databaseConfig, metrics);
+    System.out.printf("thread count is %d%n", jobConfig.getJobConfig().getWriteThreads());
+    System.out.printf("database pool size %d%n", pooledDataSource.getMaximumPoolSize());
```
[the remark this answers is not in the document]
brianburton: This is a command line program used for local testing so it doesn't use logging for basic output.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/DirectRdaLoadApp.java
```diff
@@ -89,12 +92,12 @@ public static void main(String[] args) throws Exception {
...
   }
 
-  private static DatabaseOptions readDatabaseOptions(ConfigLoader options) {
+  private static DatabaseOptions readDatabaseOptions(ConfigLoader options, int threadCount) {
     return new DatabaseOptions(
         options.stringValue("database.url", null),
         options.stringValue("database.user", null),
         options.stringValue("database.password", null),
-        10);
+        options.intValue("database.maxConnections", Math.max(10, 5 * threadCount)));
```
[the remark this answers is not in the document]
brianburton: Since this is just run for local testing/benchmarking that's not a problem.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/assembly/zip.xml
```diff
@@ -0,0 +1,32 @@
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2"
```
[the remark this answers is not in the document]
brianburton: I need to be able to upload the server jar to an EC2 instance to run it for benchmarking.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/ClaimWriterThread.java
```diff
@@ -0,0 +1,275 @@
...
+ * messages for processing and uses a callback function to report its successes and failures.
+ *
+ * <p>The thread is stopped by calling its close() method. This adds a special token to the work
+ * queue that tells the thread to flush its buffer and exit its run loop.
+ *
+ * @param <TMessage> RDA API message class
+ * @param <TClaim> JPA entity class
+ */
+public class ClaimWriterThread<TMessage, TClaim> implements Callable<Integer>, AutoCloseable {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ClaimWriterThread.class);
```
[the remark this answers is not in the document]
brianburton: Good point.  Are we using`@Slf4j` in other places?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/ClaimWriterThread.java
```diff
@@ -0,0 +1,275 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+import lombok.AllArgsConstructor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A Callable object submitted to an ExecutorService to process RDA API messages and write them to
+ * the database using an RdaSink object. Each thread has its own BlockingQueue to receive inbound
+ * messages for processing and uses a callback function to report its successes and failures.
```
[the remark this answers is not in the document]
brianburton: Absolutely necessary here to ensure claims are written on a single thread.  The failure mode here is simple.  If any thread fails the whole process stops (just as a single threaded sink would work).

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
brianburton: Since answer is an optional this isn't really expensive.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/ClaimWriterThread.java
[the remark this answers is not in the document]
KyRobbins: I'm using it in my bridge CLI tool, but I'm not aware of it being used in any other module, as we don't have lombok in anything except the new modules we added I don't think.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/ClaimWriterThread.java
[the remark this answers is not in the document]
KyRobbins: I actually realized this later after writing the comment when I saw notes about maintaining order, chronographically, of like claims, but thanks for the verification!

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
brianburton: It's all very linear.  It hasn't changed much from master and since it's mostly catch blocks to carry through the correct processed count in a ProcessingException it can't be separated out much.  I tried and the separate method just contains the about the same set of catch blocks as its caller.  I suggest we consider that as a refactoring of its own for a future ticket.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/SequenceNumberWriterThread.java
```diff
@@ -0,0 +1,212 @@
...
+   * @param sink RdaSink used to write the sequence number
+   * @param sequenceNumber value to write
+   */
+  private void writeSequenceNumber(RdaSink<TMessage, TClaim> sink, long sequenceNumber)
+      throws InterruptedException {
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("writing sequenceNumber {}", sequenceNumber);
+    }
+    sink.updateLastSequenceNumber(sequenceNumber);
+  }
```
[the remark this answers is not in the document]
brianburton: Nothing.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/ConcurrentRdaSink.java
```diff
@@ -0,0 +1,154 @@
...
+   * should manage sequence number updates itself (true) or not update sequence numbers (false).
+   *
+   * @param maxThreads number of threads to use when writing
+   * @param keyExtractor function to get a key to dedup records
+   * @param sinkFactory function to create a sink
+   * @param <T> type of objects being written
+   * @return either a simple sink or a ConcurrentRdaSink
+   */
+  public static <TMessage, TClaim> RdaSink<TMessage, TClaim> createSink(
+      int maxThreads, int batchSize, Function<Boolean, RdaSink<TMessage, TClaim>> sinkFactory) {
```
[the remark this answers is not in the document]
brianburton: Fixed.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V37__api_progress_table.sql
```diff
@@ -0,0 +1,11 @@
+/*
+ * Creates a table to track the highest sequence number imported into the database.
+ * A query of this table can be used to track the sequence number to pass as a
+ * {@code since} parameter to the RDA API.
+ */
+
+create table "pre_adj"."RdaApiProgress" (
+    "claimType"          VARCHAR(20) NOT NULL PRIMARY KEY,
+    "lastSequenceNumber" BIGINT      NOT NULL,
```
ericbuckley, line 9: with this addition, do we need to keep indexes for the sequence numbers in the FISS and MCS claims tables?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/SequenceNumberTracker.java
```diff
@@ -0,0 +1,78 @@
...
+   * Gets the current sequence number for which we know there are no unwritten records with a lower
+   * sequence number.
+   *
+   * @return the best sequence number to send to the RDA API when fetching claims
+   */
+  public synchronized long getHighestWrittenSequenceNumber() {
+    if (activeSequenceNumbers.size() > 0) {
+      // Since the set is ordered the first value is lowest unwritten value.
+      // We return one less so that caller knows to resume at our first unwritten value.
+      return activeSequenceNumbers.iterator().next() - 1;
```
ericbuckley, line 72: I'm pretty sure this doesn't matter, but I think there is a chance here of returning a value that has yet to be recorded in the DB.  There is no guarantee that we will receive a message for every possible sequence number value.  If for example RDA never sends a message for sequence number 100, but at the time of running this function the smallest number in the set for 101, this function will return 100.  Again, this seems fine if all we need is a number to know where to continue from, but if there is some logic that assumes this value DOES exist in our DB we might have a problem.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -49,33 +70,84 @@ public void close() throws Exception {
...
+   * @throws ProcessingException if the operation fails
+   */
+  @Override
+  public Optional<Long> readMaxExistingSequenceNumber() throws ProcessingException {
+    try {
+      logger.info("running query to find max sequence number");
+      String query =
+          String.format(
+              "select p.%s from RdaApiProgress p where p.claimType='%s'",
+              RdaApiProgress.Fields.lastSequenceNumber, claimType.name());
```
ericbuckley, line 86: any reason to not using a parameter for the claimType?

apps/bfd-pipeline/bfd-pipeline-rda-grpc-server/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/DirectRdaLoadApp.java
```diff
@@ -54,9 +54,12 @@ public static void main(String[] args) throws Exception {
     reporter.start(5, TimeUnit.SECONDS);
 
     final RdaLoadOptions jobConfig = readRdaLoadOptionsFromProperties(options);
-    final DatabaseOptions databaseConfig = readDatabaseOptions(options);
+    final DatabaseOptions databaseConfig =
+        readDatabaseOptions(options, jobConfig.getJobConfig().getWriteThreads());
     HikariDataSource pooledDataSource =
         PipelineApplicationState.createPooledDataSource(databaseConfig, metrics);
+    System.out.printf("thread count is %d%n", jobConfig.getJobConfig().getWriteThreads());
```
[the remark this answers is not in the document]
brianburtonadhoc: This is a command line utility so direct output to standard out is helpful.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaLoadOptions.java
```diff
@@ -21,6 +23,7 @@
  */
 public class RdaLoadOptions implements Serializable {
   private static final long serialVersionUID = 7635897362336183L;
+  private static final int HASH_CACHE_SIZE = 1000;
```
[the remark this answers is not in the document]
brianburtonadhoc: Yes I support that 100%

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V37__api_progress_table.sql
brianburtonadhoc: Good catch.  I will update the migration script to drop those indexes.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/SequenceNumberTracker.java
brianburtonadhoc: Your analysis is correct.  The purpose of this is to track safe values for the sequence number column in the progress table.  It's not to guarantee that the number is always a value received from the API but it will be a value between the value in the database when the pipeline started and the lowest unwritten value that we have received while the pipeline was running.

Any logic that depends on the value in this column actually existing in a record within the claims tables would be a pretty serious design bug.  There is no integrity constraint relating to this table to records in the claim tables.  Also there is aonly a single column holding a sequence number so no good way to create any such constraint.  No cxonstraint is needed though because this table is really just a scratch pad to allow the pipeline to track its best known resumption point as it runs and between runs.

Even if someone tried to work out a way to have a constraint like that it would be doomed to cause problems simply because the transaction that writes the sequence numbers to the progress table is decoupled from the transactions that update the claims themselves.  It's not unlikely that the sequence number value to be written at this moment might already have been made obsolete by another transaction updating the claim that originally had that sequence number.  There is no harm in that since in the worst case of the pipeline crashing at that moment and restarting it would just resume at the older point and reprocess something it had already done but not recorded in the progress table yet.

That's by design since it safely allows all of the claim update transactions to work within just the claims tables.  They cannot conflict with each other while trying to also update the same record of the progress table in the same transaction.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
brianburtonadhoc: Yes because concrete classes that derive from this are dedicated to a single claim type and the user of those classes does not need to know or care what the actual claim type is.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V37__api_progress_table.sql
brianburton: FYI - the migration file had to be renamed following a merge because it had the same version number as a migration from the merge.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaLoadOptions.java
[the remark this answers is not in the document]
ericbuckley: 👍 I created [PACA-350](https://jira.cms.gov/browse/PACA-350)

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
ericbuckley: I meant at the query level.
```
      String query =
          String.format(
              "select p.%s from RdaApiProgress p where p.claimType=:claimType",
              RdaApiProgress.Fields.lastSequenceNumber);
      final List<Long> sequenceNumber =
          entityManager.createQuery(query, Long.class)
             .setParameter("claimType", claimType.name())
             .getResultList();
```
maybe I'm over thinking it

ericbuckley (approved):

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
brianburton: Ah I see.  Good point.  I think given this string is from a constant it's not a big deal to just add it directly.

ericbuckley (dismissed):

KyRobbins (dismissed):

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/SequenceNumberTracker.java
```diff
@@ -0,0 +1,78 @@
...
+    activeSequenceNumbers.remove(sequenceNumber);
+  }
+
+  /**
+   * Gets the current sequence number for which we know there are no unwritten records with a lower
+   * sequence number.
+   *
+   * @return the best sequence number to send to the RDA API when fetching claims
+   */
+  public synchronized long getHighestWrittenSequenceNumber() {
```
christrotteradhoc, line 68: In the context of the discussion below I would suggest renaming this method to 'getSafeResumeSequenceNumber' or something along those lines. The method name as it is currently here implies more than it should and future maintainers may make assumptions based on that.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/WriterThreadPool.java
```diff
@@ -0,0 +1,237 @@
...
+  private final List<ClaimWriterThread<TMessage, TClaim>> writers;
+
+  public WriterThreadPool(
+      int maxThreads, int batchSize, Supplier<RdaSink<TMessage, TClaim>> sinkFactory) {
+    Preconditions.checkArgument(maxThreads > 0);
+    Preconditions.checkArgument(batchSize > 0);
+    sequenceNumbers = new SequenceNumberTracker(0);
+    sink = sinkFactory.get();
+    threadPool =
+        Executors.newCachedThreadPool(
```
christrotteradhoc, line 52: Can you elaborate some on your choice of a cached thread pool here as opposed to a fixed thread pool? Have we been able to load test this at production levels and ensure that the size of this pool was kept to reasonable levels? It might be useful to include somewhere in here as a comment as well the factors that led to choosing this.

christrotteradhoc (commented):
This is really impressive. Nice work! I have a small comment inline and I am ready to approve once we are ready to do the multi-staged deployment next week.

christrotteradhoc (changes requested):
Blocking so that we don't accidentally merge until we have split it up and followed the migration deployment process.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/WriterThreadPool.java
brianburton: Cached thread pool creates only as many threads as needed to execute submitted jobs and shuts down any unused/cached threads if they are idle very long.  I prefer them here because it's easy to get the count wrong when creating a fixed thread pool and have a stuck job that never launches or an unused thread that's sitting idle and wasting resources.  The count issue is particularly likely when a class is updated later to add a new type worker thread.

In this case the pool is created and all jobs are submitted at the beginning so there is no way for the pool to run away on us.  It will always have exactly one thread per job.  Which in this case is one thread per ClaimWriter or SequenceNumberWriter.

ericbuckley:
> Blocking so that we don't accidentally merge until we have split it up and followed the migration deployment process.

@christrotteradhoc we spoke about this today and we think the migration + code changes are safe to be deployed at the same time.
1. This migration is backwards compatible, as adding a new table and removing indexes will NOT prevent old versions of the code base from running (removing the indexes would cause an older version of the API to run slower in production, but being that we don't have any data and our API isn't live, this doesn't really matter)
2. We tested the hibernate validation locally and it does not assert an error when unknown tables exist in the schema

What do you think about merging the migration with the code changes?

ericbuckley (approved):

christrotteradhoc (approved):

PR #870, opened 2021-12-14
BFD-1424: Fix mtus code

JeremyIR:
**JIRA Ticket:**
[BFD-1424](https://jira.cms.gov/browse/BFD-1424)

Use the mtus code instead of mtus count in the V2 carrier FHIR extension.
---

### What Does This PR Do?

Modification of the carrierclaim transformer in V2 to use the mtus code instead of mtus count as a fhir extension value
- modified carrier unit tests to include the line item extension field with mtus code 
- updated endpoint response json files to include new carrier extension changes

Do the updated extension changes in V2 carrier claim transformer and unit tests look correct and avoid any issues for end users if shipped?

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


cbrunefearless (approved):
LGTM, Make sure Brady looks at this as well.

christrotteradhoc (changes requested):
blocking for comms

bfausett77:
The code change, unit test changes, and updates to the json examples all look good to me!  Nice job Jeremy!

bfausett77 (dismissed):
All changes look good!  Nice work Jeremy!

bfausett77 (approved):

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
```diff
@@ -4843,10 +4843,11 @@
             "value" : 1
           }
         }, {
-          "url" : "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cnt",
+          "url" : "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cd",
```
mackec1, line 4846: Since this was a bug, how are we communicating to peering partners? They may be using a lookup trying to find: carr_line_mtus_cnt.

bfausett77: we have comms for this

mackec1 (approved):
As noted as a comment in one of the files, when we change a URI definition, even though the one being 'fixed' was wrong, we may have downstream consumers using that URI to do a lookup, in which case this becomes a breaking change.

bfausett77:
> 

I understand what you are saying here.  In v1, the url was correct.  In v2, this url is being used for TWO separate data fields...so in V2, the downstream consumer isn't able to determine which is which at this point and therefore, this fix is ok to make without worrying about backwards compatibility IMHO.

mackec1 (approved):

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
christrotteradhoc: this is not backwards compatible and a client might have developed functioning code that will break with this change. the recommended approach is to leave this in place and add another valueCoding that has the correct url

bfausett77: Incorrect.  Since there is already another field that has the same system/discriminator, this makes it impossible for them to identify the correct value for the mtus_cnt system/discriminator.  We need this change as is!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2.java
```diff
@@ -266,13 +267,24 @@ private static ExplanationOfBenefit transformClaim(
...
+                                  CcwCodebookVariable.CARR_LINE_MTUS_CNT))));
+
       // CARR_LINE_MTUS_CD => ExplanationOfBenefit.item.extension
       line.getMtusCode()
           .ifPresent(
               code ->
                   item.addExtension(
                       TransformerUtilsV2.createExtensionCoding(
-                          eob, CcwCodebookVariable.CARR_LINE_MTUS_CNT, code)));
+                          eob, CcwCodebookVariable.CARR_LINE_MTUS_CD, code)));
```
christrotteradhoc, line 287: This should be unchanged for backwards compatibility

christrotteradhoc (changes requested):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2.java
JeremyIR: > This should be unchanged for backwards compatibility

What I have generates the expected proposed extension according to the Jira ticket notes. You can compare my output in eobByPatientIdPaged.json line 4847 to those notes.

christrotteradhoc (approved):

apps/bfd-server/dev/api-changelog.md
```diff
@@ -1,5 +1,27 @@
...
+{
+      "url" : "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cd",
+      "valueCoding" : {
+        "system" : "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cd",
+        "code" : "3",
+        "display":"Services"
+      }
+}
+```
+
```
christrotteradhoc, line 24: I feel like we should probably mention and show the details of the element that we are leaving behind for backwards compatibility. thoughts @bfausett77 ?

bfausett77: That makes sense.  This is more clear and 'feels' more appropriate than adding an extension to 'deprecate' the element that we are leaving behind.  Great suggestion Trotter!

christrotteradhoc (changes requested):
The code looks good but just blocking it again on the requested change for an additional update to the changelog

christrotteradhoc (dismissed):

mackec1 (approved):

apps/bfd-server/dev/api-changelog.md
```diff
@@ -2,9 +2,31 @@
 
 ## BFD-1424 Fix mtus code
 
-Update Carrier claim mapping for mtus code instead of mtus count:
-The new coding:
+Update Carrier claim extension mapping for mtus code instead of mtus count:
+The old coding:
```
bfausett77, line 6: Maybe we can indicate that we are adding a new extension with the correct url/system for MTUS Count and keeping the old extension with the MTUS Count value that had the incorrect url/system of MTUS Code.  Something to that effect here?  Does that make sense?

bfausett77: Oops except I made one mistake and what you added looks and sounds good minus the mistake I made.  Let's change to: "Added a new extension with the correct url/system for MTUS Code and kept the old extension with the MTUS Code value that had the incorrect url/system of MTUS Count"

bfausett77 (approved):
Nice work!

bfausett77 (dismissed):
Nice work!

apps/bfd-server/dev/api-changelog.md
```diff
@@ -1,5 +1,51 @@
...
+      "url" : "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cd",
+      "valueCoding" : {
+        "system" : "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cd",
+        "code" : "3",
+        "display":"Services"
+      }
+}
+```
+
+## BFD-1338 Add 2021 CPT Codes for SAMHSA Filtering
```
christrotteradhoc, line 47: looks like a stray line here probably from a merge?

christrotteradhoc (dismissed):

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2Test.java
```diff
@@ -647,6 +647,114 @@ public void shouldHaveLineItemQuantity() {
...
+        TransformerTestUtilsV2.findExtensionByUrl(
+            "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cnt",
+            eob.getItemFirstRep().getExtension());
+
+    Extension compare1 =
+        new Extension(
+            "https://bluebutton.cms.gov/resources/variables/carr_line_mtus_cnt",
+            new Quantity().setValue(1));
+
+    assertTrue(compare1.equalsDeep(ex1));
```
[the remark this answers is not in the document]
JeremyIR: That's fair. Something to add to the backlog perhaps.

lsmitchell (dismissed):

mackec1 (approved):

mackec1 (dismissed):

lsmitchell (approved):

christrotteradhoc (approved):

PR #888, opened 2022-01-05
BFD-1459: RFC for Handling of Custom URIs in FHIR Code Systems, etc.

karlmdavis:
**JIRA Ticket:**
[BFD-1459](https://jira.cms.gov/browse/BFD-1459)

As a BFD engineer, I need to ensure we're aligned on how to handle custom code system, identifier, etc. URIs in BFD, so that we continue to evolve our API in a consistent fashion.

---

### What Does This PR Do?

(see title)

* Does the RFC clearly and correctly explain the issues involved?
* Do you agree with the RFC's recommendations.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


rfcs/0010-custom-system-uris.md
```diff
@@ -0,0 +1,325 @@
...
+      <all sorts of stuff, but mostly EOB.extension discriminators>
+  identifier/
+    <several identifier.system discriminators>
+```
+
+_Most_ of the leaf URI "path" components refer to variables found in the
+  [CCW Data Dictionaries](https://www2.ccwdata.org/web/guest/data-dictionaries).
+As such, the URI structure should be extended for incorporation of future, non-CCW fields.
+For PACA, something like the following would be recommended,
+  which would ensure proper namespacing of similar-yet-actually-distinct fields:
```
ericbuckley, line 214: @karlmdavis when we have a strong inclination that like data elements in CCW are the same as those in a shared system (without production data some of these are hard to confirm, but when we compare data dictionary values between CCW and the shared systems, the descriptions are identical).  Should we reuse the existing CCW based system or create a new one for the shared system?

For example, our FISS `LOB-CD` enum description and possible values matches that of https://bluebutton.cms.gov/resources/variables/clm_fac_type_cd/.  Which of the following are you recommending?
- https://bluebutton.cms.gov/resources/variables/clm_fac_type_cd/
- https://bluebutton.cms.gov/resources/variables/fiss/lob-cd/

rfcs/0010-custom-system-uris.md
```diff
@@ -0,0 +1,325 @@
...
+Whereas a URL is also a "magic string" but one that web browsers and other software applications can use to retrieve something.
+Often, URLs start with "http://", indicating that a web browser can likely open it.
+In the vein of "a square is a rectangle but not all rectangles are squares",
+  a URL is a URI but not all URIs are URLs.
+
+BFD's data payloads include a number of custom URIs for FHIR
+  [Coding.system](https://www.hl7.org/fhir/datatypes-definitions.html#Coding.system)s,
+  [Identifier.system](https://www.hl7.org/fhir/datatypes-definitions.html#Identifier.system)s,
+  and other such data types.
+These URIs are serve two functions:
```
bfausett77, line 40: Remove 'are' here.

rfcs/0010-custom-system-uris.md
```diff
@@ -0,0 +1,325 @@
...
+  <https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/patientRead.json>.
+If an application wanted to select the `Patient.identifier` entry that had a beneficiary's MBI,
+  they would search for the entry with a `system` value of `http://hl7.org/fhir/sid/us-mbi`.
+HL7, a standards body, has said that `system` should be used for all MBIs,
+  in all FHIR-compliant APIs, and so BFD does not need a custom URI there.
+If, however, an application wants to select the `Patient.identifier` entry that contains a beneficiary's CCW beneficiary ID,
+  they would search for the entry with a `system` value of `https://bluebutton.cms.gov/resources/variables/bene_id`.
+The "CCW beneficiary ID" field is only used within/by CMS,
+  so no standards body has assigned it a standard URI,
+  and BFD must provide a custom URI for it.
```
bfausett77, line 56: This is a great way to explain it.  I think there is also a third function that these uri's/url's provide too, which is to allow for validation of the coding.code contained within the FHIR resource against the coding.system url.  There is typically a .json addition to the URL's webpage that a FHIR validator can use to perform these validations for coding systems/codes.

rfcs/0010-custom-system-uris.md
```diff
@@ -0,0 +1,325 @@
...
+From the perspective of their business decision makers,
+  the only thing that really matters is that these URIs are stable:
+  that we do not change our API in ways that are backwards-incompatible with their existing applications.
+That's about the sum total of our users' concerns:
+
+* Ensure that our URIs are documented,
+    such that they can be copy-pasted into their application code.
+* Ensure that our URIs are descriptive,
+    such that engineers building applications against our APIs can make reasonable guesses as to their meaning.
+* Ensure that our APIs continue to work with their applications.
```
bfausett77, line 108: Again, there is also the issue if their validators are unable to validate the code within the specified 'system' url.

mackec1: Per Brady's statement, which implies validators unable to validate, the reason for this is that we really don't have publicly (http) accessible services that resolve a given URI.

rfcs/0010-custom-system-uris.md
```diff
@@ -0,0 +1,325 @@
...
+        * On Mac OS X, you can use the built-in Quicktime Player or the built-in Mac OS X Mojave (and up) feature.
+    * Diagrams are often a good idea.
+        * Keep it simple! Use something like <http://asciiflow.com/>.
+* As part of implementing this proposal, will any documentation updates be needed, e.g. changelogs, Confluence pages, etc.?
+  If so, draft them now! Include the draft as a subsection or addendum.
+-->
+
+This RFC makes the recommendation that BFD should stick with a base of
+  `https://bluebutton.cms.gov/resources/` for all custom URIs,
+  _at least_ until a v3 of the API.
```
mackec1, line 154: Should we not 'bake in' versioning? i.e., https://bluebutton.cms.gov/resources/v1/. Allows us to be both backwards compatible as well as providing an avenue of future changes.

rfcs/0010-custom-system-uris.md
karlmdavis: Isn't offering a terminology service (at some point in the future) an alternative? Or would that need to be hosted at the same URLs?

rfcs/0010-custom-system-uris.md
karlmdavis: In this particular case, I think I'd lean more towards pushing the versions down, e.g. `https://bluebutton.cms.gov/resources/variables/bene_id/v2` or something vaguely like that. Most codesystems won't change from release to release, so it'd be nice to avoid forcing a bunch of unnecessary app changes with new API releases.

Thoughts?

rfcs/0010-custom-system-uris.md
bfausett77: Typically it is at the same URL, as that's how the validator knows 'where' to get the value set to compare the code against, if that makes sense.  There may be a way to 'redirect' to a separate url that has that, but I'm not sure how at this point.  Providing this functionality 'could' be a future endeavor though, as I think they only show up as warnings or information within the validators response.

rfcs/0010-custom-system-uris.md
ericbuckley: I understand the reason for option number 2, it feels safer having no production partially-adjudicated data to confirm these assumptions with.  However, as to future consolidation with additional codings.  Will that only work with CodeableConcepts?  What about extensions, would we just duplicate the value twice in the extensions list so it would show up as two different systems?

karlmdavis: Yeah, I think that'd be how you'd have to do it -- short of a breaking change, leastways.

mackec1: which is the strategy we've been implementing when we hit something like that; add in the Codable where it is expected, but also add it as an extension. Sure it's a duplication of data, but it allows backwards compatibility.

rfcs/0010-custom-system-uris.md
mackec1: I like that idea better; also, aligns more with the strategy where future functionality may be added to say R4 services, but not R2 (except in the case of an actual bug in R2). So level-setting at R4 but allowing say a /v1 to map to our v1 functionality as needed.

mackec1 (commented):
Looks good; the only question I have is this leading to having a concrete, publicly-accessible, implementation of these code system URLs?

karlmdavis:
> Looks good; the only question I have is this leading to having a concrete, publicly-accessible, implementation of these code system URLs?

In a way, yes: keeping everything under the same URL domain & base path makes hosting content at those URLs more feasible.

rfcs/0010-custom-system-uris.md
karlmdavis: I don't think we're doing anything here that _precludes_ supporting that validation in the future, so I'm going to mark this as resolved for now. If I'm misunderstanding things, though, **please** comment and unresolve this.

rfcs/0010-custom-system-uris.md
karlmdavis: Resolving this also, per previous thread.

ericbuckley (approved):

mackec1 (approved):

rfcs/0010-custom-system-uris.md
SJWalter11: The only concern on this front would be contention with the API versioning (which is at a higher level) and the ability to know if on the most current component.   Would there be some sort of map along with this lower chain versioning.   It is somewhat non-intuitive to be using a /v4 api with /v1 set of resources.   Though likely way off the map here.   I liked the higher level /version component due to this, though can see additional work to keep updated, and the developer inference that updates are constantly needed.   Not against the lower level versioning, though had this concern.

christrotteradhoc (approved):

lsmitchell (approved):

rfcs/0010-custom-system-uris.md
williamharty: I'll add one comment here.  I would recommend we stick with the status quo for now. When we release a V3 of the API,  I think there is a bigger discussion about properly defining valuesets, codesystems and extensions.  Currently these URLs (e.g. [https://bluebutton.cms.gov/resources/variables/](https://bluebutton.cms.gov/resources/variables/))  are really just webpages.  Ideally they would point at actual  ValueSet and CodeSystem definitions - likewise extensions should point to the corresponding StructureDefinition.   I realize this would be a breaking change, thus a V3 type item., but it would allow the Validator to catch valueset errors.   Currently these are basically ignored.

PR #895, opened 2022-01-11
BFD-1436: Upgrade to Junit5

lsmitchell:
**JIRA Ticket:**
[BFD-1436](https://jira.cms.gov/browse/BFD-1436)

As a BFD engineer, I want JUnit upgraded, so the BFD's software dependencies are up to date.

---

### What Does This PR Do?

- Changes the junit dependency to the new Junit5 version
- Updates every test to use new Junit5 syntax/annotations
- Standardizes all assertions to use static imports
- Upgrade the parameterized tests to match the simpler junit5 style
- Small updates to some docs that were outdated

* Verify pom adjustments follow best practices

      * [X] Yes
      * [ ] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [X] Yes
      * [ ] No

* N/A

### Submitter Checklist


lsmitchell:
For new dependencies:
**org.junit.jutpiter**: pulls in the dependencies needed for junit5, see:
https://junit.org/junit5/docs/current/user-guide/#dependency-metadata-junit-jupiter

**org.junit.vintage**: adds the packages needed to support existing junit4 tests, until/if we want to "upgrade" them to junit5 style. See:
https://junit.org/junit5/docs/current/user-guide/#dependency-metadata-junit-vintage

christrotteradhoc:
How big is the effort to convert the existing?

lsmitchell:
Based solely on the migration tips junit provides, it seems most things have annotations renamed and imports need to be adjusted, so it shouldnt be difficult but there is a significant bulk of tests to go through and do this to, so it might take 1-3 days to convert everything and test it works assuming there are no logic/restructuring issues encountered.

apps/bfd-server/pom.xml
```diff
@@ -50,9 +50,16 @@
...
+				<groupId>org.junit.jupiter</groupId>
+				<artifactId>junit-jupiter-engine</artifactId>
+				<version>${junit.version}</version>
+				<scope>test</scope>
+			</dependency>
+			<dependency>
+				<!-- Allows junit compatibility with old junit4 tests -->
+				<groupId>org.junit.vintage</groupId>
+				<artifactId>junit-vintage-engine</artifactId>
+				<version>${junit.vintage.version}</version>
```
christrotteradhoc, line 62: we shouldn't need any junit stuff here since we already have it in dependencyManagement in the root pom

apps/bfd-model/pom.xml
```diff
@@ -37,11 +37,20 @@
...
 				<scope>test</scope>
 			</dependency>
+
+			<dependency>
+				<!-- Allows junit compatibility with old junit4 tests -->
+				<groupId>org.junit.vintage</groupId>
+				<artifactId>junit-vintage-engine</artifactId>
+				<version>${junit.vintage.version}</version>
+				<scope>test</scope>
+			</dependency>
```
christrotteradhoc, line 52: we shouldn't need any junit stuff here since we already have it in dependencyManagement in the root pom

apps/pom.xml
```diff
@@ -93,6 +93,8 @@
 
 		<metrics.version>4.1.2</metrics.version>
 		<hibernate.version>5.6.1.Final</hibernate.version>
+		<junit.version>5.8.2</junit.version>
+		<junit.vintage.version>5.8.2</junit.vintage.version>
```
christrotteradhoc, line 97: assuming these are kept in lock step -- do we need two properties that could inadvertently diverge?

apps/pom.xml
```diff
@@ -297,9 +299,17 @@
...
+				<artifactId>junit-jupiter-engine</artifactId>
+				<version>${junit.version}</version>
+				<scope>test</scope>
+			</dependency>
+
+			<dependency>
+				<!-- Allows junit compatibility with old junit4 tests -->
+				<groupId>org.junit.vintage</groupId>
+				<artifactId>junit-vintage-engine</artifactId>
+				<version>${junit.vintage.version}</version>
```
christrotteradhoc, line 312: I wonder whether we should just make this a dependency in this pom so that we don't need it in every single other pom. Not something that I would suggest for practically any other dependency but this is ubiquitous. Do we have any projects that don't need junit as a test dependency?

christrotteradhoc:
> Based solely on the migration tips junit provides, it seems mostly annotations were renamed and imports need to be adjusted and things like that, so it shouldnt be difficult. However there are a significant bulk of tests to go through, so it might take 1-3 days to convert everything and test it works assuming there are no logic/restructuring issues encountered. (So many modules!)

I'm curious how close we get by just doing a quick sed script to fix the imports and annotations across all projects. I think it is worth it to check that before we go with this since we may not get back to this for a while or ever.

> 
> Since the Pr didn't mention if we wanted to "upgrade" the old tests, and there was a low-effort backwards compatibility option available, I decided to float this solution first. I did read Junit vintage (the solution for supporting junit4 tests) will be updated and supported going forward, so it should be "safe" from a security perspective.

If we do go with this then we should open a new JIRA to do the conversion

apps/pom.xml
lsmitchell: There are some submodules that didn't have junit listed as a dependency, at least.

apps/bfd-server/pom.xml
lsmitchell: I'll double check, it had been redeclared here before for some reason. (I think because this is both a child pom and a parent pom?)

lsmitchell:
Updated all existing tests to use Junit5 annotations/syntax, including the parameterized tests. The only tests left unmodified are a couple tests that used Rules, which will have a ticket made to clean up later since the junit rules were replaced by a completely different thing (extensions).

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/RifFilesProcessorTest.java
```diff
@@ -38,6 +38,7 @@
 
 /** Unit tests for {@link RifFilesProcessor}. */
 public final class RifFilesProcessorTest {
+  // TODO: Replace with junit5 extension and upgrade class to junit5
```
mackec1, line 41: do we still need this TODO?

mackec1 (approved):
LOTS of good work here; hopefully you were able to handle a bunch with global search-replace!

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/RifFilesProcessorTest.java
lsmitchell: The couple places we use Rules I left running on Junit4 until I researched more about extensions since Rules are totally replaced; however I think the rules that expect no exceptions like this one could maybe be removed since junit automatically fails if an exception is thrown, and it seems redundant to me.

apps/bfd-pipeline/bfd-pipeline-app/pom.xml
```diff
@@ -83,8 +83,15 @@
...
-			<artifactId>junit</artifactId>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<!-- Allows junit compatibility with old junit4 tests -->
+			<groupId>org.junit.vintage</groupId>
+			<artifactId>junit-vintage-engine</artifactId>
+			<version>${junit.vintage.version}</version>
```
christrotteradhoc, line 94: if we do end up merging this with this dependency, please put a comment explaining that we have it specifically (and exclusively) for the `@Rule` functionality

apps/bfd-pipeline/bfd-pipeline-rda-bridge/src/test/java/gov/cms/bfd/pipeline/bridge/RDABridgeIT.java
```diff
@@ -114,13 +116,16 @@ public void close() throws IOException {}
...
       if (message instanceof FissClaimChange) {
-        fissTransformer.transformClaim((FissClaimChange) message);
+        resultingTransform = fissTransformer.transformClaim((FissClaimChange) message);
       } else {
-        mcsTransformer.transformClaim((McsClaimChange) message);
+        resultingTransform = mcsTransformer.transformClaim((McsClaimChange) message);
       }
     }
+    // TODO: Add some better assertions here
+    assertNotNull(resultingTransform);
```
christrotteradhoc, line 128: I know you are just adding in something but shouldn't this go inside the loop at least? FYI @brianburton @KyRobbins

apps/pom.xml
```diff
@@ -298,12 +300,17 @@
...
+				<groupId>org.junit.jupiter</groupId>
+				<artifactId>junit-jupiter-engine</artifactId>
+				<version>${junit.version}</version>
+				<scope>test</scope>
+			</dependency>
+			<dependency>
+				<groupId>org.junit.jupiter</groupId>
+				<artifactId>junit-jupiter-params</artifactId>
+				<version>${junit.version}</version>
 				<scope>test</scope>
```
christrotteradhoc, line 312: can you add a comment on this dependency (i.e. 'parametrized test extension for junit' if that is what it is)

christrotteradhoc:
Great work!

apps/bfd-pipeline/bfd-pipeline-app/pom.xml
lsmitchell: Removed it, since I converted the remaining tests that needed it!

apps/bfd-pipeline/bfd-pipeline-rda-bridge/src/test/java/gov/cms/bfd/pipeline/bridge/RDABridgeIT.java
lsmitchell: Not sure, there were no assertions before so I just added this to ensure there was at least one. If someone has a suggestion, I can enhance.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/RifFilesProcessorTest.java
lsmitchell: Ended up removing it after a team discussion in engineering sync.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/RdaServerJobIT.java
```diff
@@ -112,30 +115,30 @@ public void testS3() throws Exception {
             new FissClaimStreamCaller(new FissClaimTransformer(clock, hasher));
         final GrpcResponseStream<RdaChange<PreAdjFissClaim>> fissStream =
             fissCaller.callService(fissChannel, CallOptions.DEFAULT, 1098);
-        assertEquals(true, fissStream.hasNext());
+        assertTrue(fissStream.hasNext());
```
brianburton, line 118: FYI assertEquals here was a deliberate choice since it provides a more meaningful message when the test fails than just "assertion failed".  But it's not worth rolling all of them back not that you've changed them

apps/bfd-pipeline/bfd-pipeline-rda-bridge/src/test/java/gov/cms/bfd/pipeline/bridge/RDABridgeIT.java
brianburton: Given the explicit expected argument to `@Test` I think this was verifying that no exceptions were thrown.

brianburton (approved):
This looks good.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/RdaServerJobIT.java
brianburton: I just tested the error message with junit 5 and it's the more useful message now (expected true but was false) so these changes are fine.  They'll eliminate some complaints from IDEA as well.

apps/bfd-pipeline/bfd-pipeline-rda-bridge/src/test/java/gov/cms/bfd/pipeline/bridge/RDABridgeIT.java
lsmitchell: I think that was the idea, but Junit will already fail the test when any exception is thrown, so that isn't really a needed explicit assertion. I assume there is a state at the end of the functionality being tested here that we care about testing. Given its a transform, I made the assumption we at _least_ care that the transform returned something, and in calling it we can also test no exceptions got thrown to the end of the process.

mackec1 (approved):

christrotteradhoc (approved):

StewGoin (commented):
👍 from me, this seems very reasonable from what I can see and good to get ahead of any deprecation concerns.

apps/bfd-pipeline/bfd-pipeline-rda-bridge/src/test/java/gov/cms/bfd/pipeline/bridge/RDABridgeIT.java
KyRobbins: The intent of the test was to show that no exception is thrown.  It's true that the annotation is not required, but it is a cue to the reader as to why the tests exists, which is to check that no exception was thrown.  Checking that `resultingTransform` is not null is not the true intention of the test.

PR #919, opened 2022-01-27
PACA-428: Fix null pointer on null status for MCS

KyRobbins:
**JIRA Ticket:**
[PACA-428](https://jira.cms.gov/browse/PACA-428)

---

### What Does This PR Do?

MCS claims can have a null value for their status, thus the application needs to set a default status of "Active" with an outcome of "Queued" when the status provided by RDA is null.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
```diff
@@ -96,11 +97,22 @@ private static ClaimResponse transformClaim(PreAdjMcsClaim claimGroup) {
     claim.setContained(List.of(getContainedPatient(claimGroup)));
     claim.setExtension(getExtension(claimGroup));
     claim.setIdentifier(getIdentifier(claimGroup));
-    claim.setStatus(
```
ericbuckley, line 99: Would this be a good use case for Optional?
```
String statusCode = Optional.ofNullable(claimGroup.getIdrStatusCode()).orElse("");
```

ericbuckley:
@KyRobbins can we add a test case for this

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
KyRobbins: It would be syntactically terser, I thought going with an `if()` might make it easier to understand there is an edge case we are resolving here.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
```diff
@@ -96,11 +97,22 @@ private static ClaimResponse transformClaim(PreAdjMcsClaim claimGroup) {
...
+      claim.setStatus(
+          CANCELLED_STATUS_CODES.contains(claimGroup.getIdrStatusCode().toLowerCase())
+              ? ClaimResponse.ClaimResponseStatus.CANCELLED
+              : ClaimResponse.ClaimResponseStatus.ACTIVE);
+    }
+
+    claim.setOutcome(
+        ObjectUtils.defaultIfNull(
+            OUTCOME_MAP.get(ObjectUtils.defaultIfNull(claimGroup.getIdrStatusCode(), "")),
+            ClaimResponse.RemittanceOutcome.QUEUED));
```
brianburton, line 115: This might be a little more direct:

```
    // Some status codes for MCS can be null.
    // These will be interpreted as status=active, outcome=queued.
    if (claimGroup.getIdrStatusCode() == null) {
      claim.setStatus(ClaimResponse.ClaimResponseStatus.ACTIVE);
      claim.setOutcome(ClaimResponse.RemittanceOutcome.QUEUED);
    } else {
      String statusCode = Strings.nullToEmpty(claimGroup.getIdrStatusCode()).toLowerCase();
      claim.setStatus(
          CANCELLED_STATUS_CODES.contains(statusCode)
              ? ClaimResponse.ClaimResponseStatus.CANCELLED
              : ClaimResponse.ClaimResponseStatus.ACTIVE);
      claim.setOutcome(
          OUTCOME_MAP.getOrDefault(statusCode, ClaimResponse.RemittanceOutcome.QUEUED));
    }
```

Making CANCELLED_STATUS_CODES a Map would let you eliminate the ternary operator in `setStatus` too since you could just use `getOrDefault` there.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
brianburton: Rather than Optional for a string value you can use `Strings.nullToEmpty()`

```
String statusCode = Strings.nullToEmpty(claimGroup.getIdrStatusCode());
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
brianburton: Good point about the extra `nullToEmpty()`.  My mind. was on the previous comment about using `Optional`.

I do think using a `STATUS_CODES_MAP` instead of `CANCELLED_STATUS_CODES` set would make the code clearer though.  It would let the code adapt other statuses later too.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
```diff
@@ -96,11 +99,22 @@ private static ClaimResponse transformClaim(PreAdjMcsClaim claimGroup) {
...
+      claim.setStatus(ClaimResponse.ClaimResponseStatus.ACTIVE);
+      claim.setOutcome(ClaimResponse.RemittanceOutcome.QUEUED);
+    } else {
+      claim.setStatus(
+          STATUS_MAP.getOrDefault(
+              claimGroup.getIdrStatusCode().toLowerCase(),
+              ClaimResponse.ClaimResponseStatus.ACTIVE));
+      claim.setOutcome(
+          OUTCOME_MAP.getOrDefault(
+              claimGroup.getIdrStatusCode(), ClaimResponse.RemittanceOutcome.QUEUED));
```
brianburton, line 115: Doesn't `OUTCOME_MAP` needs lowercase string key as well?

brianburton (dismissed):
Look good!

ericbuckley (approved):
looks good, thanks for the quick fix

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
```diff
@@ -33,7 +32,11 @@
   private static final String METRIC_NAME =
       MetricRegistry.name(McsClaimResponseTransformerV2.class.getSimpleName(), "transform");
 
-  private static final Set<String> CANCELLED_STATUS_CODES = Set.of("r", "z", "9");
+  private static final Map<String, ClaimResponse.ClaimResponseStatus> STATUS_MAP =
```
christrotteradhoc, line 35: this name seems somewhat misleading since it can only map the statuses for cancelled things. further logic is required by the consumer of this to fully map the status.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
```diff
@@ -96,11 +99,22 @@ private static ClaimResponse transformClaim(PreAdjMcsClaim claimGroup) {
...
+    } else {
+      claim.setStatus(
+          STATUS_MAP.getOrDefault(
+              claimGroup.getIdrStatusCode().toLowerCase(),
+              ClaimResponse.ClaimResponseStatus.ACTIVE));
+      claim.setOutcome(
+          OUTCOME_MAP.getOrDefault(
+              claimGroup.getIdrStatusCode().toLowerCase(), ClaimResponse.RemittanceOutcome.QUEUED));
+    }
+
```
christrotteradhoc, line 117: Another thought for implementing this -- replace the two static final maps with static final Function<String, ClaimResponseStatus> and static final Function<String, RemittanceOutcome> and use those here. It keeps the calculation of status and outcome encapsulated in one place and gets rid of low level logic here in the transform method.

christrotteradhoc (changes requested):
I don't see any functional issues but I've got an alternative implementation for your consideration.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
ericbuckley: I see your point.  We are only using [2 of the 4 status codes](http://hl7.org/fhir/valueset-fm-status.html), "active" and "canceled", so from this logic if it's not in that map than it needs to be "active".  However, someone new to the project may wonder whether "draft" or "entered-in-error" should be used.

@KyRobbins could we maybe change the name of this Map and/or add some documentation to let the readers know we are only using the "active" and "cancelled" states to represent our data.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/preadj/McsClaimResponseTransformerV2.java
KyRobbins: It would still require a map (or a switch, etc.) to do the logic.  I can move them to a separate method though if that would help.

christrotteradhoc (approved):

brianburton (approved):

ericbuckley (approved):

PR #922, opened 2022-01-28
PACA-350: MBI caching 1/4 - create table

brianburton:
**JIRA Ticket:**
[PACA-350](https://jira.cms.gov/browse/PACA-350)

Normalize MBI values in FISS/MCS claims tables.

---

### What Does This PR Do?

**This is PR 1 of 4 for the MBI caching changes.  This PR creates the new table and columns in preparation for the pipeline changes in the next PR.  The following description will be included in all 4 PRs.**

---

Removes the columns in `FissClaims` and `McsClaims` tables that contained the plain and hashed MBI values and adds a new table `MbiCache` where each record contains a MBI and its associated hash value.  Searches for claim by MBI will now be done using a join.

````
┌─────────────┐         ┌───────────┐          ┌─────────────┐
│  FissClaims │         │ MbiCache  │          │ McsClaims   │
│             │         │           │          │             │
│  mbiId      ├────────►│ mbiId     │◄─────────┤ mbiId       │
│             │         │ mbi       │          │             │
│             │         │ hash      │          │             │
│             │         │ oldHash   │          │             │
└─────────────┘         └───────────┘          └─────────────┘
````

This PR does the following to make this work:

- Updates the table entities to use a new entity and a `@ManyToOne` relationship.
- Updates the `IdHasher` implementation used by the RDA sinks to look up unknown hashes in the database before computing one.
- Updates the query used in the server to find claims by MBI.
- Adds support for current and old hash value for each MBI.
- Adds launch option `-o` to `server-start.sh` to control whether old MBI hashes are used in queries.

The ability to find claims using either a current or old hash value will allow the API to provide temporary backwards compatible support for old hashes if the hash function is changed.  The `oldHash` column could be populated with the old `hash` and the `hash` column polulated with the new one in every record in the `MbiCache` table.  Once backwards compatibility is no longer needed the `oldHash` column can be set to `null` again in all records.

Most of the code changes are minor ones to accomodate the change in the entity objects.  The most important classes to review are:

- Mbi is the database entity used for the MbiCache table.
- MbiCache is the class that implements MBI lookup/hashing.
- ClaimDao implements the modified logic to query for claims by MBI or MBI hash.
- Fiss/McsClaimTransformer now use the MbiCache instead of directly computing hash values.

There is a migration script `V40__preadj_mbi_caching.sql` to accomplish the necessary database changes to support the cache.

There are a lot of JSON changes in `GrpcRdaSourceIT.java`.  The test was failing intermittently because field order in the serialized JSON was changing inconsistently in different test runs.  The test now forces the fields to be sorted alphabetically by name in the JSON and that created a bunch of diffs.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

N/A

### Submitter Checklist

### CODE REVIEW

File: apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V53__preadj_mbi_caching-1.sql [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+    "oldHash" VARCHAR(64),
+    "lastUpdated" timestamp with time zone
+);
+
+/*
+ * We need these indexes to support queries by mbi, hash, and oldHash.  Note that oldHash
+ * is usually null (only set to a value during hash rotation) so the index cannot be unique.
+ */
+CREATE UNIQUE INDEX "MbiCache_mbi_idx" on "pre_adj"."MbiCache"("mbi");
+CREATE UNIQUE INDEX "MbiCache_hash_idx" on "pre_adj"."MbiCache"("hash");
```
ericbuckley, line 17: I believe postgres does allow you to have multiple nulls in a unique column, however I'm wondering if we'd ever run into collisions when copying a value from old to new or vice versa.  Would it be best to just make the mbi column unique?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V53__preadj_mbi_caching-1.sql
```diff
@@ -0,0 +1,37 @@
...
+CREATE INDEX "MbiCache_oldHash_idx" on "pre_adj"."MbiCache"("oldHash");
+
+/*
+ * Add foreign key to reference MbiCache in claims tables.
+ */
+ALTER TABLE "pre_adj"."FissClaims"
```
ericbuckley, line 23: We'll need to also disable hibernate validation before we mege this PR

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V53__preadj_mbi_caching-1.sql
brianburton: These are both NOT NULL columns.  A collision in the hash column would be a serious problem that I think we'd want to detect early.  The UNIQUE constraint would identify it for us quickly.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V53__preadj_mbi_caching-1.sql
brianburton: Since we are only adding a column hibernate validation won't care.  It only cares about missing columns or data type changes in the columns that it knows about.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V53__preadj_mbi_caching-1.sql
ericbuckley: The first part of my comment was in reference to line 18, the oldHash column

brianburton: I don't thik UNIQUE is allowed for NULLABLE columns though.

brianburton: Wow.  I'm wrong about UNIQUE on NULLABLE.  Just tried it and it worked.  That is really odd...

ericbuckley: With that said, maybe just remove the comment on line 14 since thats not the case

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V53__preadj_mbi_caching-1.sql
```diff
@@ -0,0 +1,37 @@
+/*
+ * Table to hold known MBI and their hashed equivalents.
+ */
+CREATE TABLE "pre_adj"."MbiCache" (
+    "mbiId"   bigint      NOT NULL PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    "mbi"     VARCHAR(13) NOT NULL,
```
ericbuckley, line 6: this should be VARCHAR(11)

ericbuckley (approved):

ericbuckley (approved):

KyRobbins (approved):

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V53__preadj_mbi_caching-1.sql
```diff
@@ -0,0 +1,36 @@
+/*
```
christrotteradhoc, line 1: minor request on these 3 script names. the numbering won't mean much in the future. I think tying them together with a common prefix is good to show the relationship but can you replace the number with a more descriptive tag? also, I believe that another migration that is in review is going to get in ahead of this one so the version numbers will have to go up by one.

brianburton: Thanks.  I have renamed the file in all branches and merged master into them as well.

christrotteradhoc (commented):
Looks good. I'll approve and merge when we get to this deployment.

christrotteradhoc (approved):

keithdadkins (approved):

PR #938, opened 2022-02-03
PACA-426: Allow empty strings in enum values

brianburton:
**JIRA Ticket:**
[PACA-426](https://jira.cms.gov/browse/PACA-426)

Many RDA enum fields have the option of sending an unknown value rather than a known enum. The enum fields are fixed, and we know the length of those, however the unknown values could be of a different length (however it can't be larger since the RDA's DB schema doesn't support that). We should remove the min length field check, so we can allow values, like the empty string, to be persisted.

---

### What Does This PR Do?

Changes enum string value validation to allow an arbitrary minimum length instead of hard coding it to one.  Changes transformers to use `minLength = 0` where appropriate.

* Are there any enum fields that were missed.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/McsClaimTransformer.java
```diff
@@ -1042,7 +1042,7 @@ private PreAdjMcsLocation transformMessageImpl(
     transformer.copyEnumAsString(
         namePrefix + PreAdjMcsLocation.Fields.idrLocActvCode,
         true,
-        1,
+        0,
```
KyRobbins, line 1045: I'm not totally sure about what this number is, but based on how many were updated, should it be a constant value?

brianburton: Previously the `minLength` defaulted to 1.  Now most enums need to use `0` instead.  But that's a good point.  I'll make the default `0` and just pass in `1` in the few places that need it.

brianburton: That's the minimum allowed string length.

brianburton: I have refactored the code to eliminate the explicit zeros.

KyRobbins (approved):

ericbuckley (approved):

christrotteradhoc (approved):

PR #949, opened 2022-02-10
[BFD-1446] BFD-1446:Added focal field to v2

cbrunefearless:
**JIRA Ticket:**
[BFD-1446](https://jira.cms.gov/browse/BFD-1446)

User Story:

As a DASG API customer, I want the focal insurance field populated, so I can use that field and its data.
AC:

For V2, set eob.insurance.focal to 'true' for all hard coded eob.insurance.coverage elements
Additional Detail.

This is a Boolean field and should be set to either true or false. The definition of the field is this: "Coverage to be used for adjudication". My guess is that there will only be one insurance per claim. If this is the case then the focal should always be set to true. If there is more than one, then we need to determine if that insurance/coverage was used for adjudication or not. However, this is only for PDE claims, since it appears this is the only claim type that sets any values within the eob.insurnace[N]. This is also ONLY A FIX FOR V2.

### What Does This PR Do?

Added focal field to v2 when insurance is added

Make sure the transformers are changed to incorporate the focal field when insurance is added.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X]  No
      * [ ] Yes
      * [X]  No

      * [ ] Yes
      * [X]  No

* N/A

### Submitter Checklist

* [X]  I have named this PR and branch so they are [automatically linked](https://confluence.atlassian.com/adminjiracloud/integrating-with-development-tools-776636216.html) to the (most) relevant Jira issue. Ie: `BFD-123: Adds foo`
* [X]  This PR is reasonably limited in scope, to help ensure that:
* [X]  This PR includes any required documentation changes, including `README` updates and changelog / release notes entries.
* [X]  All new and modified code is appropriately commented, such that the what and why of its design would be reasonably clear to engineers, preferably ones unfamiliar with the project.
* [X]  All tech debt and/or shortcomings introduced by this PR are detailed in `TODO` and/or `FIXME` comments, which include a JIRA ticket ID for any items that require urgent attention.
* [X]  Reviews are requested from both:
* [X]  Any deviations from the other policies in the [DASG Engineering Standards](https://github.com/CMSgov/cms-oeda-dasg/blob/master/policies/engineering_standards.md) are specifically called out in this PR, above.


apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/DMEClaimTransformerV2Test.java
```diff
@@ -298,6 +298,7 @@ public void shouldReferenceCoverageInInsurance() {
 
     InsuranceComponent compare =
         new InsuranceComponent()
+            .setFocal(true)
             .setCoverage(new Reference().setReference("Coverage/part-a-567834"));
```
mackec1, line 302: Chris,  can you check with Brady on this; he ran into this the other day vis-a-vis Part A vs Part B DME claims. I read through some stuff as did he, and not sure if a consensus was that you can (or can't) have DME claims under Part A...I think you can, but it may involve checking some codes to see where this actually falls.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -1818,7 +1818,7 @@ static void mapEobCommonClaimHeaderData(
         .setType(createC4BBClaimCodeableConcept());
 
     // BENE_ID + Coverage Type => ExplanationOfBenefit.insurance.coverage (ref)
-    eob.addInsurance().setCoverage(referenceCoverage(beneficiaryId, coverageType));
+    eob.addInsurance().setFocal(true).setCoverage(referenceCoverage(beneficiaryId, coverageType));
```
christrotteradhoc, line 1821: The PR description and ticket A/C says to set this to true if there is only one insurance per claim and if there is more than one than do something else (doesn't really say what, except to reach out to Brady). We need to clarify the A/C and PR description with the correct behavior in unambiguous terms.

christrotteradhoc (changes requested):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
bfausett77: At this point, there is always just one insurance coverage documented, since they are hard coded by claim type.  If we get to a point where this is no longer hard coded, and the example above for DME we may have to do this, then we will need to determine the logic/specific data points to use to determine if focal should not default to true.  Does that make sense?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/DMEClaimTransformerV2Test.java
bfausett77: I think this is still an ongoing conversation and I don't think there has been a firm decision yet.  I feel like this should be it's own ticket/spike to figure this out.  At that point, there may be a case, depending upon the HCPCS Code, or some other data point, that we may go from part-a to part-b, so there would be some logic here to determine which medicare coverage would be displayed.  However, the focal would still be set to true in both cases.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
christrotteradhoc: Is it correct to say that we would like the A/C for this ticket to be updated to this:

- For V2, set focal to true on all insurance elements

?

bfausett77: I think that would probably make sense.  Maybe say, "For V2, set eob.insurance.focal to 'true' for all hard coded eob.insurance.coverage elements"

christrotteradhoc (approved):

apps/bfd-server/dev/api-changelog.md
```diff
@@ -704,3 +704,8 @@ Future updates may add `Coding.display` values for additional fields.
...
 	* The "FIXME: check if this field is non-nullable and if not remove the 'if' check" issues were addressed by comparing the fields to their definition in the rif-layout-and-fhir-mapping.xlsx file. Most fields were found to be non-nullable and so the "if" check was removed.
  
+## BFD-1446 Populate Focal Insurance Field
+
+ * Following FHIR mapping changes were made:
+
+  * For V2, set eob.insurance.focal to 'true' for all hard coded eob.insurance.coverage element.  This is a Boolean field and should be set to either true or false. The definition of the field is this: "Coverage to be used for adjudication". My guess is that there will only be one insurance per claim. If this is the case then the focal should always be set to true. If there is more than one, then we need to determine if that insurance/coverage was used for adjudication or not. I can work with the engineer on determining this, IF there is more than one insurance. However, this is only for PDE claims, since it appears this is the only claim type that sets any values within the eob.insurnace[N]. This is also ONLY A FIX FOR V2
```
bfausett77, line 711: I would remove the verbiage surrounding our internal conversations regarding this field.  This may be better here:  "The field 'eob.insurance.focal' is a Boolean field and should be set to either true or false. The definition of the field is this: "Coverage to be used for adjudication".  For V2 we are setting all instances of eob.insurance to have a focal of true.  There may be changes to this field in the future if there are multiple instances of eob.insurance."

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
lsmitchell: How are we going to remember this going forward? There should probably be a comment referring to this decision so if we adjust this in the future, we remember not to apply true to more than one item.

Perhaps a test to ensure there is only one item, with this caveat attached, so if we add another the test will break and bring attention to the implementer to fix/resolve the focal issue.

apps/bfd-server/dev/api-changelog.md
lsmitchell: We've also been including what the new section looks like with the change in these changelogs.

lsmitchell (approved):

christrotteradhoc (approved):

apps/bfd-server/dev/api-changelog.md
christrotteradhoc: this needs to be resolved on master

PR #956, opened 2022-02-15
BFD-1473 - RFC: Separate Flyway from BFD Pipeline Application

christrotteradhoc:
**JIRA Ticket:**
[BFD-1483](https://jira.cms.gov/browse/BFD-1483)

RFC: Separate Flyway from BFD Pipeline Application

---

### What Does This PR Do?

RFC for separating Flyway from BFD Pipeline Application.

* If the answer to any of the questions below is **Yes**, then **you must** supply a link to the associated Security Impact Assessment (SIA), security checklist, or other similar document in Confluence here: 

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [X] Yes
      * [ ] No

* N/A

### Submitter Checklist


christrotteradhoc:
I might get a couple of diagrams to add courtesy of burling, but I think otherwise the pieces are here and I'm interested in getting comments.

mackec1 (commented):
Excellent structure and presentation of background / facts / arguments. I guess the one item that I see always impacting deployments, is the fact that we have an uber-build monolith. Is there any chance that we can parition things into functional entities (allowing for some entities being cross-functional). For example, BFD is made up of:
- Ops
- Database
- Application FHIR services
- Pipeline 

The RFC touches upon these in the context of orchestrating a build/deploy but is there an option to literally carve off portions of the current BFD into something more bite-size?

christrotteradhoc:
> The RFC touches upon these in the context of orchestrating a build/deploy but is there an option to literally carve off portions of the current BFD into something more bite-size?

I definitely think we need to spend some time on that question and some ideas have been put on the table by @keithdadkins and @mjburling (mostly around pre-building artifacts and preemptively running parts of the pipeline when master has a new commit). Those would probably land in their own RFC but I'm open to ideas on how to modify this one to take them into account without changing the scope.

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+* A new Flyway migration will be developed that creates all non-user roles and sets privileges correctly on all existing
+database objects.
+
+### Proposed Solution: Unresolved Questions
+[Proposed Solution: Unresolved Questions]: #proposed-solution-unresolved-questions
+
+Do we need to continue to invoke Hibernate validation at all? What benefit do we derive from this check that is not
+already derived from running the unit and integration tests? What sorts of problems would not be caught by the tests
+that would be caught by hibernate validation?
+
```
brianburton, line 340: Invoking Hibernate validation in the migrator app takes very little time and provides a sanity check to ensure the migration did not fail in some odd way that the app did not otherwise detect.

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+The proposed solution is to remove the invocations of Flyway and Hibernate validation from the current BFD applications
+and instead run Flyway and Hibernate validation as a step in the deployment that must complete successfully before
+continuing with the deployment of the BFD Pipeline application and BFD FHIR Server.
+
+By moving the Flyway migrations to a step before all other applications are deployed, it no longer is necessary to
+deploy a schema change separately from its accompanying application change or have forward-compatible changes (since the
+new application will only ever start up after the new schema has been deployed). Backward-compatibility remains a
+requirement because the old applications will still coexist with the new schema for a brief period of time.
+
```
brianburton, line 270: Thinking of the recent MbiCache PR sequence:

1. SQL only PR to create MbiCache table and add foreign key columns to claims tables
2. Combined PR with sql to populate new table from existing database records plus pipeline code update to start populating the new table and columns for newly arriving records
3. Code only PR that updates api to query using the new table and column instead of old columns and changes pipeline to stop updating the old columns.
4. SQL only PR to drop the old columns

I'm not sure how well this new process would work for this sort of migration.  The first sql only migration could be combined with the pipeline code changes so that's a win.  But when could the sql to populate the new table from existing records be executed?  It needs for the old pipeline to be stopped before it can reliably do that.  In this proposal the old pipeline will still be running while the migration is running.  This could even create conflicts between simultaneous updates to records by both the migration script and the pipeline app.

Also the cleanup migration would still have to take place in its own PR.  So we can't reliably run the sql from PR 2 and still need 3 and 4 to be separate PRs.

Would it be possible to:

1. Stop the old pipeline before kicking off the migrator app.
2. Add support for a POST deployment migration that could perform cleanup changes that only make sense once the new code is fulling running.

With these changes the old 4-PR sequence could be just 2 PRs (1+2 & 3+4 from the old sequence).

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+This makes debugging more complex. There will be new places (log files, possibly directories) where debugging
+information will be kept.
+
+### Proposed Solution: Notable Alternatives
+[Proposed Solution: Notable Alternatives]: #proposed-solution-notable-alternatives
+
+Instead of running Flyway and Hibernate validation in a new application, Flyway could be invoked from the command line
+instead. These leaves the question as to where to invoke Hibernate validation. If it continues to be invoked in the
+applications, it is possible to encounter validation errors if auto-scaling of an old application occurs just after a
+new schema has been deployed. One possible answer is to not invoke Hibernate validation at all (see open questions).
```
brianburton, line 359: I like the idea of using a custom migration app to trigger the Flyway migration and do the hibernate validation.  It should be a very simple app and it ensures everything is done with the same configuration that the pipeline and api apps will be using.

brianburton:
I like this proposal a lot but have some questions about how it will work out in practice.

rfcs/0011-separate-flyway-from-pipeline.md
christrotteradhoc: Thanks for giving this a read.

We could fairly easily stop the pipeline application before running the migrator app but it means that the pipeline would not ingest data during the migration which could be 24 hours for some long running synchronous migrations that we envision happening. I was assuming that would be a non-starter for the RDA job.

There are ways to get around that though. We could make it an option of the deployment as to whether the pipeline application would be stopped or not during the deployment but that seems error prone to me. Another possibility would be to run that type of long running job in a post-deployment migration which you have also alluded to as another desirable feature. That feature is currently sitting in the Future Possibilities section and these are good use cases for it so I think we should keep it in mind. Luckily that would build on top of this proposal and not be in conflict with it.

I appreciate putting a real work change to the test but I think we should also consider whether this particular complex change is representative of typical migration changes that we expect to encounter most often. I believe the previous two that were done for PACA where we followed the BFD process to the letter (including the hibernate validation off/on steps for absolute safety even though we were not totally sure that they were needed) are real world examples of going from 4 PRs to 1.

StewGoin (dismissed):
From review and discussion: the proposed restructure and segmentation of functionality, etc... proposed in this RFC looks good to me from an ISSO perspective. Nothing about it seems to rise to the level of needing an additional external SIA document as this RFC + the expected contents of the PR(s) to implement it will be sufficient for assessing security impact.

brianburton (approved):
I think this is a good strategy and well worth pursuing.

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+a basic testing requirement of any change to the system). The interactions between old and new components though is
+important in understanding this proposal and leads to some classifications of migrations based on the compatibility of
+the schema change with old and new versions of the application:
+
+* A backward-compatible database migration is one where the new schema is compatible with the old applications
+(otherwise it is considered to be backward-incompatible)
+* A forward-compatible database migration is one where the old schema is compatible with the new applications
+(otherwise it is considered to be forward-incompatible)
+* A fully-compatible database migration is one that is both backward-compatible and forward-compatible
+* A fully-incompatible database migration is one that is both backward-incompatible and forward-incompatible
```
karlmdavis, line 87: Love this! ❤️

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+helpful to define some terminology, identify the relevant system components, and differentiate between different types
+of migrations.
+
+### Background on BFD Applications
+
+It is assumed that readers of this proposal are generally familiar with two open source tools that BFD depends on for
+database operations:
+
+* Flyway -- for executing database schema migration scripts
+* Hibernate -- for object-relational mapping between the applications and the database
```
karlmdavis, line 51: Minor suggestion: I'd add the links, just to be helpful.

```suggestion
* [Flyway](https://flywaydb.org/) -- for executing database schema migration scripts
* [Hibernate](https://hibernate.org/orm/) -- for object-relational mapping between the applications and the database
```

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+* A fully-compatible database migration is one that is both backward-compatible and forward-compatible
+* A fully-incompatible database migration is one that is both backward-incompatible and forward-incompatible
+
+Frequently, database migrations include both a schema change and an accompanying application change although there are
+some database migrations that consist only of a schema change. Strictly speaking, a database migration that does not
+contain a schema change is just an application change like any other application change. It is useful though for the
+purposes of this RFC for application changes deployed independently of a schema change to also be considered database
+migrations. Of interest in the next section is the fact that as defined above, a database migration that consists only
+of an application change is fully-compatible because it must work with both old and new schema (which are the same) to
+be a valid change at all.
```
karlmdavis, line 95: Moderate suggestion: A nuance here, and above, that I think would be helpful to clarify is that "old applications" and "new applications" are not... unbounded ranges. You really just need migrations that are forward- and backward-compatible with one deployed version. (Unless/until we move to a more complex deployment strategy that includes >2 application versions running simultaneously.)

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+
+#### Updating ONLY the application to start using a column or table that already exists
+* This is fully-compatible.
+
+#### Dropping a column from the schema ONLY (that the application does not reference)
+* This is fully-compatible.
+
+Of note here is that a migration that is fully-incompatible (like renaming a column in the schema and the application)
+can be accomplished in a manner that is fully-compatible by breaking it into multiple migrations (which would have to
+be deployed separately).
```
karlmdavis, line 130: This is a delightful way of explaining this.

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+Deployment design:
+
+![Proposed Deployment Sequence](./resources/0011-proposed-deployment-sequence.svg)
+
+<details><summary>Diagram Notes</summary>
+Like the previous diagram, this illustrates the deployment sequence with some broad definitions. This adds the `bfd-db-migrator` participant that includes a blocking deployment stage for each environment that must complete **before** proceeding to apply updates to the `bfd-pipeline` and `bfd-server` resources for each environment.
+</details>
+
+* A new migrator service account and associated migrator database role that has all required privileges for running
+migrations and hibernate validation will be created.
```
karlmdavis, line 311: Moderate question: Do we know if this ends up being the DB owner?

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+
+* The Jenkins APP AMI build phase will build a new migrator AMI that includes the new application.
+
+* The migrator AMI and migrator service account will be used to run the migrator application on a new EC2 instance as
+part of a new Jenkins pipeline stage that will be inserted prior to each of the three stages that deploy to TEST,
+PROD-SBX, PROD.
+
+* The Jenkins pipeline will not proceed to the next stage (deploying the other applications) until the migrator
+application has exited with a status indicating success. In the event that this application status indicates that not
+all operations were successful, the Jenkins deployment will halt and be marked as a failure.
```
karlmdavis, line 321: Minor suggestion: You might note here that DB migrations are generally run transactionally, so this should not result in downtime.

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,382 @@
...
+
+* The new application will produce log files that contain the Flyway and Hibernate logging that is currently captured in
+the BFD Pipeline application logs. These new application logs will be sent to Splunk and Cloudwatch.
+
+* The migrator EC2 instance could be powered off or terminated once the migrator application exits and all artifacts
+have been gathered.
```
karlmdavis, line 327: Minor question: This is not a suggestion, but I'm curious if something other than EC2, e.g. Lambda, might be a better fit for this use case?

karlmdavis (dismissed):
Chris, this is fantastic work! 💯 

I have several minor & moderate comments & questions, but nothing that needs to block this RFC.

rfcs/0011-separate-flyway-from-pipeline.md
christrotteradhoc: I'm not 100% sure but that was my assumption. We should verify this as part of the implementation. I will add it to the open questions.

rfcs/0011-separate-flyway-from-pipeline.md
christrotteradhoc: Possibly and we did briefly discuss it but didn't come to a conclusion to use something else. I'll add this to the open questions.

brianburton:
This may be old news to some of us but I found this doc about schema updates in Postgresql very informative: https://gist.github.com/jcoleman/1e6ad1bf8de454c166da94b67537758b

rfcs/0011-separate-flyway-from-pipeline.md
```diff
@@ -0,0 +1,389 @@
...
+will focus on optimizing the Standard Jenkins deployment since it is the preferred deployment and is most commonly used.
+Optimizing the Jenkins deployment can also make it a more viable option for certain complex migrations that otherwise
+would be candidates for a cloned deployment or a downtime deployment.
+
+### Constraints on migrations that are deployed via Jenkins deployment
+
+During a Jenkins deployment the old API server continues to run and serve traffic during the deployment of the new
+applications until the new API server is fully deployed and available. This means that it is possible that the old API
+server will attempt to read from the database after the new schema is in place. This leads to a constraint on BFD
+migrations:
```
lsmitchell, line 183: Perhaps a strange thought, but this is creating a constraint because we can't just clone the DB and bring up a new DB instance with all the data in the new schema configuration, right? Upon reading this, my initial naive thought was "we clone the BFD servers to maintain uptime, why not clone the new DB too and delete the old one once its BFD servers are replaced fully?" I assume it's because the data contained is too massive to clone, and would be prohibitive on a time/money scale, so its not a consideration here.

Not sure if that point is worth mentioning for completeness; but the rest of the RFC seems to explicitly spell out many parts of the migration strategy, so maybe?

lsmitchell (approved):
Great proposal! 

One question: After the flyway migration step, and before the new code is fully deployed and brought up, there seem to be cases where a backwards incompatible change could mean the freshly adjusted DB is read by the still-active old servers before they get replaced. Especially if something weird goes on with the BFD app deployment/AWS/etc. Is this typically an expected short downtime? Is this something that would happen today anyway? (Or am I missing something where this isn't an issue?)

ericbuckley (approved):

mackec1 (approved):
LGTM; might have more meat (or at least some more documentation) when we start leveraging this RFC for ongoing db migration work. Again, I really liked the RFC's structure!

rfcs/0011-separate-flyway-from-pipeline.md
christrotteradhoc: We can do that but it is a different type of deployment. I've referred to it above as a cloned deployment and it does have fewer constraints but has additional costs -- additional hardware but also a window where writes would have to be made to both clusters (which becomes more relevant when RDA goes live), and is currently not an automated solution (although we would like to get there eventually which would make that concern less important).

christrotteradhoc:
> One question: After the flyway migration step, and before the new code is fully deployed and brought up, there seem to be cases where a backwards incompatible change could mean the freshly adjusted DB is read by the still-active old servers before they get replaced. Especially if something weird goes on with the BFD app deployment/AWS/etc. Is this typically an expected short downtime? Is this something that would happen today anyway? (Or am I missing something where this isn't an issue?)

The premise is that all migrations even after this change is in place must be backwards-compatible to avoid this scenario and have zero downtime.

PR #980, opened 2022-02-28
BFD-1566: Implement CCW RIF Filtering for 2022 Enrollment Issues

karlmdavis:
**JIRA Ticket:**
[BFD-1566](https://jira.cms.gov/browse/BFD-1566)

**User Story or Bug Summary:**

In some of the CCW data some benes were coming in with late 2021 enrollment data, which was overwriting newer 2022 entries. This attempts to remedy this issue by adding a filter to skip and save the late 2021 data until CCW comes up with a comprehensive fix to how to send older updates after the year has change.

---

### What Does This PR Do?

- Added filtering for non-null, non-2022 bene UPDATES to be moved into a table for later processing
- Added logic to blow up on any load that attempts to INSERT non-2022 data, as this will cause additional issues that need to be resolved related to claims
- Added supporting classes for a new table to hold the skipped 2021 rif file data
- Added a load option for turning on/off the filtering, filteringNonNullAndNon2022Benes, which will default to true (on)
- Added various IT tests to test the new filtering in various configurations
- Fixed existing IT tests to work with the new filtering logic (or disable it if needed to retain coverage)
- Changed various test support architecture to allow for testing with the new flag and load strategies
- Added the raw CSV data to the RifRecordEvent to support testing
- Adjusted RifFileType to support returning the column header enums, in order to programmatically create headers for test files
- Cleaned up SAMPLE_B data that was no longer needed, along with tests and variables that used this data
- Updated the query checker for the new table/fields
- Added a tag in the FHIR response to indicate a patient has records that exist in the skipped records table
- Adjusted the JPA queries in `PatientResourceProvider` and `R4PatientResourceProvider` to use SQL `JOIN`s rather than N+1 queries

### What Should Reviewers Watch For?

If you're reviewing this PR, please check for these things in particular:

* Verify all PR security questions and checklists have been completed and addressed.

### What Security Implications Does This PR Have?

Submitters should complete the following questionnaire:

* If the answer to any of the questions below is **Yes**, then **you must** supply a link to the associated Security Impact Assessment (SIA), security checklist, or other similar document in Confluence here: **N/A**

    * Does this PR add any new software dependencies? 
      * [ ] Yes
      * [X] No
    * Does this PR modify or invalidate any of our security controls?
      * [ ] Yes
      * [X] No
    * Does this PR store or transmit data that was not stored or transmitted before?
      * [X] Yes
      * [ ] No

[...]

PR #982, opened 2022-03-02
BFD-1566: Add migration script for new table

lsmitchell:
**JIRA Ticket:**
[BFD-1566](https://jira.cms.gov/browse/BFD-1566)

In some of the CCW data some benes were coming in with late 2021 enrollment data, which was overwriting newer 2022 entries. This adds the migration for the table which will hold the skipped records, to be implemented in #980 .

---

### What Does This PR Do?

Adds the migration script for a new table that will hold skipped records.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V58__Add_rif_records_skipped_Table.sql
+++ b/apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V58__Add_rif_records_skipped_Table.sql
@@ -1,0 +1,39 @@
+/**
+ * Adds a new table for storing CCW RIF records that have been temporarily skipped.
+ *
+ * Please note that every record in this table is a piece of as-yet-unaddressed technical debt, with an
+ * exorbitantly high "interest rate": the older these records get, the harder it will be to figure out what
+ * to do with them, and the harder it will be to implement any such process. Ideally, this table is only ever
+ * populated for VERY brief periods of time.
+ * 
+ * Added as part of <https://jira.cms.gov/browse/BFD-1566>.
+ */
+
+CREATE TABLE skipped_rif_records (
+  -- Sequence-generated PK for these records, as no natural PK is feasible.
+  record_id            bigint                     NOT NULL PRIMARY KEY,
+
+  -- The timestamp associated with the CCW data set manifest that this record is from.
+  rif_file_timestamp   timestamp with time zone   NOT NULL,
+
+  -- The `RifFileType` (e.g. `BENEFICIARY`, `CARRIER`, etc.) of the RIF file that this record is from.
+  rif_file_type        varchar(48)                NOT NULL,
+
+  -- TODO: add DML_IND
+  
+  -- The `bene_id` of the beneficiary that this record is of / associated with.
+  bene_id              bigint                     NOT NULL,
+
+  -- The RIF/CSV row or rows representing the record (i.e. beneficiary or claim) that was skipped.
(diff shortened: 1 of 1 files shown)


apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V58__Add_rif_records_skipped_Table.sql
```diff
@@ -0,0 +1,40 @@
...
+  rif_file_timestamp   timestamp with time zone   NOT NULL,
+
+  -- The `RifFileType` (e.g. `BENEFICIARY`, `CARRIER`, etc.) of the RIF file that this record is from.
+  rif_file_type        varchar(48)                NOT NULL,
+
+  -- The `DML_IND` of the RIF record(s).
+  dml_ind              varchar(6)                 NOT NULL,
+
+  -- The `bene_id` of the beneficiary that this record is of / associated with.
+  bene_id              bigint                     NOT NULL,
```
mackec1, line 26: Logan, not sure how you guys worked the bene_id out, but the current db treats it as a varchar(15). Tratings it as a bigint is what the schema changes are heading towards, but we aren't there yet. Leaving as a varchar is not wrong by any stretch, and we can always cast as needed when reconciling this table with future claims processing.

mackec1 (approved):

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V58__Add_rif_records_skipped_Table.sql
christrotteradhoc: the thinking was that since this table doesn't really need an FK (nice to have one day though) that we save a future migration effort and make it bigint now and follow it up with an FK only migration once we have the beneficiaries table switched over.

karlmdavis: It depends on how we plan to use it:
1. If we aren't going to use it at all, for now, then having it be a `bigint` likely makes the most sense, as that's where the rest of the schema is headed and we avoid future conversion work.
2. If we're going to query the table separately, same answer as #1: keep it as a `bigint`.
3. If we're going to query it in `JOIN`s, might make more sense to make it a `varchar` for now, to keep the joins simpler.

Thoughts?

karlmdavis: Trotter's point re: FKs I think it more important than the query strategy, and I agree with him: no FK for now.

mackec1 (approved):

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V58__Add_rif_records_skipped_Table.sql
mackec1: The other possibility is to also add a 'temporary' column for bene_id varchar (i.e., bene_id_string) with an index. When we get to the full bigint support in the schema, we simply drop the added column.

christrotteradhoc (approved):

christrotteradhoc (approved):

karlmdavis (approved):
LGTM

PR #1009, opened 2022-03-15
BFD-1456 - create fake drug code for use with unit and integration tests

cbrunefearless:
**JIRA Ticket:**
[BFD-1456](https://jira.cms.gov/browse/BFD-1456)

User Story:

As a BFD engineer, I want a dummy NDC drug code to be used for tests, so that build tests still pass when the NDC drug codes change.
AC:

A fake drug code is used for unit and integration tests 
Note - failing tests are the dme and part d event transformers
A test is written that ensures that the fake drug code is not present in production software 

---

### What Does This PR Do?

This PR helps create a fake drug code to get passed the normal problem of drug codes constan

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist

*[X] All tech debt and/or shortcomings introduced by this PR are detailed in `TODO` and/or `FIXME` comments, which include a JIRA ticket ID for any items that require urgent attention.


apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -282,7 +282,13 @@
...
 			<version>2.4.7</version>
 		</dependency>
-	</dependencies>
+	
+   <dependency>
+     <groupId>io.sprucehill</groupId>
+     <artifactId>spring-cors-filter</artifactId>
+     <version>1.0.0</version>
+   </dependency>
+ </dependencies>
```
christrotteradhoc, line 291: I don't think this is needed

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -460,4 +461,11 @@ public HealthCheckRegistry healthCheckRegistry() {
     HealthCheckRegistry healthCheckRegistry = new HealthCheckRegistry();
     return healthCheckRegistry;
   }
+
+  @Bean
+  public IDrugCodeProvider drugCodeProvider(
```
christrotteradhoc, line 466: needs a doc

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2.java
```diff
@@ -273,6 +281,11 @@ private static ExplanationOfBenefit transformClaim(
                       TransformerUtilsV2.createExtensionCoding(
                           eob, CcwCodebookVariable.CARR_LINE_MTUS_CNT, code)));
 
+      String drugCode = null;
+
+      if (line.getNationalDrugCode().isPresent()) {
+        drugCode = drugCodeProvider.retrieveFDADrugCodeDisplay(line.getNationalDrugCode().get());
+      }
```
christrotteradhoc, line 288: I suggest encapsulating this in a method that can be reused across these transformers and invoking it inline below where drugCode is needed.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DMEClaimTransformerV2.java
```diff
@@ -24,7 +25,8 @@
 import org.hl7.fhir.r4.model.Quantity;
 
 /** Transforms CCW {@link DMEClaim} instances into FHIR {@link ExplanationOfBenefit} resources. */
-final class DMEClaimTransformerV2 {
+public final class DMEClaimTransformerV2 {
```
christrotteradhoc, line 28: did the access on this class need to change to public?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/FDADrugUtils.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.Set;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FDADrugUtils implements IDrugCodeProvider {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(FDADrugUtils.class);
+  /** Stores the PRODUCTNDC and SUBSTANCENAME from the downloaded NDC file. */
+  private static Map<String, String> ndcProductMap = null;
```
mackec1, line 20: This is more a style argument, so not a criticism at all because we see it in many places in BFD. So we have essentially a Singleton pattern here which could be instantiated in any number of ways. Ideally we'd have DI (dependency injection ) but we could also treat it as singleton which could be instantiated on an as needed basis (i.e., a factory method) anywhere in our codebase. The reason I bring it up is that you had to jump through a bunch of hoops passing around this class; granted, via its interface, but it still pollutes method signatures which ends up being passed down the stack any number of times.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerUtils.java
```diff
@@ -333,7 +326,7 @@ static CareTeamComponent addCareTeamPractitioner(
     // If no match was found, add one to the EOB.
     if (careTeamEntry == null) {
       careTeamEntry = eob.addCareTeam();
-      careTeamEntry.setSequence(eob.getCareTeam().size() + 1);
+      careTeamEntry.setSequence(eob.getCareTeam().size());
```
christrotteradhoc, line 329: this doesn't seem related to this change

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerUtils.java
```diff
@@ -3023,13 +3017,13 @@ public static String retrieveIcdCodeDisplay(String icdCode) {
...
-    if (!drugCodeLookupMissingFailures.contains(icdCode)) {
-      drugCodeLookupMissingFailures.add(icdCode);
+    /*  if (!drugCodeProvider.drugCodeLookupMissingFailures.contains(icdCode)) {
+      drugCodeProvider.drugCodeLookupMissingFailures.add(icdCode);
       LOGGER.info(
           "No ICD code display value match found for ICD code {} in resource {}.",
           icdCode,
           "DGNS_CD.txt");
-    }
+    } */
```
christrotteradhoc, line 3026: this looks like it is incorrect

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v1/eobByPatientIdPaged.json
```diff
@@ -934,12 +934,11 @@
...
           "valueCoding" : {
             "system" : "http://hl7.org/fhir/sid/ndc",
-            "code" : "500904610",
-            "display" : "ACETAMINOPHEN AND CODEINE PHOSPHATE - ACETAMINOPHEN; CODEINE PHOSPHATE"
+            "code" : "000000000"
           }
         } ],
         "sequence" : 1,
-        "careTeamLinkId" : [ 2 ],
+        "careTeamLinkId" : [ 1 ],
```
christrotteradhoc, line 941: I don't think this should be changing here, right?

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v1/eobByPatientIdPaged.json
```diff
@@ -1349,7 +1348,7 @@
         }
       } ],
       "careTeam" : [ {
-        "sequence" : 2,
+        "sequence" : 1,
```
christrotteradhoc, line 1351: I don't think this should be changing here, right?

apps/bfd-server/bfd-server-war/src/main/config/server-start.sh
```diff
@@ -193,6 +193,7 @@ BFD_PORT="${serverPortHttps}" \
 	"-DbfdServer.db.username=" \
 	"-DbfdServer.db.password=" \
 	"-DbfdServer.db.schema.apply=true" \
+	"-DbfdServer.include.fake.drug.code=true" \
```
christrotteradhoc, line 196: I know none of the other ones are comments but I think we should write a comment in here about why we are passing this, that it only applies to testing, etc

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/IDrugCodeProvider.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.Map;
+import java.util.Set;
+
+public interface IDrugCodeProvider {
+  String retrieveFDADrugCodeDisplay(String claimDrugCode);
+
+  Map<String, String> readFDADrugCodeFile();
+
+  Set<String> drugCodeLookupMissingFailures = new HashSet<>();
+}
```
christrotteradhoc, line 13: given the pivot away from conditional beans I suggest getting rid of the interface. either way -- these methods and the class needs documentation.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/FDADrugUtils.java
```diff
@@ -0,0 +1,152 @@
...
+      }
+    } catch (IOException e) {
+      throw new UncheckedIOException("Unable to read NDC code data.", e);
+    }
+    return ndcProductHashMap;
+  }
+
+  private void appendFDATestCode(Map<String, String> ndcProductHashMap) {
+    String line =
+        "0000-0000_000000zz-0zz0-0z00-zzz0-0z00zzz00000\t0000-0000\tFAKE DRUG\tFake Diluent\t\tfake\tFAKE SOLUTION\tFake\t0\t\tFAK\tFAK000000\tFake Company\tWATER\t1\tmL/mL\t\t\tN\t00000000";
```
christrotteradhoc, line 137: if we end up just using part of this string do we need this? can we just put the right fake things into the map and be done?

apps/bfd-server/bfd-server-war/pom.xml
cbrunefearless: Done

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -64,6 +64,7 @@ public class SpringConfiguration {
   public static final String PROP_DB_PASSWORD = "bfdServer.db.password";
   public static final String PROP_DB_CONNECTIONS_MAX = "bfdServer.db.connections.max";
   public static final String PROP_DB_SCHEMA_APPLY = "bfdServer.db.schema.apply";
+  public static final String PROP_INCLUDE_FAKE_DRUG_CODE = "bfdServer.include.fake.drug.code";
```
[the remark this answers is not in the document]
cbrunefearless: done

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
cbrunefearless: done

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DMEClaimTransformerV2.java
cbrunefearless: changed

apps/bfd-server/bfd-server-war/src/main/config/server-start.sh
cbrunefearless: Done

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/FDADrugUtils.java
cbrunefearless: done

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2.java
cbrunefearless: changed it to take an optional...

apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -282,7 +282,7 @@
 			<artifactId>spark-sketch_2.12</artifactId>
 			<version>2.4.7</version>
 		</dependency>
-	</dependencies>
+ </dependencies>
```
christrotteradhoc, line 285: this file should go back to be un-touched by this PR

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -64,6 +64,9 @@ public class SpringConfiguration {
   public static final String PROP_DB_PASSWORD = "bfdServer.db.password";
   public static final String PROP_DB_CONNECTIONS_MAX = "bfdServer.db.connections.max";
   public static final String PROP_DB_SCHEMA_APPLY = "bfdServer.db.schema.apply";
+  /** The {@link String} for the parameter to use the fake drug code implementation */
```
christrotteradhoc, line 67: I suggest a bit more detail here about the parameter. Something along these lines:

Boolean property that is used to enable the fake drug code (00000-0000) that is used for integration testing. When this property is set to the string 'true', this fake drug code will be appended to the drug code lookup map to avoid test failures that result from unexpected changes to the external drug code file <<ideally put a link here to the relevant code that reads that file and generates the map>>. This property defaults to false and should only be set to true when the server is under test in a local environment.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -460,4 +463,16 @@ public HealthCheckRegistry healthCheckRegistry() {
     HealthCheckRegistry healthCheckRegistry = new HealthCheckRegistry();
     return healthCheckRegistry;
   }
+
+  /**
+   * @param includeFakeDrugCodeText the {@link String} to use
```
christrotteradhoc, line 468: this is missing the javadoc description

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -460,4 +463,16 @@ public HealthCheckRegistry healthCheckRegistry() {
...
+   * @return the {@link FDADrugUtils} for the application, uses the current prod implementation or
+   *     the fake drug code for test.
+   */
+  @Bean
+  public FDADrugUtils drugCodeProvider(
+      @Value("${" + PROP_INCLUDE_FAKE_DRUG_CODE + ":false}") String includeFakeDrugCodeText) {
```
christrotteradhoc, line 474: a little bit of searching tells me that we can probably have a boolean property instead of a string property with conversion. can you try to get that to work?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2.java
```diff
@@ -273,6 +279,7 @@ private static ExplanationOfBenefit transformClaim(
                       TransformerUtilsV2.createExtensionCoding(
                           eob, CcwCodebookVariable.CARR_LINE_MTUS_CNT, code)));
 
+      String drugCode = drugCodeProvider.retrieveFDADrugCodeDisplay(line.getNationalDrugCode());
```
christrotteradhoc, line 282: I think it makes sense to inline this (here and below) in the method call since that is more consistent with the other ones and avoids backwards referencing by 50 lines.

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v1/eobByPatientIdAll.json
```diff
@@ -239,7 +239,7 @@
         }
       } ],
       "careTeam" : [ {
-        "sequence" : 2,
+        "sequence" : 1,
```
christrotteradhoc, line 242: I still don't understand these changes or the ones for care team link. Do the JSONs need to be regenerated?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -1456,7 +1458,70 @@ public void searchForEobsByExistingPatientAndType() throws FHIRException {
...
+            Optional.of(false),
+            new FDADrugUtils(true)),
+        filterToClaimType(searchResults, ClaimTypeV2.PDE).get(0));
+  }
+
+  /**
+   * Verifies that {@link
+   * gov.cms.bfd.server.war.r4.providers.ExplanationOfBenefitResourceProvider#findByPatient(ca.uhn.fhir.rest.param.ReferenceParam)}
+   * works as expected for a {@link Patient} that does exist in the DB, with filtering by claim
+   * type.
```
christrotteradhoc, line 1470: I believe this is the test for this AC:

`A test is written that ensures that the fake drug code is not present in production software `

Assuming that is correct, I don't think this description really reflects that and I don't really understand how the test is validating that either. Can we have a more explicit test on the drug code provider class?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/FDADrugUtils.java
christrotteradhoc: This is using dependency injection now (see the changes in SpringConfiguration) but I believe the problem is that all of the consuming classes were static so they cannot inject the bean. Open to concrete suggestions on how Chris could make that better in this PR without refactoring the entire thing.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
cbrunefearless: done

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerUtils.java
cbrunefearless: i took this out dunno since there is a check inside the drugutils

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/FDADrugUtils.java
christrotteradhoc: I think best answer here is a refactor to make the transformer classes actual instantiated objects that can inject dependencies. This is also a problem with the MetricsRegistry. That I believe it outside the scope of this PR though.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PartDEventTransformerV2.java
```diff
@@ -44,7 +48,7 @@ static ExplanationOfBenefit transform(
       throw new BadCodeMonkeyException();
     }
 
-    ExplanationOfBenefit eob = transformClaim((PartDEvent) claim);
+    ExplanationOfBenefit eob = transformClaim((PartDEvent) claim, drugCodeDisplayLookup);
```
mackec1, line 51: Probably need to do some Javadoc (in multiple files); at least certainly the methods that you've added the drug display code to.

mackec1 (approved):
Yikes....lots of work on this! It looks reasonable; I assume you and Karl are going to get together to vett this further?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/FdaDrugCodeDisplayLookup.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  /**
+   * Constructs an {@link FdaDrugCodeDisplayLookup}
+   *
+   * @param includeFakeDrugCode whether to include the fake testing drug code or not
+   */
+  private FdaDrugCodeDisplayLookup(boolean includeFakeDrugCode) {
+    readFDADrugCodeFile(includeFakeDrugCode);
+  }
```
karlmdavis, line 84: Nice to move this to eager loading, rather than lazy loading (which I _think_ it was before?). Thanks.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/FdaDrugCodeDisplayLookup.java
```diff
@@ -0,0 +1,179 @@
...
+   *
+   * @param claimDrugCode - NDC value in claim records
+   * @return the fda drug code display string
+   */
+  public String retrieveFDADrugCodeDisplay(Optional<String> claimDrugCode) {
+    /*
+     * Handle bad data (e.g. our random test data) if drug code is empty or length is less than 9
+     * characters
+     */
+    if (!claimDrugCode.isPresent() || claimDrugCode.isEmpty() || claimDrugCode.get().length() < 9) {
```
karlmdavis, line 98: Aren't the first two checks here redundant with each other?

```
!claimDrugCode.isPresent() || claimDrugCode.isEmpty() || ...
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/FdaDrugCodeDisplayLookup.java
```diff
@@ -0,0 +1,179 @@
...
+      return null;
+    }
+
+    String claimDrugCodeReformatted = null;
+
+    claimDrugCodeReformatted =
+        claimDrugCode.get().substring(0, 5) + "-" + claimDrugCode.get().substring(5, 9);
```
karlmdavis, line 105: Has anyone checked to see how many of the drug codes aren't at least 9 characters long, both in the lookup file and in the DB?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/FdaDrugCodeDisplayLookup.java
```diff
@@ -0,0 +1,179 @@
...
+
+    // log which NDC codes we couldn't find a match for in our downloaded NDC file
+    if (!drugCodeLookupMissingFailures.contains(claimDrugCode.get())) {
+      drugCodeLookupMissingFailures.add(claimDrugCode.get());
+      LOGGER.info(
+          "No national drug code value (PRODUCTNDC column) match found for drug code {} in"
+              + " resource {}.",
+          claimDrugCode.get(),
+          "fda_products_utf8.tsv");
+    }
```
karlmdavis, line 120: This leads to **so much** log noise, and no one has ever done anything with the data. I'd remove it or at least downgrade it to `debug`.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/FDADrugUtilsTest.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+
+public class FDADrugUtilsTest {
+
+  /** Return Fake Drug Code when parameter is true */
+  @Test
+  public void shouldReturnFakeDrugCodeWhenConstructorSetToTrue() {
+    FdaDrugCodeDisplayLookup drugUtils = FdaDrugCodeDisplayLookup.createDrugCodeLookupForTesting();
+    String drugCodeDisplay = drugUtils.retrieveFDADrugCodeDisplay(Optional.of("000000000"));
```
karlmdavis, line 15: This `"000000000"` value should be a constant.

karlmdavis (commented):
Excellent refactoring! I've left some specific comments.

A more general comment, though, is that the ridiculous size of this PR is a strong indication that it's time to start wrapping our transformation configuration and context variables into other objects, e.g. `TransformationContext` and `TransformationConfiguration` or something like that.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/FdaDrugCodeDisplayLookup.java
cbrunefearless: we can look at this in another PR i think

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/FDADrugUtilsTest.java
```diff
@@ -8,11 +8,14 @@
 
 public class FDADrugUtilsTest {
 
+  public static final String FAKE_DRUG_CODE_NUMBER = "000000000";
```
christrotteradhoc, line 11: This should be referenced from the Lookup class where we already have a constant (and I believe this has a difference value btw and should not but let me know if that is incorrect).

christrotteradhoc (approved):

christrotteradhoc (approved):

mackec1 (approved):
LGTM

PR #1020, opened 2022-03-22
BFD-1536 create bene_id_numeric column in beneficiaries table

mackec1:
**JIRA Ticket:**
[BFD-1536](https://jira.cms.gov/browse/BFD-1536)

This ticket marks the beginning of a series of tickets to change the BFD database schema to:
- change bene_id from varchar(15) to bigint

Prior to claims table changes, we create a bene_id_numeric which can be used by parent claims tables as a foreign key.

### What Does This PR Do?
This PR encompasses three flyway migration scripts to setup the beneficiaries table in preparation for doing claims table(s) migrations.

In general terms, the three flyway migrations will:

1. Alter the beneficiaries table to add a new column, bene_id_numeric.
2. Populate the new column by casting the current varchar bene_id to a bigint; this update will take some time (2+ hours) but due to the nature of an update, will not lock readers (SELECT) from accessing data in the table. In addition, the script will register a pre-insert db trigger, that will be invoked on any INSERT operation on the beneficiaries table.
3. Creates a unique index on the new column, bene_id_numeric; again, this will take some time, but the index is built concurrently which does not lock the beneficiaries table.

The reason this needs to be done using 3 scripts is that flyway precludes mixing transactional and non-transactional DDL; in this case, creating an index concurrently is non-transactional.

Reviewers should assert the validity of the DDL (SQL) as well as verify that the operations presented do not affect current BFD services; that is, no operation grabs an exclusive lock on the beneficiaries table for more than 1sec.

NA

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* NA

* N/A

### Submitter Checklist

* [X] I have named this PR and branch so they are [automatically linked]

### DIFF
--- a/apps/bfd-server/bfd-server-launcher/src/main/java/gov/cms/bfd/server/launcher/DataServerLauncherApp.java
+++ b/apps/bfd-server/bfd-server-launcher/src/main/java/gov/cms/bfd/server/launcher/DataServerLauncherApp.java
@@ -204,7 +204,7 @@ public final class DataServerLauncherApp {
     final String accessLogFileName =
         System.getProperty("bfdServer.logs.dir", "./target/server-work/") + "access.log";
     final String requestLogFormat =
-        "%{remote}a - \"%u\" %t \"%r\" \"%q\" %s %{CLF}S %D"
+        "%{remote}a - \"%u\" %t \"%r\" \"%q\" %s %{CLF}S %{ms}T"
             + " %{BlueButton-OriginalQueryId}i"
             + " %{BlueButton-OriginalQueryCounter}i"
             + " [%{BlueButton-OriginalQueryTimestamp}i]"
--- a/ops/terraform/modules/stateless/main.tf
+++ b/ops/terraform/modules/stateless/main.tf
@@ -352,7 +352,7 @@ module "bfd_server_alarm_all_eob_6s-p95" {
     datapoints       = "15"
     statistic        = null
     ext_statistic    = "p95"
-    threshold        = "6000000.0"
+    threshold        = "6000.0" # milliseconds
     alarm_notify_arn = data.aws_sns_topic.cloudwatch_alarms.arn
     ok_notify_arn    = data.aws_sns_topic.cloudwatch_ok.arn
   }


apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/schema/DatabaseSchemaManager.java
```diff
@@ -102,6 +104,8 @@ private static Flyway createFlyway(DataSource dataSource) {
         placeholders.put("logic.sequence-start", "start");
         placeholders.put("logic.sequence-increment", "increment");
         placeholders.put("logic.perms", "");
+        placeholders.put("logic.psql-only", "");
+        placeholders.put("logic.hsql-only", "-- ");
```
christrotteradhoc, line 108: I like the clarity that this change introduces. 👍

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V60__populate_bene_id_numeric.sql
```diff
@@ -0,0 +1,32 @@
...
+${logic.psql-only} CREATE TRIGGER beneficiaries_insert_trigger
+${logic.psql-only}     BEFORE INSERT ON public.beneficiaries
+${logic.psql-only}     FOR EACH ROW
+${logic.psql-only}       EXECUTE FUNCTION public.beneficiaries_pre_insert();
+
+-- update will be fairly long running; minimal effect readers (select)
+UPDATE public.beneficiaries
+${logic.hsql-only} SET bene_id_numeric = convert(bene_id, SQL_BIGINT)
+${logic.psql-only} SET bene_id_numeric = bene_id::bigint
+WHERE bene_id is not null;
```
christrotteradhoc, line 32: I thought in our last sync that we had planned to wrap this update in an exclusive lock as a best practice to protect against write conflicts.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V61__set_bene_id_numeric_constraint.sql
```diff
@@ -0,0 +1,4 @@
+-- add a constraint using the indexed column bene_id_numeric;
+-- psql concurrency enabled which treats this as non-transactional
+CREATE UNIQUE INDEX ${logic.index-create-concurrently} beneficiaries_bene_id_numeric_idx
+    ON public.beneficiaries (bene_id_numeric);
```
christrotteradhoc, line 4: I thought we were going to forgo the concurrent index build in favor of a single script since this takes only a few minutes and simplifies the migration. Did something come up in testing that changed that?

mackec1: Yup, will remove the concurrency; building index will lock out writes, but not reads.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V60__populate_bene_id_numeric.sql
christrotteradhoc: the lock will be held in this transaction which is doing the updates though

mackec1: OK, I was wrong here; you were correct that one can explicitly lock the table as part of the flyway transaction, but you still have to wrap the DDL within a transaction block. So I've made that change and tested.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V59__Alter_beneficiaries_bene_id_numeric.sql
```diff
@@ -0,0 +1,9 @@
+--
+-- The goal here is to create a column, bene_id_numeric, on the
+-- beneficiaries table; an ALTER TABLE will lock the table from
+-- table reads (SELECT) which would definitely affect BFD services
+-- but this is simply a meta change and will complete in ms.
+--
+-- Also, it should be noted that unless explicitly disabled, each flyway
+-- script executes within a transaction.
+ALTER TABLE public.beneficiaries ADD bene_id_numeric bigint;
```
christrotteradhoc, line 9: at this point can we just make it one script?

mackec1: doubtful; even though the alter table runs in a transaction, the alter table by definition creates an exclusive lock cuz it basically has to add the column to every row in the table; it's somewhat analogous to 'vacuum full'. Here is my little cheat-sheet that I use:

mackec1: I've been trying to operate under the assumption, that at most a table would be locked from a read for < 1sec

christrotteradhoc: but if you move this inside of your ex-lock in the other script then it already has the lock

mackec1: It would, except I shouldn't have an exclusive lock in the other script; it needs to be:
LOCK TABLE ONLY public.beneficiaries IN SHARE UPDATE EXCLUSIVE MODE;
which allows SELECT (reads) to still happen.

ALTER TABLE is a killer with respect to having minimal impact on reads; by definition, it creates an EXCLUSIVE LOCK, which prevents reads from happening until the lock is released. Testing in the clone last yesterday afternoon confirmed this.

Now I can merge everything into one flyway migration, but it complicates the script a bit; I'll have to have two seperate transaction, each with its own pessimistic locks. I'll run a test now.

christrotteradhoc: ah, ok, thanks. looks like the alter is actually getting an 'ACCESS EXCLUSIVE' lock and not just an 'EXCLUSIVE' lock (which would allow reads). https://www.postgresql.org/docs/9.1/explicit-locking.html

I think I am fine with the strategy that you have now

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V61__populate_bene_id_numeric.sql
```diff
@@ -0,0 +1,23 @@
+-- lock the beneficiaries table for UPDATE EXCLUSIVE; we need
+-- to wrap in a transaction since lock can only be acquired in a transaction.
+${logic.psql-only} BEGIN;
+-- set a short lock timeout...no one should be updating table (yet)
+${logic.psql-only} set lock_timeout = 100;
+-- try to grab a lock on the table
+${logic.psql-only} LOCK TABLE ONLY public.beneficiaries IN SHARE UPDATE EXCLUSIVE MODE;
+-- once we have the lock, turn off the lock timeout since the rest of
+-- the DDL will take some time; read (SELECT) will still have access 
+${logic.psql-only} set lock_timeout = 0;
```
christrotteradhoc, line 10: can be removed

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V60__Alter_beneficiaries_bene_id_numeric.sql
```diff
@@ -0,0 +1,27 @@
+-- add a column to beneficiaries table; an ALTER TABLE by definition
+-- creates an exclusive lock on the table.
```
christrotteradhoc, line 2: should be 'access exclusive'

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V60__Alter_beneficiaries_bene_id_numeric.sql
```diff
@@ -0,0 +1,27 @@
+-- add a column to beneficiaries table; an ALTER TABLE by definition
```
christrotteradhoc, line 1: top level comment

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V61__populate_bene_id_numeric.sql
```diff
@@ -0,0 +1,23 @@
+-- lock the beneficiaries table for UPDATE EXCLUSIVE; we need
```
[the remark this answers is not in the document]
mackec1: done.

christrotteradhoc (dismissed):

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V60__Alter_beneficiaries_bene_id_numeric.sql
```diff
@@ -0,0 +1,35 @@
...
+
+-- create a database trigger which will be used to imbue a bigint value
+-- into the bene_id_numeric column  during a table record INSERT operation.
+-- The trigger is defined as a pre-insert trigger, which means that prior to
+-- new record being written, the database engine will execute the trigger,
+-- which in turn, fills the bene_id_numeric with a value based on a 'cast'
+-- of the bene_id (varchar).
+
+-- HSQL always gets created/initialized from scratch so no need to drop trigger
+${logic.hsql-only} CREATE TRIGGER beneficiaries_insert_trigger BEFORE INSERT ON public.beneficiaries
```
ericbuckley, line 14: Out of curiosity, why is the trigger needed for HSQL?  Will there be some future test cases that depend on it?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V61__populate_bene_id_numeric.sql
```diff
@@ -0,0 +1,35 @@
...
+${logic.psql-only} set lock_timeout = 100;
+
+-- try to grab a lock on the table
+${logic.psql-only} LOCK TABLE ONLY public.beneficiaries IN SHARE UPDATE EXCLUSIVE MODE;
+
+-- update will be fairly long running; our acquired lock does not impact readers (select)
+UPDATE public.beneficiaries
+${logic.hsql-only} SET bene_id_numeric = convert(bene_id, SQL_BIGINT)
+${logic.psql-only} SET bene_id_numeric = bene_id::bigint
+WHERE bene_id is not null;
```
ericbuckley, line 28: If bene_id is nullable, is there any chance that a row in that table will go from NULL to a value before this series of migrations is complete?  I ask because the trigger in migration V60 is only acting on INSERTS and not UPDATES.

ericbuckley (approved):
comments in the migrations were super helpful, great work

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V60__Alter_beneficiaries_bene_id_numeric.sql
mackec1: good point....while certainly not really needed, it's there in case someone asked the reciprocal question...why don't you have it. also, if someone has an IT where they are inserting a record into the beneficiaries table, then we'd want to at least make things work like postgres.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V61__populate_bene_id_numeric.sql
mackec1: well if bene_id is null we are in a heap of trouble...it's the primary key of the beneficiaries table; things like 'where exists' could have been used as well.

JeremyIR (approved):

PR #1027, opened 2022-03-24
BFD-1595-SNF-Claims-Migration

mackec1:
**JIRA Ticket:**
[BFD-1595](https://jira.cms.gov/browse/BFD-1595)

This PR represents ticketed work to migrate the current SNF_CLAIMS and SNF_CLAIM_LINES tables into a new structure as part of the ongoing schema migration work.

### What Does This PR Do?
The PR contains a single flyway migration script that:
- creates 2 new database tables in the current PUBLIC schema:
  * snf_claims_new
  * snf_claim_lines_new
 - migrates data from current table as follows:
   * data from current SNF_CLAIMS to SNF_CLAIMS_NEW 
   * data from current SNF_CLAIM_LINES to SNF_CLAIM_LINES_NEW
- creates various db artifacts such primary key definition, data constraints and indexes on the _new_ (migrated) table

Verify validity of flyway script:
- flyway script is able to operate using current database migration constraints such as permissions to:
  - create tables
  - create views
  - create indexes
- well organized and commented DDL
- verify that no unintended consequences of DDL; no impact on current production services (i.e., table or record locking of current table data)
- new table structure and completeness

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

- This PR should not be executed until after weekend CCW data feed processing has been verified; this PR needs to create new tables based on latest claims data.
- ideally, this PR would be executed before noon, on Monday after weekend CCW data has been processed.
- The data migration associated with this ticket may run upwards of 24 hours.

* Weekend CCW data processing has completed and been verified.

### Submitter Checklist


apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V62__new_schema_snf_claims.sql
```diff
@@ -0,0 +1,843 @@
...
+-- target table. PSQL can derive the table structure based on
+-- the data input (i.e., column name, data type). Thus, for HSQL,
+-- we need to explicitly define the table structure prior to loading data.
+--
+-- For HSQL, explicitly define/create a new SNF_CLAIMS_NEW table in
+-- the current PUBLIC schema
+${logic.hsql-only}	create table public.snf_claims_new (
+	${logic.hsql-only}	clm_id								bigint NOT NULL,
+	${logic.hsql-only}	bene_id								bigint NOT NULL,
+	${logic.hsql-only}	clm_grp_id							bigint NOT NULL,
```
christrotteradhoc, line 27: I didn't realize we were doing clm_grp_id now. We don't have the pre-work done for the application for that column. Is that going to make the corresponding application change too complex?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V62__new_schema_snf_claims.sql
```diff
@@ -0,0 +1,843 @@
...
+	${logic.hsql-only} clm_pps_cptl_excptn_amt,
+	${logic.psql-only} cast(clm_pps_cptl_excptn_amt as numeric(10,2)),
+	${logic.hsql-only} clm_pps_cptl_fsp_amt,
+	${logic.psql-only} cast(clm_pps_cptl_fsp_amt as numeric(10,2)),		
+	${logic.hsql-only} clm_pps_cptl_ime_amt,
+	${logic.psql-only} cast(clm_pps_cptl_ime_amt as numeric(10,2)),	
+	${logic.hsql-only} clm_pps_cptl_outlier_amt,
+	${logic.psql-only} cast(clm_pps_cptl_outlier_amt as numeric(10,2)),	
+	${logic.hsql-only} clm_pps_old_cptl_hld_hrmls_amt,
+	${logic.psql-only} cast(clm_pps_old_cptl_hld_hrmls_amt as numeric(10,2)),
```
christrotteradhoc, line 526: I was not expecting all of the additional datatype changes to be a part of this. These are all new types, right?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V62__new_schema_snf_claims.sql
mackec1: So CLM_GRP_ID will have zero minimal impact; it is defined as a numeric right now. The same holds true for things like LINE_NUM...currently defined as NUMERIC but is actually a SMALLINT (SHORT) and should be defined in db as such.

So the pre-work that you did is not impacted at all; in the Java code, we are passing around numeric data type for CLM_GRP_ID.

The original 'charter' of the schema work was to make things as small as possible, leading to storage savings and performance gains. So CLM_GRP_ID is currently defined as a numeric. However, If precision is not required, you should not use the NUMERIC type because calculations on NUMERIC values are typically slower than [integers](https://www.postgresqltutorial.com/postgresql-integer/), floats, and double precisions.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V62__new_schema_snf_claims.sql
mackec1: not really; the only thing that changes on the money fields is going from (12,2) to (10,2); minor space saving(s) on a field that would never need to support $1,111,111,111.11. I know health costs are going up but I certainly hope we don't hit a Dr's bill that hits almost a 10 billion $$$'s.

christrotteradhoc: I'm not sure that going from 12,2 to 10,2 gets us any savings. Leading zeroes will not be stored according to this: https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL:

"Numeric values are physically stored without any extra leading or trailing zeroes. Thus, the declared precision and scale of a column are maximums, not fixed allocations. (In this sense the numeric type is more akin to varchar(n) than to char(n).) The actual storage requirement is two bytes for each group of four decimal digits, plus three to eight bytes overhead."

So I think this might be undesirable since it moves us further away from parity with CCW

mackec1: OK, np, will leave them (12,2)

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V63__new_schema_snf_claims.sql
```diff
@@ -0,0 +1,826 @@
+-- NEW_SCHEMA_SNF_CLAIMS.SQL
+-- flyway migration for SNF_CLAIMS and SNF_CLAIM_LINES tables into
+-- a new schema structure that:
+--   1) changes data type of CLM_ID, CLM_GROUP_ID, BENE_ID to BIGINT.
+--   2) organizes parent claim table (SNF_CLAIMS) such that common
+--      claims data columns are organized at top of table structure.
```
christrotteradhoc, line 6: there are a few other columns whose datatypes are changing so they should also be mentioned here

mackec1: added more comments and explicitly mentioned those columns that went from NUMERIC to an integral data type.

lsmitchell (approved):

christrotteradhoc (approved):

PR #1029, opened 2022-03-24
BFD-1529: Populate Part B Termination Code Correctly

JeremyIR:
**JIRA Ticket:**
[BFD-1529](https://jira.cms.gov/browse/BFD-1529)

As a BFD engineer, I want the correct value for the Part B termination code in v2 coverage coded extension, so I can use that field.

---

### What Does This PR Do?

Use the correct part B termination code instead of partA.

Are we using the correct partB termination code? 

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


mackec1 (approved):
Ship it!

mackec1 (approved):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -184,7 +184,7 @@ private static Coverage transformPartB(MetricRegistry metricRegistry, Beneficiar
         coverage, CcwCodebookVariable.MS_CD, beneficiary.getMedicareEnrollmentStatusCode());
 
     addCoverageCodeExtension(
-        coverage, CcwCodebookVariable.B_TRM_CD, beneficiary.getPartATerminationCode());
+        coverage, CcwCodebookVariable.B_TRM_CD, beneficiary.getPartBTerminationCode());
```
christrotteradhoc, line 187: We are missing a unit test and a changelog entry.

christrotteradhoc: Can we change the partATerminationCode in the test data so that it is different than partBTerminationCode? That should make the test fail with the existing code and your change here will cause that test to pass. It should also necessitate changing the tests that validate partATerminationCode.

For the api changelog I'm not sure that I agree. The only reason why the JSON didn't change is because our test data has the same values in these fields. That is not necessarily true for real data (I assume, otherwise they would be the same field) so consumers of the API will potentially see different, correct data showing up after this change. fyi @bfausett77

lsmitchell (approved):

mackec1 (approved):

mackec1 (approved):

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
```diff
@@ -699,11 +699,11 @@ static void assertPartAMatches(Beneficiary inBeneficiary, Coverage inCoverage) {
     currSegment = MedicareSegment.PART_A;
     assertNotNull(coverage);
     assertNotNull(beneficiary);
+    assertEquals("cancelled", coverage.getStatus().toCode());
 
     verifyCoverageClass("Part A");
     verifyMeta();
     verifyExtensionsPartA();
-    verifyCoverageStatus();
```
christrotteradhoc, line 706: why was this removed and the direct assertion added? do we still use this method in other places where we expect the status to be active? if so adding a method parameter to the method would be more consistent

apps/bfd-server/dev/api-changelog.md
```diff
@@ -1,5 +1,72 @@
 # API Changelog
 
+## BFD-1529: Populate Part B Termination Code Correctly
+
+For Coverage Claims, the Part A coverage termination code was being provided by the API for Part B events. 
+In the test dataset, Part A and Part B coverage termination codes are the same, 
+however in production, this will not always be the case.
+So in addition to providing the Part B termination code correctly, 
+Part A will have a different termination code than Part B
```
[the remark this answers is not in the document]
bfausett77: I think the information in the changelog can be even more simple.  I don't think we even need to provide an example like we are below.  The change is literally just ensuring that the part-b termination code that we have for the patient is being displayed instead of the part-a termination code.  So I would just have the API ChangeLog say:

"Previously the Part B coverage termination code was always the same as the Part A coverage termination code.  The data will now reflect the proper Part B termination code."  Thoughts?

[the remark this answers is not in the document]
bfausett77: @lsmitchell @christrotteradhoc

bfausett77 (dismissed):
Looks good.  Thanks for the changes in the 'change log'.  Nice work!

apps/bfd-server/dev/api-changelog.md
```diff
@@ -1,5 +1,10 @@
 # API Changelog
 
+## BFD-1529: Populate Part B Termination Code Correctly
+
+Previously the Part B coverage termination code was always the same as the Part A coverage termination code.
+The data will now reflect the proper Part B termination code
```
christrotteradhoc, line 6: nitpick -- missing a period

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
```diff
@@ -146,7 +146,7 @@ private static void verifyExtensionsPartA() {
   @Test
   public void verifyCoverageStatusPartA() {
     transformCoverage(MedicareSegment.PART_A, false);
-    verifyCoverageStatus();
+    assertEquals("cancelled", coverage.getStatus().toCode());
```
christrotteradhoc, line 149: can't this be `verifyCoverageStatus("cancelled")` now ?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
```diff
@@ -576,8 +576,18 @@ private static void verifyCodedExtension(String url, String code, String display
...
+    if (status != null && !"".equals(status)) {
+      if ("active".equals(status)) {
+        assertEquals("active", coverage.getStatus().toCode());
+      } else if ("cancelled".equals(status)) {
+        assertEquals("cancelled", coverage.getStatus().toCode());
+      } else if ("draft".equals(status)) {
+        assertEquals("draft", coverage.getStatus().toCode());
+      } else if ("entered-in-error".equals(status)) {
+        assertEquals("entered-in-error", coverage.getStatus().toCode());
+      }
```
christrotteradhoc, line 589: can we just make this `assertEquals(status, coverage.getStatus().toCode())` ?

mackec1 (dismissed):
I'll assume you'll add the 'period' to your change log entry.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
christrotteradhoc: I was not very clear, sorry. I was thinking the the entire method body can be replaced with just one assertEquals.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
```diff
@@ -576,8 +576,18 @@ private static void verifyCodedExtension(String url, String code, String display
     assertTrue(compare.equalsDeep(ex));
   }
 
-  private static void verifyCoverageStatus() {
-    assertEquals("active", coverage.getStatus().toCode());
+  private static void verifyCoverageStatus(String status) {
+    if (status != null && !"".equals(status)) {
```
lsmitchell, line 580: Generally I think this would be written ` if (status != null && status.length > 0)` just to be slightly easier to read. (Apache common utils also has a nice util StringUtils.isNotBlank too for doing this type of null+empty check)

lsmitchell (approved):
Approved pending period adjustment

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
```diff
@@ -576,8 +576,10 @@ private static void verifyCodedExtension(String url, String code, String display
     assertTrue(compare.equalsDeep(ex));
   }
 
-  private static void verifyCoverageStatus() {
-    assertEquals("active", coverage.getStatus().toCode());
+  private static void verifyCoverageStatus(String status) {
+    if (status != null && status.length() > 0) {
```
christrotteradhoc, line 580: do we actually need this condition? I would think that if a test passed in null or an empty string here then this method should assert that the coverage code is such. If that is wrong then the test should not assert it as so.

The risk I see here is that if a test does pass null, this will not assert anything and silently just work.

lsmitchell: yup with the refactor of the method this was no longer needed

lsmitchell (approved):

christrotteradhoc (approved):

PR #1043, opened 2022-04-06
BFD-1596: code change to work with new schema: snf_claims_new and snf_claim_lines_new

mackec1:
**JIRA Ticket:**
[BFD-1596](https://jira.cms.gov/browse/BFD-1596)

BFD code changes to use new, modified schema tables, _snf_claims_new_ and _snf_claim_lines_new_.

### What Does This PR Do?
This PR changes how BFD services provide SNF claims data to requesters. The changes germane to this functionality:
* modify the Excel workbook that drives the RifLayoutsProcessor generation of ORM models; in particular, modifies the SNFClaims worksheet to match new db table content.
* Maintains support for yet-to-be migrated claims tables while at the same time, allowing for changes to the SNF claims table(s) data types and table structure.

Reviewers should:
* verify that code changes to match SNF claims data table structure (i.e., code matches schema)
* check comments to ensure they adequately explain the ongoing migration effort, and that changes associated with this PR are directed towards to producing valid FHIR responses for SNF claims.
* Adequately identifies code changes that may be considered transient (i.e, places where code will be reverted when all BFD data tables have been migrated and cod cleanup becomes necessary).
* Assert that code changes and associated db schema changes do not introduce a regression and continue to produce expected FHIR responses for both stu3 and R4.

* check FHIR responses match current Master API codebase responses, especially those that include SNF claims data.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* Database migrations tied to BFD-1595 (https://github.com/CMSgov/beneficiary-fhir-data/tree/cmac/BFD-1595-SNF-Claims-Migration)

* https://github.com/CMSgov/beneficiary-fhir-data/pull/1027

### Submitter Checklist


apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/MappingSpec.java
```diff
@@ -311,7 +332,9 @@ public String getLineEntityLineNumberField() {
    *     that should be used for the identifying yearMonth, if any
    */
   public String getEntityBeneficiaryMonthlyField() {
-    if (!isBeneficiaryEntity) throw new IllegalStateException();
+    if (!isBeneficiaryEntity) {
+      throw new IllegalStateException();
+    }
```
christrotteradhoc, line 337: these types of changes seem to be the only type in this file. any reason why this is changing on this PR?

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -141,6 +141,23 @@ public static Long parseLong(String longText) {
...
 
+  /**
+   * @param shortText the number string to parse
+   * @return the specified text parsed into an {@link Long}
+   */
+  public static short parseShort(String shortText) {
+    /*
+     * Might seem silly to pull this out, but it makes the code a bit easier
+     * to read, and ensures that this parsing is standardized.
+     */
```
christrotteradhoc, line 152: I suggest getting rid of this low-value comment here (I know that it got copy pasted from the integer one)

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -141,6 +141,23 @@ public static Long parseLong(String longText) {
     }
   }
 
+  /**
```
christrotteradhoc, line 144: Can you put a description here as that will be in the checkstyle rules later

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -164,11 +177,16 @@ public static Long parseLong(String longText) {
...
   public static Optional<Long> parseOptionalLong(String longText) {
-    if (longText.isEmpty()) {
-      return Optional.empty();
-    } else {
-      return Optional.of(parseLong(longText));
-    }
+    return longText.isEmpty() ? Optional.empty() : Optional.of(parseLong(longText));
+  }
+
+  /**
```
christrotteradhoc, line 183: description

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -295,11 +303,7 @@ public static Instant parseTimestamp(String timestampText) {
    *     Optional if not
    */
   public static Optional<Instant> parseOptionalTimestamp(String timestampText) {
-    if (timestampText.isEmpty()) {
-      return Optional.empty();
-    } else {
-      return Optional.of(parseTimestamp(timestampText));
-    }
+    return timestampText.isEmpty() ? Optional.empty() : Optional.of(parseTimestamp(timestampText));
```
christrotteradhoc, line 306: I see a lot of these changes in here -- I think we ought to add this to a future style guide discussion before changing them everywhere. It's a personal preference that we should discuss once and adopt a standard for. Without that I think changing them everywhere is premature.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -165,7 +165,7 @@ public static IdDt buildPatientId(Beneficiary beneficiary) {
    *     Beneficiary}
    */
   public static IdDt buildPatientId(Long beneficiaryId) {
-    return new IdDt(Patient.class.getSimpleName(), beneficiaryId.toString());
+    return new IdDt(Patient.class.getSimpleName(), beneficiaryId);
```
christrotteradhoc, line 168: I remember this change from a previous PR. Why the change back?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -404,6 +404,26 @@ static Identifier createClaimIdentifier(
     return identifier;
   }
 
+  /**
```
christrotteradhoc, line 407: description

christrotteradhoc (commented):
nothing major, but a few questions

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/MappingSpec.java
mackec1: Apart from being grievously affronted by the lack of brackets? Personally. I believe this is an absolute line in the sand as far as code is concerned. Coming from background in other langs (espec c/c++) I have seen too many errors because devs miss things when not absolutely scoped. So my linter barks at this sort of stuff...

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
mackec1: ok, if I remove this, I think we should remove them all; there's not a whole lot of value to the comment, and there's nothing wrong with code that handles particular inputs.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
mackec1: I 'sort of' agree.....it goes back to my firm belief of brackets & scoping (and another one of multiple return statements). So I could change the code to include the brackets which will result in 5 ilnes of change:

if (timestampText.isEmpty()) {
    return Optional.empty();
} else {
    return Optional.of(parseTimestamp(timestampText));
}

Now as a lazy developer, I'd rather have one line than 5!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
mackec1: This is the proverbial "more than one way to skin a cat". Now for something like this I prefer to let the class itself handle things; to wit, here are the 2 pertinent constructors:

IdDt(String theResourceType, Long theIdPart)
IdDt(String theResourceType, String theId)

So in this case, I think it's better to pass the Long and let the class handle it rather than do some gyration in the caller to 'fit it in'.

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/MappingSpec.java
christrotteradhoc: I guess that last part is what I don't understand -- this class has no changes other than this cleanup that I can see.

I'm all for cleaning up stuff but I think the approach should be:

1. propose a style rule
2. get buy-in from the team and document the result in the style guide
3. modify the formatter and linter accordingly so that the change can be applied globally and not regress later
4. apply the change to the entire codebase at once in a single standalone PR

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
christrotteradhoc: This is probably academic at this point since I agree with this change. However this is what it looked like originally and you modified it at the very end of 1532 because you found something in your testing. I just want to be certain that there is not something here of concern as I didn't understand originally why you wanted to change it.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
mackec1: agreed; however I did have to introduce code (new parser for bigint, smallint, and integer); and I will always use the ternary operator for these simple if-then-else. So my conundrum was add my changes using ternary and then field questions as to why didn't I use a verbose if-then-else.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
mackec1: no, did not find anything other than doing some QC of code and I knew from previous work (BFD1076) that the FHIR IdDt class supported Long.

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/RifLayoutsProcessor.java
```diff
@@ -1892,6 +1910,20 @@ private static List<RifField> createDetailsForAdditionalDatabaseFields(
         addlDatabaseFields.add(lastUpdated);
         continue;
       }
+      if (additionalDatabaseField.contentEquals("BENE_ID_NUMERIC")) {
+        RifField lastUpdated =
```
lsmitchell, line 1914: I see the same error in the other fields, but this var shouldnt be named lastUpdated since this is bene_id_numberic.

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/RifLayoutsProcessor.java
```diff
@@ -2203,26 +2235,27 @@ private static TypeName selectJavaFieldType(
...
-    else if (type == RifColumnType.TIMESTAMP && columnLength.orElse(0) == 20)
-      return ClassName.get(Instant.class);
+    if (type == RifColumnType.CHAR) {
+      if (columnLength.orElse(Integer.MAX_VALUE) == 1) {
+        return isColumnOptional ? ClassName.get(Character.class) : TypeName.CHAR;
+      } else {
+        return ClassName.get(String.class);
+      }
+    } else if (type == RifColumnType.DATE) return ClassName.get(LocalDate.class);
+    else if (type == RifColumnType.TIMESTAMP) return ClassName.get(Instant.class);
```
lsmitchell, line 2245: Here's some cursed non-bracketed if statements you could have justification to fix 😄

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/RifLayoutsProcessor.java
```diff
@@ -2203,26 +2235,27 @@ private static TypeName selectJavaFieldType(
...
       return ClassName.get(BigDecimal.class);
     else if (type == RifColumnType.NUM
         && columnScale.orElse(Integer.MAX_VALUE) == 0
         && !isColumnOptional) return TypeName.INT;
-    else if (type == RifColumnType.NUM
-        && columnScale.orElse(Integer.MAX_VALUE) == 0
-        && isColumnOptional) return ClassName.get(Integer.class);
-    else throw new IllegalArgumentException("Unhandled field type: " + type.name());
+    else if (type == RifColumnType.SMALLINT) {
+      return isColumnOptional ? ClassName.get(Short.class) : TypeName.SHORT;
```
lsmitchell, line 2252: What is this even doing? This method could particularly use an @returns in the doc cause I don't know whats even happening here. So we return "Short" or "short" based on.. the column being optional? Not a condemnation of this change since its following the convention of the method, but its just written confusingly in the first place.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -141,6 +137,19 @@ public static Long parseLong(String longText) {
     }
   }
 
+  /**
+   * @param shortText the number string to parse
+   * @return the specified text parsed into an {@link Long}
```
lsmitchell, line 142: Should be updated to say short instead of long

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
```diff
@@ -178,12 +188,8 @@ public static Optional<Long> parseOptionalLong(String longText) {
    * @return the specified text parsed into a {@link BigDecimal}
    */
   public static BigDecimal parseDecimal(String decimalText) {
-    /*
-     * Might seem silly to pull this out, but it makes the code a bit easier
-     * to read, and ensures that this parsing is standardized.
-     */
     if (decimalText.isEmpty()) {
-      return new BigDecimal(0);
+      return BigDecimal.valueOf(0);
```
lsmitchell, line 192: 🍰  Can also use BigDecimal.ZERO

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
lsmitchell: IMO ternaries are fine for simple logic splits like this, but is certainly a good thing to put on the list of things to discuss as part of a greater style guide.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -398,8 +398,18 @@ private <T> List<T> findClaimTypeByPatient(
...
+    // we may need to change the expected type for beneficiaryId in the entity Predicate.
+    // Once all claims have been migrated we can modify the ClaimTypeV2 to specifically
+    // return a long data type and this extra data type checking can be removed.
+    java.lang.Class javaClass = claimType.getEntityBeneficiaryIdAttribute().getJavaType();
+
     Predicate wherePredicate =
-        builder.equal(root.get(claimType.getEntityBeneficiaryIdAttribute()), patientId);
+        builder.equal(
+            root.get(claimType.getEntityBeneficiaryIdAttribute()),
+            javaClass.getName().equals("long") ? Long.parseLong(patientId) : patientId);
```
lsmitchell, line 411: I think getJavaType was returning class.getName under certain conditions, wasnt it? Will that return "Long" or the lowercase version? (And if it returns the same value, why do we return the same value in two different ways within getJavaType?)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/SNFClaimTransformer.java
```diff
@@ -345,7 +345,10 @@ private static ExplanationOfBenefit transformClaim(SNFClaim claimGroup) {
           claimLine.getTotalChargeAmount(),
           claimLine.getNonCoveredChargeAmount(),
           BigDecimal.valueOf(claimLine.getUnitCount()),
-          claimLine.getNationalDrugCodeQuantity(),
+          claimLine.getNationalDrugCodeQuantity().isPresent()
+              ? Optional.of(
+                  BigDecimal.valueOf(claimLine.getNationalDrugCodeQuantity().get().longValue()))
+              : Optional.empty(),
```
lsmitchell, line 351: IMO, This is near the line of complexity where i'd recommend breaking it out of a ternary and into a statement, then pass the result of the statement to this line.

lsmitchell (commented):
Given the larger code changes in RifLayoutsProcessor, I'd be interested in ensuring the new code changes are properly covered by tests; could you run coverage on that class at least and ensure there are tests for the various code paths in the method(s) you touched, particularly on the new code?

I like a lot of the cleanup in this PR! Hopefully future style guide discussion and work can calcify and automate some similar styles so we can clean this up more thoroughly in the codebase.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
mackec1: Depends on the actual db column type AND whether the db column can be nullable. There is a very simple rule to be applied here:

If the db column is defined as an integral data data **AND that column cannot be null**, then use a native primitive; in this case 99% of the time, beneficiaryId is not nullable and therefore ORM models need to treat it as a long, period.

However, we have one table, medicare_benficiaryid_history where bene_id is defined as nullable; so for that we need a Long data type.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
mackec1: good call (and more efficient). changing to use .ZERO.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
lsmitchell: Thanks for context on the temp migration code! Looking forward to this being simplified once we're done with the migrations. (Maybe the temporary code should be denoted as such for easier removal later?)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/SNFClaimTransformer.java
mackec1: agreed; will break apart retrieval of the long value prior to the ternary.

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/RifLayoutsProcessor.java
mackec1: So the TypeName that it returns is a Java Poet ORM data type to be applied to that column in the entity model. I've added a return to try to explain a bit more. And yes your are right on many counts....it is confusing (to say the least) and the isColumnOptional drives whether the column can be a primitive or a Java class.

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/RifLayoutsProcessor.java
mackec1: yea, already got beat up for fixing some of the others; I basically operated on the premise that if I had to modify a method, I was not going to perpetuate bad coding practices. I'll take a further look at this one, to see it meets my gag criteria.

mackec1: I did add those lines to my gag reflex fixing on non-bracketed code.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/parse/RifParsingUtils.java
mackec1: added description.

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/RifLayoutsProcessor.java
mackec1: As an aside, a comment on your request to add tests coverage.....this is where things could get really tricky and really create some scope creep. Where you asked to add test coverage, we run into the proverbial chicken-egg; the RifLayoutsProcessor is part of the bfd-model-codegen which not only doesn't have any tests, it lives only within the constraints of maven to create ORM models based on an Excel spreadsheet. I'd be in favor of a separate ticket to implement some sort of testing, but literally it'll be bootstrapping an entire test system not just adding a test or two to assert some newly added or changed code.

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/MappingSpec.java
christrotteradhoc: what's the plan on this comment? Looks like all other comments resolved and this is about ready to go

mackec1: reverted back to no brackets.

christrotteradhoc (dismissed):

apps/bfd-model/bfd-model-codegen/src/main/java/gov/cms/bfd/model/codegen/RifLayoutsProcessor.java
lsmitchell: Ok, we should capture this as tech debt then.

lsmitchell (approved):

christrotteradhoc (approved):

PR #1074, opened 2022-04-25
BFD-1663: Propose bfd-server startup check (RFC)

mjburling:
**JIRA Ticket:**
[BFD-1663](https://jira.cms.gov/browse/BFD-1663)

Propose bfd-server startup check.

---

### What Does This PR Do?

This RFC proposes a minor change to existing bfd-server on-host health checking to satisfy the requirements of a reliable startup check.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [x] Yes
      * [ ] No

* N/A

### Submitter Checklist


mariosmith90 (approved):

christrotteradhoc (dismissed):
Great work here

StewGoin (commented):
One question prior to approval -- is there a reason to go with an iptables DROP vs. a REJECT? I'm curious if a REJECT might be a better behavior since it gives a packet response vs. just ignoring.

mjburling:
> One question prior to approval -- is there a reason to go with an iptables DROP vs. a REJECT? I'm curious if a REJECT might be a better behavior since it gives a packet response vs. just ignoring.

Given the simplicity of the solution, I honestly hadn't given it a second thought. I've read recently that VPC ACLs behave similarly to `DROP` rules, but the fact that we're proposing them here is a happy accident building off of the spike work @mariosmith90 and I did to inform the proposal. I don't believe the CLB's tcp ping's behavior will change based on meeting dropped vs rejected packets, but I am curious. The implementation *is available* to deploy from #1077 to the shared integration test environment (once things have calmed down there). I'll meditate on this before getting back to you. 

What's more, it might be nice from the debugging perspective if things go awry...

StewGoin:
I'm good with whichever path forward makes sense, I just know from past experience that _not_ doing TCP according to spec can sometimes have unintended consequences and unless DROP provides an advantage to us it's probably worth doing a REJECT instead for troubleshooting, etc... like you said :)

mjburling:
Thanks for the helpful feedback, @StewGoin. We all really appreciate it. tl;dr; there are advantages to playing by the rules and adopting `REJECT` over `DROP`. I'll adjust the proposal here and the implementation. Thanks!

<details><summary>More time to kill?</summary>

After giving this some more thought and running through a few different scenarios, I have to agree that `REJECT` is the intended behavior here–while it doesn't _really_ matter for **new** application instances, this will also bring better support for service restarts, which isn't really an *extreme* edge case when considering exceptional, out-of-band fixes that may have an operator making manual adjustments and restarting the service. This isn't really supported today, but there is appetite for developing automated workflows that will make these exceptional changes easier, more predictable, and safer.

> not doing TCP according to spec can sometimes have unintended consequences

So, it _does_ matter for new instances, too. The health check is needlessly waiting on 5-second timeouts when we could otherwise clearly communicate that the service does exist, it just hasn't fully started and won't accept traffic. That's in keeping with the spirit of the startup check. Let's send the right signal.

</details>

christrotteradhoc (dismissed):

keithdadkins:
I would not have the bfd service account manage firewall rules, have a separate service do that instead. This can easily be accomplished with a few systemd units:
- One to enable/restore iptables on boot (base rules, established traffic, loopbacks, ssh, etc)
- And a second to run your health-check script forwarding/rejecting 7443 as desired

These units can run privileged, so no need to elevate bfd's service account privs. And systemd allows us to take advantage of service dependencies, survive reboots, etc. Ie., healthchecker depends on bfd, iptables depends on network. KISS

mjburling:
> [...]
> This can easily be accomplished with a few systemd units:
> 
> * One to enable/restore iptables on boot (base rules, established traffic, loopbacks, ssh, etc)
> * And a second to run your health-check script forwarding/rejecting 7443 as desired
> [...] KISS

I considered this very implementation! Ultimately, I opted for something simpler. But, maybe simplicity is in the eye of the beholder? I've reduced and added these comments to the future possibilities section. Thanks!

> And systemd allows us to take advantage of service dependencies, survive reboots, etc.

Explicit dependencies *are* great, but I don't think we can make the dependency relationships any clearer than how they are already defined within the same function today. Surviving reboots might be nice too, but this proposal doesn't include any provisions for reboot survivability; when outside the initial 430s ASG grace period, bfd-server application instances are terminated after 2-consecutive unhealthy checks over 20 seconds as reported by the CLB. A rebooted instance would certainly be out of service long enough to trigger replaceent, systemd units or otherwise.

keithdadkins (approved):

christrotteradhoc (approved):

PR #1075, opened 2022-04-25
BFD-1676: Introduce parallel performance regression test suite

malessi:
**JIRA Ticket:**
[BFD-1676](https://jira.cms.gov/browse/BFD-1676)

As a BFD engineer, I want to know if my new changes have significantly impacted the performance of the system so that I can avoid slowing down the endpoint responses.

---

### What Does This PR Do?

This PR:  

* Adds two new Python load test suites/files -- `v1/baseline_tests.py` and `v2/baseline_tests.py`
  * These files include all of the single tests (excluding those from `v2/partially_adjudicated.py` for `v2/baseline_tests.py`) from their respective API versions. However, when running one of these two new baseline files these tests run in parallel (with equal weighting to each endpoint)
  * Some work is done to attempt to bypass caching by setting `Cache-Control` headers and by shuffling input data per Locust `User`
  * Some expected SLA values were added to `common/validation.py` for both V1 and V2 baseline tests to ensure SLAs for these tests
* Refactors the way YAML configuration data is loaded
  * YAML configuration is read and merged with CLI arguments with the CLI taking priority, allowing users to re-use values in existing configuration files
  * Users can now specify a path to a _readonly_ (unless the file is `./config.yml`) YAML configuration file 
    * I expect this to be useful for pipeline integration, as the pipeline could provide connection details via the CLI, but could read users, spawn rate, and more via `configPath` without them being hardcoded in a `Jenkinsfile` or other source code
    * If the path is not `config.yml`, then the file is only read from and _not_ written to (the merged values are written to `config.yml`). I felt this behavior was preferable, especially if a developer is creating and testing a particular configuration as they would otherwise have to discard changes to the config file between runs
  * > **Note:** These changes to the configuration reading were made so that values that might vary depending on the target machine (such as user count, spawn rate, runtime, etc.) would not need to "hardcoded" in a `Jenkinsfile` or recorded in an external document  
* Adds YAML configuration files for both V1 and V2 baseline tests running against a `c5xlarge` target under the `config/` directory
  * Running these tests on a test instance would look something like `python3 runtests.py --configPath="config/config.baseline_v1_c5xlarge.yml --testFile="v1/baseline_tests.py" ...`
* Modifies the `README` to include some examples of how to run the tests with the configuration refactors, and also to fix some inaccuracies related to database CLI arguments

* Are these new test suites representative of what we feel is "baseline"? Are these suites suitable for identifying performance regressions?
* Is any additional documentation required?

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/ops/ccs-ops-misc/load_test/README.md
+++ b/ops/ccs-ops-misc/load_test/README.md
@@ -88,6 +88,8 @@ Essentially, all the items you would set up in the config file are set in a sing
 
 **--testFile** : (Required) : The path to the test file we want to run.
 
+**--configPath** : (Optional) : The path to a YAML configuration file that will be read from (but not _written_ to) for the values specified here. The values in this configuration file will be merged with values from the CLI, with the CLI values taking priority. The resulting merged values will then be written to the root `config.yml` file. If not provided, defaults to `config.yml` (the root YAML configuration file). 
+
 **--serverPublicKey** : (Optional) : To allow the tests to trust the server responses, you can add the path to the public certificate here. This is not required to run the tests successfully, and may not be needed as a parameter at all. If not provided, defaults to an empty string (does not cause test issues.)
 
 **--testRunTime** : (Optional) : How long the test will run. This uses values such as 1m, 30s, 1h, or combinations of these such as 1m 30s. If not provided, defaults to 1m. **Note**: We automatically adjust the run time so that the test runs for the specified amount of time *after* spawning all clients (see `--maxClients` and `--clientsPerSecond` below). For example, with the defaults of a one-minute test, 100 clients, and a spawn rate of 5 clients per second, then the script will spend twenty seconds ramping up its clients and *then* run for the one minute specified. It is optional whether or not to use the `--resetStats` flag to drop the statistics covering this ramp-up period.
--- a/ops/ccs-ops-misc/load_test/README.md
+++ b/ops/ccs-ops-misc/load_test/README.md
@@ -75,13 +75,13 @@ A single-line test will look like this (placeholders to replace in brackets):
 If you have existing configuration in `./config.yml`, you can also run the tests via:
 
 ```
-python3 runtests.py
+python3 runtests.py --testFile="<your-test-file>"
 ```
 
 Or, if you have some YAML configuration in a different file (note that these values will be saved to the root `./config.yml`, so subsequent runs can omit the `configPath` if you are not changing anything):
 
 ```
-python3 runtests.py --configPath="config/<your-config-here>.yml" <other-cli-args-here>...
+python3 runtests.py --configPath="config/<your-config-here>.yml" --testFile="<your-test-file>" <other-cli-args-here>...
 ```
 
 Essentially, all the items you would set up in the config file are set in a single line. There are some optional, and some required parameters here:
--- a/ops/ccs-ops-misc/load_test/README.md
+++ b/ops/ccs-ops-misc/load_test/README.md
@@ -72,6 +72,18 @@ A single-line test will look like this (placeholders to replace in brackets):
 
     python3 runtests.py --homePath="<home_directory_path>" --clientCertPath="<home_directory_path>/bluebutton-backend-test-data-server-client-test-keypair.pem" --databaseUri="postgres://<db-username>:<db-password>@<db-host>:<port>/<db-name>" --testHost="https://test.bfd.cms.gov" --testFile="./v2/eob_test_id_count.py" --testRunTime="1m" --maxClients="100" --clientsPerSecond="5"
 
+If you have existing configuration in `./config.yml`, you can also run the tests via:
+
+```
+python3 runtests.py
+```
+
+Or, if you have some YAML configuration in a different file (note that these values will be saved to the root `./config.yml`, so subsequent runs can omit the `configPath` if you are not changing anything):
+
+```
+python3 runtests.py --configPath="config/<your-config-here>.yml" <other-cli-args-here>...
+```
+
 Essentially, all the items you would set up in the config file are set in a single line. There are some optional, and some required parameters here:
 
 **--homePath** : (Required) : The path to your home directory on the local box, such as /home/logan.mitchell/
--- a/ops/ccs-ops-misc/load_test/README.md
+++ b/ops/ccs-ops-misc/load_test/README.md
@@ -70,7 +70,7 @@ The tests are run with parameters that specify all the test params and test file
 
 A single-line test will look like this (placeholders to replace in brackets):
 
-    python3 runtests.py --homePath="<home_directory_path>" --clientCertPath="<home_directory_path>/bluebutton-backend-test-data-server-client-test-keypair.pem" --databaseHost="<AWS_DB_host>" --databaseUsername="<db_username_from_keybase>" --databasePassword="<db_password_from_keybase>" --testHost="https://test.bfd.cms.gov" --testFile="./v2/eob_test_id_count.py" --testRunTime="1m" --maxClients="100" --clientsPerSecond="5"
+    python3 runtests.py --homePath="<home_directory_path>" --clientCertPath="<home_directory_path>/bluebutton-backend-test-data-server-client-test-keypair.pem" --databaseUri="postgres://<db-username>:<db-password>@<db-host>:<port>/<db-name>" --testHost="https://test.bfd.cms.gov" --testFile="./v2/eob_test_id_count.py" --testRunTime="1m" --maxClients="100" --clientsPerSecond="5"
 
 Essentially, all the items you would set up in the config file are set in a single line. There are some optional, and some required parameters here:
 
@@ -78,11 +78,7 @@ Essentially, all the items you would set up in the config file are set in a sing
 
 **--clientCertPath** : (Required) : The path to the PEM file we copied/modified earlier. Should be located in your home directory like this: ~/bluebutton-backend-test-data-server-client-test-keypair.pem
 
-**--databaseHost** : (Required) : The database host, used to pull data. Get this from the database endpoint we noted in the earlier step "Getting The Database Instance"
-
-**--databaseUsername** : (Required) : The username for connecting to the database, can be found in Keybase. Make sure to use the correct one for the environment you're connecting to.
-
-**--databasePassword** : (Required) : The password for connecting to the database, can be found in Keybase. Make sure to use the correct one for the environment you're connecting to.
+**--databaseUri** : (Required) : The URI used for connecting to the database server. Needs to include username, password (both can be found in Keybase, make sure to use the correct one for the environment you're connecting to), hostname and port.
 
 **--testHost** : (Required) : The load balancer or single node to run the tests against. The environment used here should match the same environment (test, prod-sbx, prod) as the database, so we pull the appropriate data for the environment tested against. Note that when using a single node, you should specity the Ip AND the port for the application.
 
--- a/ops/ccs-ops-misc/load_test/README.md
+++ b/ops/ccs-ops-misc/load_test/README.md
@@ -114,6 +114,12 @@ Essentially, all the items you would set up in the config file are set in a sing
 
 There is one special type of test that requires a data setup script to be run beforehand; this is the coverage contract test. This test runs through every list of pages for a set of contracts, but the urls for hitting all the pages of whatever contracts need to be set up before the test runs. This can be done by calling the data setup script common/write_contract_cursors.py similar to this:
 
+> **Note:** You _must_ generate the contract URLs via the below script if you are running any of the following test suites:
+> * `v1/patient_test_coverageContract.py`
+> * `v1/baseline_tests.py`
+> * `v2/patient_test_coverageContract.py`
+> * `v2/baseline_tests.py`
+
     python3 ./common/write_contract_cursors.py --contracts="Z9997,Z9998,Z9999" --year="2020" --month="01" --count="5" --version="v2"
 
 The script will write out the results to homePath specified in the config (and expect them there during runtime.)
--- a/ops/ccs-ops-misc/load_test/common/url_path.py
+++ b/ops/ccs-ops-misc/load_test/common/url_path.py
@@ -1,9 +1,10 @@
 import time
+from typing import Dict
 from urllib.parse import urlencode
 
 '''
 Creates a query path from a base path (i.e. /v2/fhir/Coverage) and a dictionary of query parameters
 '''
-def create_url_path(path: str, query_params: dict[str, str]):
+def create_url_path(path: str, query_params: Dict[str, str]) -> str:
   params = urlencode(query_params)
   return "{}?{}".format(path, params)
--- a/ops/ccs-ops-misc/load_test/common/url_path.py
+++ b/ops/ccs-ops-misc/load_test/common/url_path.py
@@ -1,0 +1,9 @@
+import time
+from urllib.parse import urlencode
+
+'''
+Creates a query path from a base path (i.e. /v2/fhir/Coverage) and a dictionary of query parameters
+'''
+def create_url_path(path: str, query_params: dict[str, str]):
+  params = urlencode(query_params)
+  return "{}?{}".format(path, params)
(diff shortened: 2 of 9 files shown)


ops/ccs-ops-misc/load_test/README.md
```diff
@@ -70,24 +70,34 @@ The tests are run with parameters that specify all the test params and test file
...
 **--homePath** : (Required) : The path to your home directory on the local box, such as /home/logan.mitchell/
 
 **--clientCertPath** : (Required) : The path to the PEM file we copied/modified earlier. Should be located in your home directory like this: ~/bluebutton-backend-test-data-server-client-test-keypair.pem
 
-**--databaseHost** : (Required) : The database host, used to pull data. Get this from the database endpoint we noted in the earlier step "Getting The Database Instance"
-
-**--databaseUsername** : (Required) : The username for connecting to the database, can be found in Keybase. Make sure to use the correct one for the environment you're connecting to.
-
-**--databasePassword** : (Required) : The password for connecting to the database, can be found in Keybase. Make sure to use the correct one for the environment you're connecting to.
+**--databaseUri** : (Required) : The URI used for connecting to the database server. Needs to include username, password (both can be found in Keybase, make sure to use the correct one for the environment you're connecting to), hostname and port.
```
lsmitchell, line 93: Thanks for fixing this!

ops/ccs-ops-misc/load_test/README.md
```diff
@@ -70,24 +70,34 @@ The tests are run with parameters that specify all the test params and test file
...
 
 **--testHost** : (Required) : The load balancer or single node to run the tests against. The environment used here should match the same environment (test, prod-sbx, prod) as the database, so we pull the appropriate data for the environment tested against. Note that when using a single node, you should specity the Ip AND the port for the application.
 
 **--testFile** : (Required) : The path to the test file we want to run.
 
+**--configPath** : (Optional) : The path to a YAML configuration file that will be read from (but not _written_ to) for the values specified here. The values in this configuration file will be merged with values from the CLI, with the CLI values taking priority. The resulting merged values will then be written to the root `config.yml` file. If not provided, defaults to `config.yml` (the root YAML configuration file).
```
lsmitchell, line 99: A little specific but it may be written to, if you use the root path or use the default, right? We read from this which may also be the place we write to. Not sure if we can reword this to be more specific, or if its worth doing so, since its relatively clear as is.

ops/ccs-ops-misc/load_test/common/url_path.py
```diff
@@ -0,0 +1,13 @@
+import time
```
lsmitchell, line 1: Is this being used?

ops/ccs-ops-misc/load_test/common/url_path.py
```diff
@@ -0,0 +1,13 @@
...
+from typing import Dict
+from urllib.parse import urlencode
+
+'''
+Creates a query path from a base path (i.e. /v2/fhir/Coverage) and a dictionary of query parameters
+'''
+def create_url_path(path: str, query_params: Dict[str, str] = {}) -> str:
```
lsmitchell, line 8: Huh didnt know you could specify types for params in python, I like it.

ops/ccs-ops-misc/load_test/common/validation.py
```diff
@@ -9,13 +9,21 @@
...
     SLA_PATIENT : [1000, 3000, 5000, 8000],
     SLA_EOB_WITH_SINCE : [100, 250, 1000, 3000],
-    SLA_EOB_WITHOUT_SINCE : [500, 1000, 3000, 6000]
+    SLA_EOB_WITHOUT_SINCE : [500, 1000, 3000, 6000],
+    # The following values were selected for 5 concurrent users against
+    # a target c5xlarge AWS instance running bfd-server 
+    SLA_V1_BASELINE : [10, 325, 550, 10000],
+    # The following values were selected for 3 concurrent users against
+    # a target c5xlarge AWS instance running bfd-server
+    SLA_V2_BASELINE : [10, 325, 550, 10000],
```
lsmitchell, line 26: these will eventually be pulled from the new relic recorded average or whatever we use for metrics; maybe add a TODO here

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
```diff
@@ -0,0 +1,131 @@
...
+server_public_key = setup.loadServerPublicKey()
+setup.disable_no_cert_warnings(server_public_key, urllib3)
+setup.set_locust_env(config.load())
+
+mbis = data.load_mbis()
+eob_ids = data.load_bene_ids()
+last_updated = data.get_last_updated()
+cursor_list = data.load_cursors("v1")
+
+class BFDUser(HttpUser):
```
lsmitchell, line 22: Could you add some documentation for this class please?

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
```diff
@@ -0,0 +1,131 @@
+import random
```
lsmitchell, line 1: Maybe we want to call this the regression suite or something instead of baseline tests? The baseline will be used to validate them, but they are intended to be used as regression tests. (same with the v2 name)

lsmitchell (commented):
Great work! A couple minor things, but this is looking really nice.

ops/ccs-ops-misc/load_test/common/url_path.py
malessi: Good catch! I must have accidentally imported that automatically when writing this, I'll remove it.

ops/ccs-ops-misc/load_test/common/url_path.py
malessi: Yeah, I didn't either until working on this. Albeit I think they're just type hints opposed to actual compiler enforced types, so I think a caller could still pass a non-`str` type for `path` or a `list` for `query_params` and you there'd be no complaints from Python until runtime.

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
```diff
@@ -0,0 +1,131 @@
...
+        copied_cursor_list = cursor_list.copy()
+        random.shuffle(copied_cursor_list)
+
+        self.eob_ids = copied_eob_ids
+        self.mbis = copied_mbis
+        self.cursor_list = copied_cursor_list
+
+    def get_eob(self) -> int:
+      if len(self.eob_ids) == 0:
+            errors.no_data_stop_test(self)
```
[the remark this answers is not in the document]
lsmitchell: This is a good point; for instance we have many fewer contract data pieces than generic bene ids; this means the test may end earlier than we want if we equally weight all tests.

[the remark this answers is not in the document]
andrew-kramer: Also, I think we probably consume those ids faster for some tests than others. There are like five or six that use bene_ids, a couple that use contract ids, and a couple more with MBIs. It may also depend on how quickly the requests come back, since we're firing all of them as quickly as we can.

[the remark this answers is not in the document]
malessi: This is a good point, and I didn't really consider that the list of contract URLs is less than the list of `bene_id`s, so in a prolonged test run of the suite we may run out of contract URLs to test against. As it stands, I was seeing approximately 25 requests per second per concurrent user for V2 and about double that for V1, so to your point @andrew-kramer over a 3 minute run we end up only sending about 7500 to 10000 requests _total_.

Note that each Locust `User` _copies_ the initial list of each type of input data (MBIs, `bene_id`s and contract URLs) in its `on_start(...)`. This means each concurrent user has their own set of independent data, by the way.

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
```diff
@@ -0,0 +1,131 @@
...
+import common.validation as validation
+from locust import HttpUser, task, events, tag
+
+client_cert = setup.getClientCert()
+server_public_key = setup.loadServerPublicKey()
+setup.disable_no_cert_warnings(server_public_key, urllib3)
+setup.set_locust_env(config.load())
+
+mbis = data.load_mbis()
+eob_ids = data.load_bene_ids()
```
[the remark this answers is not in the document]
malessi: Yeah, I'm definitely in favor of that. I'll go ahead and rename all `eob_ids` to `bene_ids` and also rename `get_eob_id()` to `get_bene_id()`, too.

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
malessi: Yeah, I think that's a better name for these suites. I'll rename them to the following:

* `v1/baseline_tests.py` -> `v2/regression_suite.py`
* `v2/baseline_tests.py` -> `v2/regression_suite.py`

ops/ccs-ops-misc/load_test/common/validation.py
malessi: Sounds good, I wanted to add some context as to why I chose those SLAs. I'll add a `TODO` above the dictionary here stating that we'll eventually be pulling these numbers from metrics.

ops/ccs-ops-misc/load_test/README.md
malessi: Yep, you're correct. If `configPath` is `./config.yml` (the default config path, or the root configuration file) we will read _and_ write to it. If the `configPath` is _not_ `./config.yml` (some other YAML file) then we will _read_ from `configPath` and then write to `./config.yml` so that the file at `configPath` isn't overwritten unexpectedly.

I think I'll take a variation of what I wrote above here and rewrite that explanation, because looking at it now it's definitely a bit confusing.

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
malessi: Yeah, I'll add some documentation to both regression suites and to their class-level methods.

ops/ccs-ops-misc/load_test/runtests.py
```diff
@@ -91,48 +99,60 @@ class configData: pass
...
             workerThreads = arg
         elif opt == "--resetStats":
-            configData.resetStatsAfterClientSpawn = True
+            configData["resetStatsAfterClientSpawn"] = True
         else:
             print(helpString)
             sys.exit()
 
+    ## Read the specified configuration file and discard any falsey key
+    storedConfigData = {k: v for k, v in config.load_from_path(configData["configPath"]).items() if v}
```
andrew-kramer, line 130: I get a KeyError here if I run without a configPath set, but configPath isn't listed as a required field. We should do one or the other -- my opinion is, it should probably not be required.

malessi: Good catch! That's definitely a mistake on my end. I intended it to be optional, I just didn't catch that I was reading from `configData` prior to `configData["configPath"]` being set by merging with `defaultConfigData`. 

I've fixed this so that we use `Dict.get(...)` to fallback to `defaultConfigData["configPath"]` if `--configPath` was not passed-in. I've tested this locally and `--configPath` is now optional as expected.

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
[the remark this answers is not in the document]
andrew-kramer: Important note: in my own testing, if you run a Locust Task and it never makes a request to the API (i.e. you return immediately or after discovering that there's no data), then it doesn't count that run in the statistics. So it would be ok to do this:

```python
@task
def some_test(self):
    if self.bene_ids:
        # There is some data in self.bene_ids, so run the test
        self.client.get("url", params={ "_id": self.bene_ids.pop() })
    else:
        # There is no data, so just return and this test won't count
        return
```

If there is a bene_id there, it will run and count the test. If not, then the `if` will fail and the test will return without calling `client.get`. Locust will run other tests for which there is data until it runs out of that, too.

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
malessi: @lsmitchell I've added documentation comments to both files; let me know if you were thinking of adding external documentation or if you think that the file comments I've added are sufficient.

lsmitchell (dismissed):
Nice work!

ops/ccs-ops-misc/load_test/v1/baseline_tests.py
lsmitchell: That should be fine, just wanted some small explanation of what the tests are doing in the file.

christrotteradhoc:
This looks really great! Nice improvements here. Do you have any data on the stability of the tests from one run to the next that can help inform how much variation is normal? I think a section in the README or somewhere else that explains the steps and gives guidance on evaluating the results of a performance regression run would be really helpful.

malessi:
> 

Yep, I do have some data (specifically for the V2 suite, but I have no reason to believe that the V1 suite would be any less consistent). [Here's a Google Sheet](https://docs.google.com/spreadsheets/d/1ZWFTwaN7TXDRxqqZ_dM_8ZRTl2uIS6337S7xIwaynms/edit?usp=sharing)
 with the statistics from 3 back-to-back test runs against the same target and with the same values (users, spawn rate, runtime). 

I've created charts comparing responses/second, 50%s and total average response times across the 3 runs. It's apparent that the results are _very_ consistent once the target machine is warmed-up (runs 2 and 3 have near-identical data) whereas a "cold" run tends to be slower in the beginning.

Here's the charts (I don't know why the first run has less data-points, the run didn't fail and it ran for the same amount of time as runs 2 and 3):

Combined Responses_Second for V2 Regression Suite Against c5xlarge Target

Combined 50%s for V2 Regression Suite Against c5xlarge Target

Combined Total Avg Response for V2 Regression Suite Against c5xlarge Target

malessi:
> You might want to switch the data acquisition sql back to TABLE_SAMPLE and see the results; the holdover was to pull a static number of records from the DB which means your second and third tests might be using cached data. (We want to use table sampled results anyway, and I think the thing that was preventing us from doing so has been fixed now)

That's a good idea, the results definitely seem like some form of caching on the target machine and/or database. I'll use `TABLESAMPLE` for as much of the data as I can and see if anything changes.

christrotteradhoc:
@malessi thanks for providing that data. The information that I am referring to is guidance on how much variation is expected from run to run for the purpose of correctly evaluating the absence of regression. However if this is going to be included in another story then that works for me.

malessi:
> @malessi thanks for providing that data. The information that I am referring to is guidance on how much variation is expected from run to run for the purpose of correctly evaluating the absence of regression. However if this is going to be included in another story then that works for me.

Gotcha -- given now that we may not want to use `TAMPLESAMPLE`ing when running the regression suite (whereas we'd probably want it for the _load tests_) I'd say that those charts are fairly representative of the variation expected between subsequent runs using _"known" input data_ (same bene IDs/MBIs/contact URLs from run-to-run). 

However, I'd like to gather more data and have a method of storing and loading said data before adding any documentation about variance here -- at least until [BFD-1688](https://jira.cms.gov/browse/BFD-1688) is done (more likely whenever we actually implement the PoC from 1688).

lsmitchell (approved):

andrew-kramer (approved):
Looks great! Thanks, Mitch!

mackec1 (approved):
The only thing that I wondered about was why we have two distinct code paths for version (v1 vs. v2) when the v[1|2] is the only diffference in the request URL.

malessi:
> 

This is going to be _mostly_ resolved by the changes in [BFD-1655](https://github.com/CMSgov/beneficiary-fhir-data/pull/1083).

andrew-kramer:
As Mitch mentioned, it's mostly resolved in BFD-1655. Also, there are a few important differences - at least one endpoint uses the v1/Patient/{id} format, whereas there's no test for that endpoint in v2.

christrotteradhoc (approved):

PR #1083, opened 2022-04-29
BFD-1655: Add paginated endpoints to the data pool

andrew-kramer:
**JIRA Ticket:**
[BFD-1655](https://jira.cms.gov/browse/BFD-1655)

As a BFD engineer, I want paginated endpoints to add any next page of a result to the data pool, so that performance tests can be expanded to test more pages of a response rather than only a single page like they do currently.

---

### What Does This PR Do?

Tests store and use next paginated URL.

* When a test encounters a paginated response, store the `next` URL in a pool.
* If there is no more initial data, test against paginated endpoints instead.
* Remove the old pre-fetched method of generating contract cursors, in favor of using the paginated responses at run-time.
* Refactor tests to separate URL creation from tests and standardize setup.
* Re-enable `TABLESAMPLE` in the database queries, which was necessary to ensure efficiency and randomization in the shared pools of `bene_id`.
* Lint load_test files.

*Apologies to the reviewers -- I tried to keep the scope smaller, but changes to the way we generate and utilize URLs were fundamental to the tests themselves. Creating a base class will simplify things in the future, but it definitely makes for a large commit now. The alternative -- adding 20-30 lines in each of the 20ish tests -- would have been even messier (in my opinion) and much more difficult to maintain. I linted the files I touched (which ended up being nearly all of them).*

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* You *could* merge this first, but this branch is based off of branch `alessio/BFD-1676__parallel-perf-test-suite` (see #1075). It would make way more sense to merge that first.

### Submitter Checklist


ops/ccs-ops-misc/load_test/v1/patient_test_id.py
```diff
@@ -1,43 +1,17 @@
...
 
-bene_ids = data.load_bene_ids()
-client_cert = setup.getClientCert()
-setup.set_locust_env(config.load())
 
-class BFDUser(HttpUser):
     @task
-    def patient_by_id(self):
-        if len(bene_ids) == 0:
-            errors.no_data_stop_test(self)
```
lsmitchell, line 20: I noticed this got abstracted; what happens now when a test runs out of benes? The way we handle running out of data is probably going to want to be different depending on if we're running the regression suite (where we probably want to continue with the other endpoints?) and a single test (where we want to stop the test).

lsmitchell (commented):
Very nice test refactor! Removes a lot of the boilerplate, which was actually one of the TODOs in the readme.

A few things, but overall looks good!

ops/ccs-ops-misc/load_test/common/db.py
```diff
@@ -59,6 +61,30 @@ def get_hashed_mbis(uri):
     return [str(r[0]) for r in _execute(uri, beneQuery)]
```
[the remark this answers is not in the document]
andrew-kramer: I saw that comment on Mitch's PR, and I totally agree!

And you mean remove `WHERE "mbi_hash" IS NOT NULL` from the `get_hashed_mbis` query?

ops/ccs-ops-misc/load_test/v1/patient_test_id.py
andrew-kramer: I just tested this, and it looks like the tests continue for the allotted time without sending any requests. Obviously not really what we want. I'll correct that.

ops/ccs-ops-misc/load_test/common/db.py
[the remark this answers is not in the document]
lsmitchell: The commented code in get_bene_ids would replace the where clause (line 41). Anywhere else we are getting bene ids should also use tablesample so we can pull a randomized set of data during the query.

ops/ccs-ops-misc/load_test/v1/regression_suite.py
```diff
@@ -6,173 +6,140 @@
...
     def coverage_test_id_count(self):
-        self.get(f'/v1/fhir/Coverage', params={'beneficiary': f'{self.get_bene_id()}', '_count':'10'},
-                name='/v1/fhir/Coverage search by id / count=10')
+        def make_url(instance):
+            return create_url_path('/v1/fhir/Coverage', {
+                'beneficiary': instance.bene_ids.pop(),
+                '_count': '10'
+            })
+
+        self.run_task(name='/v1/fhir/Coverage search by id / count=10', url_callback=make_url)
```
malessi, line 27: I notice that this pattern of defining a `make_url` function which returns the string created by `create_url_path` is almost universal across all of the test methods. I'm wondering if we could reduce a lot of this boilerplate (since it's so ubiquitous) by adding an overload for `run_task` that does this and takes the parameters that `get` normally would, i.e.:

```python
# Name can be whatever you want
def run_url_path_task(self, path: str, query_params: Dict[str, str] = {}, headers: Dict[str, str] = {}, name: str):
  def make_url(instance):
     ...
  self.run_task(name=name, url_callback=make_url)
```
Then, this would become something more like:

```python
  self.run_url_path_task('/v1/fhir/Coverage', {
                'beneficiary': instance.bene_ids.pop(),
                '_count': '10'
            }, name='/v1/fhir/Coverage search by id / count=10')
```

ops/ccs-ops-misc/load_test/common/bfd_user_base.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+    # If a child class is going to use a set of data during the test, such as 
+    # Beneficiary IDs, MBI numbers, etc. then it has to be defined in this class
+    # field so that we can pre-load that data from the database.
+    DATA_REQUIRED = [
+        # 'BENE_IDS',
+        # 'MBIS',
+        # 'CONTRACT_IDS'
+    ]
```
malessi, line 30: _super_ nit: Instead of strings here, maybe we could use an [enum](https://docs.python.org/3/library/enum.html)? I don't know if that would be more "pythonic", but it would potentially let us type hint this field (as something other than `str[]`) and allow for autocomplete and correctness checking. Also would eliminate the possibility of typoing one of the three data types and then wondering why the list of data is empty.

> Not sure if we do "nits" here, but in the past I've always prefixed comments on reviews that I'm fine with being ignored with "nit:".

ops/ccs-ops-misc/load_test/common/bfd_user_base.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def load_data(self, flag_name: str, load_function: Callable, *args) -> List:
+        '''If the given DATA_REQUIRED flag exists on this class, pre-load the
+        data from the database.'''
+        if hasattr(self, 'DATA_REQUIRED') and flag_name in self.DATA_REQUIRED:
+            data_list = data.load_data_segment(load_function, *args).copy()
+            random.shuffle(data_list)
+            return data_list
+        else:
+            return []
```
malessi, line 146: I _think_ there's a difference in behavior with how these changes load the data from the database with respect to how the current code does it. Before, the code would load the data _once_ from the database regardless of how many concurrent Locust `User`s there were, whereas with these changes it seems that each Locust `User` would independently query the database to get its data.

I'm not sure this is something we should be concerned with, _especially_ if we start using `TABLESAMPLE` and want to make sure our input data isn't being cached as much as possible, though.

malessi (commented):
Seconding Logan's comment: these changes look great! I'm happy that a lot of the code duplication is being reduced, _and_ that some of the tests are becoming less complicated to setup (i.e. the contract tests).

I also have a few things to add, but none of it is necessary for you to address in order for me to approve these changes.

ops/ccs-ops-misc/load_test/common/bfd_user_base.py
lsmitchell: Indeed, with the distributed mode it would load once and split the data amongst the users. The data gathering part of the tests is expensive (in terms of both database weight and time) so we'd ideally not want to be calling for data multiple times, especially when the early users may have started hitting the service as we're still querying data. it would be ideal to gather the data once, if indeed we are doing it multiple times here.

andrew-kramer: I moved the loading of those IDs and the code that relies on them into children of BFDUserBase. It should greatly reduce the amount of test duplication also -- each test should now be defined in a single place.

This sets us up for making the tests run based solely on Locust tags, but I'll leave that for a future item if we decide to do it.

andrew-kramer: I've checked pretty thoroughly, and I'm pretty sure every worker thread is loading up its own separate database query. With `TABLESAMPLE` enabled, it's almost guaranteed to be totally different results. Personally, I think that's sufficient, and my preference is to rely on that for the `bene_id` and `hashed_mbi`. I'll take out the part that splits up the results between workers, because we won't even have the same results, so why discard perfectly useful `bene_id`s?

It's a little different for the contracts -- there are only a few, so we end up with the same data in each thread. They're randomized, but I think we'll still hit the same endpoints multiple times. For these, it probably is worthwhile to divide the results between the workers.

I'm definitely open to suggestions. I think if we want to truly guarantee separate datasets for the different threads, then we'd need to implement something with messaging between the master and worker threads. But I don't think anything like that is necessary for the testing we're doing so far.

ops/ccs-ops-misc/load_test/common/db.py
```diff
@@ -22,49 +23,76 @@ def _execute(uri, query):
...
     Return a list of bene IDs from the adjudicated beneficiary table
     """
-    beneQuery = (
+    bene_query = (
         'SELECT "bene_id" '
         'FROM "beneficiaries" '
-        # 'TABLESAMPLE SYSTEM (0.25) '  # when data is cleaned up, use this to access random benes
-        # there are tons of benes with null ref year which return 404 if we use them
+        f'TABLESAMPLE SYSTEM ({table_sample_pct}) '
         'WHERE "rfrnc_yr" IS NOT NULL '
```
lsmitchell, line 40: We shouldnt need the rfrnc_yr not null check here anymore; the application shoudnt error even if theyre null now.

lsmitchell (approved):
With regards to the table sample, maybe a followup cleanup item for after this PR, I think per Karl's input we may want to use table sample for the _load_ tests but not regression suite, where we might use a more curated section of data. We may want to have a way to control if we are running in table sample mode or curated sample mode.

Nice work, and thanks for the adjustments.

malessi (approved):
Seconding Logan's comment: great work on these changes! The updates you've made address everything I previously commented on.

ops/ccs-ops-misc/load_test/README.md
```diff
@@ -110,7 +110,37 @@ Essentially, all the items you would set up in the config file are set in a sing
...
+**--year** : (Required) : The year to search by for contracts. This will be applied to all contracts from --contracts. This should be in YYYY format.
+
+**--month** : (Required) : The month to search by for contracts. This will be applied to all contracts from --contracts. This must be in MM format (i.e. 01 for January, 05 for May, etc)
+
+**--count** : (Required) : How many results per page to request. A lower number means you'll get more pages with smaller sized results, higher number means less pages with larger results. This can help test various sized payloads, or give tests more urls to use.
+
+**--version** : (Required) : The endpoint version to setup the data for. The only valid values for this (at time of writing) are "v1" or "v2". The resulting file will be prepended with this version so that multiple cursor files can be set up for different endpoints.
+
+The script will output a file named after the --version in the configured home directory, where the contract tests are expecting it. Make sure you run the script again if you need to set up data for a different endpoint version.
+>>>>>>> be38e90f64c4c2e8b2dd378e9e5e3bed5227f065
```
malessi, line 143: Looks like there was a merge conflict here that slipped through -- I'm assuming that this section should be removed (excluding `--resetStats` documentation on line 114, obviously).

andrew-kramer: And I fail at VS Code... Sorry about that!

malessi (dismissed):
Everything looks good to me now!

lsmitchell (dismissed):

ops/ccs-ops-misc/load_test/common/bene_tests.py
```diff
@@ -0,0 +1,178 @@
...
+
+class BeneTestUser(BFDUserBase):
+    '''Locust tests that require a pool of beneficiary IDs.'''
+
+    # Mark this class as abstract so Locust knows it doesn't contain Tasks
+    abstract = True
+
+    # Should we use the Table Sample feature of Postgres to query only against a portion of the
+    # table?
+    USE_TABLE_SANPLE = True
```
mjburling, line 18: While consistent, I'm only curious. Is _sanple_ intentional throughout this changeset?

ops/ccs-ops-misc/load_test/common/config.py
```diff
@@ -1,73 +1,85 @@
...
-    configFile.write("clientCertPath: \"%s\"\n" % fileData["clientCertPath"])
-    configFile.write("serverPublicKey: \"%s\"\n" % fileData["serverPublicKey"])
-    configFile.write("dbUri: \"%s\"\n" % fileData["dbUri"])
-    configFile.write("testHost: \"%s\"\n" % fileData["testHost"])
-    configFile.write("testRunTime: \"%s\"\n" % fileData["testRunTime"])
-    configFile.write("testNumTotalClients: \"%s\"\n" % fileData["testNumTotalClients"])
-    configFile.write("testCreatedClientsPerSecond: \"%s\"\n" % fileData["testCreatedClientsPerSecond"])
-    configFile.write("resetStatsAfterClientSpawn: \"%s\"" % fileData["resetStatsAfterClientSpawn"])
-    configFile.close()
+def save(file_data):
```
mjburling, line 6: I realize this is an improvement over what already existed, but I think this function could be made a _little_ safer without much work:
1. I think the zeitgeist (_or_ [PEP-343](https://peps.python.org/pep-0343)) would have you wrap the file i/o `open(...)` in a `with` block, e.g.
```python
with open('config.yml', 'w', encoding='utf-8') as config:
    # do stuff with the scoped `config` variable, but worry not about
    # having to explicitly issue `.close()`
```
Otherwise, I think you might consider wrapping it in a try/except/finally or simply try/finally to always close the file.

2. I don't know if I've ever really been _in_ the python game, but last time I checked, `ruamel` seemed to be a feature-complete for yaml marshalling/unmarshalling. I would think that `PyYAML==6.0` is also capable. I'd imagine something like the following would be more auspicious than manually coercing some string into a yaml config file, e.g.

```python
def save(file_data): # assuming file_data is indeed a dict
    with open('config.yml', 'w') as config:
        yaml.dump(file_data, config, default_flow_style=False)
```

ops/ccs-ops-misc/load_test/common/bfd_user_base.py
```diff
@@ -0,0 +1,183 @@
...
+        paginated URL Pool.
+        '''
+
+        # First, see if we can generate a URL using the callback
+        try:
+            url = url_callback()
+        except IndexError:
+            url = None
+
+        if url is None and name in self.url_pools and self.url_pools[name]:
```
mjburling, line 122:
```suggestion
        if url is None and self.url_pools.get(name):
```

ops/ccs-ops-misc/load_test/common/bfd_user_base.py
```diff
@@ -0,0 +1,183 @@
...
+    def on_stop(self):
+        '''Run tear-down tasks after the tests have completed.'''
+
+        # Report the various response time percentiles against the SLA
+        if hasattr(self, 'VALIDATION_GOALS') and self.VALIDATION_GOALS:
+            validation.check_sla_validation(self.environment, self.VALIDATION_GOALS)
+
+
+    def get_by_url(self, url: str, headers: Dict[str, str] = None,
+            name: str = ''):
```
mjburling, line 77: Curious about the explicit `None` here instead of an empty dictionary. That would save us the extra conditional one-liner below involving `safe_headers`, but... maybe I'm not seeing the whole picture.
```suggestion
    def get_by_url(self, url: str, headers: Dict[str, str] = {},
            name: str = ''):
```

mjburling (commented):
I gave it a once-over for ~60 minutes and I came up with a few comments/suggestions. I'm trying to avoid purely stylistic feedback so far. I'll take another pass at this as time permits later this afternoon.

ops/ccs-ops-misc/load_test/common/bfd_user_base.py
andrew-kramer: This one came from the linter, but after doing some research, it looks like it's a safety issue. If you use a default empty dictionary and then make changes to the dictionary, then subsequent invocations can/will receive the modified dictionary instead of a new empty one.

https://stackoverflow.com/questions/5587582/numpy-array-subclass-unexpedly-shares-attributes-across-instances

ops/ccs-ops-misc/load_test/common/bene_tests.py
andrew-kramer: Ugh... I guess I just cam't spell.

ops/ccs-ops-misc/load_test/common/config.py
andrew-kramer: You're exactly right. We're already using PyYAML in the load function, so I went with that.

ops/ccs-ops-misc/load_test/common/bfd_user_base.py
mjburling: No kidding? That's not very intuitive (to me) and probably pretty annoying to debug if encountered in the wild. Thanks for the explanation!

mjburling (commented):
Just a few questions/comments on docstrings and function names. This is looking a lot more pythonic. Thank you!

malessi (approved):
The changes since my last review look good to me!

mjburling (approved):
🍰

PR #1092, opened 2022-05-03
PACA-559: Add retry logic to handle InterruptedException during gRPC channel shutdown.

brianburton:
**JIRA Ticket:**
[PACA-559](https://jira.cms.gov/browse/PACA-559)

Clean up the gRPC errors in the logs during a deployment.

When the ETL pipeline process is being shut down we are seeing stack traces in the logs from `InterruptedException`s inside of the `GrpcRdaSource#close` method. These exceptions happen when a thread is sent an interrupt by the pipeline manager to trigger a shutdown. Normally the interrupt is caught inside of the RdaSink objects and handled cleanly there but under some circumstances they can instead be caught while waiting for the gRPC connection to be fully closed.

---

### What Does This PR Do?

Adds logic to catch `InterruptedException`s thrown by `ManagedChannel#awaitTermination` method inside of the `GrpcRdaSource#close` method.  The first such exception triggers a retry which should succeed since the interrupted flag on the current thread would be cleared by the first call.  As a precaution, if the exception is thrown again during the retry the close method falls back to calling `shutdownNow` and exiting.  This last resort is the recommended course of action so it is a safe ultimate fallback.

* Are the retry actions logically sound.
* Are the tests sufficient to verify the logic works as expected.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
```diff
@@ -262,7 +266,18 @@ public void close() throws Exception {
         channel.shutdown();
       }
       if (!channel.isTerminated()) {
-        channel.awaitTermination(5, TimeUnit.SECONDS);
+        try {
+          channel.awaitTermination(1, TimeUnit.MINUTES);
```
ericbuckley, line 270: is this wait period just for the sinks to finish writing to the database?

ericbuckley, line 270: is this wait period just for the sinks to finish writing to the database?

brianburton: No this is waiting for the gRPC stub to finish closing its connection to the RDA API server and shut down any worker threads it was using.

brianburton: I bumped the timeout while making this change because I felt 5 seconds was just too little.  The choice of one minute was arbitrary but is well within the time that the pipeline scheduler allows jobs to consume.  The actual time taken will generally be much shorter since the call will return as soon as the shutdown is completed.

ericbuckley (approved):

KyRobbins (approved):

lsmitchell (approved):

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
```diff
@@ -262,7 +266,18 @@ public void close() throws Exception {
...
+        } catch (InterruptedException ex) {
+          LOGGER.info("caught InterruptedException while closing ManagedChannel - retrying once");
+          try {
+            channel.awaitTermination(1, TimeUnit.MINUTES);
+          } catch (InterruptedException ex2) {
+            LOGGER.info(
+                "caught second InterruptedException while closing ManagedChannel - calling shutdownNow");
```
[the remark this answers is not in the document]
brianburton: A warn could unecessarily trigger an alarm and this situation is harmless.  I'm just noting it in the log so we can look for it in case it happens often.

PR #1134, opened 2022-05-26
BFD-1600: Add bfd-db-migrator launch playbook

mjburling:
**JIRA Ticket:**
[BFD-1600](https://jira.cms.gov/browse/BFD-1600)

Introduce the ansible portions of the bfd-db-migrator infrastructure configuration.

---

### What Does This PR Do?
This _largely_ adds the "launch" playbook (and dependency), but there are some minor adjustments to the "build" playbook and migrator role that seemed appropriate to introduce at this time.

#### Future Work
Ultimately, this is still begging to be tidied up further, but the `bfd-db-migrator` infrastructure has been implicitly re-prioritized over the past six weeks several times and it's badly needed. I've tried to keep the ansible for the migrator and its respective playbooks similar in shape to the existing services, `bfd-pipeline` and `bfd-server`, however there are some innovations that should probably be applied throughout:
1. The split between build and launch ansible playbooks doesn't appear to be necessary–the [play tagging](https://docs.ansible.com/ansible/2.9/user_guide/playbooks_tags.html) is largely implemented already to achieve a similar split in functionality between `pre-ami` and `post-ami` tags. It wouldn't take much more than some careful consideration a fair amount of redundancy.
2. There are no associated _notifies_ with the `Restart awslogs Service` handler, but it's still included. What's more, the reference to `awslogs` is available in Amazon Linux instead of today's `awslogsd` for Amazon Linux 2. I've excluded it from the `bfd-db-migrator` solution, but I've left it alone in the `bfd-pipeline` and `bfd-server` services.
3. The cloudwatch configuration at the playbook level is somewhat awkward in forcing us to encode redundant configuration to the _launch_ playbook. This could be tidier and more obvious if it were packaged in the role itself.

Additionally, the terraform solution is still forthcoming, but both that in-progress solution and the changes in this PR have been successfully leveraged against the `test` environment. After this exists in the trunk, the terraform solution and adjustments should follow shortly thereafter. Full utilization of this service should be implemented shortly after that with updates to the Jenkins stage definitions.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

    5. There isn't too much of a burden on reviewers.
    6. Any problems it causes have a small "blast radius".
    7. It'll be easier to rollback if that becomes necessary.

### DIFF
--- a/ops/ansible/roles/bfd-db-migrator/defaults/main.yml
+++ b/ops/ansible/roles/bfd-db-migrator/defaults/main.yml
@@ -2,7 +2,6 @@
 env: test
 db_migrator_dir: /opt/bfd-db-migrator
 db_migrator_user: bb-migrator
-db_migrator_jvm_args: -Xmx64g
 db_migrator_tmp_dir: /tmp
 migrator_monitor_enabled: true
 migrator_monitor_heartbeat_interval_seconds: 300
--- a/ops/ansible/playbooks-ccs/build_bfd-db-migrator.yml
+++ b/ops/ansible/playbooks-ccs/build_bfd-db-migrator.yml
@@ -9,8 +9,8 @@
   gather_facts: yes
   vars:
     ansible_ssh_pipelining: no
-    env: "test"
     bfd_migrator_dir: /opt/bfd-db-migrator
+    env: "test"
 
   tasks:
     - name: Include global variables
--- a/ops/ansible/playbooks-ccs/build_bfd-db-migrator.yml
+++ b/ops/ansible/playbooks-ccs/build_bfd-db-migrator.yml
@@ -10,6 +10,7 @@
   vars:
     ansible_ssh_pipelining: no
     env: "test"
+    bfd_migrator_dir: /opt/bfd-db-migrator
 
   tasks:
     - name: Include global variables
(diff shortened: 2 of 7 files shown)


ops/ansible/roles/bfd-db-migrator/defaults/main.yml
```diff
@@ -2,7 +2,6 @@
 env: test
 db_migrator_dir: /opt/bfd-db-migrator
 db_migrator_user: bb-migrator
-db_migrator_jvm_args: -Xmx64g
```
mackec1, line 5: curious as to why you removed the mem config.

mjburling: tl;dr; It's unused.

The values and shape of the `bfd-db-migrator` role were initially copied from `bfd-pipeline` because the migrator is going to be taking over for some of the functionality that the pipeline is currently doing, so I reasoned that the solution for its role would be similar in many ways. As it turns out, the pipeline doesn't actually make use of that value, and because pipeline was effectively the template, it's not used in the migrator, either.

christrotteradhoc (approved):

PR #1165, opened 2022-06-10
BFD-1294 Add payload size to the structured access log (access.json)

christrotteradhoc:
**JIRA Ticket:**
[BFD-1294](https://jira.cms.gov/browse/BFD-1294)

Adds the http response payload size (in bytes) to the access.json, making it possible to deprecate and eventually remove the access.log file entirely once other dependencies are adjusted to no longer rely on access.log.

---

### What Does This PR Do?

BFD currently writes to two access logs -- access.log and access.json (which are respectively unstructured and structured) although the preference is to only have one log (the structured one). In order to remove the unstructured log, all data available in that log must be made available in the structured log file. Currently only the response payload size in bytes is missing from the structured log file. Making that available in the structured log file is the subject of this PR.

In order to accomplish that, this PR moves the code that performs the writes to access.json from a servlet filter (the RequestResponseLoggingFilter) to a new BFD implementation of the Jetty RequestLog interface. This interface has access to the underlying Jetty response object which exposes a method for retrieving the number of bytes in the http output payload which can then be added to access.json via the logging MDC. In order to preserve the unstructured log (which is needed for a time while its dependents are switched over to the structured log), additional code is included in the BFD implementation of the RequestLog interface to write the access.log since only one RequestLog interface may be registered with Jetty.

With the writing (logging) of the access.json no longer taking place in the RequestResponseLoggingFilter, it is renamed to the RequestResponsePopulateMdcFilter to more accurately represent its new purpose -- to populate request and response information in the MDC.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
@@ -20,19 +20,17 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
 /**
- * Ensure that every request-response pair lands in the application's NDJSON-formatted HTTP access
- * log. Also adds a bunch of data to the logging {@link MDC} and also ensures that the {@link MDC}
- * is completely cleared after every request. This {@link Filter} must be declared before all others
- * in the {@code web.xml}.
+ * Ensure that every request-response pair adds data to the logging {@link MDC} which will be
+ * present on all log messages, particularly the log messages written to the access log via Jetty.
+ * This {@link Filter} must be declared before all others in the {@code web.xml}.
  *
  * <p>(Note: We don't use or extend Logback's builtin <code>MDCInsertingServletFilter</code>, as it
  * includes more properties than we really need. It also doesn't fully clear the {@link MDC} after
  * each request, only partially.)
  */
-public final class RequestResponseLoggingFilter implements Filter {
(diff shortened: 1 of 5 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
```diff
@@ -43,18 +41,17 @@ public final class RequestResponseLoggingFilter implements Filter {
...
       throws IOException, ServletException {
+    /*
+     * This should be cleared by Jetty via the request log handling but just in case we have a situation where
+     * that handler does not fire (say, due to a Jetty defect) clear it now before the request starts.
+     */
+    MDC.clear();
+
+    handleRequest(request);
     try {
-      handleRequest(request);
```
lsmitchell, line 47: Is this not needed inside the try anymore? It seems to be the same call.

lsmitchell (approved):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
christrotteradhoc: the outer try was just to have the finally block that clears MDC. The MDC clearing is now done in DataServerLauncherApp.java and so the finally block was moved to that new location.

mjburling (approved):

PR #1235, opened 2022-07-19
PACA-698 (DRAFT): Fix RDA API since parameter starting value

brianburton:
**JIRA Ticket:**
[PACA-698](https://jira.cms.gov/browse/PACA-698)

The RDA API team made a mistake in their documentation of the "since" parameter. When specified, the API will return the next available sequence number. We need to adjust our logic such that when reading from the progress table, we use that value as is when constructing the since parameter, not N+1.

---

### What Does This PR Do?

Modifies `GrpcRdaSource` to use the correct starting sequence number for both resuming at a point in the database and for starting from a configured starting point.  Also modifies `RdaService` to mimic the behavior of the real API server by starting at the next sequence number after the `since` parameter value.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
@@ -236,7 +236,9 @@ public class GrpcRdaSource<TMessage, TClaim> implements RdaSource<TMessage, TCla
   /**
    * Uses the hard coded starting sequence number if one has been configured. Otherwise asks the
    * sink to provide its maximum known sequence number as our starting point. When all else fails we
-   * start at the beginning.
+   * start at the beginning. The RDA API assumes we are sending the highest sequence number that we
+   * already have so they will start at that plus 1. For a configured starting sequence number we
+   * subtract one but for a value from the database we can just pass it through unchanged.
    *
    * @param sink used to obtain the maximum known sequence number
    * @return a valid RDA change sequence number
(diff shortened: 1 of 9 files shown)


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/GrpcRdaSource.java
```diff
@@ -236,17 +236,19 @@ private void logOrRethrowDroppedConnectionException(
...
+   * subtract one but for a value from the database we can just pass it through unchanged.
    *
    * @param sink used to obtain the maximum known sequence number
    * @return a valid RDA change sequence number
    */
   private long getStartingSequenceNumber(RdaSink<TMessage, TClaim> sink)
       throws ProcessingException {
     if (startingSequenceNumber.isPresent()) {
-      return startingSequenceNumber.get();
+      return startingSequenceNumber.map(x -> Math.max(MIN_SEQUENCE_NUM, x - 1)).get();
```
ericbuckley, line 249: I'm wondering if it makes more sense to leave this as it was and make our variable align directly with how the RDA API is interpreting it.

brianburton: No the `Math.max()` call ensures the value can't be lower than zero.

KyRobbins (approved):

ericbuckley (approved):

mackec1 (approved):

PR #1242, opened 2022-07-20
BFD-1776: Create high volume load test suite

mharvey:
**JIRA Ticket:**
[BFD-1776](https://jira.cms.gov/browse/BFD-1776)

As a BFD engineer, I want a load test suite, so that the BFD team can better understand and manage production load.

### What Does This PR Do?

* Add `high_volume_suite.py`
* Add `config/config.high_volume_c5xlarge.conf`

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


ops/ccs-ops-misc/load_test/high_volume_suite.py
```diff
@@ -0,0 +1,374 @@
...
+
+    # See https://docs.locust.io/en/stable/extending-locust.html#test-data-management
+    # for Locust's documentation on the test data management pattern used here
+    global MASTER_BENE_IDS
+    MASTER_BENE_IDS = data.load_from_parsed_opts(
+        environment.parsed_options,
+        db.get_bene_ids,
+        use_table_sample=True,
+        data_type_name="bene_ids",
+    )
```
lsmitchell, line 35: We should keep in mind that adjustments to the data load method will affect both suites.  (I assume @malessi will be creating a new method to pull specific data especially for the regression suite, so we dont inadvertently affect this one in the future?)

lsmitchell:
Also checking that this disables the metrics gathering for this suite too, right?

ops/ccs-ops-misc/load_test/config/config.high_volume_c5xlarge.conf
```diff
@@ -0,0 +1,6 @@
+# Configuration to run high volume load tests
+
+run-time = "10m"
+users = 10
+spawn-rate = 5
+table-sample-percent = 0.5
```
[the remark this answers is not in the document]
mharvey: This is a misunderstanding of the units here on my part. I thought this was a percentage expressed as decimal, but it's raw percent, so this is 0.5%, not 50%. .5% is actually pretty quick. 😛 
Tweaking these numbers is for sure part of BFD-1783

ops/ccs-ops-misc/load_test/high_volume_suite.py
```diff
@@ -0,0 +1,374 @@
...
+    @task
+    def eob_test_id(self):
+        """Explanation of Benefit search by ID"""
+        self.run_task_by_parameters(
+            base_path="/v2/fhir/ExplanationOfBenefit",
+            params={"patient": self.bene_ids.pop(), "_format": "application/fhir+json"},
+            name="/v2/fhir/ExplanationOfBenefit search by id",
+        )
+
+    @tag("patient", "patient_test_coverage_contract", "v2")
```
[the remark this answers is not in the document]
mharvey: Yeah, it runs for 10 minutes. With the parameters as set, it gets out about 32,984 requests in that time.

malessi:
> Also checking that this disables the metrics gathering for this suite too, right?

The metrics gathering is opt-in via the `--stats-config` flag for any Locust test suite in this project.

As long as we don't pass `--stats-config` then nothing gets gathered.

ops/ccs-ops-misc/load_test/high_volume_suite.py
mharvey: We might need to make some adjustments to data loading that are specific to the high volume suite as well. We can decouple them when the time comes/as needed.

ops/ccs-ops-misc/load_test/high_volume_suite.py
```diff
@@ -0,0 +1,374 @@
...
+    @task
+    def patient_test_id_include_mbi_last_updated(self):
+        """Patient search by ID with last updated, include MBI"""
+        self.run_task_by_parameters(
+            base_path="/v2/fhir/Patient",
+            params={
+                "_id": self.bene_ids.pop(),
+                "_format": "application/fhir+json",
+                "_IncludeIdentifiers": "mbi",
+                "_lastUpdated": f"gt{self.last_updated}",
```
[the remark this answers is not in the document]
malessi: Yeah, this is still using the existing method of calculating `last_updated` via `data.py`'s `get_last_updated()`. Basically, it seems like `self.last_updated` will always be 2 weeks prior to the current date.

This is something that should be removed from `BFDUserBase` and handled per-suite depending on their needs.

@mharvey perhaps usages of `self.last_updated` in this new load suite should be replaced with some specific date like Logan's suggested? To be fair, I'm not sure _which_ date that should be, I think @lsmitchell would know better than I. I can handle removing `last_updated` from `BFDUserBase` and the required changes to the regression suite in a future ticket.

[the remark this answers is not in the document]
lsmitchell: yeah for either test its probably fine to use 1-3 weeks before 2022-07-11 (test and see how much comes back from that i guess) since thats when the last records were created/loaded, and im assuming thats the date that gets used for the record's last updated date.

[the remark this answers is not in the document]
mharvey: I handled this by overriding the value for `self.last_updated` in the `__init__` for `HighVolumeUser`

ops/ccs-ops-misc/load_test/config/config.high_volume_c5xlarge.conf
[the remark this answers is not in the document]
malessi: FYI, at this point I feel like these `config/*.conf` files are going to be left unused and probably removed entirely in the future. 

These `*.conf` files are a relic of what I believed _at the time_ to be the way we would pull configuration values for the upcoming regression suite pipeline integration. That was at least six weeks ago at this point, and with BFD-1936 laying the groundwork for using SSM parameters I think we will probably end up going that direction with this load suite when the time comes to integrate it into the pipeline.

Just providing some context, no need to worry about removing these just yet.

ops/ccs-ops-misc/load_test/high_volume_suite.py
```diff
@@ -0,0 +1,376 @@
...
+        data_type_name="contract_data",
+    )
+
+    global MASTER_HASHED_MBIS
+    MASTER_HASHED_MBIS = data.load_from_parsed_opts(
+        environment.parsed_options,
+        db.get_hashed_mbis,
+        use_table_sample=True,
+        data_type_name="hashed_mbis",
+    )
```
malessi, line 51: One thing we should probably think about is whether we want to load data _once_ or _per-`HighVolumeUser`_. 

If we load _once_ (as the high volume suite is now), we may want to think about distributing data in some manner that ensures each `HighVolumeUser` isn't just hitting the same endpoint (i.e. 2 `HighVolumeUser`s get a `list` of `bene_id`s and, since they're the same list, they start sending requests to the _same_ endpoints at the same time). I can think of one fairly naive, but _very_ easy solution and that is to [`shuffle()`](https://docs.python.org/3/library/random.html#random.shuffle) each `list` of data in `HighVolumeUser`'s `__init__()` _after_ copying, i.e.:

```python
self.bene_ids = MASTER_BENE_IDS.copy()
shuffle(self.bene_ids)
...
```

Shuffling should ensure that, although the _initial_ data is the same, different IDs are being used between `HighVolumeUser`s at all times. I guess we could also split data between `HighVolumeUser`s in some fashion, but that would probably get fairly complex for little benefit.

The other option is to load data per-`HighVolumeUser` during Locust's [`on_start()`](https://docs.locust.io/en/stable/writing-a-locustfile.html#on-start-and-on-stop-methods). With `TABLESAMPLE` this should all but guarantee each `HighVolumeUser` gets a distinct set of data. The downside to this would be that we incur some setup time per-`HighVolumeUser`, _but_ if we limited it to something relatively small (like 1000 rows, or perhaps base it upon runtime) then I don't believe this would be too bad at all.

malessi (commented):
There's a few small things I've seen, and a bit of a discussion item about the data loading strategy. Other than that, everything looks good!

ops/ccs-ops-misc/load_test/high_volume_suite.py
lsmitchell: I had some code in the run tests script to split the data amongst the users, but not sure how easy itd be to do with how things are generated now. Shuffle would at least help a little with avoiding caching, assuming we pulled enough data that we arent reusing too many of the same bene ids.

malessi: I think splitting the **global** (loaded once) data across `HighVolumeUser`s would be more complexity than it's worth; honestly, I don't think that'd get us anything that loading 1000 (we could scale this with `spawned_runtime`)  or so pieces of data per `HighVolumeUser` in `on_start()` wouldn't.

Personally, I'd lean toward trying to load small amounts of data per-`HighVolumeUser` first. If that's too slow, or if it's not what we want for some reason, then just load globally and shuffle. If that isn't getting us what we want, then I feel like it'd be worth the effort to explore properly splitting a larger dataset across multiple `HighVolumeUser`s.

mharvey: There may be some considerations around this based on how we wind up running the tests also.
It might not be too bad of an effort if we go with some small number of users per table sample and then just have it request another batch when it exhausts the current batch. (This could be terrible though! I haven't looked into exact implementation of that and if Locust would make it difficult.)

mharvey: I added the shuffle. Doesn't seem to have much of a perceptible impact on performance either way, at least not with a limited number of users (10), but seems fine to keep in.

lsmitchell (approved):

malessi (approved):
Looking good to me now! Thanks for addressing everything.

mjburling (approved):
Thanks, @mharvey. Looks just fine! 🎉

PR #1257, opened 2022-07-25
BFD-1936: Get Locust Regression Suite Running in Deployment Pipeline

malessi:
**JIRA Ticket:**
[BFD-1936](https://jira.cms.gov/browse/BFD-1936)

As a BFD engineer, I want to set up performance regression tests in Jenkins to run in CI/CD, so that they are able to automatically run post deploy in TEST. 

---

### What Does This PR Do?

This PR can logically be separated into the changes/additions made to the Locust tests directly and changes/additions made to infrastructure-as-code to support running the regression suite in the deployment pipeline:

#### Locust Tests Changes

* The naive method of adjusting total test runtime to account for user ramp-up has been replaced with a more robust solution using the `UserInitAwareLoadShape` [load shape class](https://docs.locust.io/en/stable/custom-load-shape.html)
  * In short, this "load shape" does not begin counting down the runtime until _all_ users have spawned 
    * This means that specifying a `spawned-runtime` of `30s` will guarantee that the tests will run for _at least_ 30 seconds. The test run may take longer depending on the time it takes to spawn the users
  * Locust's built-in `--run-time` parameter is **incompatible** with load shapes in general, and so `--spawned-runtime` was introduced to replicate its functionality. In almost all cases, `--run-time` can be replaced with `--spawned-runtime` 
  * All existing tests now use this load shape, and future Locustfiles should include a load shape class extending from it
* `pscyogreen` has been introduced as a new dependency in order to make `psycopg2` compatible with `gevent`, which Locust uses
  * Without `psycogreen` running in distributed mode is impossible as `psycopg2` operations are blocking and cause Locust workers to fail to heartbeat to the master Locust runner. In turn, the master runner marks the "zombie" workers as dead, and destroys them
  * A modest speed-up is also gained from the inclusion
* Fixes for several invalid database queries
* Overhaul of Python requirements
  * The `requirements.txt` has been split into `*requirements.in` and `*requirements.txt` files
  * The `*.in` file specifies the **top-level** dependencies for the project, whereas the `*.txt` files are compiled using [`pip-compile`](https://github.com/jazzband/pip-tools) from said `*.in` files.
    * Using `pip-compile` allows for easier understanding of which dependencies are direct dependencies of the project, and also gives us additional security by using `pip`'s [Hash Checking Mode](https://pip.pypa.io/en/stable/topics/secure-installs/)
    * If you're familiar with npm, think of `*requirements.in` as `package.json` and `*requirements.txt` as `package-lock.json`
  * Even more specifically, developer dependencies and application dependencies have been split into `dev-requirements.in` and `requirements.in`, respectively, along with their corresponding, compiled `*requirements.txt`s
    * Developer dependencies are those that are _not_ deployed with the application (in this case, to AWS Lambda) -- these are dependencies like formatters and linters 

#### Infrastructure Changes

* Terraform module has been introduced describing the `bfd-<env>-server-regression` Lambda and its components
* A small Jenkins pipeline for building and pushing Lambda Docker images of the Locust regression suite has been created under `bfd-server-regression-build`
  * The corresponding `Jenkinsfile`, `Dockerfile`, and more is found under `../load_test/lambda/server-regression/`
* Changes have been made to the main deployment pipeline `Jenkinsfile` to deploy the `server-regression` Lambda to `test`, `prod-sbx`, and `prod`, and to run the `server-regression` Lambda in `test` and `prod-sbx` (but _not_ `prod`, yet)

* Is more documentation required?
* Is new code easy to understand and readable?
* Do the changes make sense?

      * [x] Yes
      * [ ] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


ops/terraform/services/server/server-regression/backend.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  backend "s3" {
+    bucket         = "bfd-tf-state"
+    key            = "services/server/terraform.tfstate"
```
mjburling, line 4: I know our opinion of this changed a few times, but we've ultimately landed on a distinct states for `server` and `server-regression`. I'll migrate this.

ops/terraform/services/server/server-regression/docs/server-regression.plantuml
```diff
@@ -0,0 +1,49 @@
...
+participant "deployment pipeline"
+database "AWS ECR"
+participant "parameter store"
+participant "AWS Lambda"
+participant "sqs queue"
+note over "bfd-server-regression-build","deployment pipeline": distinct Jenkins pipeline jobs ran asynchronously
+operator -> github: Submit change set
+github --> operator
+operator -> "deployment pipeline": Trigger Deployment
+operator -> "bfd-server-regression-build": Trigger Image Build
```
mjburling, line 15: As discussed yesterday, I think it's important to illustrate that these are (at least currently) two distinct processes or sequences.

ops/ccs-ops-misc/load_test/lambda/server-regression/.dockerignore
```diff
@@ -0,0 +1,7 @@
+*
+!common/
+!config/
+!v1
+!v2
```
mharvey, line 5: `common` and `config` have a trailing slash, but `v1` and `v2` don't. Does this cause different behavior in the `.dockerignore`?

andrew-kramer (approved):
This looks great! A lot of really positive changes here. It's huge, but I think overall the changes to the test suite especially will make things run more as Locust intends.

I had a couple comments, but no action items.

StewGoin (approved):
Looks good to me, the new dependencies appear to be safe based on discussion and inspection.

ops/ccs-ops-misc/load_test/common/db.py
```diff
@@ -52,12 +57,19 @@ def get_hashed_mbis(uri: str, table_sample_pct: Optional[float] = None) -> List:
...
-        f'SELECT "mbi_hash" FROM "beneficiaries" {table_sample_text} WHERE "mbi_hash" IS NOT NULL '
+    # Handle possible hash collisions from duplicate hashes in beneficiaries_history
+    # by only taking MBI hashes that are distinct in both beneficiaries _and_
+    # benficiaries_history
+    mbi_query = (
+        f"SELECT beneficiaries.mbi_hash FROM beneficiaries {table_sample_text} "
+        "    INNER JOIN beneficiaries_history "
+        "        ON beneficiaries.mbi_hash = beneficiaries_history.mbi_hash "
+        "    GROUP BY beneficiaries.mbi_hash "
+        "    HAVING count(beneficiaries_history.mbi_hash) = 1 "
```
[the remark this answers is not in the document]
mjburling: I'd be curious of the performance of _this_ suggestion. Indeed, the alternative was taken because the original query performed poorly on non-trivially sized data sets–ceteris paribus, the original query takes greater than 7 minutes to return data in production.

ops/ccs-ops-misc/load_test/lambda/server-regression/.dockerignore
malessi: [There definitely is](https://stackoverflow.com/questions/17888695/difference-between-gitignore-rules-with-and-without-trailing-slash-like-dir-an):

>Yes, these rules are different.
>
>    * `/dir` will match a file, directory, link, anything named `dir`
>    * `/dir/` will match only a directory named `dir`
>    * `/dir/*` will match all files, directories and anything else inside a directory named `dir` (but not the `dir` directory itself).

This is definitely _confusing_ in the sense that these lines are not following a single convention for un-ignoring a directory. _However_, for now this will not cause any problems, and I would say that both styles are essentially interchangeable as far as what we are using the `.dockerignore` for.

ops/ccs-ops-misc/load_test/common/db.py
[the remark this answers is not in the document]
malessi: So @andrew-kramer after watching @mjburling test the proposed query and the existing query (that is slow on large datasets like `prod`) we found that the existing query is somewhat faster by about 2x (with a `LIMIT` of 100,00, like in the code, we saw the existing query return in 6s and the proposed query return in 22s).

In the future we are going to rip this out entirely and stick with a known set of contiguous beneficiary IDs once the data is in a better state. When that happens we should see _significantly_ better performance here, since it will just be a simple `SELECT`.

mjburling (approved):

PR #1264, opened 2022-07-28
BFD-1908 - pipeline support to denote synthetic data

mackec1:
**JIRA Ticket:**
[BFD-1908](https://jira.cms.gov/browse/BFD-1908)

The BFD pipeline needs to behave differently when loading synthetic data than it does when loading non-synthetic data. In particular, the functionality for filtering back-dated beneficiaries that was introduced in early 2022 to avoid some data inconsistencies should not be active during synthetic data loading but should be active for non-synthetic data loads. 

### What Does This PR Do?
Modify pipeline to disable back-dated beneficiary filtering for synthetic data based on a new XML attribute in the data set manifest; this optional attribute will have a default value of _false_ (not synthetic data).

- a new Java class, DataSetManifestFactory will be introduced that extends manifest data processing via an embedded XML Schema Definition (XSD); this is necessary since the current JAXB processing will trap malformed XML but does not perform XML data validity checking. This allows us to ensure manifest XML files are **BOTH** well-formed and valid. For example, invalid XML value for _syntheticData_="falsee" will now throw an Excepetion because the XSD expects a valid boolean'.
- All Java files that instantiated a DataSetManifest via contained JAXB unmarshalling code have been modified to use the DataSetManifestFactory instead.
- CcwRifLoadJob modified to log the value of the syntheticData XML attribute prior to processing a manifest file.
- unit and Integration Tests (IT) have been modified to use DataSetManifestFactory; in addition new tests have been added to explicitly check for malformed XML or invalid XML manifests.
- When the _syntheticData_ attribute is set to 'true', the BFD pipeline should disable back-dated beneficiary filtering for the data files referenced in the manifest regardless of the configuration parameter setting for back-dated beneficiary filtering.
- Current pipeline processing aborts if a manifest JAXB exception is encountered; in order to maintain current pipeline processing logic, a JAXBException will always be thrown from the DataSetManifestFactory if XML parsing error(s) occur. In the case of XML data validation, the JAXBException may wrap an underlying UnmarshallException or SAXParserException as a linked exception.

Replace me.

* Is user input properly sanitized & handled?
* Does this make any backwards-incompatible changes that might break end user clients?
* Can you find any bugs if you run the code locally and test it manually?

* Unit and Integration Tests adequately cover new functionality.
* Javadoc is provided for classes, methods, variables.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
```diff
@@ -113,6 +134,11 @@ public int getSequenceId() {
     return sequenceId;
   }
 
+  /** @return the {@link boolean} denoting if the data is synthetic {@link DataSetManifest} */
```
lsmitchell, line 137: Missing doc description

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.xml.XMLConstants;
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.Unmarshaller;
+import javax.xml.transform.stream.StreamSource;
+import javax.xml.validation.Schema;
+import javax.xml.validation.SchemaFactory;
+import org.xml.sax.SAXException;
+
+public class DataSetManifestFactory {
```
lsmitchell, line 14: Missing class doc

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
```diff
@@ -0,0 +1,100 @@
...
+  /** The {@link String} pipeline manifest XML Schema Definition (XSD). */
+  public static final String MANIFEST_XSD =
```
lsmitchell, line 16: Should this be externally loaded from an xml? Seems strange to have it in-code.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -578,6 +581,8 @@ private boolean isBackdatedBene(RifRecordEvent<?> rifRecordEvent) {
    * @return {@code true} if the bene should be filtered/skipped
    */
   private boolean isBackdatedBene(Beneficiary bene) {
+    // no filtering of synthetic data
+
```
lsmitchell, line 585: This seems like it was moved out, so should this comment be removed/moved?

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestTest.java
```diff
@@ -80,6 +137,34 @@ public void jaxbUnmarshallingForSampleB() throws JAXBException {
     assertEquals(RifFileType.BENEFICIARY, manifest.getEntries().get(0).getType());
   }
 
+  /**
+   * Verifies that {@link DataSetManifest} can be unmarshalled, as expected; the sample-d XML
+   * contains the syntheticData attribute which is set to true whci causes the data set to be
```
lsmitchell: minor typo "whci"

lsmitchell (commented):
Looks good to me overall, a few little things

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
mackec1: possibly...however BFD to date has no support for URL based file services. For this case, the XSD is so limited it is probably not worth the effort to have an XSD resource, available to both testing (build) and prod pipeline execution. In fact I debated whether to even do the right thing of using an XSD when I could have simply hard-coded checking the value of the 'syntheticData' attribute with a simple 'if' statement and throw a JAXBException if it wasn't 'true' or 'false'.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
mackec1: removed.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
mackec1: added class doc.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
christrotteradhoc: Can't we just load it from the classpath as a resource? Having this in the file system would allow us to pass it along to mitre and ccw if they ever wanted to validate the manifests that they generate, and we could add it to the synthetic automation as an extra layer of protection so that we don't blow up the pipeline with a bad manifest.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
```diff
@@ -0,0 +1,106 @@
...
+      SchemaFactory sf = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+      Schema schema = sf.newSchema(new StreamSource(xsdStream));
+      JAXBContext jaxbContext = JAXBContext.newInstance(DataSetManifest.class);
+      Unmarshaller jaxbUnmarshaller = jaxbContext.createUnmarshaller();
+      jaxbUnmarshaller.setSchema(schema);
+      manifest = (DataSetManifest) jaxbUnmarshaller.unmarshal(manifestStream);
+    } catch (SAXException e) {
+      throw new JAXBException(e.getMessage(), e);
+    } catch (Exception e) {
+      throw new JAXBException(e.getMessage(), e);
```
christrotteradhoc, line 102: I think we should avoid constructing JAXBExceptions ourselves -- it could be confusing for debugging purposes since most will assume that JAXBExceptions originate from JAXB.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
```diff
@@ -0,0 +1,106 @@
...
+   * @return the {@link DataSetManifest}
+   * @throws JAXBException (any errors encountered will be bubbled up).
+   */
+  public DataSetManifest parseManifest(String manifestFilename) throws JAXBException {
+    try {
+      InputStream manifestStream =
+          Thread.currentThread().getContextClassLoader().getResourceAsStream(manifestFilename);
+      return parseManifest(manifestStream);
+    } catch (Exception e) {
+      throw new JAXBException(e.getMessage(), e);
```
christrotteradhoc, line 75: I think we should avoid constructing JAXBExceptions ourselves -- it could be confusing for debugging purposes since most will assume that JAXBExceptions originate from JAXB.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
mackec1: Depends....I didn't write the AC, but here is the pertinent part:

The XML binding library should handle the error cases above for us by **throwing a JAXBException which is already handled in the BFD pipeline.**

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
mackec1: The pipeline is built to 'exit gracefully' (?oxymoron) on a JAXBException.....the AC was written to preserve that concept....

During unmarshalling, at least 3 diff exceptions are possible:

1. SAXException occurs when the syntheticData attribute value is validated vs. the XSD.
2. UnmarshallException when the unmarshall op fails
3. JAXBException - malformed XML, etc.

Since the first two are not currently handled by the pipeline AND the pipeline exception handling / shutdown is tied to look for a JAXBException, I left things as they were in the pipeline per the AC.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -436,8 +437,9 @@ private List<RifRecordLoadResult> process(
           Object recordInDb = entityManager.find(record.getClass(), recordId);
           timerIdempotencyQuery.close();
 
-          // Log if we have a non-2022 enrollment year INSERT
-          if (isBackdatedBene(rifRecordEvent)) {
+          // Log if we have a non-2022 enrollment year INSERT and we are
+          // NOT processing synthetic data.
+          if (!isSyntheticData && isBackdatedBene(rifRecordEvent)) {
```
christrotteradhoc, line 442: Since this condition it getting more complex and appears three times we may want to factor it out into a new well named method that encapsulates the concepts

```java
private boolean shouldBeSkipped(RifRecordEvent event) {
   // Don't skip any events when processing synthetic data
   if (event.getFileEvent().getParentFilesEvent().isSyntheticData())
      return false;
   else
      return isBackdatedBene(rifRecordEvent);
}
```

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
christrotteradhoc: It is not an AC, just an implementation note and it did not anticipate usage of XSD (which is a great idea btw) so we may need to tweak the caller to handle SAXException or wrap everything in some BFD flavor of exception.

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/RifFilesEvent.java
```diff
@@ -16,21 +16,25 @@ public final class RifFilesEvent {
   private final Instant timestamp;
   /** The {@link RifFileEvent}s contained in this {@link RifFilesEvent}. */
   private final List<RifFileEvent> fileEvents;
+  /** boolean indicator denoting if data is synthetic (true). */
+  private final boolean syntheticData;
```
brianburton, line 20: Sometimes using a boolean for something like this can become troublesome in the future.  It's also less clear at the point of call when looking at the parameter list to know what `true` means.  Using an enum for the type of data could allow more nuanced treatment later without much refactoring.  For now it could just have values like Synthetic and Genuine.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestTest.java
```diff
@@ -30,22 +34,78 @@ public void jaxbUnmarshallingForSampleA() throws JAXBException {
...
+                  Thread.currentThread()
+                      .getContextClassLoader()
+                      .getResourceAsStream("manifest-invalid-sample-a.xml");
+
+              DataSetManifestFactory.newInstance().parseManifest(manifestStream);
+            },
+            "SAXParseException message was expected");
+
+    UnmarshalException unmarshalException = (UnmarshalException) thrown.getLinkedException();
+    assertNotNull(unmarshalException);
```
brianburton, line 80: This would be a good place for `assertInstanceOf`:

```
assertNotNull(unmarshalException);
UnmarshalException unmarshalException = assertInstanceOf(UnmarshalException.class, thrown.getLinkedException());
```

That way if the exception is not an instance of `UnmarshalException` you'd get a meaningful error message instead of a `ClassCastException`.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
mackec1: loaded from file; however, it's not like we can actually publish this XSD since BFD is brain-dead for static file resources.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
mackec1: done.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestFactory.java
mackec1: changed RifLoader to handle both JAXBException and SAXException.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
mackec1: sort of; I do not want to exec _event.getFileEvent().getParentFilesEvent().isSyntheticData()_ as noted in your example. The processing is in a for loop, which would result is something being fetched (i.e., determination of synthetic data) hundreds and hundreds and hundreds of thousands of times, when we only need it once (cuz it'll be for the who parentFilesEnvent).

apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/RifFilesEvent.java
mackec1: Yea, agreed but not my call; AC specified boolean with the backwards compatibility such that if the syntheticData is missing, then it defaults to false. So yes, I could do the same with an enum, but thats adds mores stuff to the XSD.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestTest.java
mackec1: Brian, thx, I glossed over that and your idea is perfect. Implemented with the assertInstanceof

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -456,19 +455,16 @@ private List<RifRecordLoadResult> process(
...
+              // don't care about return status here...just want the logging
+              isBackdatedBene(rifRecordEvent);
             }
             tweakIfBeneficiary(entityManager, loadedBatchBuilder, rifRecordEvent);
             entityManager.persist(record);
           } else if (rifRecordEvent.getRecordAction().equals(RecordAction.UPDATE)) {
             loadAction = LoadAction.UPDATED;
             // Skip this record if the year is not 2022 and its an update.
-            if (isBackdatedBene(rifRecordEvent)) {
+            if (!isSyntheticData && isBackdatedBene(rifRecordEvent)) {
```
christrotteradhoc, line 467: I think with the logging moved into the isBackDatedBene() method, this will begin printing a message that was not there before and could be confusing since this is the UPDATE case and the logging references inserts.

mackec1: reverted logging back to orig.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
christrotteradhoc: ok, thats fine although I would be surprised if the java compiler wasn't smart enough to inline that function and figure that out for us.

christrotteradhoc (approved):

lsmitchell (approved):

PR #1266, opened 2022-07-29
BFD-1909: Create script for generating synthea characteristics file

lsmitchell:
**JIRA Ticket:**
[BFD-1909](https://jira.cms.gov/browse/BFD-1909)

As a BFD Engineer, I want to know the makeup and information about an automated synthea generation, so that I can better use the data created.

---

### What Does This PR Do?

Adds a script which runs queries and exports a csv describing any number of benes between two bene_ids and the number of contracts and claim types associated with those benes.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/ops/ccs-ops-misc/synthetic-data/scripts/generate-characteristics-file.py
+++ b/ops/ccs-ops-misc/synthetic-data/scripts/generate-characteristics-file.py
@@ -1,0 +1,31 @@
+import sys
+import psycopg2
+import re
+from pathlib import Path
+
+def generate_characteristics_file(args):
+    """
+    Generates a beneficiary characteristics file for a given
+    synthea load, and exports it as a csv.
+    """
+    
+    # TODO: What do we need? Synthea start props?
+    # TODO: Check/Find the queries we use for this and do the thing. we'll know what info we need too
+
+    end_state_properties_file = Path(args[0]).read_text()
+    db_string = args[3]
+    generated_benes = args[2]
+    synthea_prop_filepath = args[1]
+    
+
+    
+
+## Runs the program via run args when this file is run
+if __name__ == "__main__":
+    # 5 args:
+    # arg1: previous end state properties file location
+    # arg2: number of items to be generated
+    # arg3: file system location of synthea properties file to edit
+    # arg4: db string for target environment DB, in this format: postgres://<dbName>:<db-pass>@<aws db url>:5432/fhirdb
+    # arg5: (optional) skip validation, useful if re-generating a bad batch, True or False
+    validate_and_update(sys.argv[1:])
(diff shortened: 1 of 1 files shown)


lsmitchell:
Example output from script, using bene id ranges -10000010009985 to -10000010050001

image

Compare to 

https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-model/bfd-model-rif-samples/dev/synthea_releases/01-20-2022/characteristics.csv

which was an old example characteristics file; same rows, row order, and ordering of benes (highest to lowest, which reads better given they are negative numbers). Data in the files is different, but useful for ensuring the form and format is the same.

ops/ccs-ops-misc/synthetic-data/scripts/generate-characteristics-file.py
```diff
@@ -0,0 +1,166 @@
...
+    return raw_query_response
+    
+def get_table_count(table_name, bene_id_start, bene_id_end, db_string):
+    """
+    Gets the table count for each bene in the specified range for the specified
+    database, and returns a dictionary with the bene id as the key and the
+    table count as the value.
+    """
+    
+    query = "SELECT t1.bene_id, count(t2.*)"\
```
mackec1, line 96: Curious on this logic; do we even need to query the beneficiaries table here. All the parent claims tables will a) have bene_id in them and b) the db has a constraint such that you cannot have a claim record without there being a beneficiary record that has the bene_id.

mackec1 (commented):
Had a comment on one the queries...

ops/ccs-ops-misc/synthetic-data/scripts/generate-characteristics-file.py
lsmitchell: I just implemented the sql queries as they were historically run, but actually looking at this...

We could get back the non-0 count results per table by just doing

```
SELECT bene_id, count(*)
FROM public.{table}
WHERE bene_id <= {start} and bene_id > {end}
GROUP BY bene_id
ORDER BY bene_id desc
```

but it doesnt give us the 0-count bene ids in the result (since those bene ids dont exist in that table), so it would require being a bit smarter about using the response (default to 0 if bene id doesnt exist in the table's response dict), unless there is some non-join sql adjustment to that query that im not thinking of that could ensure we have the full set of the requested bene ids in the result set.

Might be worth the query speed to just do some additional logic in the code when reading out of the dict results and remove a join; I'll tinker with it.

mackec1: so I think even the orig query would not get the 0-count bene_id(s). The orig did a left-join which will only return rcds where the bene_id was in both tables; using a left outer join would be necessary to find 0-count rcds.

lsmitchell: Was able to use the single query and a simple ternary on the place where the data is read to replace the join; good call on that! Speed was increased ~40%!

lsmitchell: The previous query did return the 0-count benes:
image
(and the new one doesnt) but due to the above I was able to use the python code to remove the need for the expensive join, so not a big deal!

mackec1 (approved):
LGTM

christrotteradhoc (approved):
Nice work

PR #1298, opened 2022-08-09
BFD-2067: Reduce Time in Jenkins Build Apps Stage

mjburling:
**JIRA Ticket:**
[BFD-2067](https://jira.cms.gov/browse/BFD-2067)

Remove redundant tests, test artifacts, javadocs, fingerprinting, and archiving from the Build Apps Stage.

---

### What Does This PR Do?
By removing these additional, largely redundant steps from the Build Apps stage, it now takes between 1m 30s and 1m 45s on average compared to the existing range between ~6m across the previous 10 builds. In short:
- **without** tests and javadoc, builds take 3-5m to execute (a 2m to 3m savings)
- The fingerprinting and archiving steps take ~1m 30s so eliminating this further reduces by the same amount

Again, these pieces are largely redundant, because deployments to our path-to-production environments (`test`, `prod-sbx`, `prod`) are executed from the trunk/master branch, where changes can only be accepted upon successful test execution within GitHub Actions.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/build.groovy
+++ b/apps/build.groovy
@@ -29,15 +29,7 @@ def build(boolean verboseMaven) {
 	dir ('apps') {
 		quietFlags = verboseMaven ? '' : '--quiet --batch-mode'
 
-		sh "mvn ${quietFlags} --threads 1C --update-snapshots -Dmaven.test.failure.ignore clean verify"
-		/*
-		 * Fingerprint the output artifacts and archive the test results.
-		 *
-		 * Archiving the artifacts here would waste space, as the build deploys them to the local Maven repository.
-		 */
-		fingerprint '**/target/*.jar,**/target/*.war,**/target/*.zip'
-		junit testResults: '**/target/*-reports/TEST-*.xml', keepLongStdio: true
-		archiveArtifacts artifacts: '**/target/*.jar,**/target/*.war,**/target/*.zip,**/target/*-reports/*.txt', allowEmptyArchive: true
+		sh "mvn ${quietFlags} --threads 1C --update-snapshots -Dmaven.test.failure.ignore -DskipITs -DskipTests -Dmaven.javadoc.skip=true clean verify"
 	}
 
 	return new AppBuildResults(
--- a/apps/build.groovy
+++ b/apps/build.groovy
@@ -29,7 +29,7 @@ def build(boolean verboseMaven) {
 	dir ('apps') {
 		quietFlags = verboseMaven ? '' : '--quiet --batch-mode'
 
-		sh "mvn ${quietFlags} --threads 1C --update-snapshots -Dmaven.test.failure.ignore -DskipITs -DskipTests -Dmaven.javadoc.skip=true clean verify"
+		sh "mvn ${quietFlags} --threads 1C --update-snapshots -DskipITs -DskipTests -Dmaven.javadoc.skip=true clean verify"
 	}
 
 	return new AppBuildResults(


cbrunefearless (dismissed):
lgtm

mackec1 (dismissed):

apps/build.groovy
```diff
@@ -29,15 +29,7 @@ def build(boolean verboseMaven) {
...
-		sh "mvn ${quietFlags} --threads 1C --update-snapshots -Dmaven.test.failure.ignore clean verify"
-		/*
-		 * Fingerprint the output artifacts and archive the test results.
-		 *
-		 * Archiving the artifacts here would waste space, as the build deploys them to the local Maven repository.
-		 */
-		fingerprint '**/target/*.jar,**/target/*.war,**/target/*.zip'
-		junit testResults: '**/target/*-reports/TEST-*.xml', keepLongStdio: true
-		archiveArtifacts artifacts: '**/target/*.jar,**/target/*.war,**/target/*.zip,**/target/*-reports/*.txt', allowEmptyArchive: true
+		sh "mvn ${quietFlags} --threads 1C --update-snapshots -Dmaven.test.failure.ignore -DskipITs -DskipTests -Dmaven.javadoc.skip=true clean verify"
```
christrotteradhoc, line 32: do we still need to pass maven.test.failure.ignore?

mjburling: Fear had me keep it. I can get rid of that. Thanks!

apps/build.groovy
```diff
@@ -29,15 +29,7 @@ def build(boolean verboseMaven) {
...
 		quietFlags = verboseMaven ? '' : '--quiet --batch-mode'
 
-		sh "mvn ${quietFlags} --threads 1C --update-snapshots -Dmaven.test.failure.ignore clean verify"
-		/*
-		 * Fingerprint the output artifacts and archive the test results.
-		 *
-		 * Archiving the artifacts here would waste space, as the build deploys them to the local Maven repository.
-		 */
-		fingerprint '**/target/*.jar,**/target/*.war,**/target/*.zip'
-		junit testResults: '**/target/*-reports/TEST-*.xml', keepLongStdio: true
```
lsmitchell, line 39: Is there any type of configuration outside of github which could break tests (and the environment) which we would lose visibility of by removing this step? (Alternatively, could we just run tests in _one_ environment during deploys instead?) I can't immediately think of too many cases of non-code config breaking the tests, outside of maybe database or aws config changes between last merged commit and deploy. Keeping our merge > deploy cycle short will minimize these risks, but still something to think about.

mackec1 (approved):

christrotteradhoc (approved):

apps/build.groovy
mjburling: Thanks for the feedback!

> Is there any type of configuration outside of github which could break tests (and the environment) which we would lose visibility of by removing this step?

In some earlier discussions about eliminating the tests as part of the Jenkins pipeline, there was some discussion of whether it has ever been the case that we've had a failure in Jenkins but the same change set passed in GitHub Actions. Ignoring intermittent failures in flakey ITs, the _collective_ recollection suggested that this happened around the time of upgrading Java/JDK versions...

Further, others said that ensuring that testing with two different CI mechanisms is not a common requirement and has really just been accidental in this case.

Also of note, we weren't relying on the tests to do anything for us in Jenkins. Historically, we ignored failures and proceeded to deploy regardless of the test outcome. 

> (Alternatively, could we just run tests in one environment during deploys instead?)

The way we currently think about this monolithic pipeline, there isn't an obvious solution to that. I _could_ see a path forward to making the tests optional, defaulting to no tests, and running these things with some set schedule. I'm not sure of the value proposition, though...

> Keeping our merge > deploy cycle short will minimize these risks [..]

Most definitely!

PR #1318, opened 2022-08-22
BFD-1778: Refine how performance metrics are compared and stored to make the Regression Suite more consistent and useful

malessi:
**JIRA Ticket:**
[BFD-1778](https://jira.cms.gov/browse/BFD-1778)

As a BFD engineer I would like to figure out what the baseline should be for the performance regression tests so that I can set up realistic validation failure thresholds.

---

### What Does This PR Do?

In short, this PR makes a bunch of changes to how the regression suite loads data, how individual stats/metrics are compared, and more in order to make the regression suite more consistent and its outputs more useful and meaningful. With these changes, the regression suite should automatically be able to tell us if a given deployment regresses in performance in some fashion, and the data output from each run should be useful enough to help us to determine _where_ we regressed in performance.

This PR:
- Adds more metadata to aggregated stats that is stored and used to verify that comparisons can be made:
  - The requested runtime (via `--run-time` or `--spawned-runtime`)
  - The Locust `HttpUser` classes that `Task`s were ran from
  - A SHA256 hexadecimal hash string that is the hash of all of the comparable metadata about stats
    - This hash is also used as a partition in the Glue Catalog Table, and replaces most of the verification checks when loading comparable statistics
  - The overall comparison result between the current run and the "baseline"
    - This overall result is `NOT_APPLICABLE` by default -- when loading baseline aggregated stats, only aggregated stats with a `compare_result` of `FAILED` are excluded
- Adds the aggregated totals reported by Locust to the aggregated stats that are stored, and uses them as an additional set of data to compare against
- Allows for storing aggregated stats under multiple tags
- Adds a flag to aggregated stats' metadata, `compare_result`, that indicates the overall result of comparison between the current run and the user-specified baseline
  - This is used, as stated above, to filter out aggregated stats from runs that had failing comparisons -- this way, we can analyze failing runs by looking at their performance metrics without it affecting the baseline
- Changes how stats are compared against a baseline:
  - The percent ratio thresholds, along with whether a stat is "better" if its value is smaller or larger, are now encoded in a JSON "stats comparisons metadata" file. 
    - For the regression suite, this file is `./config/regression_suites_compare_meta.json`
    - The percent ratio threshold is now per-stat instead of global -- this allows us to specify tighter baselines for certain statistics while allowing more room for stats that are sometimes inconsistent (such as response time percentiles)
    - Additionally, percent ratio thresholds per-stat can be defined differently for the aggregated totals and for tasks. This allows us to tighten the comparisons on aggregated totals (as they should not change much), while leaving some room for each task (which can, depending on the stat, fluctuate somewhat)
  - A new percent ratio threshold, `warn_threshold`, has been added to indicate when the value of a stat _may_ be abnormally "worse" in comparison to the baseline
    - This does not set the run as failed (return an exit code of `1` and set `compare_result` to `FAILED`)
- Changes the data loaded by the regression suite to use a consistent set of 19,999 synthetic beneficiaries that exist in each environment
  - This greatly reduces data load time for all environments, and since each environment now uses the same data there are no erroneous performance differences between each environment (i.e. `PROD` being significantly slower than the other environments)
  - Performance is now more consistent, however it is not _perfectly_ consistent between runs. There is some variance, especially between tasks' response time percentiles and total requests-per-second. Generally speaking, the total requests-per-second is within 100 to 120 rps, and individual task's RTPs can vary from 50% to 200% between runs
- Refactors `--stats-config`'s custom `key=value` string parsing into distinct command-line arguments
- Adds a limit to the number of aggregated stats loaded for a baseline during compare, controlled by `--stats-compare-load-limit`. Defaults to `5`
  - Stats are loaded in order of most recent to least recent
  - This is used solely for average comparisons -- previous is unaffected, as it only loads the most recent, valid aggregated stats as the baseline
- Changes the default comparison type for the deployment pipeline to make the baseline be the average of the previous 5 runs under the tag matching the current branch name
- Enables running the regression suite during deployment against `PROD`
- Enables the regression suite to pause the deployment pipeline upon failure until an operator manually aborts or accepts the failure and allows for the deployment to proceed
  - This allows for deployments to continue in certain scenarios where false negatives are expected
- The results of individual stat comparisons are now logged in JSON format to the console regardless of their status
  - In practice this may increase log size by 1 - 2 KB, but allows us to see the exact values of each stat versus the baseline even if they passed comparison
- Performance stats from a deployment regression suite run will now be stored under two tags:
  - The name of the branch
  - The build number of the current Jenkins build plus the name of the branch, i.e.: `build21___master`
- Disables hardcoded SLA validation for each regression suite 

* With the changes made, will regression failures be meaningful?
* Do the chosen percent ratio thresholds make sense for each stat?
* Do the changes make sense?
* Are all of the AC associated with BFD-1778 addressed adequately?

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/utils/locust_tests/common/stats/stats_compare.py
```diff
@@ -1,212 +1,332 @@
...
 
     stat: str
     """The name of the stat"""
     percent: float
-    """The percent value of this stat's value versus a previous run's, i.e. this stat is percent% times worse than
-    the previous run or average of all runs. A value of 100 means no change (i.e. current stat is 100%
-    of a previous snapshot of the same stat), whereas values lower than 100 generally indicate
-    positive change and values greater than 100 indicate negative change"""
+    """The percent value of this stat's value versus a previous run's, i.e. this stat is percent%
+	times worse than the previous run or average of all runs. A value of 100 means no change (i.e.
```
mackec1, line 74: Just curous....would a negative val=ue (if supported) represent an improvement?

malessi: When I wrote this I didn't actually realize that what `percent` is is actually the "percent ratio" between the current stat and previous stat. So, you can think of it as:

$$\frac{current}{previous}\times 100 = percent$$
or
$$\frac{previous}{current}\times 100 = percent$$
(depending on whether or not the stat's value is "better" for smaller values [i.e. response times] or for larger values [number of requests])

Since it's a ratio, anything below 100% means that the current stat's value is _better_ than the baseline's value. As in, if the current `total_reqs_per_second` is 100 and the baseline's value was 95:

$$\frac{previous}{current}\times 100 = \frac{95}{100}\times 100 = 95\\%$$

So percent ratios under 100% indicate improvement, a percent ratio of _exactly_ 100% indicates no change, and a percent ratio greater than 100% indicates negative change. It shouldn't be possible for `percent` be negative, since it's really just a ratio between the two stats.

mackec1 (approved):
Lots to digest here; looks reasonable, thx for the clarification on perf regression vs. improvement.

apps/utils/locust_tests/common/stats/stats_compare.py
```diff
@@ -1,212 +1,332 @@
...
+            Tuple[List[StatComparison], Dict[str, List[StatComparison]]]: A Tuple where the first
+            entry is the list of comparison results for the aggregated totls and the second entry
+            is a dictionary of tasks to their list of comparison results
     """
     prev_tasks = {task.task_name: task for task in previous.tasks}
     cur_tasks = {task.task_name: task for task in current.tasks}
     common_tasks = prev_tasks.keys() & cur_tasks.keys()
 
-    failed_tasks_with_percents = {}
+    tasks_exceeded_results = {}
```
lsmitchell, line 263: this name isn't immediately clear on what it is; could this be named to be slightly clearer? (I think calling what I think are the results or runs "tasks" is throwing me a bit when the documentation references results)

apps/utils/locust_tests/common/stats/stats_config.py
```diff
@@ -36,160 +37,234 @@ class StatsComparisonType(str, Enum):
...
-    database: Optional[str]
+    stats_store_s3_database: Optional[str]
     """Name of the Athena database that is queried upon when comparing statistics.
     Also used as part of the file path when storing stats in S3"""
-    table: Optional[str]
+    stats_store_s3_table: Optional[str]
     """Name of the table to query using Athena if store is s3 and compare is set.
-    Also used as part of the file path when storing stats in S3"""
-    compare: Optional[StatsComparisonType]
+    jAlso used as part of the file path when storing stats in S3"""
```
lsmitchell, line 68: Errant 'j'

apps/utils/locust_tests/v2/regression_suite.py
```diff
@@ -30,29 +30,26 @@ def _(environment: Environment, **kwargs):
...
         environment.parsed_options,
-        db.get_hashed_mbis,
-        use_table_sample=False,
+        db.get_regression_hashed_mbis,
         data_type_name="hashed_mbis",
     )
 
 
-validation.set_validation_goal(validation.ValidationGoal.SLA_V2_BASELINE)
+set_comparisons_metadata_path("./config/regression_suites_compare_meta.json")
```
lsmitchell, line 52: Can/should we not have both goals? We can check for regressions, but should also make sur ethe SLAs arent violated too.
Edit: looking at the config json, maybe they are checked, just in the configurable checks?

apps/utils/locust_tests/config/regression_suites_compare_meta.json
```diff
@@ -0,0 +1,162 @@
...
+      "better_if": "SMALLER",
+      "warn_percent": 150,
+      "fail_percent": 500
+    },
+    {
+      "name": "total_reqs_per_second",
+      "better_if": "LARGER",
+      "warn_percent": 150,
+      "fail_percent": 200
+    },
```
lsmitchell, line 38: With the way that locust works, this is essentially a side effect of the response time, isn't it? Seems like we'd be capturing two metrics that basically say the same thing with requests/second.

lsmitchell (commented):
Between the linting and various fixes across all the files, was a little hard to keep a track on the major changes, but I think I got it. A few comments/questions, but looks good overall!

apps/utils/locust_tests/v2/regression_suite.py
malessi: We definitely could; however, the numbers for V1 and V2 SLAs are fairly arbitrary though since I just picked numbers that were high enough so that random failures didn't happen during actual runs. 

The reason why I removed this is because, assuming the regression suite is ran how it _should_ be ran (by comparing against a previous run [baseline]), the performance metrics of the current run will be validated against a baseline to ensure performance is not noticeably worse than normal. I felt the additional, arbitrary SLA validation was redundant in the usual regression suite case.

I think if we adjusted the SLAs to be accurate (instead of just arbitrarily high) we could re-enable this validation, but as it stands I don't think there's much value added on top of the baseline validation.

Note that we still always validate that no response times are greater than 10 seconds -- this is true for every Locustfile in this project.

apps/utils/locust_tests/common/stats/stats_config.py
malessi: Good catch, I'll remove it.

apps/utils/locust_tests/common/stats/stats_compare.py
malessi: So `tasks_exceeded_results` is short for "all `StatCompareResult`s that exceeded the passed-in `threshold`% for each Locust `Task`s keyed by the `Task`'s name". Whenever I use "task" in a Locust context, I am talking about a single test case within a test suite (Locustfile), i.e.:

```python
@tag("coverage", "coverage_test_id_count")
@task
    """Coverage search by ID, Paginated"""
    self.run_task_by_parameters(
        base_path="/v1/fhir/Coverage",
        params={"beneficiary": self.bene_ids.pop(), "_count": "10"},
        name="/v1/fhir/Coverage search by id / count=10",
    )
```

Does that help with your understanding? I do agree that this particular naming is probably a bit too terse or difficult to understand without the context I have from writing it. I'll think about a different name and/or adding a comment that matches what I wrote above.

apps/utils/locust_tests/common/db.py
```diff
@@ -32,6 +39,44 @@ def _execute(uri: str, query: str) -> List:
...
+
+
+def get_regression_bene_ids(uri: str) -> List[str]:
+    bene_query = __get_regression_query('SELECT "bene_id" FROM "beneficiaries"')
+    return [str(r[0]) for r in _execute(uri, bene_query)]
+
+
+def get_regression_hashed_mbis(uri: str) -> List[str]:
+    mbi_query = __get_regression_query('SELECT "mbi_hash" FROM "beneficiaries"')
+    return [str(r[0]) for r in _execute(uri, mbi_query)]
```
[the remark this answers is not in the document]
malessi: These functions are used by each regression suite to load data -- each of them calls `__get_regression_query()` with a `SELECT` query string as the value for `select_query`:

```python
def _get_regression_query(select_query: str) -> str:
    return " ".join(
        [
            select_query,
            f'WHERE "bene_id" BETWEEN {REGRESSION_BENE_RANGE_START} AND {REGRESSION_BENE_RANGE_END}',
            'ORDER BY "bene_id" ASC',
            f"LIMIT {LIMIT}",
        ]
    )
```

`__get_regression_query()` is fairly simple, it just concatenates the `SELECT` query with: 

- A `WHERE` clause that limits the query to the range of 19,999 synthetic beneficiaries that exist in each environment
- An `ORDER BY` clause that orders the results by beneficiary ID in ascending order (so that the data is always consistently ordered)
- And a `LIMIT` clause that puts a hard 100,000 result limit on the returned results

So for the two functions here, `get_regression_bene_ids()` and `get_regression_hashed_mbis()`, both are already limited by the range of 19,999 synthetic beneficiaries and by the global limit of 100,000 results (via the `LIMIT` constant in `db.py`).

Does that answer your question? Or, are you talking about a different kind of limit here?

Jenkinsfile
```diff
@@ -278,11 +278,7 @@ try {
 						if (hasRegressionRunSucceeded) {
 							println 'Regression suite passed, proceeding to next stage...'
 						} else {
-							// TODO: Regression suite is currently inconsistent, so failing builds automatically
-							// would cause false negatives. Uncomment the line below when the regression suite is
-							// more consistent
-
-							// error('Regression suite failed, check the CloudWatch logs above for more details')
+							error('Regression suite failed, check the CloudWatch logs above for more details')
```
mjburling, line 281: Curious on your thoughts, not necessarily looking for any changes here or elsewhere.

I wonder if we want to initially (or always) yield to an operator when the regression suite _fails_? Instead, we might alert via slack message and enter a waiting state for the operator to make a decision as to whether the deployment should proceed. It's not difficult to imagine scenarios where we're deploying changes that are more time-sensitive, and being forced to re-deploy due to an unrelated regression suite failure would be pretty frustrating.

ops/terraform/services/server/deploy.groovy
```diff
@@ -91,9 +91,9 @@ def runServerRegression(Map args = [:]) {
         'suite_version': 'v2',
         'spawn_rate': 10,
         'users': 10,
-        'spawned_runtime': '30s',
-        'compare_tag': "build${lastSuccessfulBuildID}__${sanitizedBranchName}",
-        'store_tag': "build${currentBuildId}__${sanitizedBranchName}"
+        'spawned_runtime': '1m',
```
mjburling, line 94: Curious if you could comment on the change to double the time. Was 30s insufficient? Obviously, if it's necessary, let's continue, otherwise I'm just concerned about adding another 30s to each environment's deployment (or 180s over our standard, monolithic, three-environment deployment).

Jenkinsfile
malessi: This is something I've been thinking about as well, and probably needs additional work. I was thinking about something more automated -- maybe if a run fails against the baseline, we run again to be _very_ sure that there wasn't some one-off problem that caused an erroneous failure? Then, if that "safetynet" run fails, we actually fail the deployment.

I also really like the idea of having an operator come in and make a decision to continue even if we fail as a sort of override and safeguard against potential false negatives.

I think both strategies are mutually exclusive from each other, and could be used together. Either way, I also feel a little uneasy now about straight-up failing. I think we should create a card for improving this.

ops/terraform/services/server/deploy.groovy
malessi: I recall seeing some inconsistent results with 30 seconds even after switching to the consistent dataset. I'll do some more runs against test with 30 seconds and see if I'm off on that.

apps/utils/locust_tests/common/db.py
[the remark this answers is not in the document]
lsmitchell: That answers it, it's basically implicitly limited by the amount of data. I think its fine as-is then, given i doubt we want to adjust the data we're using for now.

Jenkinsfile
```diff
@@ -278,7 +278,12 @@ try {
 						if (hasRegressionRunSucceeded) {
 							println 'Regression suite passed, proceeding to next stage...'
 						} else {
-							error('Regression suite failed, check the CloudWatch logs above for more details')
+							try {
+								input 'Regression suite failed, check the CloudWatch logs above for more details. Should deployment proceed?'
+								echo "Regression suite failure in '${bfdEnv}' has been accepted by operator. Proceeding to next stage..."
+							} catch(err) {
+								error "Operator opted to fail deployment due to regression suite failure in '${bfdEnv}'"
+							}
```
malessi, line 286: @lsmitchell @christrotteradhoc @mjburling FYI for those in the Surge Status meeting and for Burling -- I've updated the handling for regression suite failures in the deployment pipeline to not immediately fail. Instead, we wait for an operator to explicitly abort or proceed via the [`input`](https://www.jenkins.io/doc/pipeline/steps/pipeline-input-step/) step, similarly to the "manual approval" stage. This allows us to avoid false negatives automatically causing deployment failures.

Let me know if any of you think this is insufficient. Also, (and this can be done in the future), should we be sending a notification when this happens so that there's better awareness? I know we monitor deployments fairly closely right now, so I don't think it's necessary at the moment but it's something I was thinking about.

malessi: Also, this block is duplicated across the two other stages (for `prod-sbx` and `prod`) -- not just for `test`.

mjburling: This is exactly the sort of functionality that I had in mind! Is that all it takes.

apps/utils/locust_tests/config/regression_suites_compare_meta.json
malessi: I suppose it is, although I don't think it hurts to include this particular metric in our baseline comparisons and in the aggregated stats we store. It's been a good "at-a-glance" metric for when we run look at logs to see if a run is overall good or not, too.

Jenkinsfile
mjburling: Seems so. Thanks for putting that together!

apps/utils/locust_tests/common/stats/stats_compare.py
lsmitchell: Yup that helps! If they are essentially test case results I'd have preferred they be named something that indicates that, but not going to ask you go and adjust the whole codebase over it. I think just adjusting the variable names like you suggested here would go a long way in describing it to the uninitiated!

apps/utils/locust_tests/v2/regression_suite.py
lsmitchell: I was mostly concerned that because our "failure" rate is pretty high on some metrics, we could slowly degrade and eventually slip out of SLA territory, even if the original baseline was well within them. It might make sense to avoid actively checking SLAs until we get the more accurate ones though, since like you said we don't have real numbers for now, but I think checking SLAs in these tests in general still has value once we can accurately do it.

malessi: I agree that we should reintroduce SLA checks once we have more accurate numbers (rather than numbers that are just convenient).

Note that dhe failure threshold is high for some because they do tend to vary substantially, but those that tend to vary a lot don't seem to have much effect on more important metrics like average and median response times and requests per second. Since they vary a lot, if they hit their high failure thresholds we can be more confident that something is definitely wrong.

lsmitchell (dismissed):
Thanks for explaining and fixes!

ops/terraform/services/server/deploy.groovy
malessi: I've done a few 30 second runs, and I feel like we should be OK if we stick with `30s`. One of the runs _did_ hit warning thresholds for some of the more _flexible_ statistics (i.e. `0.99` response time percentiles), but has since been just fine and been fairly consistent.

I'll change it back to `30s` and if, in the future, we run into issues with that we can always increase it.

P.S. Maybe we should think about getting these values into SSM so we don't need to open a PR to adjust them in the future?

lsmitchell (approved):

mjburling (approved):
💯

mjburling (approved):

PR #1326, opened 2022-08-26
BFD-2116: Enable base service deployments

mjburling:
**JIRA Ticket:**
[BFD-2116](https://jira.cms.gov/browse/BFD-2116)

Enable base service deployments.

---

### What Does This PR Do?
This PR introduces environment-specific `base` service deployment stages to the existing build/deliver/deploy pipeline. Additionally, because the `base` service relies on `yq`, included in an updated `bfd-cbc-build` image. Embarassingly, the pace of configuration changes has been so sluggish that this has been done manually for each environment. Let's fix that!

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,6 +52,7 @@ def scriptForApps
 def scriptForDeploys
 def migratorScripts
 def serverScripts
+def baseScripts
 def canDeployToProdEnvs
 def willDeployToProdEnvs
 def appBuildResults
@@ -135,7 +136,7 @@ try {
 		containers: [
 			containerTemplate(
 				name: 'bfd-cbc-build',
-				image: 'public.ecr.aws/c2o1d8s9/bfd-cbc-build:jdk11-mvn3-an29-tfenv',
+				image: 'public.ecr.aws/c2o1d8s9/bfd-cbc-build:jdk11-mvn3-an29-tfenv-aeaa61fa6', //TODO: consider a smarter solution for resolving this image
 				command: 'cat',
 				ttyEnabled: true,
 				alwaysPullImage: false, // NOTE: This implies that we observe immutable container images
@@ -154,6 +155,9 @@ try {
 					// Load the child Jenkinsfiles.
 					scriptForApps = load('apps/build.groovy')
 					scriptForDeploys = load('ops/deploy-ccs.groovy')
+
+					// terraservice deployments...
+					baseScripts = load('ops/terraform/services/base/deploy.groovy')
 					migratorScripts = load('ops/terraform/services/migrator/Jenkinsfile')
 					serverScripts = load('ops/terraform/services/server/deploy.groovy')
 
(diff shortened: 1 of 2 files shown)


ops/terraform/services/base/deploy.groovy
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env groovy
+
+// TODO: Consider further generalizing this and hoisting into a new ops/jenkins/global-pipeline-libs/terraform.groovy
+
+/* Deploys base terraservice
+ * @param args a {@link Map} must include `bfdEnv`, `serviceDirectory`
```
christrotteradhoc, line 6: technically args should contain 'env' and 'directory', right?

mjburling: You got me. Let me correct that. Thanks!

Jenkinsfile
```diff
@@ -228,8 +232,22 @@ try {
 				}
 			}
 
+			bfdEnv = 'test'
+			stage('Deploy base to TEST') {
```
malessi, line 236: Genuinely curious: does everything work correctly here with `bfdEnv` being set outside of a `stage` block/scope? Previously, this variable was set within the `stage`. I know in a Declarative Pipeline, code within a `pipeline` block but _not_ in a `stage` would blow up.

mjburling: I was curious about this, too. There isn't any magic going on here as far as I know, to such an extent that I had wondered if we wouldn't be better served by wrapping the stages in `withEnv` blocks and so on. All that said, I've tested this against `test`, and it works.

christrotteradhoc (approved):

Jenkinsfile
```diff
@@ -135,7 +136,7 @@ try {
 		containers: [
 			containerTemplate(
 				name: 'bfd-cbc-build',
-				image: 'public.ecr.aws/c2o1d8s9/bfd-cbc-build:jdk11-mvn3-an29-tfenv',
+				image: 'public.ecr.aws/c2o1d8s9/bfd-cbc-build:jdk11-mvn3-an29-tfenv-aeaa61fa6', //TODO: consider a smarter solution for resolving this image
```
keithdadkins, line 139: Couldn't you tag whichever image we should be currently using `latest`?

keithdadkins (approved):
Looks good! We (bfd-engineers/me) would benefit from a good TLDR CI/CD walkthrough during an engineering sync sometime soon.

Jenkinsfile
mjburling: tl;dr; yes and no. We need to give this just a bit more thought.

We could! However, tag mutability is a double-edged sword. It cuts down on complexity when referencing a `latest` tag, but it complicates things if when changes are introduced that aren't appropriate for all workloads that refer to it. I think the use of tag mutability is still considered somewhat of an anti-pattern.

A reference to a literal, specific image tag isn't _much_ better, because we're left with headspace/complexities surrounding updates to that reference when a new tag is available. A seemingly obvious solution to this would involve storing the desired image tag somewhere, e.g. AWS SSM Parameter. This requires a little more thought, because the awscli doesn't exist in the entrypoint to our Jenkins pipelines... instead, we're looking at potentially storing this in a truly global variable in our Jenkins project.

PR #1356, opened 2022-09-07
BFD-1885: Adjust load test validation to only be safeguards

lsmitchell:
**JIRA Ticket:**
[BFD-1885](https://jira.cms.gov/browse/BFD-1885)

Adjusts the load test validation to only be safeguards and not check actual SLOs.

---

### What Does This PR Do?

Originally, we wanted to add SLO validation to the load test to ensure the application is performant or at least does not die under load. However, we discussed implementing these SLOs and decided there was little value in these checks being automated and failing a run, so this PR removes those automated validations. Instead we will check the results manually when we run the suite under varying loads.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/utils/locust_tests/high_volume_suite.py
+++ b/apps/utils/locust_tests/high_volume_suite.py
@@ -52,11 +52,6 @@ def _(environment: Environment, **kwargs):
         data_type_name="hashed_mbis",
     )
 
-## Load tests won't have validation EXCEPT our safeguard to ensure the database isn't falling over and bringing down the environment
-## If we want to test destroying the environment with massive calls, remove this.
-validation.set_validation_goal(validation.ValidationGoal.SLA_SAFEGUARD)
-
-
 class TestLoadShape(UserInitAwareLoadShape):
     pass
 
--- a/apps/utils/locust_tests/high_volume_suite.py
+++ b/apps/utils/locust_tests/high_volume_suite.py
@@ -52,8 +52,9 @@ def _(environment: Environment, **kwargs):
         data_type_name="hashed_mbis",
     )
 
-
-validation.set_validation_goal(validation.ValidationGoal.SLA_V2_BASELINE)
+## Load tests won't have validation EXCEPT our safeguard to ensure the database isn't falling over and bringing down the environment
+## If we want to test destroying the environment with massive calls, remove this.
+validation.set_validation_goal(validation.ValidationGoal.SLA_SAFEGUARD)
 
 
 class TestLoadShape(UserInitAwareLoadShape):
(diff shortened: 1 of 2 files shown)


mackec1 (dismissed):
looks reasonable.

apps/utils/locust_tests/high_volume_suite.py
```diff
@@ -52,8 +52,9 @@ def _(environment: Environment, **kwargs):
         data_type_name="hashed_mbis",
     )
 
-
-validation.set_validation_goal(validation.ValidationGoal.SLA_V2_BASELINE)
+## Load tests won't have validation EXCEPT our safeguard to ensure the database isn't falling over and bringing down the environment
+## If we want to test destroying the environment with massive calls, remove this.
+validation.set_validation_goal(validation.ValidationGoal.SLA_SAFEGUARD)
```
malessi, line 57: So, I have some concerns about this, and some comments:

1. I am not sure if you are aware, but the validation goals are only ever checked at the _end_ of a test run. From your inline comment here it seems this is meant to immediately stop the test run if we ever exceed this safeguard. As the validation logic is written today, this will not happen -- instead, if the 50%, 95%, or 99%s exceed 3000 ms an error will be logged and the return code of the Locust process will be set to `1`
2. 3000 ms is exceptionally low -- running a scaling load test earlier today showed that several endpoints hit 3000 ms on their 99%s prior to ever hitting scaling without there being any indication that the infrastructure was falling over or any requests failing.

Considering (1) and (2) I think some changes need to be made here. If we want the 50%s, 95%s, and 99%s to be checked continuously some additional code changes need to be made, and I feel that the 3 second safeguard should be increased to _at least_ 10 seconds to ensure we're not ending perfectly valid load test runs prematurely.

lsmitchell: Thanks Mitch! I had originally set a safeguard that should be checked on each test's end (to be able to shut things down if they get out of hand) but I think the spirit of it was lost at some point when the validation style changed. I'll go back to trying to implement that again instead of doing this.

lsmitchell: Just noticed the failsafe is already set up in the bfd_user_base as part of each test, and is set to 10 seconds, so I should just need to remove the validation here.

mjburling (approved):

apps/utils/locust_tests/high_volume_suite.py
malessi: Oops, that's my bad! I assumed we wanted _additional_ safeguards on the 50%s, 95%s, and 99%s so I didn't even mention it. I'm glad that the existing failsafe is sufficient!

malessi (approved):

mackec1 (approved):

PR #1366, opened 2022-09-14
PACA-716: Add database schema notes to README.

brianburton:
**JIRA Ticket:**
[PACA-716](https://jira.cms.gov/browse/PACA-716)

Add RDA Schema info (feel free to use https://confluence.cms.gov/display/BCDA/PACA+Schema+for+RDA+Data as a starting point) to the [bfd-model-rda README](https://github.com/CMSgov/beneficiary-fhir-data/tree/master/apps/bfd-model/bfd-model-rda).

---

### What Does This PR Do?

Adds RDA schema info to the README.md file.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


ericbuckley (approved):

KyRobbins (approved):

ericbuckley (approved):

apps/bfd-model/bfd-model-rda/README.md
```diff
@@ -17,3 +17,155 @@ uses Spring to build up its JPA configuration and Spring does not have the same
...
+The RDA ETL pipeline jobs consume data by calling a gRPC service in the RDA API.  This service
+returns a stream of claim messages.  Each message represents either a new claim or an update
+to an existing claim.  Every message includes a sequence number.  These sequence numbers
+can be used to track progress within the stream and to resume from a specific location
+in the stream in a later RPC call.
+
+The `rda_api_progress` table is used to store the last processed sequence number for each claim type.
+The pipeline jobs update the table as they process incoming claims.  The jobs can be forced to
+replay data by stopping the job, changing the `last_sequence_number` value for the appropriate
+claim type, and restarting the pipeline job.
```
lsmitchell, line 151: Is this different from the way non-RDA claims are processed, or am I just misssing information about how the pipeline works? (For instance, we can replay non-RDA loads by running in idempotent mode without touching a database directly)

lsmitchell (approved):
Very informative!

apps/bfd-model/bfd-model-rda/README.md
brianburton: The RDA pipeline is always idempotent in the sense that reingesting the same claims multiple times will not really change any claim data except last updated timestamps.  But the progress table is only consulted when the pipeline jobs start and then updated continuously as they ingest claims.  As a consequence if you change the value while the job is running you would just see your change overwritten without it doing what you want.

The alternative would be to query the current value from the progress table before updating it (which happens millions of times per run) prior to each update and maybe abort the process if the value is different than expected.  We did not implement it that way since reingestion is expected to be extremely rare in practice (like only in response to a data consistency incident or something) and it would be somewhat slower to do it that way.

lsmitchell: I see, so RDA has a more granular record of what's been loaded, in which case it's unlikely we'd need to rerun the entire load anyway. Any error/failure would only need to resume at the last progress update, so its not like the non-RDA pipeline. Thanks for explaining!

PR #1381, opened 2022-09-21
BFD-1883: Create a Jenkins job for running arbitrary load testing

malessi:
**JIRA Ticket:**
[BFD-1883](https://jira.cms.gov/browse/BFD-1883)

As a BFD Engineer, I want to be able to run the load test suite from Jenkins so that they can be run
on demand. I also want to be able to run the load tests with an arbitrary amount of users so I can
test various scenarios against the servers.

---

### What Does This PR Do?

This PR adds a new Jenkins pipeline named `bfd-run-server-load` that is capable of running many
different load testing scenarios, such as:

- Load tests that ramp up infinitely until a scaling event
  - The number of instances in the scaling group to target is configurable
- Static load tests that run with a set number of users for a set period of time ignoring scaling events
- And more...

The flexibility allows for an engineer to modify parameters and see how the BFD Server performs
under the load conditions provided. In all, the following parameters can be modified:

- SDLC Environment
- The host to test
- The number of _initial_ worker **nodes**/Lambdas (_not_ simulated users) to start with before the Locust master starts
- The amount of time to wait between spawning nodes/Lambdas
- The maximum number of nodes/Lambdas to spawn
- The maximum number of _simulated users_ for **Locust** to spawn
  - This controls the load actually generated on the target
  - In general, a ratio of 10:1 (10 users to 1 node/lambda) should be followed for the best performance
- The spawn rate of _simulated users_ in number of users per second
- The test runtime limit
  - This is used as a safeguard to ensure that runaway load testing does not occur
  - Separate from the Lambda runtime limit
- The amount of time to "coast" after a scaling event was encountered
  - Allows for testing the performance during and immediately after scaling
  - Set to `0` if the amount of load required to scale (and nothing more) is desired
- The target number of instances that have been scaled up to before load testing stops
- Whether to stop when a scaling event is encountered
- Whether to stop when the node/Lambda limit is reached or not
  - Use this as a safeguard against runway Lambda spawning during non-static load tests

By default, _for the Jenkins job_, these parameters are set to values that will continue to ramp up
the test until a scaling event is encountered at which point the test will "coast" for 10 seconds
and stop.

Note that builds/runs of this pipeline _do not_ communicate with the controller directly. The
operator is required to indicate when the load test has finished (or allow the built-in timeout to
lapse).

The statistics and logs generated from a load test run are made available in CloudWatch in realtime,
under the following log groups with each log stream being the instance ID it was gathered from:

- /bfd/{env}/bfd-server-load/server-load-controller.log
- /bfd/{env}/bfd-server-load/load_exceptions.csv
- /bfd/{env}/bfd-server-load/load_failures.csv
- /bfd/{env}/bfd-server-load/load_stats.csv
- /bfd/{env}/bfd-server-load/load_stats_history.csv

This service is built to be as safe as possible, having a variety of safeguards built-in to ensure that unintended or runaway load testing does not occur:

- A hardcoded "stop signal" is constantly polled for. If encountered, all nodes _and_ the controller stop immediately
- When the controller stops, it unconditionally sends a stop signal to ensure nodes stop as well
- When the Jenkins pipeline build ends, it sends a stop signal to ensure that the controller and nodes stop
- When the Jenkins pipeline build ends, the controller instance is immediately destroyed
- At any time an operator can opt to abort the Jenkins pipeline build causing the above safeguards to fire
- A default timeout of `TEST_RUNTIME_LIMIT` plus 10 minutes will stop the Jenkins build automatically if there is no user interaction during that time
- The default parameters _excluding the Jenkins job itself_ are set to values that will immediately stop the test if not overriden

* Does this Jenkins job satisfy our load testing needs? Are all scenarios covered by this job?
* Does the solution make sense?
* Are there any additional safeguards that should be added to the job?

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


ops/ansible/playbooks-ccs/build_bfd-server-load.yml
```diff
@@ -0,0 +1,46 @@
...
+    # TODO: Consider removing `env` when SSH/sudoer users have been resolved in SSM BFD-2068
+    env: "test"
+
+  tasks:
+    - name: Download BFD Repository
+      git:
+        repo: https://github.com/CMSgov/beneficiary-fhir-data.git
+        dest: /beneficiary-fhir-data
+        # use hash from `git_commit` when on stable master branch, otherwise use git_branch name
+        version: "{{ git_branch if git_branch != 'master' else git_commit }}"
```
mjburling, line 21: I don't like that this is necessary, but I must have iterated on this a bit to make it more palatable for myself. I'm not sure if anyone else has an opinion, but I suppose my preference is for _x if y is equal_ rather than _x if x is not equal_. What do you think?
```suggestion
        version: "{{ git_commit if git_branch == 'master' else git_branch }}"
```

ops/ansible/playbooks-ccs/launch_bfd-server-load.yml
```diff
@@ -0,0 +1,41 @@
...
+        #server_load_initial_worker_nodes:
+        #server_load_node_spawn_time:
+        #server_load_max_spawned_nodes:
+        #server_load_max_spawned_users:
+        #server_load_user_spawn_rate:
+        #server_load_test_runtime_limit:
+        #server_load_coasting_time:
+        #server_load_warm_instance_target:
+        #server_load_stop_on_scaling:
+        #server_load_stop_on_node_limit:
```
mjburling, line 24: The comment on variable injection via user-data is probably fine, but you might consider dropping the other commented-out lines of variables.

mjburling (commented):
It's bigger change set and I've had my fair share recently. I left a few comments for you to consider, but the ansible and terraform components are _largely_ in-line with what has come before. As discussed, I'm running this against `test` to be sure that we aren't seeing any regressions in the app ami construction stage.

ops/terraform/services/server/server-load/iam.tf
```diff
@@ -1,3 +1,18 @@
...
+# AWS CloudWatch agent needs extra IAM permissions for x-ray
+data "aws_iam_policy" "cloudwatch_agent_xray_policy" {
+  arn = "arn:aws:iam::aws:policy/AWSXRayDaemonWriteAccess"
+}
+
+# TODO: This should be removed as of BFD-1786
+data "aws_iam_policy" "ansible_vault_ro" {
+  arn = "arn:aws:iam::${local.account_id}:policy/bfd-ansible-vault-pw-ro-s3"
+}
+
```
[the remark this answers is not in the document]
malessi: Good point, didn't even cross my mind. Thanks for the suggestion!

ops/ansible/playbooks-ccs/build_bfd-server-load.yml
malessi: I don't feel _too_ strongly either way -- I think if this is the way the other Ansible `build*` playbooks are doing this check then this playbook should also follow that convention. I think, at-a-glance, `x if y == <something>` is more readable than `x if x != <something>` in this instance.

ops/ansible/playbooks-ccs/launch_bfd-server-load.yml
malessi: As usual, I don't feel strongly either way -- this was more-or-less me "cargo culting" the existing Ansible playbooks I built upon. It was already a bit difficult to keep track of everywhere these variables were referenced, so this list is very likely to end up out-of-sync.

I'll remove this list but keep the comment indicating where the variables for the task are set.

mjburling (approved):

ops/ansible/roles/bfd-server-load/test/ansible.cfg
```diff
@@ -0,0 +1,17 @@
...
+# environments.
+allow_world_readable_tmpfiles = True
+
+# More human readable errors when they happen
+stdout_callback = yaml
+
+# Provides summary profiling information
+# NOTE: callback_whitelist is deprecated in ansible >2.9 in favor of callbacks_enabled
+# callbacks_enabled = profile_tasks
+callback_whitelist = profile_tasks
```
lsmitchell, line 17: Given the above comment, should this be using callbacks instead?

ops/ansible/roles/bfd-server-load/test/run-tests.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+
+# Stop immediately if any command returns a non-zero result.
+set -eou pipefail
+
+function help() {
+    echo "This script runs our test cases locally, via Docker."
```
lsmitchell, line 7: Is this script only for running the load tests locally, or is this a helper for testing the load test's architecture?

lsmitchell (commented):
So many configurations! I can't say I'm super familiar with all the ansible, docker, and other infrastructure parts, but nothing outside the comments I noted _looks_ off to me.

ops/ansible/roles/bfd-server-load/test/run-tests.sh
malessi: This script is used only for testing the Ansible playbook(s) for this `server-load` service, and was copied from the same testing scripts from other Ansible roles. So, yep, it's just a helper for testing the Ansible infrastructure-as-code for the load tests.

apps/utils/locust_tests/services/server-load/node/README.md
```diff
@@ -0,0 +1,36 @@
...
+#### Option 1 -- Specifying the `services/server-load/node/node.Dockerfile` from the Root
+
+This option is _preferred_ to option 2, as it is generally better form to run `docker build` from
+the directory of the build context itself.
+
+Ensure you current working directory is `/apps/utils/locust_tests` and run:
+
+```bash
+docker build -f services/server-load/node/node.Dockerfile -t "<your-tag>" --platform linux/amd64 .
+```
```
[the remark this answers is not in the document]
malessi: A [tag](https://docs.docker.com/engine/reference/commandline/tag/) is just a tag that is used to identify a particular Docker image. In this case, `<your-tag>` would likely be replaced with our private [ECR](https://aws.amazon.com/ecr/)'s hostname plus some identifying string. For example, `../node/build-push.sh` sets the tag as the following:

```bash
PRIVATE_REGISTRY_URI="$(aws ecr describe-registry --region "$AWS_REGION" | jq -r '.registryId').dkr.ecr.${AWS_REGION}.amazonaws.com"
IMAGE_NAME_NODE="${PRIVATE_REGISTRY_URI}/bfd-mgmt-server-load-node"
...
DOCKER_TAG="${DOCKER_TAG_OVERRIDE:-"$GIT_SHORT_HASH"}"
...
docker build . \
...
  --tag "${IMAGE_NAME_NODE}:${DOCKER_TAG}" \
```

So, `build-push.sh` will set the Docker image's tag to something like:

```
<PRIVATE_REGISTRY_URI>/bfd-mgmt-server-load-node:<GIT_SHORT_HASH>
```

Hopefully that helps with your understanding.

apps/utils/locust_tests/services/server-load/common/boto_utils.py
```diff
@@ -0,0 +1,118 @@
+"""This module contains various utility functions shared between the node and controller that are
+related to boto3"""
+import json
+import re
+from typing import Any, Dict, List
+
+
+def get_ssm_parameter(ssm_client, name: str, with_decrypt: bool = False) -> str:
```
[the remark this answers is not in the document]
malessi: That's definitely something we could talk about -- I'm not sure how much value vs. effort there is to try and avoid small repetitions like this. I could see the effort being worth it for `boto3` APIs that require [exponential backoff](https://docs.aws.amazon.com/general/latest/gr/api-retries.html) or operations that require require multiple `boto3` clients and resources.

apps/utils/locust_tests/services/server-load/README.md
```diff
@@ -0,0 +1,74 @@
...
+- Whether to stop when the node/Lambda limit is reached or not
+  - Use this as a safeguard against runway Lambda spawning during non-static load tests
+
+By default, _for the Jenkins job_, these parameters are set to values that will continue to ramp up
+the test until a scaling event is encountered at which point the test will "coast" for 10 seconds
+and stop.
+
+Note that builds/runs of this pipeline _do not_ communicate with the controller directly. The
+operator is required to indicate when the load test has finished (or allow the built-in timeout to
+lapse).
```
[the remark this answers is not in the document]
malessi: Much like when we get failures in the regression suite during deployment, once the controller instance has been deployed during a load test Jenkins run the pipeline will await the operator to click either `Proceed` or `Abort` to indicate that the load test is done (those buttons can be seen either in log output or in the overall stage view). Whoever is running the Jenkins job will then monitor the controller logs via CloudWatch or by SSHing into the controller and `tail`ing the `server-load-controller.log` until the load test is finished, and then once the load test is finished the operator can go back to the Jenkins pipeline and click one of those two buttons to cleanup.

Basically, it's on the operator to watch the logs generated by the load test run to see when it finishes. If, for whatever reason, they don't happen to be monitoring the load test run it will timeout on its own 10 minutes after the load test runtime limit.

apps/utils/locust_tests/services/server-load/README.md
```diff
@@ -0,0 +1,74 @@
...
+- SDLC Environment
+- The host to test
+- The number of _initial_ worker **nodes**/Lambdas (_not_ simulated users) to start with before the
+  Locust master starts
+- The amount of time to wait between spawning nodes/Lambdas
+- The maximum number of nodes/Lambdas to spawn
+- The maximum number of _simulated users_ for **Locust** to spawn
+  - This controls the load actually generated on the target
+  - In general, a ratio of 10:1 (10 users to 1 node/lambda) should be followed for the best
+    performance
```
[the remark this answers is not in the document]
malessi: Yeah, I can add this to the description of the simulated users parameter.

We _could_ do that, but I figure it's easier (and more flexible) to just let the operator decide how many nodes to spawn, when to spawn them, how many users per-node, etc. I think it'll allow for more scenarios in the future.

ops/ansible/roles/bfd-server-load/test/ansible.cfg
malessi: It might? Frankly, everything under `bfd-server-load/test/` was copied from the equivalent `test/` directory from the `bfd-db-migrator` Ansible role which also copied from the other Ansible roles within the `ops/ansible/roles/` directory. This appears to have just never been updated along the way, and as far as I can tell has not caused any problems.

@mjburling Should we be concerned with changing this, or just leave it be for now and save it for a future PR?

mjburling: Good question! In short, no. `callback_whitelist` is correct for ansible 2.9.

The configuration here was _largely_ copied from the other roles, where we _eagerly_ added it to assist whoever would be asked to pursue ansible upgrades. We haven't gotten there yet, and we're **still** on ansible 2.9.

lsmitchell (approved):

PR #1406, opened 2022-10-06
PACA-783 (DRAFT - OBSOLETE): phase feature flag

brianburton:
**JIRA Ticket:**
[PACA-783](https://jira.cms.gov/browse/PACA-783)

Add a feature flag to limit the claims returned by the BFD API based on phase (both major and minor), so we can control the level of maturity a claim is at before sending to the user. This is particularly useful if we want to guarantee that the CWF has validated the MBI before sending the claim to a DCE.

---

### What Does This PR Do?

- Adds a new system property, `bfdServer.include.pac.phase.range`, in `SpringConfiguration` that allows a maven-like semantic version number range (e.g. `[1.3,]` to accept any phase greater than or equal to 1.3) used to filter which partially adjudicated claims can be sent to clients.
- Defines a new `ResourceFilter` interface used by `AbstractR4ResourceProvider` to filter claims prior to transforming them.
- Modifies `SpringConfiguration` to create a `ResourceFilter` instance using the system property.  If the system property is unset than all claims will be returned.  If it is set but not a valid range string an error is logged and no claims will be returned.
- Adds classes for working with semantic version strings.  These might also be used in future to control which version of RDA API server the pipeline will interact with.
- Modifies the claim entities and tables to store phase and phase sequence number.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/PhaseResourceFilter.java
```diff
@@ -0,0 +1,35 @@
...
+   * as a semantic version number and ensures that the number falls within the range allowed by this
+   * filter.
+   *
+   * @param resourceType used to interrogate the entity object
+   * @param entity the entity object
+   * @return true if the claim should be returned to clients
+   */
+  @Override
+  public boolean shouldRetain(ResourceTypeV2<?, ?> resourceType, Object entity) {
+    final var phaseVersionString = resourceType.createPhaseVersionString(entity);
```
KyRobbins, line 31: I can see why the `ClaimTypeV2` subclasses were used here, it seems convenient, though Claim/ClaimResponse doesn't play a part in that logic, which is a big part of those interface subclasses.  Wondering if a common interface on the entity would be simpler, then do like `<E extends PhasedEntity>`.

I think this would also have a small side benefit of being able to ditch all the wildcard type declarations we've previously had to use for `ClaimTypeV2` in the provider, etc.

KyRobbins (commented):
Very neat approach, I am wondering how it would differ from doing a `greaterThan` predicate in the DAO, both in complexity and performance.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__add_rda_phase_columns.sql
```diff
@@ -0,0 +1,9 @@
+/*
+ * Adds new columns to the FISS and MCS claims tables to store the phase number
+ * and phase sequence number for use in filtering claims.
+ */
+
+ALTER TABLE rda.fiss_claims ADD phase smallint;
+ALTER TABLE rda.fiss_claims ADD phase_seq_num smallint;
+ALTER TABLE rda.mcs_claims ADD phase smallint;
+ALTER TABLE rda.mcs_claims ADD phase_seq_num smallint;
```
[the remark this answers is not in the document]
brianburton: Not sure yet.  Right now I feel like it's not really necessary and the update statements would take a long time to run.  But before this PR is done maybe we should add them anyway.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractResourceTypeV2.java
```diff
@@ -90,6 +103,27 @@ public ResourceTransformer<TResource> getTransformer() {
...
+   * {@inheritDoc}
+   *
+   * <p>If either phase or phase sequence number are null in the claim they are represented as zero
+   * in the version string.
+   *
+   * @param entityObject object of type TEntity
+   * @return a valid semantic version number
+   */
+  @Override
+  public String createPhaseVersionString(Object entityObject) {
```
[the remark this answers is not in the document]
brianburton: No it had to be an Object because of java type reasons around the fact we're processing captured wildcard types in the resource provider.  Based on your other suggestion though this might become cleaner even with it being Object.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/PhaseResourceFilter.java
brianburton: Thanks for the suggestion.  I was following the original pattern but I agree now might be a good time to revisit that.  I'll take a look.

brianburton: I have simplified the implementation by using an interface to recognize entities to filter.  It removed the need to modify the resource type stuff.  Can't do the generic type suggestion (`<E extends PhasedEntity>`) though because of the wildcard being used in `createBundleFor()`.  I like this implementation though since the filters can be applied to anything and just filter what they can and pass through anything else.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__add_rda_phase_columns.sql
```diff
@@ -0,0 +1,9 @@
+/*
+ * Adds new columns to the FISS and MCS claims tables to store the phase number
+ * and phase sequence number for use in filtering claims.
+ */
+
+ALTER TABLE rda.fiss_claims ADD phase smallint;
```
ericbuckley, line 6: Should we turn these into NOT NULLs in PACA-717?

brianburton: Good question.  According to the `record_source.proto` file these values are optional in the RDA API.  We could default them to 0 or something if they are not populated there.  Also we'd definitely have to backfill the claim tables before doing so.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/PhaseResourceFilter.java
brianburton: OIC - disregard the backfill part of my reply.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__add_rda_phase_columns.sql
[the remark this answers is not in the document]
brianburton: Based on discussion with Eric since PACA-717 will truncate these tables anyway we won't need to backfill existing data.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__add_rda_phase_columns.sql
brianburton: Disregard the backfill part of my reply...

ericbuckley:
@brianburton the SemanticVersionRange is a really flexible concept, however looking forward we really only have a need (or potential need) for greater than or equal to statements (eg  >=2.0).  I believe those could be easily implemented as where clauses in the query.  Did you consider the performance implications for filtering in memory vs RDS?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__add_rda_phase_columns.sql
ericbuckley: Ah I see that it's also optional in the metadata table, https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V95__add_rda_record_source.sql, so never mind I thought RDA had made those fields required.

brianburton:
> @brianburton the SemanticVersionRange is a really flexible concept, however looking forward we really only have a need (or potential need) for greater than or equal to statements (eg >=2.0). I believe those could be easily implemented as where clauses in the query. Did you consider the performance implications for filtering in memory vs RDS?

For the phase we might not need an upper bound but we would need one for PACA-199 (RDA API version check).  This class is meant to be useable for both purposes.  Then again if we ever need to not return fully adjudicated claims maybe an upper bound would become necessary.

I considered the alternatives and felt that filtering in memory was a better fit because:
- The number of claims being excluded will be small compared to the total number of claims returned.
- Our assumption has always been that we would all claims so this does not lower performance from its current level.
- The complexities of implementing this in a CriteriaQuery are much greater.  Consider it's not just a single `greaterThan` since there are two columns.  It would wind up being something like this in sql: `(phase is not null AND phase_seq_num is not null) AND ((phase > 1) OR (phase = 1 AND phase_seq_num >= 10))`.  Now imagine the logic to construct that with chained calls to QueryBuilder.  The potential query performance improvement doesn't seem to justify the extra code complexity.

brianburton:
This PR was superceded by an alternative implementation.

PR #1448, opened 2022-11-01
BFD-2207 : support synthea pre-validation in ETL Pipeline

mackec1:
**JIRA Ticket:**
[BFD-2207](https://jira.cms.gov/browse/BFD-2207)

Whenever the ETL Pipeline detects a Synthea-generated manifest, it will perform a pre-validation step to ensure that the associated manifest and its RIF files will not create a situation where duplicate key(s) and/or hash collisions will occur when writing to the targeted database.

### What Does This PR Do?
This PR adds code and pre-validation processing to the ETL Pipeline whenever Synthea-generated 0_manifest.xml is encountered. It does this by:

- the bfd-manifest.xsd has been modified to include a new optional element, _syntheaEndStateProperties_.
- a new Flyway script that creates a new function, _synthea_load_pre_validate_, is introduced.
- the CcwRifLoadJob, when encountering a _0_manifest.xml_ to run that contains the _isSynthetic=true_ attribute, AND the _0_manifest.xml_ contains an optional XML element, _syntheaEndStateProperties_, will trigger a pre-validation step that asserts that the contents of the manifest should not encounter any duplicate key and/or hash collisions during processing.
  - the job fill be marked as a go/no go as follows:
    - calls the database function,  _synthea_load_pre_validate_, to perform validation of the values (contained in XML elements)
    - if the pre-validation function returns a zero from the database function call, then the pre-validation has passed and the CcwRifLoadJob can continue.
    - if  the pre-validation function returns a non-zero (> 1) from the database function call, then the pre-validation has failed and the CcwRifLoadJob will abort.

* verify that the XML Schema Definition (XSD) can assert not only the validity of the 0_manifest.xml data, but does so in a backwards compatible manner (i.e., where no _syntheaEndStateProperties_ are contained in the manifest).
* verify that the database function, _synthea_load_pre_validate_, does what it is intended to do; that is, range-checking of new synthetic beneficiaries (bene_id) and claims (clm_id); it also asserts that there should not be any MBUI or HICN hash collisions.
* Verify that the CcwRifLoadJob changes are backwards compatible and the efficacy of the changes are straight-forward, easy to understand, and most importantly terminate the RIF load job when an error has been detected.

   * [ ] Yes
    * [x] No
   * [ ] Yes
   * [x] No
   * [ ] Yes
   * [x] No

   * [ ] Yes
   * [x] No

* N/A

### Submitter Checklist


apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__Synthea_Pre_Validation.sql
```diff
@@ -0,0 +1,324 @@
...
+                            (
+                                select count (bene_id) from beneficiaries
+                                where (hicn_unhashed = p_hicn_start or mbi_num = p_mbi_start)
+                                limit 1    
+                            );
+                        END IF;
+
+    -- add support for fi_cntl_num here; not the most efficient step since
+    -- fi_doc_clm_cntl_num is not an indexed field (therefore causing a
+    -- full table scan).
```
lsmitchell, line 247: Is this a TODO?

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/synthea_pre_validation_debug.sql
```diff
@@ -0,0 +1,256 @@
+-- This SQL mimics the function SYNTHEA_LOAD_PRE_VALIDATE, but does so
+-- by breaking apart the large 'union' of values into individual named
+-- values so one can see when/why a Synthea pre-valiation may have failed.
```
lsmitchell, line 3: Good idea! I suppose we can manually run this if something fails, but I wonder if this shouldnt just get run if we fail so it prints out the reason? Unless this takes much longer or some reason we wouldnt want to.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
```diff
@@ -475,4 +507,262 @@ public String toString() {
...
+          && clmGrpIdStart < 0
+          && pdeIdStart < 0
+          && carrClmCntlNumStart < 0
+          && fiDocCntlNumStart < 0
+          && clmIdStart < 0
+          && hicnStart != null
+          && mbiStart != null);
+    }
+
+    /** @see java.lang.Object#toString() */
```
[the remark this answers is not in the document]
mackec1: Yea, I wavered between following what was already in the file vs. the inherit; changed them all.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
```diff
@@ -475,4 +507,262 @@ public String toString() {
       return builder.toString();
     }
   }
+
+  @XmlAccessorType(XmlAccessType.FIELD)
+  public static final class SyntheaEndStateProperties {
+    @XmlElement(name = "bene_id_start", required = true)
+    protected long beneIdStart;
+
+    @XmlElement(name = "bene_id_end", required = true)
```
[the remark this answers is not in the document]
mackec1: changed....however, this is sort of a by-product of bad practice. In theory, we should be generating Java POJO(s) via JAXB consumption of the XSD (maven plugin). In that scenario, you don't get element-level comments.

Added doc to elements.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifestTest.java
```diff
@@ -245,6 +249,140 @@ public void manifestSyntheticIdRoundtrip() {
         s3Key, manifestId.computeS3Key(CcwRifLoadJob.S3_PREFIX_PENDING_SYNTHETIC_DATA_SETS));
   }
 
+  /**
+   * Utility method to invoke {@link DataSetManifestFactory} parse factory that converts an XML stream
+   * of data into {@link DataSetManifest}.
+   * @param xmlStream
+   * @return
```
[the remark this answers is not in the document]
mackec1: done.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/synthea_pre_validation_debug.sql
mackec1: Well I suppose we could, although that becomes kinda heavy-weight (as written) since it performs the entire gamut of checks vs. the actual function which operates on a 'fail fast' model. The thought was to provide some visibility into the number of failures that could be encountered from the end_state.properties data, so it was intended as more of a debugging tool.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__Synthea_Pre_Validation.sql
mackec1: removed; artifact from the original query which was a giant 'union' which didn't work so well in HSQL.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/resources/bfd-manifest.xsd
```diff
@@ -12,6 +12,22 @@
             <xs:attribute name="exportType" type="xs:string" use="optional" />
           </xs:complexType>
         </xs:element>
+        <xs:element name="syntheaEndStateProperties" minOccurs="0" maxOccurs="1">
```
christrotteradhoc, line 15: I thought we were in agreement that this interface should not be coupled to synthea or end-state properties (a synthea concept) and instead be based around the more general concept of pre-validation of a load. Having the pipeline perform optional pre-validation is a better separation of concerns than having it be coupled to synthea concepts which could change. The coupling to synthea concepts should only be present in the jenkins job which has knowledge of synthea and the pipeline.

mackec1: Well sort of; so I can certainly change the name of the element (remove any trace of synthea), as well as remove any biding between the isSynthetic attribute and the checking of the pre-validation elements. However, it is VERY unlikely that upstream sources of a manifest have the same level of constraints or 'pre-knowledge' of meta-data to imbue into a manifest.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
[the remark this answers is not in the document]
lsmitchell: Yeah if they were generated we'd skip them for documentation, but for checkstyle purposes as long as we hand-write something it'll need docs. It would be nice to switch to generating more of this boilerplate code in the future.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/resources/bfd-manifest.xsd
christrotteradhoc: The notion of synthetic data is not synthea specific so I don't see an issue there. This is less about leaving the possibility of another upstream open (although, who knows for sure?) and more about not tightly coupling all of the various components together when they don't need to be.

christrotteradhoc: Taking this conversation a bit further, the individual elements I believe could also use some renaming to support a more intuitive _pipeline_ interface for validation. For instance, clm_grp_id_start in this context doesn't say where the clm_grp_ids for _this_ load start, it says where the clm_grp_ids for the next synthea generated data should start. A more intuitive validation interface would describe this load itself so I suggest naming this clm_grp_id_max or clm_grp_id_end and passing the value from the end state properties minus 1. This keeps the knowledge of synthea as much out of the pipeline as possible.

mackec1: Agreed (mostly); where things get a little hairy when trying to completely synthea-agnostic, is the following:
- the db function that is called performs range checking, much of which is specific to Synthea data; for example a range check of bene_id (_p_beg_bene_id, p_end_bene_id_) results in a where clause _bene_id <= p_beg_bene_id AND bene_id > p_end_bene_id_. This however is ONLY valid for synthetic data which use negative bene_id values; if fact for that same range check for non-synthetic data it would be: _bene_id > p_beg_bene_id AND bene_id <= p_end_bene_id_.
- ideally, for any validation that involves range checking, both the _beg_<whatever column> and _end_<whatever column> shod be passed in.

So I guess one could derive the LT vs. GT value checking by checking the sign of the passed in variable (i.e., p_beg_bene_id):
- p_beg_bene_id < 0 : where bene_id <= p_beg_bene_id
- p_beg_bene_id > 0 : where bene_id > p_beg_bene_id

This is starting to feel like significant AC creep; I'd prefer get this in under a mostly Synthea required functionality and a subsequent ticket to make more agnostic.

christrotteradhoc: My main concern is the interface -- the manifest XML naming and I think that should be part of this so that the dependent work can be completed and not have to be changed later. The internal method naming is secondary and could be changed in a later ticket if that is burdensome. I would not expect any additional validation support beyond what is needed for this use case but the things that are needed for this use case should be well named in the interface (manifest).

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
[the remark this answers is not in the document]
mackec1: but it won't build without making changes to maven; take the branch, change one of the items to {@inheritdoc}  and see if it builds.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__Synthea_Pre_Validation.sql
```diff
@@ -0,0 +1,322 @@
...
+--    p_carr_clm_cntl_num_start => -694670632,
+--    p_fi_doc_cntl_num_start => 423014560,
+--    p_hicn_start => 'T11100073A',
+--    p_mbi_start => '1S00EU2GA73');
+--
+-- NOTE: The final SQL is not as performant as a version using only PSQL;
+--       Unfortunately, HSQL constraints precluded aspects of the PSQL query
+--       so the RSLT variable had to be incremented in a sequenctial manner.
+--
+${logic.psql-only}  CREATE OR REPLACE FUNCTION synthea_load_pre_validate (
```
christrotteradhoc, line 99: can you explain some about why a stored procedure was chosen as opposed to just directly executing the select queries from the pipeline?

mackec1: 1. SQL is significantly long....
  - so it'll need to be compiled in the db engine every time it is invoked
  - it'll take up 100+ lines of code in the .java source
  - as a function it'll be compiled once, when it it created in the db
 2.  the function actually allows for a 'fail fast' mode of processing (i.e., select progression of queries...first one that fails we're out). Yes, that same could be done .java source, but then that would mean making 10+ queries from Java, each of which will need to be compiled before executing.
 3. it allows for distribution of responsibilities....this is entirely a database pre-validation so let the db do that.

mackec1: I can try....it'll mean doing everything that the SQL function does but in Java source. I do know from running Logan's .py (prepare_and_run_synthea) that the pure db function vs. separate queries ran about 1x faster.

So if I am going to muck up that java code to just verify those numbers, then I might as well just put it all in Java.

christrotteradhoc: Well, I think the answer would be that it saves a few milliseconds or seconds at best which would put this in the category of an unnecessary premature optimization. This is only going to run 4 times a year.

mackec1: And I may have underestimated the amount of Java code needed. So the queries themselves will probably up to 100+ lines in Java; then add in the code for CallableStatement and setting variables times the number of calls we'll need to make, probably adds up to 200-300 lines of code in the CcWRifLoadJob.

christrotteradhoc: well there are almost 600 lines between the stored procedure and its supporting test. I really don't think that SPL is any less dense than Java and definitely not to the point of justifying the complexity of splitting code across two systems and languages.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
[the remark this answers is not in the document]
lsmitchell: Not sure what issue you're seeing but.. I grabbed your branch and changed the line to this:
image
and it built fine?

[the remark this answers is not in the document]
mackec1: not sure what I hit when I first tried the @inhertDoc but it is OK now

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
```diff
@@ -814,50 +787,107 @@ public void setMbiStart(String value) {
...
-          && fiDocCntlNumStart < 0
-          && clmIdStart < 0
-          && hicnStart != null
-          && mbiStart != null);
+    public long getClmIdEnd() {
+      return clmIdEnd;
+    }
+
+    /**
+     * Sets the value of the clmIdEnd property.
```
lsmitchell, line 817: Not a huge deal, but I had been writing getters and setters as "Sets the {@link #clmIdEnd)."

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadPreValidateSynthea.java
```diff
@@ -0,0 +1,292 @@
...
+   *     provide the various Synthea end-state property values that can be used to perform the
+   *     pre-validation.
+   * @return {@link boolean} asserting that the pre-validation succeeded (true) or failed (false).
+   */
+  @Override
+  public boolean isValid(Object obj) throws Exception {
+    if (appState == null) {
+      throw new BadCodeMonkeyException("Invalid state; missing PipelineApplicationState");
+    }
+    if (!(obj instanceof DataSetManifest)) {
```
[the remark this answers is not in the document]
mackec1: well trotter wanted this to be completely non-Synthea so I made it entirely up to an implementation to define what/how isValid should work. I replaced the Object with a DataSetManifest but not sure it is now as flexible as it once was. Easy enough to change back to an Object if needed.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadPreValidateSynthea.java
```diff
@@ -0,0 +1,292 @@
...
+        ps.close();
+        ps = null;
+        rs = null;
+        if (!isValid) {
+          LOGGER.warn("pre-validation failed for CHECK_HICN_MBI_HASH");
+          break;
+        }
+      }
+    } catch (Exception e) {
+      isValid = false;
```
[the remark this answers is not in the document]
mackec1: yea, I logged it; not sure we need to throw an exception since the pre-val failure by definition will shut down the pipeline anyways.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
mackec1: done.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadJobIT.java
```diff
@@ -507,6 +518,41 @@ public void skipDataSetTestForFutureManifestDate() throws Exception {
...
+  @Test
+  public void skipDataSetTestForSyntheaPreValidationFailed() throws Exception {
+    validateLoadAtLocations(
+        CcwRifLoadJob.S3_PREFIX_PENDING_SYNTHETIC_DATA_SETS,
+        CcwRifLoadJob.S3_PREFIX_COMPLETED_SYNTHETIC_DATA_SETS,
+        false);
+
+    // job ran and now we should have some Synthea data in the db; we'll re-run
+    // the same dataset but this time specifying Synthea end-state information.
+    Exception exception =
```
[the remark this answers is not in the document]
mackec1: all tests for the PreValidation were moved to RifLoaderIT since all we really care about is vetting data and that class is more suitable spot for the tests.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetManifest.java
```diff
@@ -36,7 +36,7 @@ public final class DataSetManifest implements Comparable<DataSetManifest> {
   @XmlAttribute(name = "sequenceId", required = true)
   private int sequenceId;
 
-  /** A boolean denoting if this is synthetic data (true) or not (false). */
+  /** A boolean denoting if this is synthetic data (true) or not (false}. */
```
lsmitchell, line 39: think this one was touched by mistake

lsmitchell (approved):

cbrunefearless (approved):
lgtm

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V97__Synthea_Pre_Validation.sql
mackec1: moved to Java code.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/resources/bfd-manifest.xsd
mackec1: So the interface as now a full-fledged Java interface that leaves the validation up to the implementation class the implements the interface. The only mention of the word 'synthea' is when instantiating the implementation that is associated with Synthea data. All pre-validation manifest content has been changed to be more inclusive to non-synthea interfaces.

PR #1478, opened 2022-11-17
BFD-2254: mdc resources returned count fix

cbrunefearless:
**JIRA Ticket:**
[BFD-2254](https://jira.cms.gov/browse/BFD-2254)

Bug description:

All FHIR requests are supposed to set the `resources_returned_count` in the MDC to the number of resources returned in the response. However, the EOB calls have some cases where zero is returned even though there are in fact more than zero resources returned in the payload.

Specifically, (using V2 as an example) the number of resources is being set in the MDC [here](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java#L1567) using the size of the bundle but the bundle size is not itself set until later on [here](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java#L1588).

AC:

The issue described above should be corrected to ensure that the EOB endpoint (v1 and v2) always sets the correct value for resources_returned_count
A BFD Insights query running in PROD should be used to perform a validation by ensuring that starting from the application of the fix and going forward for at least 24 hours, that all EOB calls with resources_returned_count set to zero have an output size <= the largest possible empty EOB response (currently believed to be around 350 bytes). See notes for such a query.
BFD insights [documentation](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/insights/terraform/projects/bfd/README.md) should be updated to note the window during which the data was incorrect.
Notes:

This should return zero when the defect is fixed (replace the timestamp with a timestamp after the fix is applied):

---

### What Does This PR Do?

Fixes the mdc resources returned count

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -1586,6 +1581,10 @@ public static Bundle createBundle(
                 ? Date.from(transactionTime)
                 : Date.from(maxBundleDate));
     bundle.setTotal(resources.size());
+
+    // Add number of resources to MDC logs
+    LoggingUtils.logResourceCountToMdc(bundle.getTotal());
```
christrotteradhoc, line 1586: This is going to be incorrect when there are more resources than a page holds and there should be a test for that case.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2Test.java
```diff
@@ -641,4 +663,92 @@ public void addCareTeamMemberWithNpiOrgShouldCreateCareTeamEntry() {
...
+                FdaDrugCodeDisplayLookup.createDrugCodeLookupForTesting(),
+                NPIOrgLookup.createNpiOrgLookupForTesting()),
+            claim);
+    IParser parser = fhirContext.newJsonParser();
+    String json = parser.encodeResourceToString(genEob);
+    List<IBaseResource> eobs = new ArrayList<IBaseResource>();
+    eobs.add(parser.parseResource(ExplanationOfBenefit.class, json));
+
+    Bundle bundle = TransformerUtilsV2.createBundle(paging, eobs, Instant.now());
+    assertEquals(1, bundle.getTotal());
```
lsmitchell, line 705: Might want to make total return 2-3 if its not a big lift, just to ensure this works with many (using 1 doesnt check that multiple are counted correctly)

christrotteradhoc, line 705: Should have an assert for the MDC value

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/TransformerUtilsTest.java
```diff
@@ -266,6 +289,94 @@ public void createReferenceForIdentifierTypeTest() {
...
+    eobs.add(parser.parseResource(ExplanationOfBenefit.class, json));
+
+    Bundle bundle = TransformerUtils.createBundle(paging, eobs, Instant.now());
+    assertEquals(1, bundle.getTotal());
+  }
+
+  /**
+   * Verifies that {@link gov.cms.bfd.server.war.stu3.providers.TransformerUtils#createBundle
+   * (OffsetLinkBuilder paging, List<IBaseResource> resources, Instant transactionTime)} sets bundle
+   * with paging size correctly} correctly.
```
mackec1, line 336: minor nit...comment seems a bit whacked.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2Test.java
```diff
@@ -641,4 +663,92 @@ public void addCareTeamMemberWithNpiOrgShouldCreateCareTeamEntry() {
...
+   * (OffsetLinkBuilder paging, List<IBaseResource> resources, Instant transactionTime)} sets bundle
+   * with paging size correctly} correctly.
+   */
+  @Test
+  public void createBundleWithPaging() throws IOException {
+
+    RequestDetails requestDetails = mock(RequestDetails.class);
+    Map<String, String[]> pagingParams = new HashMap<String, String[]>();
+    pagingParams.put(Constants.PARAM_COUNT, new String[] {"1"});
+    pagingParams.put("startIndex", new String[] {"1"});
```
lsmitchell, line 719: Is this only making one page? Shouldnt there be multiple pages to properly test this works with paging?

lsmitchell (commented):
Minor questions/comments about the tests.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
cbrunefearless: Moved

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2Test.java
cbrunefearless: Done

lsmitchell (approved):
Thanks for adding additional testing!

christrotteradhoc (approved):
Thanks for getting this one turned around quickly Chris!

PR #1485, opened 2022-11-21
BFD-2233: Add the result of validation to the statistics output from the locust tests and use it when creating the baseline for regression

malessi:
**JIRA Ticket:**
[BFD-2233](https://jira.cms.gov/browse/BFD-2233)

As a BFD Engineer, I want the aggregated statistics output (in JSON format) from the regression
suite to properly encode whether validation passed or failed for the corresponding test run so that
only truly valid runs are included in the rolling baseline. As of now, only the result of comparison
against the baseline is stored and checked against.

---

### What Does This PR Do?

This PR adds the result of static _validation_ to the aggregated statistics JSON generated when the
Locust tests are ran and stat storage is specified. A new property, `metadata.validation_result`, is
now included indicating one of three states:

- `NOT_APPLICABLE` -- indicates that validation was not ran; this is the default state
- `FAILED` -- indicates that validation failed
- `PASSED` -- indicates that validation passed

As of now, validation will _always_ check if there are **any** failures (i.e. `5xx` responses). The
validation logic will also check hardcoded percentile SLAs _if specified by the test suite_; for the
regression and load test suites this SLA validation is disabled.

When building a baseline to compare against, the `validation_result`s of candidate aggregated
statistics are checked and only those with `PASSED` or `NOT_APPLICABLE` values are added to the
baseline. Any with `FAILED` values are excluded, therefore removing any possibility of invalid test
runs polluting the baseline. Specifically, this means that if we see regression failures during a
regression suite run we do not need to take any action to exclude those results from the baseline
any longer.

Out-of-band of these changes, the QuickSight Dashboards' Datasets have all been updated to
**exclude** runs that have failed validation. As of writing, there is only one such run in the
`test` environment. Additionally, every JSON file stored in S3 has been updated to have a
`validation_result` of `NOT_APPLICABLE` as it is not possible for us to know if they had or had not
passed validation.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist


lsmitchell (approved):
Only comment would be that "Not Applicable" sounds like it _wouldnt_ be applied to the baseline; i.e. the results are not applicable to the baseline. Would it just be easier to have two values where we default to passed, since NA is functionally equivalent to PASSED? (Other than being a default, i dont see NA being used for anything functionally separate from passed)

PR #1488, opened 2022-11-21
BFD 2223: Mapping Coverage Period Start/End Dates in V2

JeremyIR:
**JIRA Ticket:**
[BFD-2223](https://jira.cms.gov/browse/BFD-2223)

Map period start and end dates in V2

---

### What Does This PR Do?

Create mapping for part A,B,D period dates in V2 via transformer

are the part A,B,D period dates properly mapped and adequately tested?

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


williamharty (approved):
endpoint test resources look good.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -85,9 +85,9 @@ private static Coverage transformPartA(MetricRegistry metricRegistry, Beneficiar
 
     setCoverageStatus(coverage, beneficiary.getPartATerminationCode());
 
-    beneficiary
-        .getMedicareCoverageStartDate()
-        .ifPresent(value -> TransformerUtilsV2.setPeriodStart(coverage.getPeriod(), value));
+    TransformerUtilsV2.setPeriodStart(
+        coverage.getPeriod(), beneficiary.getPartACoverageStartDate());
+    TransformerUtilsV2.setPeriodEnd(coverage.getPeriod(), beneficiary.getPartACoverageEndDate());
```
lsmitchell, line 90: We had checks before to only add these if the values existed; do we want the same here? What happens if nulls get added here? If nothing gets output as before that's fine; but we should double check we dont print out the period with empty values or something in the final output.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
```diff
@@ -512,6 +512,12 @@ public void verifyRelationshipPartD() {
     verifyRelationship();
   }
 
+  @Test
```
lsmitchell, line 515: Missing docs here; i know the rest don't really have them but please add docs at least for new tests. Less for me to fix later.

lsmitchell (commented):
I think we're missing a test case or two here; since these can be null, and will be for much of our data, we should have a test that ensures when the fields are passed as null the mapping handles it without issue and does not output anything for those fields in the fhir response.

PR #1550, opened 2023-01-11
BFD-2102: Implement Jenkins availability pipeline to populate our BFD Server availability SLI

malessi:
**JIRA Ticket:**
[BFD-2102](https://jira.cms.gov/browse/BFD-2102)

As a BFD Engineer, I want an SLI to be captured indicating service availability for the BFD Server so that I can use it for our availability SLOs

---

### What Does This PR Do?

This PR adds two new Jenkins pipelines, one that builds a Docker image, `bfd-availability-checker`,
based on Alpine called `bfd-availability-checker-build` and another that is used to check the
availability of the BFD Server called `bfd-availability-checker`.

The `bfd-availability-checker` Docker image is a small Alpine image that includes `aws-cli` and
`curl`, both needed for use in the `bfd-availability-checker` Jenkins pipeline. As this image is
required for use in the aforementioned pipeline, and we do not yet have a means for defining an
`imagePullSecret` for our Jenkins' k8s cluster, this image is stored in our public ECR.

The main pipeline this PR adds, `bfd-availability-checker`, is a Jenkins pipeline that is intended
to be run once per-minute. This pipeline runs a simple `curl` against each environment's
`/metrics/ping` endpoint looking for a `200 OK` response. If a `200 OK` is returned from the
request, a value of `1` is submitted to the `availability/success` metric under the
`bfd-{ENV}/bfd-server` namespace. Conversely, if a non-`200` response is returned _or_ if `curl`
itself fails a value of `1` is submitted to the `availability/failure` metric in CloudWatch, and the
corresponding Jenkins pipeline build will be marked as `UNSTABLE`.

With these new metrics we will be able to calculate uptime percentages for the BFD Server and thus
start measuring our availability SLOs more accurately.

Note that this pipeline is not currently running at this time. Declarative pipelines are
_post-processed_, therefore the `cron` trigger defined in the `Jenkinsfile` is not active until the
**next** build of `bfd-availability-checker` which I will do once this PR is merged.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist


lsmitchell (commented):
No particular objection, but I don't get why it's creating/discarding a pod every minute instead of just keeping one alive for some set time (24h?) and running a thing every minute within it? I meant to ask yesterday when we were talking about it.

malessi:
> No particular objection, but I don't get why it's creating/discarding a pod every minute instead of just keeping one alive for some set time (24h?) and running a thing every minute within it? I meant to ask yesterday when we were talking about it.

@lsmitchell: just to preface, I'm not much of a fan of using Jenkins like this myself. Jenkins isn't necessarily the best tool for short, frequently-ran jobs like this, but it's the most _convenient_ tool for us at the moment. On the other hand, I don't think creating three very lightweight containers every minute is a _huge_ concern. If I'd have to guess, the overhead of Jenkins orchestrating the pipeline is much, much more significant than starting and tearing down the `alpine`-based containers.

It would be possible to rewrite this pipeline such that it is no longer short-lived. For example, the `cron` trigger could be changed to every hour or every day, and the `Run availability check` stage could run continuously during that period executing availability checks on a shorter interval (i.e. 1 minute, 30 seconds). At the time of creating this pipeline, I didn't do this because it felt more "wrong", as pipelines and jobs should do one thing (build, run some steps, etc.) and exit; adding a mini "scheduler" into a pipeline complicates this and makes the process more brittle. Additionally, since Jenkins handles the scheduling and the process is so short if the pipeline fails for any reason that failure does not affect subsequent availability checks.

However, I have spent a bit of time refactoring the pipeline according to the above paragraph and I believe it is also a valid way to implement this process. Comparing the two:

| Method | Pros | Cons |
| --- | --- | --- |
| As-is | - Short-lived, single-purpose pipeline<br>- Easy to reason about<br>- Self-scheduling using Jenkins' `cron` trigger<br>- Failures with the pipeline _itself_ are also short-lived and only affect the single run; the next run of the pipeline will be unaffected by a previous failure | - A lot of unnecessary orchestration overhead for a process that is basically a two-line shell script (`curl` then `aws cloudwatch put-metric-data`) -- whether this overhead is _meaningful_ is something we have to decide  |
| Long-running `Run availability check` stage | - Significantly reduced overhead (assuming it's a concern), even when running the pipeline overall on an hour interval -- this is the **biggest** benefit with this solution<br>- Similar to (1) but less calls to SSM<br>- Availability checks could execute on an interval shorter than 1 minute<br> | - _Slightly_ more complicated code<br>- Breaks the "do one thing then exit" semantics that pipelines should follow<br>- Requires our own "scheduler" (in quotes because it's a fairly simple loop) within a scheduled (via the `cron` trigger) pipeline<br>- More prone to failure -- if a run fails midway we would require additional failure handling to restart the pipeline and not lose any data |

I am going to give this some thought as I am not entirely sure which one is preferable right now.

malessi:
After doing some more thinking I believe keeping the pipeline as-is for the time-being will work fine for us. After adjusting each parallel run (each `matrix` `axis`, i.e. environment) to run on the same pod instead of separate pods the pipeline now runs consistently in about 30 seconds including scheduling, SCM checkout, and pipeline steps. By a _significant_ margin the longest step is downloading the TLS certificate from SSM, so I would have to assume that the overhead of starting the pod and running the pipeline is minimal.

If for some reason we find that the current solution does not work for us I have kept the refactored "long-running" version.

mjburling (approved):

lsmitchell (approved):
Thanks for the extra thought; I'll defer to your much more detailed and well-thought out opinion here.

PR #1565, opened 2023-01-24
PACA-911: Implement bad claim handling configuration

KyRobbins:
**JIRA Ticket:**
[PACA-911](https://jira.cms.gov/browse/PACA-911)

---

### What Does This PR Do?

To better support future situations of bad claims coming from RDA API, this PR adds in the ability to set a configurable limit to the number of bad claims that can be skipped before the job is stopped.

Claims that are skipped due to a transformation error will be stored in the `message_errors` table and retried on every restart of the job process prior to continuing normal claim ingestion.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaSink.java
```diff
@@ -83,16 +84,24 @@ default int writeMessage(String dataVersion, TMessage object) throws ProcessingE
...
     throw new UnsupportedOperationException();
   }
 
   /**
-   * Write all of the objects to the data store and return the number of objects actually written.
+   * Checks if the error limit has been exceeded.
+   *
+   * @throws ProcessingException If the error limit was reached.
+   */
+  void checkErrorCount() throws ProcessingException;
```
brianburton, line 101: Since this is only called from within `writeError` it does not need to be in the `RdaSink` interface.  It's just an implementation detail.

Alternatively if it's left in the interface it should be called right after `writeError` by users of the `RdaSink`.  I can see an argument for `writeError` not throwing if it successfully writes the error since the write did not itself fail. I think it should be one or the other though.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -142,15 +146,36 @@ public void updateLastSequenceNumber(long lastSequenceNumber) {
...
+   * @throws ProcessingException If the error limit was reached.
+   */
+  @Override
+  public void checkErrorCount() throws ProcessingException {
+    var query =
+        entityManager.createQuery(
+            "select error from MessageError error where status = :status", MessageError.class);
+    query.setParameter("status", MessageError.Status.UNRESOLVED);
+    int errorCount = query.getResultList().size();
+
```
brianburton, line 175: It would be better to select the count rather than the records here.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -271,20 +296,24 @@ private void updateLastSequenceNumberImpl(long lastSequenceNumber) {
...
   /**
-   * Transforms all of the messages in the collection into {@link RdaChange} objects and returns a
+   * Transforms all the messages in the collection into {@link RdaChange} objects and returns a
    * {@link List} of the converted changes.
    *
    * @param apiVersion appropriate string for the apiSource column of the claim table
    * @param messages collection of RDA API message objects of the correct type for this sync
    * @return the converted claims
-   * @throws DataTransformer.TransformationException if any message in the collection is invalid
+   * @throws ProcessingException if there was an issue transforming the message
```
brianburton, line 305: Maybe "if any message in the collection triggered an error"

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/concurrent/ClaimWriterThread.java
```diff
@@ -266,20 +269,22 @@ private void reportError(Exception error) throws InterruptedException {
...
+        final Optional<TClaim> claim =
+            sink.transformMessage(entry.getApiVersion(), entry.getObject());
+
+        if (claim.isPresent()) {
+          allMessages.add(entry.getObject());
+          uniqueClaims.put(claimKey, claim.get());
         }
-        throw transformationException;
+      } catch (DataTransformer.TransformationException transformationException) {
+        sink.writeError(entry.getApiVersion(), entry.getObject(), transformationException);
```
brianburton, line 287: Won't `sink.transformMessage()` have already called `writeError` internally?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -298,26 +327,25 @@ List<RdaChange<TClaim>> transformMessages(String apiVersion, Collection<TMessage
...
-      try {
-        writeError(apiVersion, message, transformationException);
-      } catch (IOException e) {
-        transformationException.addSuppressed(e);
-      }
-      throw transformationException;
+      writeError(apiVersion, message, transformationException);
     }
+
+    return Optional.empty();
```
brianburton, line 348: I'd either move this into the `catch` or return a variable that's set within the `try`.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaSink.java
brianburton: Never mind I see it being called from `retrieveAndProcessObjects`.

KyRobbins: This is actually also called from `StandardGrpcRdaSource` to check it before running a job: https://github.com/CMSgov/beneficiary-fhir-data/pull/1565/files/29ae213fd33ca7453a40a97268865c011e13334f#diff-1ca50d94e4cca6e1b14974692b5ce9bcb0324bf5955bd2b90ac4e17d6b4f5b2eR179

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -298,26 +327,28 @@ List<RdaChange<TClaim>> transformMessages(String apiVersion, Collection<TMessage
...
     } catch (DataTransformer.TransformationException transformationException) {
       metrics.transformFailures.increment();
-      try {
-        writeError(apiVersion, message, transformationException);
-      } catch (IOException e) {
-        transformationException.addSuppressed(e);
-      }
-      throw transformationException;
+      writeError(apiVersion, message, transformationException);
+      result = Optional.empty();
```
brianburton, line 348: Previously the transformation exceptions were being logged when they were thrown.  Now they won't be.  I think we should add a logging statement either here in the catch block or else in the `writeError` method to be sure the log shows all of the encountered errors.

brianburton (approved):
Looks good.  Nice work.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -142,15 +146,36 @@ public void updateLastSequenceNumber(long lastSequenceNumber) {
...
+   */
+  @Override
+  public void checkErrorCount() throws ProcessingException {
+    var query =
+        entityManager.createQuery(
+            "select error from MessageError error where status = :status", MessageError.class);
+    query.setParameter("status", MessageError.Status.UNRESOLVED);
+    int errorCount = query.getResultList().size();
+
+    if (errorCount > errorLimit) {
```
daryll, line 176: silly nit-pick...
do we wanna consider when we've reached the limit rather than past it ?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -142,15 +146,36 @@ public void updateLastSequenceNumber(long lastSequenceNumber) {
...
+   * Checks if the allowed number of unresolved {@link MessageError}s in the database was exceeded.
+   *
+   * @throws ProcessingException If the error limit was reached.
+   */
+  @Override
+  public void checkErrorCount() throws ProcessingException {
+    var query =
+        entityManager.createQuery(
+            "select error from MessageError error where status = :status", MessageError.class);
+    query.setParameter("status", MessageError.Status.UNRESOLVED);
```
daryll, line 173: curious...does every claim (or batch) processed make this query ?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -142,15 +146,36 @@ public void updateLastSequenceNumber(long lastSequenceNumber) {
...
+    int errorCount = query.getResultList().size();
+
+    if (errorCount > errorLimit) {
+      throw new ProcessingException(new IllegalStateException("Error limit reached"), 0);
```
daryll, line 177: silly nit-pick...
do we wanna consider when we've reached the limit rather than past it ?
(errorCount >= errorLimit)

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/DLQGrpcRdaSource.java
```diff
@@ -216,20 +216,20 @@ Processor dlqProcessingLogic(
...
-        try {
-          sink.shutdown(Duration.ofMinutes(5));
-        } catch (Exception ex) {
-          if (processResult.getException() != null) {
-            processResult.getException().addSuppressed(ex);
-          } else {
-            processResult.setException(ex);
-          }
+      try {
+        sink.shutdown(Duration.ofMinutes(5));
```
daryll, line 222: this number in minutes seems hidden.
can we make it a static variable ?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
KyRobbins: This method (and subsequent query) should only be executed at the start of each job, and when `writeError()` has been invoked, which only occurs when a transformation error happened.  In 99.99% of cases, we won't get any transformation errors, so the query will only get executed once when the pipeline job first starts.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
KyRobbins: My definition of `errorLimit` here is that it's the maximum allowed number of errors.  If you were to set `errorLimit` to `0`, it would mean "Stop running if there are any errors".  Using `errorCount >= errorLimit` would cause it to fail even when there are no errors in the database.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/DLQGrpcRdaSource.java
KyRobbins: This is just logic carried over from the existing code, but I can swap that for a constant, as well as the `StandardGrpcRdaSource`, which has the same invocation in it.

daryll:
LGTM!

many thx!

brianburton (approved):

mackec1 (approved):

PR #1571, opened 2023-01-26
BFD-1862: Configure alerts and pages when disk space starts running low

malessi:
**JIRA Ticket:**
[BFD-1862](https://jira.cms.gov/browse/BFD-1862)

As an on-call/BFD Engineer I want to know when one or more of our BFD Server instances' disk
utilization exceeds expected numbers so that I can address it immediately and avoid a possible
incident.

---

### What Does This PR Do?

This PR adds a new Lambda, `bfd-{ENV}-manage-disk-usage-alarms`, that is invoked on `LAUNCH` and
`TERMINATE` AutoScaling Group Notifications from an SNS topic specific to the Lambda. When a valid
event/notification is received, the Lambda will create metric alarms for all `InService` EC2
Instances in the current environment's ASG that _do not_ already have disk usage metric alarms.
Additionally, the Lambda will cleanup "orphaned" alarms for instances that are no longer
`InService`. These metric alarms transition to the `ALARM` state when the disk usage of a given EC2
Instance exceeds 95% full; we can discuss whether this is an appropriate threshold or if it should
be lower. Similarly to our existing SLO alarms, transitions to the `ALARM` state will post a
notification in Slack, and once BFD-2244 is complete it will also send an alert to VictorOps.

Out-of-band of these changes the new `disk_usage_alarms` Terraform module has been applied to the
`test` environment and verified to work as expected.

Note that this solution was chosen due to some surprising behavior from both AutoScaling
Notifications and AutoScaling EventBridge events:

- The latter requires that a [lifecycle hook](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lifecycle-hooks.html)
  is added to each ASG that, essentially, blocks the deployment of instances to the `InService`
  state until the hook is finished. Assuming the execution of the Lambda failed, it could delay or
  potentially stop instances from joining the ASG during scale-out. Also, the other, non-lifecycle
  hook actions did not appear to do anything, as well.
- The former does not fire `TERMINATE` events whenever an instance is detached manually (or
  otherwise) from the ASG. Also, while it is possible to discern whether an instance is being launched to
  the `WarmPool` or to the ASG itself, it requires additional code that would likely be brittle.
  Still, `TERMINATE` not covering detach makes relying on the `LAUNCH`/`TERMINATE` events alone a
  non-starter. AutoScaling Notifications are still used by this solution, but the Lambda only uses them as an invocation signal and nothing more.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist


mackec1 (approved):

JeremyIR (approved):

ops/terraform/modules/resources/disk_usage_alarms/main.tf
```diff
@@ -0,0 +1,192 @@
...
+  ]
+}
+EOF
+}
+
+resource "aws_iam_policy" "autoscaling" {
+  name        = "bfd-${var.env}-${local.lambda_name}-autoscaling"
+  description = "Permissions for bfd-${var.env}-${local.lambda_name} to describe ASGs"
+  # Unfortunately AWS does not support anything but wildcarding for the resource defition for the
+  # DescribeAutoScalingGroups action
```
mjburling, line 94: As far as the copy is concerned, there's a minor typographical error:
```suggestion
  # Unfortunately AWS does not support anything but wildcarding for the resource definition for the
  # DescribeAutoScalingGroups action
```
Regarding the content, I trust that this doesn't support `Condition` elements, either?

ops/terraform/modules/stateless/main.tf
```diff
@@ -305,4 +305,9 @@ module "bfd_dashboards" {
 module "cw_alarms_slack_notifier" {
   source = "../resources/cw_alarms_slack_notifier"
   env    = var.env_config.env
-}
```
mjburling, line 308: ❤️ Your editor configuration has improved. Thank you!

mjburling (commented):
Well done!

I think this is perfectly reasonable, given some of the limitations you've discovered. The 95% utilization is a _fine_ starting point, considering there's virtually nothing in place today short of an operator (attempting) connection to the instance for discovery.

ops/terraform/modules/resources/disk_usage_alarms/main.tf
malessi: > Regarding the content, I trust that this doesn't support `Condition` elements, either?

I hadn't thought of using this, so I gave it a try by adding a `Condition` to restrict the `Describe` action to only ASGs with the `Name` tag corresponding to `bfd-${env}-fhir` (using both `autoscaling:ResourceTag/Name` and `aws:ResourceTag/Name`). Unfortunately, when trying to run the Lambda with the conditional policy, the Lambda failed due to lacking permission each time. Therefore, I don't think it is possible to apply a `Condition` to `autoscaling:Describe*` actions.

For what it's worth, the only thing I can think of applying a `Condition` on is the `Name` tag so that we would only be able to `Describe` the corresponding environment's AutoScaling Group. Is there another `Condition` you had in mind to restrict this further?

Also, there is some conflicting documentation on this with respect to `autoscaling:Describe*` actions (see [this examples page](https://docs.aws.amazon.com/autoscaling/ec2/userguide/security_iam_id-based-policy-examples.html) which lists _multiple_ examples stating that "you must specify them [`Describe*` actions] in a separate statement **without conditions**", but [this page](https://docs.aws.amazon.com/autoscaling/ec2/userguide/control-access-using-iam.html#policy-auto-scaling-condition-keys) shows that `autoscaling:*` actions _do_ support `Conditions` on `ResourceTag`s and the like) so I don't exactly know if the restrictions on `Condition` for `autoscaling:Describe*` is specific to certain conditions like `aws:ResourceTag`/`autoscaling:ResourceTag` or `Condition` as a whole.

mjburling: Seems like this is a bit of a wart. Thanks for taking a closer look. I also appreciate the in-line comment calling it out. Let's just leave it alone.

mjburling (approved):

PR #1578, opened 2023-02-01
BFD-2316: Resolve Issues with DB Timeout

billie-rose:
**JIRA Ticket:**
[BFD-2316](https://jira.cms.gov/browse/BFD-2316)

As a BFD Endpoint User, I would like to fetch the pertinent patient data using the part D contract id and year-month without experiencing an ambiguous error.

---

### What Does This PR Do?

Resolve the DB timeout issue occurring as a result of using an aggregate `COUNT` vs `EXISTS` when verifying that records exist.

This PR is a direct resolution to [BFD-2257: Investigate bulk request DB timeout (Patient)](https://jira.cms.gov/browse/BFD-2257). We found that the DB was timing out during the preflight `COUNT` check for querying patients by part d contract id and year-month. Using `COUNT` to verify that >= 1 record exists can be problematic with exceptionally large tables, due to having to scan the table for the result. We can work around this by using `EXISTS` instead, which will stop execution as soon as a single record is found.

This PR was tested against a data load similar to production and demonstrated a substantial performance increase.

Sample:
```
/v1/fhir/Patient?_IncludeIdentifiers=true&_has:Coverage.extension=https://bluebutton.cms.gov/resources/variables/ptdcntrct12|S5601&_has:Coverage.rfrncyr=https://bluebutton.cms.gov/resources/variables/rfrnc_yr|2021&_count=4000&_format=application/fhir+json
```

Pre-PR:
`30054680709` nanoseconds execution time and DB timeout.

Post-PR
`177466692` nanoseconds execution time. No timeout.

Similar results were seen for `v2`.

This PR also resolves the Log WARN: `No matching query type for query: select count(*) as col_0_0_ from beneficiary_monthly beneficiar0_ where beneficiar0_.year_month=? and beneficiar0_.partd_contract_number_id=?`. 

**Note** that JPA doesn't readily support the `EXISTS` clause when it is not associated with a `Root`, hence why we must select an arbitrary `BeneficiaryMonthly` instead of `true` or `false`.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


mackec1 (commented):
billie, will there be a flyway script that accompanies the code changes?

billie-rose:
Just code :) Still need to update QueryLoggingListener and test a bit more.

mjburling (changes requested):
This looks really promising. No specific changes at this time, but I'd just like to isolate this a bit ahead of merging it in by deploying these changes to the `test` environment, and running some additional regression and load suites there. What do you think, @billie-rose?

billie-rose:
> This looks really promising. No specific changes at this time, but I'd just like to isolate this a bit ahead of merging it in by deploying these changes to the `test` environment, and running some additional regression and load suites there. What do you think, @billie-rose?

Makes sense to me; it would make the most sense to run the tests against the 2277 DB or one similar to it in size. What do you think?

mjburling:
> > This looks really promising. No specific changes at this time, but I'd just like to isolate this a bit ahead of merging it in by deploying these changes to the `test` environment, and running some additional regression and load suites there. What do you think, @billie-rose?
> 
> Makes sense to me; it would make the most sense to run the tests against the 2276* DB or one similar to it in size. What do you think?

Good point–I'd forgotten that we still had pieces of the 2276 environment available, that _could_ satisfy some of the things I'd like to look at... 🤔

I think that the existing tests in unit/integration are _probably_ answering a number of the questions I might have, but as far as the updates to bfd insights process... I'd just like to see what's there.

ops/terraform/services/server/insights/api-requests/glue.tf
```diff
@@ -1158,17 +1158,17 @@ module "glue-table-api-requests" {
...
     },
     {
-      "name"    = "mdc_jpa_query_bene_count_by_year_month_part_d_contract_id_duration_nanoseconds",
+      "name"    = "mdc_jpa_query_bene_exists_by_year_month_part_d_contract_id_duration_nanoseconds",
       "type"    = "string",
       "comment" = ""
     },
     {
-      "name"    = "mdc_jpa_query_bene_count_by_year_month_part_d_contract_id_record_count",
+      "name"    = "mdc_jpa_query_bene_exists_by_year_month_part_d_contract_id_record_count",
```
mjburling, line 1171: I'll have to meditate on this further, but we might want to reconsider what merely _appears_ to be a rename. I believe this is somewhat destructive in nature, e.g. a deletion followed by a creation. Granted, it's merely `test`, but I can't query the data that should otherwise be available because the glue table doesn't include the old count fields.

That said, the solutions looks **very** promising with the limited data you've produced so far!

billie-rose: I feel as though it may be a bandaid to retain these fields for historic reasons (and I think you already know this), so is the solution to add these back in and address that concern in a different task? or do you think it's not too problematic to keep these fields in the glue.tf when they don't appear elsewhere in the codebase?

mackec1 (approved):

ops/terraform/services/server/insights/api-requests/glue.tf
mjburling: As discussed, we elected to keep the old fields and expand the comment field. Both fields are available–this looks great!

mjburling (approved):

PR #1609, opened 2023-02-21
PACA-199: Implement an RDA API Version check

KyRobbins:
**JIRA Ticket:**
[PACA-199](https://jira.cms.gov/browse/PACA-199)

---

### What Does This PR Do?

Implements a feature to allow for Version checking of the RDA API.  If the version of the RDA API does not meet the configured requirements of the job, the job will not ingest the data.

This acts as a safeguard to prevent the pipeline from ingesting data that it is potentially not designed to correctly process yet.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -579,6 +586,8 @@ static RdaLoadOptions readRdaLoadOptionsFromEnvironmentVariables(IdHasher.Config
     readEnvParsedOptional(ENV_VAR_KEY_RDA_JOB_STARTING_MCS_SEQ_NUM, Long::parseLong)
         .ifPresent(jobConfig::startingMcsSeqNum);
     readEnvBooleanOptional(ENV_VAR_KEY_PROCESS_DLQ).ifPresent(jobConfig::processDLQ);
+    jobConfig.rdaVersion(
+        RdaVersion.builder().versionString(readEnvStringRequired(ENV_VAR_KEY_RDA_VERSION)).build());
```
brianburton, line 590: We should default this to `"~" + RdaService.RDA_PROTO_VERSION` or something so that build scripts don't break if they fail to include this env var.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
```diff
@@ -0,0 +1,181 @@
...
+        CompatabilityLevel level = CompatabilityLevel.PATCH;
+
+        // Array includes: { Level, Major, Minor, Patch }
+        Integer[] versionParts = new Integer[] {0, 0, 0, 0};
+        int stage = 0;
+
+        // Go through each character and parse out the version parts
+        for (int i = 0; i < versionString.length() && stage < versionParts.length; ) {
+          char c = versionString.charAt(i);
+
```
brianburton, line 124: I think a regex would be easier to understand here.  Something like 

```
var regex = Pattern.compile("(?<type>[\\^~]?)(?<major>\\d+)(\\.(?<minor>\\d+))?(\\.(?<patch>\\d+))?");
var matcher = regex.matcher(whatever);
if (!matcher.matches()) {
  throw ...
}
```
Then you get the values right from the matched groups.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
```diff
@@ -0,0 +1,181 @@
...
+      case MAJOR:
+        compatable = rdaVersion.major == major;
+        break;
+      case MINOR:
+        compatable = rdaVersion.major == major && rdaVersion.minor == minor;
+        break;
+      default:
+        compatable =
+            rdaVersion.major == major && rdaVersion.minor == minor && rdaVersion.patch == patch;
+    }
```
brianburton, line 71: Not sure this logic matches the NPM semver spec.

https://semver.npmjs.com

For example `~` says: "include everything greater than a particular version in the same minor range" whereas with this logic `~0.10.3` would accept `0.10.2`.  I think for MINOR we should add `rdaVersion.patch >= patch`.

brianburton: MAJOR should probably do the same kind of thing too I think.  Check both minor and patch to be `>=` the set value.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
KyRobbins: Now defaulting to using the `RdaSerice.RDA_PROTO_VERSION` defined version, matching MAJOR version (^)

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersionTest.java
```diff
@@ -0,0 +1,101 @@
...
+        Arguments.arguments("^1.0.0", "0.0.1", false),
+        Arguments.arguments("^1.0.0", "0.1.0", false),
+        Arguments.arguments("^1.0.0", "1.0.0", true),
+        Arguments.arguments("^1.1.0", "1.0.0", false),
+        Arguments.arguments("~0.0.1", "0.0.1", true),
+        Arguments.arguments("~0.0.1", "0.1.0", false),
+        Arguments.arguments("~0.0.1", "1.0.0", false),
+        Arguments.arguments("~0.1.0", "0.0.1", false),
+        Arguments.arguments("~0.1.0", "0.1.0", true),
+        Arguments.arguments("~0.1.0", "1.0.0", false),
```
brianburton, line 76: Shouldn't there be a case like `Arguments("~0.1.0", "0.1.1", true)`?  What about checking that `^0.1.0` and `0.2.0` is true?

brianburton (approved):
Looks good to me!

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
```diff
@@ -1,8 +1,21 @@
...
 
+  private static final String TYPE_GROUP = "type";
+  private static final String MAJOR_GROUP = "major";
+  private static final String MINOR_GROUP = "minor";
+  private static final String PATCH_GROUP = "patch";
+  private static final Pattern VERSION_PATTERN =
+      Pattern.compile(
+          String.format(
+              "^(?<%s>[~^])?(?<%s>\\d+)\\.(?<%s>\\d+)\\.(?<%s>\\d+)$",
+              TYPE_GROUP, MAJOR_GROUP, MINOR_GROUP, PATCH_GROUP));
```
daryll, line 22: do we need to be concerned w/ "SNAPSHOT" as well in this regex ??

https://cmsgov.slack.com/archives/C01KG2D14TH/p1677265519295269?thread_ts=1677264955.121689&cid=C01KG2D14TH

KyRobbins: A snapshot version in production would be highly undesired, having the process fail to parse it would seem sufficient.

daryll (dismissed):
looks good
thk u!

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/AbstractGrpcRdaSource.java
```diff
@@ -218,6 +223,21 @@ protected int submitBatchToSink(
...
+   * ingest data from.
+   *
+   * @param apiVersionString A string representing the RDA API version being checked.
+   * @throws IllegalStateException If the apiVersionString represents an API version that can't be
+   *     ingested.
+   */
+  protected void checkApiVersion(String apiVersionString) {
+    if (!rdaVersion.allows(apiVersionString)) {
+      throw new IllegalStateException(
+          String.format("Can not ingest data from API running version '%s'", apiVersionString));
```
lsmitchell, line 237: Could this not be handled with normal control flow instead of an exception? It seems like this is a standard validation and not really an "exception" case.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
```diff
@@ -0,0 +1,172 @@
...
+    RdaVersion rdaVersion = builder().versionString(versionString).build();
+
+    boolean compatable;
+
+    switch (compatability) {
+      case MAJOR:
+        compatable =
+            rdaVersion.major == major
+                && ((rdaVersion.minor == minor && rdaVersion.patch >= patch)
+                    || rdaVersion.minor > minor);
```
lsmitchell, line 84: Per semver rules, shouldnt a mismatch on minor and patch not cause breaking changes? Is this just being extra careful? Seems like if we're matching on major, we'd just check the major was good, and even if they are beyond on the minor/patch it should still work theoretically.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
```diff
@@ -0,0 +1,172 @@
...
+          }
+        } else {
+          throw new IllegalArgumentException(
+              String.format("Invalid RdaVersion format '%s'", versionString));
+        }
+
+        return new RdaVersion(major, minor, patch, level);
+      } else {
+        throw new NullPointerException("versionString can not be null");
+      }
```
lsmitchell, line 169: 🍰 Could reduce nesting by checking this first and just throwing the exception if versionString=null instead of having nesting all the other logic and having this be the else. (Same with the other validations)

mackec1 (approved):
Looks good!

lsmitchell (commented):
Fairly straightforward, just a couple questions/comments. 

Also didn't see any explicit new tests to check for the error case in various places where the version is illegal. I see adding the arbitrary version to various existing tests, but not where we pass an explicitly bad one and expect the exception.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
brianburton: We were following thn npm version range syntax:  https://semver.npmjs.com

lsmitchell: But this can be called on a non-range version too, per the docs? In which case even the above spec just uses normal semver (per https://semver.org/ , linked in that npm semver page). Seems like it'd need to discern if a range is being used, then use the more 'restrictive' check if so, else just use normal semver where a major would accept all minors and patches as compatible.

Just basing this off the test and hardcoded proto version not using a range ~/^ indicator.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/AbstractGrpcRdaSource.java
KyRobbins: Having the RDA API version not match the desired range is not something we currently expect to happen frequently, which arguably makes it an exceptional circumstance, and one that we want everything to stop for.

The higher logic is processing claims and returning a processing result, which can contain a processed claims count or an exception, which returning an exception vs throwing one wouldn't be much of a difference (since the logic later throws that exception).

We could add a new code path that handles it without exceptions, but that would likely mean adding logic to several layers on the way back up, which arguably greatly increases complexity.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
brianburton: We chose to allow either an exact match (no `^` or `~`) or a range match.  The default config setting is using `~` so if no environment variable is set to override that we will be using a range.

KyRobbins: This logic, and the way we intend to use it, would be such that if we define `^0.12.1`, the code will continue to work all the way up to `1.0.0`.  It will fail if we get a `1.x.x` or if we get anything below `0.12.1` (even `0.12.0`), which is the desired behavior.

It's unlikely we'll use anything but the `^`, which matches only higher or equal versions within the same major, but we wanted to make it a little more flexible / easily extendable just in case.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
KyRobbins: It's a styling choice, it can be argued that "fast fail" creates isolated path branches that can lead to potential bugs introduced, where "happy-path first" increases nesting depth.  I've always been more partial to the latter.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
lsmitchell: Ok I see, there's an implicit default exact match. Could we document the above information in one of the relevant locations?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/RdaVersion.java
lsmitchell: Fair enough; I've always seen the fast fail method pushed for but it is ultimately a stylistic choice.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/AbstractGrpcRdaSource.java
brianburton: My $.02 would be that server incompatibility is a very legitimate exception condition.  Very much like data transformation errors are.  We have explicit checks for both cases but we expect them to happen very rarely and use exceptions to make the code cleaner and easier to understand.  Similar case can be made for `Integer.parseInt()` throwing rather than returning an error code.

I disagree that fatal errors throwing exceptions makes the code harder to understand.

Now if we were doing functional programming here and using error monads that would be a different story.  But then the language would provide tools to facilitate that.

KyRobbins: How would one define "ability for the code that's processing to be able to handle the error scenario with the context it has"?  If the statement here is that an event that should cause the entire pipeline job to halt is "recoverable", then what wouldn't be a recoverable?

lsmitchell: My concern was more the graceful handling of the error and being able to return something meaningful to the caller, not necessarily the recoverability of it. I agree that this is a fatal error and isnt recoverable to the pipeline; my only quibble was making sure the way the failure handled continues being understandable to future maintainers.

This is mostly stemming from my real headache of currently getting pipeline to exit gracefully when it fails; it uses exceptions often for its failure states but due to layered complexity over time the exception never makes it to the place the writer thought it would and doesnt do what was probably intended. Over time since exceptions invisibly bubble up to a place the flow is not easy for someone to understand/notice when adding new layers of functionality.

Anyway, not going to necessarily die on this hill or block the PR. Just pointing out some issues I'm seeing today to prevent possible issues in the future.

lsmitchell (approved):

brianburton (approved):
Excellent work!

ops/ansible/roles/bfd-pipeline/templates/bfd-pipeline-service.sh.j2
```diff
@@ -42,6 +42,9 @@ export RDA_JOB_WRITE_THREADS='{{ data_pipeline_rda_job_write_threads }}'
 {% if data_pipeline_rda_process_dlq is defined %}
 export RDA_JOB_PROCESS_DLQ='{{ data_pipeline_rda_process_dlq }}'
 {% endif %}
+{% if data_pipeline_rda_version is defined %}
+export RDA_JOB_RDA_VERSION='{{ data_pipeline_rda_version }}'
+{% endif %}
```
mjburling, line 47: Overall, I think this PR's description could better serve everyone with a little extra detail. Perhaps the minor deviation from the stated requirements under PACA-199 (_The RDA Version will be a required environment variable, with no default_) might be a good seam to look a little further. To be clear, I think your solution is better 😉 

Checking my understanding here of the proposed change... are the following statements accurate?
* the environment variable `RDA_JOB_RDA_VERSION` functions as an external, operational override
* as an override value, it will not have a _sensible_ default
* when this environment variable is absent, the application yields to the version encoded in the application's `RdaService` `RDA_PROTO_VERSION` field

If the above is accurate, I have just two lingering questions if you wouldn't mind entertaining them:
* glancing at the history of the `RDA_PROTO_VERSION` field, it doesn't look like this has been updated as frequently as the overall changes supporting updates/upgrades i.e. obviously, this change set updates the value from 0.10.0 to 0.12.0, but the RDA-based pipeline jobs already support 0.12.0. Overall, does this change set introducing the RDA API Version check make `RDA_PROTO_VERSION` a more meaningful setting?
* I'm largely ambivalent, but I do favor solutions that allow for configuration like this one. I'm curious what  a _most likely_ scenario might be where we would want to override the value of the RdaService's `RDA_PROTO_VERSION` field with the content of `RDA_JOB_RDA_VERSION`. Is this potentially to support incremental patch versions and the like?

KyRobbins: You are right on the `RDA_JOB_RDA_VERSION` being an override feature, the primary mechanism is the internal hardcoded value.  This was a bit of an evolution on the concept after the ticket was written, with the argument being that the code is really tied to a specific RDA version (with certain transformations needing to exist, etc.), so hardcoding it seemed more appropriate, but we wanted to leave a potential hook if the need arrises for some reason.

As for the `RDA_PROTO_VERSION` hardcoded in the app, previously that was used for local testing, acting as the version used by the mock RDA API we use.  The actual value this was set to did not actually affect anything I don't think, we just tried to keep it in sync with the actual API for testing reasons.  Now it's being used as the default value for `RDA_JOB_RDA_VERSION` (which again, is the preferred way to get this value), so it's value is more important now.  It would not be necessary to update it on every API update, as you noticed we skipped over 0.11.0 because that update didn't involve any actual changes to the API messages.

The way we use it is also more like version ratcheting, where the API can go ahead and release 0.14.0 whenever, as long as it's not a breaking change (which minor changes should not be breaking), and we should still be happy consuming it with our service set to allow `^0.12.0`.  If catastrophe happened on the upstream API though and they slipped back to like `0.11.0`, it would stop that from being consumed.

To be honest I don't have a good use case off hand for the external override of the env variable, more one of those, just in case you need it, situations.  It also makes it a little more consistent configuration wise, while we are _technically_ hardcoding the value, the configuration is loaded just like every other configuration, which also just makes it a bit more flexible of a design.

mjburling: OK. Thanks, Kyle. I'll meditate on how I'd want to capture some of this as _knowledge_. I may come back to edit the content of this PR's body, but I'lll go ahead and approve for now!

mjburling (approved):

PR #1619, opened 2023-02-27
BFD-2180: Upgrade to AWS SDK V2

billie-rose:
**JIRA Ticket:**
[BFD-2180](https://jira.cms.gov/browse/BFD-2180)

As a BFD Engineer, I want to use more modern versions of the AWS SDK so that I can avoid using deprecated methods and more easily find relevant documentation, examples, and functionality.

I suggest [viewing this PR with the `?w=1` flag](https://github.com/CMSgov/beneficiary-fhir-data/pull/1619/files?w=1) to avoid seeing the `pom.xml` whitespace changes.

---

### What Does This PR Do?

Update all references to AWS SDK to its latest version (V2)

* Introduction of a new dependency - Cloudwatch2 versus Cloudwatch, which is required for our present AWS SDK usage.

      * [X] Yes
      * [ ] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [X] Yes
      * [ ] No

Note: Although this is technically "just" an upgrade to an existing library, it does overhaul quite a few things and is worth looking into security-wise as if it were a new dependency.

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineApplicationIT.java
```diff
@@ -205,19 +206,20 @@ public void smallAmountOfRifData() throws IOException, InterruptedException {
...
-              manifest,
-              manifest.getEntries().get(0),
-              StaticRifResource.SAMPLE_A_BENES.getResourceUrl()));
-      s3Client.putObject(
-          DataSetTestUtilities.createPutRequest(
-              bucket,
-              manifest,
-              manifest.getEntries().get(1),
-              StaticRifResource.SAMPLE_A_CARRIER.getResourceUrl()));
+      DataSetTestUtilities.putObject(s3Client, bucket, manifest);
```
mackec1, line 209: This seems a little weird; we have a single manifest that has 2 entries (BENEFICIARY and CARRIER). Not sure why we push the manifest twice; a single manifest can support 'n' entries (which is the case here) but we 'push' 2 manifests to the bucket. If this is intended, then maybe a comment to that effect would be appropriate.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/extract/subset/DataSetSubsetter.java
```diff
@@ -51,6 +47,14 @@
...
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.transfer.s3.S3TransferManager;
+import software.amazon.awssdk.transfer.s3.internal.DefaultS3TransferManager;
+import software.amazon.awssdk.transfer.s3.model.CompletedFileDownload;
+import software.amazon.awssdk.transfer.s3.model.DownloadFileRequest;
+import software.amazon.awssdk.transfer.s3.model.FileDownload;
+import software.amazon.awssdk.transfer.s3.progress.LoggingTransferListener;
 
 /**
  * Given some RIF data sets, extracts a random subset of beneficiaries and their claims. This is
```
mackec1, line 60: This class is definitely a candidate for /dev/nul; the CSV writer support is sketchy at best, especially with the recent change(s) in how the models are created (DSL vs. excel). Sorry you had to deal with this when it is not even used.

billie-rose: Do we think this should be removed as part of this PR or as a larger effort for cleanup related to its functionality?

mjburling: > Do we think this should be removed as part of this PR or as a larger effort for cleanup related to its functionality?

This PR is on the larger side, but I think you've done well in keeping it focused. I've opened [BFD-2594](https://jira.cms.gov/browse/BFD-2594) as a placeholder to address some of this obsolete code.

ops/terraform/modules/resources/prefix_list/main.tf
```diff
@@ -4,7 +4,7 @@
 // See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ec2_managed_prefix_list
 // for help on referencing an existing prefix list.
 
-// Provision the prefix list with optinal entry blocks.
```
mjburling, line 7: While I don't disagree with the correction, I don't believe it belongs _here_ 😉

billie-rose: was on the fence about this, but you're right

apps/bfd-pipeline/pom.xml
```diff
@@ -15,6 +15,11 @@
 		A parent POM for the bfd-pipeline-* projects, which migrate data from the CCW data warehouse to the Beneficiary FHIR Data systems.
 	</description>
 
+	<properties>
+		<awssdk.core.version>2.20.50</awssdk.core.version>
+		<awssdk.transfer.manager.version>2.20.50</awssdk.transfer.manager.version>
```
mjburling, line 20: Personally, I'm still a little unclear on some of the finer points with managing poms. Could this instead inherit from the `bfd-parent`'s `pom.xml` and use `${aws-java-sdk-bom.version}`, e.g.

```suggestion
		<awssdk.core.version>${aws-java-sdk-bom.version}</awssdk.core.version>
		<awssdk.transfer.manager.version>${aws-java-sdk-bom.version}</awssdk.transfer.manager.version>
```

Or... generally inherit from `${aws-java-sdk-bom.version}` skipping `awssdk.core.version` and `awssdk.transfer.manager.version` all together?

billie-rose: I am as well, lol. I did consider that, but wasn't sure if it was always the case that a new release and bumping the `bom` version would also bump the other dependencies. I figured being explicit was probably better, but if it's the case that they'll always share the same version number, totally down to change it.

billie-rose: But. to answer your question - yes, I believe it can all be inherited.

billie-rose: I'm just going to update it...it's not like we wouldn't discover any errors at the time of building if the dependencies didn't exist.

mjburling: [This](https://github.com/aws/aws-sdk-java-v2#importing-the-bom) it seem like you could go ahead and import the bom at the top level pom and just specify the services we want in the child modules **without** the identifying the version, and it will take care of itself. Is that your read?

billie-rose: ...that's very obvious in hindsight. updated.

StewGoin (dismissed):
👍 from me -- I'm almost always a fan of being on the more recent SDKs, etc...

apps/bfd-pipeline/bfd-pipeline-app/pom.xml
```diff
@@ -1,228 +1,244 @@
...
+    <description>
+        An application that pulls CCW RIF data via S3, transforms it, and pushes it to
+        the Blue Button API backend database server.
+    </description>
+
+    <properties>
+        <micrometer.cloudwatch.version>1.10.4</micrometer.cloudwatch.version>
+    </properties>
+
+    <dependencies>
```
brianburton, line 23: This should live in the root pom.xml.

apps/bfd-pipeline/bfd-pipeline-app/pom.xml
```diff
@@ -1,228 +1,244 @@
...
+            <!-- Used to transmit metrics to JMX for local testing. -->
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-jmx</artifactId>
+        </dependency>
+        <dependency>
+            <!-- Used to transmit metrics to CloudWatch for alerts and dashboards. -->
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-cloudwatch2</artifactId>
+            <version>${micrometer.cloudwatch.version}</version>
+        </dependency>
```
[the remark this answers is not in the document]
brianburton: Actually this already exists in the root pom (line 238) it's just the wrong artifactId now. Just edit that one, move the version up there too and remove the version from here.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
```diff
@@ -203,6 +203,30 @@
...
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3-transfer-manager</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>aws-core</artifactId>
+        </dependency>
```
brianburton, line 224: This may not be necessary here (see comment in bfd-pipeline-shared-utils pom below).

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadJob.java
```diff
@@ -370,32 +379,28 @@ private boolean dataSetIsAvailable(DataSetManifest manifest) {
...
+    /*
+     * Pull the object names from the keys that were returned, by
+     * stripping the timestamp prefix and slash from each of them.
+     */
+
+    Set<String> namesForObjectsInPage =
+        s3ObjectListingPaginator.stream()
+            .flatMap(s -> s.contents().stream())
+            .peek(s -> LOGGER.debug("Found file: '{}', part of data set: '{}'.", s, manifest))
+            .map(k -> k.key().substring(dataSetKeyPrefix.length()))
```
brianburton, line 401: This flatmap is publishing s3 objects now rather than their keys.  That's bad since it could result in the object itself being logged.  I suggest changing it to this:

```
        s3ObjectListingPaginator.stream()
            .flatMap(s -> s.contents().stream())
            .map(o -> o.key())
            .peek(k -> LOGGER.debug("Found file: '{}', part of data set: '{}'.", k, manifest))
            .map(k -> k.substring(dataSetKeyPrefix.length()))
            .collect(Collectors.toSet());
```

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetQueue.java
```diff
@@ -137,16 +138,24 @@ public void updatePendingDataSets() {
...
-    if (!s3TaskManager.getS3Client().doesObjectExist(options.getS3BucketName(), manifestS3Key)) {
+    HeadObjectRequest headObjectRequest =
+        HeadObjectRequest.builder().bucket(options.getS3BucketName()).key(manifestS3Key).build();
+    try {
+      s3TaskManager.getS3Client().headObject(headObjectRequest);
+    } catch (NoSuchKeyException | NoSuchBucketException e) {
+      LOGGER.warn(
+          "Unable to find keyspace {} in bucket {} while scanning for manifests.",
+          manifestS3Key,
+          options.getS3BucketName());
```
brianburton, line 154: We weren't loggin this before.  Why log it now as a warning?

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/s3/SharedS3Utilities.java
```diff
@@ -119,143 +182,140 @@ public static Bucket createTestBucket(AmazonS3 s3Client) {
...
+      Consumer<S3Object> s3ObjectConsumer) {
+    ListObjectsV2Iterable listObjectsV2Paginator =
+        s3Client.listObjectsV2Paginator(listObjectsV2Request);
 
-      s3BucketListRequest.setContinuationToken(s3ObjectListing.getNextContinuationToken());
-    } while (s3ObjectListing.isTruncated());
-    s3Client.deleteBucket(bucketName);
+    listObjectsV2Paginator.stream()
+        .flatMap(s -> s.contents().stream())
+        .forEach(k -> s3ObjectConsumer.accept(k));
```
brianburton, line 292: Nice addition.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/S3Utilities.java
```diff
@@ -14,9 +15,19 @@ public final class S3Utilities {
...
+
+  /**
+   * Create an Async S3 client.
+   *
+   * @param options the {@link ExtractionOptions} to use
+   * @return the {@link S3AsyncClient} client to use
+   */
+  public static S3AsyncClient createS3AsyncClient(ExtractionOptions options) {
+    return SharedS3Utilities.createS3AsyncClient(options.getS3Region());
+  }
```
brianburton, line 32: Why not just inline these and remove this class?  It appears to be obsolete to me.

apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineApplicationIT.java
billie-rose: Right - I *did* push it twice (on accident) in a previous commit but that was before opening the PR. I'm only seeing it pushed once, and then the individual data files thereafter, like Brian is saying as well.

apps/bfd-pipeline/bfd-pipeline-rda-grpc-apps/src/main/java/gov/cms/bfd/pipeline/rda/grpc/apps/RdaServerApp.java
```diff
@@ -78,7 +78,7 @@ private Config(String[] args) {
               .fissClaimJsonFile(config.readableFileOption("file.fiss").orElse(null))
               .mcsClaimJsonFile(config.readableFileOption("file.mcs").orElse(null))
               .s3Bucket(config.stringOption("s3.bucket").orElse(null))
-              .s3Region(config.enumOption("s3.region", Regions::fromName).orElse(null))
+              .s3Region(Region.of(config.stringOption("s3.region").orElse(null)))
               .s3Directory(config.stringOption("s3.directory").orElse(""))
```
brianburton, line 82: I think this should be something like:

```
.s3Region(config.stringOption("s3.region").map(Region::of).orElse(null))
```

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/S3DirectoryDao.java
```diff
@@ -134,9 +139,18 @@ public List<String> readFileNames() {
...
+      eTag =
+          s3Client
+              .getObject(GetObjectRequest.builder().bucket(s3BucketName).key(s3Key).build())
+              .response()
+              .eTag();
+    } catch (NoSuchKeyException ex) {
+      var fileNotFound = new FileNotFoundException(fileName);
+      fileNotFound.addSuppressed(ex);
+      throw fileNotFound;
+    }
```
brianburton, line 153: Please do not undo abstractions.  This should go into `readS3ObjectMetaData`.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/server/S3DirectoryDaoIT.java
```diff
@@ -234,23 +231,24 @@ public void testGetObjectMetaDataForMissingFile() throws Exception {
...
    * @param bucket the bucket to receive the file
    * @param objectKey the key for the object
    * @param fileData a string uploaded as a file
    * @return eTag assigned to the file by S3
    * @throws IOException pass through if anything fails
    */
-  private String uploadFileToBucket(
-      AmazonS3 s3Client, Bucket bucket, String objectKey, String fileData) throws IOException {
+  private String uploadFileToString(
+      S3Client s3Client, String bucket, String objectKey, String fileData) throws IOException {
```
brianburton, line 242: How does one upload a file to a string?  I think the name fell victim to a search/replace. :-)

brianburton (changes requested):
This was a large effort and looks like it was well done!  

Just needs a few tweaks.  Please restore the abstractions in S3DirectoryDao and take a look at some of the suggested pom changes.

apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineApplicationIT.java
mjburling: > Right - I _did_ push it twice (on accident) in a previous commit but that was before opening the PR. I'm only seeing it pushed once, and then the individual data files thereafter, like Brian is saying as well.

I've taken an even closer look... and I agree! This isn't an issue any longer. I'm probably just using GitHub _wrong_–it was leading me to believe there were still two pushes of the manifest. Sorry for _my_ confusion, friends! Resolving...

billie-rose:
@brianburton Thank you for your review! I *think* I got everything you commented on - can you take a quick look for me please?

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/S3DirectoryDao.java
```diff
@@ -406,53 +412,61 @@ public InputStream openStream() throws IOException {
...
+    /* Gather information needed to prepare the S3TransferManager */
+    String bucketLocation =
+        s3Client
+            .getBucketLocation(GetBucketLocationRequest.builder().bucket(s3BucketName).build())
+            .locationConstraintAsString();
+    // bucketLocation is null if the location is us-east-1, so we need to handle that result
+    // specifically
+    Region region =
+        Region.of(
+            StringUtils.isNotBlank(bucketLocation) ? bucketLocation : Region.US_EAST_1.toString());
```
brianburton, line 455: Not a comment on this PR.

Wow!.  What a mess in the AWS API.  Very disappointing.  Let's build in a NPE vulnerability with a specific region...

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/S3DirectoryDao.java
```diff
@@ -406,53 +412,61 @@ public InputStream openStream() throws IOException {
...
+    DownloadFileRequest downloadFileRequest =
+        DownloadFileRequest.builder()
+            .getObjectRequest(getObjectRequest)
+            .destination(tempDataFile)
+            .addTransferListener(LoggingTransferListener.create())
+            .build();
+
+    FileDownload downloadFile = s3TransferManager.downloadFile(downloadFileRequest);
+    return downloadFile.completionFuture().join().response();
   }
```
brianburton, line 471: This method previously handled AWS not found errors by throwing `FileNotFoundException`.  We need to preserve that behavior in a manner appropriate to the new API.

Catch the `CompletionException` and use an instanceof on the exception's cause to determine whether to switch to a `FileNotFound`.

Add the throws back in and change the end of the method to this:

```
    try {
      FileDownload downloadFile = s3TransferManager.downloadFile(downloadFileRequest);
      return downloadFile.completionFuture().join().response();
    } catch (CompletionException e) {
      final var cause = e.getCause();
      if (cause instanceof NoSuchKeyException || cause instanceof NoSuchBucketException) {
        final var fileName = convertS3KeyToFileName(s3Key);
        final var fileNotFound = new FileNotFoundException(fileName);
        fileNotFound.addSuppressed(e);
        throw fileNotFound;
      } else {
        throw e;
      }
    }
```

billie-rose: I originally didn't include this since we were handling it in the other helper method which was already being called prior to this one (therefore the file definitely exists by the point this method is called), but I can understand your point and do agree in the end (unless the above changes your mind). I can make the change though - a defensive approach is probably best anyways.

billie-rose: All set - I fully agree with you after all, let me know if this looks good.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetQueue.java
billie-rose: I switched this to a debug statement - I think it's valuable info and I *know* I didn't write this original log statement, so it did exist in the codebase at one point in time. I think having it as a debug is okay? If you feel strongly about it I can remove it, though.

brianburton (approved):
Looks good!  Nice work.

mjburling (approved):

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/S3DirectoryDao.java
```diff
@@ -406,53 +413,73 @@ public InputStream openStream() throws IOException {
...
-      throws IOException {
+  private GetObjectResponse downloadS3Object(String s3Key, Path tempDataFile)
+      throws FileNotFoundException {
+    /* Gather information needed to prepare the S3TransferManager */
+    String bucketLocation =
+        s3Client
+            .getBucketLocation(GetBucketLocationRequest.builder().bucket(s3BucketName).build())
+            .locationConstraintAsString();
+    // bucketLocation is null if the location is us-east-1, so we need to handle that result
+    // specifically
```
lsmitchell, line 454: Is it only us-east-1 that has null for bucketLocation? I know we chatted once about being able to flip to west for fallover purposes, so wanted to make sure this wasnt hardcoded to east if it would happen in other configurations.

lsmitchell (commented):
Some questions/comments; also some subtle differences in the implementation with the new API; did we attempt some sample data loads and ensure everything is working how we expected?

Nice work getting all these miscellaneous and subtle changes working for the updated API.

lsmitchell (commented):
Late review @billie-rose didnt get to get through this before it was merged.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/extract/s3/DataSetQueue.java
```diff
@@ -137,16 +138,24 @@ public void updatePendingDataSets() {
...
       DataSetManifest.DataSetManifestId manifestId, String manifestS3Key) {
-    DataSetManifest manifest;
-
     /*
-     * If the keyspace we're scanning doesnt exist, bail early (This can happen if we're loading synthetic data,
+     * If the keyspace we're scanning doesnt exist, bail early (This can happen if
+     * we're loading synthetic data,
      * as it checks the regular incoming folder for the manifest first.)
      */
-    if (!s3TaskManager.getS3Client().doesObjectExist(options.getS3BucketName(), manifestS3Key)) {
```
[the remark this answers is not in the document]
billie-rose: Yes :(

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/server/S3DirectoryDao.java
billie-rose: Correct, it's `locationConstraintAsString` specifically which returns null when the region is us-east-1. I'll update the comment to make this clearer

PR #1620, opened 2023-02-28
BFD-2329: Fix some issues during pipeline shutdown and add logging

lsmitchell:
**JIRA Ticket:**
[BFD-2329](https://jira.cms.gov/browse/BFD-2329)

Fixes some of the issues that cause the pipeline to not shut down correctly when encountering an error. However, this PR does not completely fix the issue; there is still a thread blocking the final shutdown even after fixing these issues. Also adds lots of logging to help diagnose what's happening around and during shutdown when we have an unrecoverable error.

---

### What Does This PR Do?

- Adjusted job failure to stop the pipeline instead of throwing an exception; this was one reason the pipeline would lock up on an error, the exception halted execution but didn't trigger the shutdown logic, so it was just dead in the water
- Fixed a bug with the "waiting for jobs to stop" message logic where it would never repeat as intended, even after the elapsed time
- Reduced the "waiting for jobs to stop" message timer to a minute to help know the pipeline hasnt locked up and give more frequent feedback during shutdown
- Added thread and job info to the "waiting for jobs to stop" message during shutdown for better visibility into whats holding it up
- Adds try/catches around RIF processing calls to stop the metrics (and add logging) in the event of an unrecoverable error while parsing the RIF
- Adjusted a confusing message when a job is (intentionally) cancelled before running during shutdown to more accurately report the job was cancelled and did not 'fail'
- Shut down the metrics loggers when the pipeline fails to reduce noisy metric logging during shutdown (and avoid confusion where it looks like the pipeline is still processing when it's not)
- Added a line in the S3 task manager shutdown to safely stop all downloads (so we can more gracefully shut down and avoid hanging threads)
- Remove noisy log about scanning for keyspaces related to synthetic data scans which occurred even when the application is running correctly
- Adds some handles to help clean up/debug during an error scenario (s3 task manager and job executor's thread pool)
- Add more logging in the event of a pipeline shutdown to help debug what is happening and what went wrong
- Add mockito dependency for pipeline (IT tests)

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -502,11 +568,12 @@ public PipelineJobOutcome call() throws Exception {
             "PipeLineJobOutcome interrupt failed with the the following: " + e.getMessage(), e);
         throw new InterruptedException("Re-firing job interrupt.");
       } catch (Exception e) {
+        // This will print the job failure + exception stacktrace
         handleJobFailure(jobRecord.getId(), e);
-
-        // Wrap and re-throw the failure.
-        throw new Exception("Re-throwing job failure.", e);
+        // If we've blown up without explicit recovery possible, stop the pipeline
+        stop();
```
lsmitchell, line 574: This was causing one of the primary issues; throwing this exception was killing the pipeline's execution, but without shutting the pipeline down, so it would just hang (with its metrics reporters still going looking as if it was still working).

With this adjustment we shut the pipeline down explicitly in the event of an error, however the pipeline is stuck on waiting for a thread in this case while still prevents proper shutdown. (More investigation needed to discern the remaining thread that's causing the job executor to be stuck from here)

apps/bfd-pipeline/bfd-pipeline-app/src/main/resources/logback.xml
```diff
@@ -20,9 +20,10 @@
 
 	<!-- At 'trace', Hibernate will log SQL parameter values. -->
 	<logger name="org.hibernate.type" level="info" />
-	
-	<!-- Log each SQL statement run by Flyway, to give us a rough progress indicator. Note that this may need to be revisited when/if our Flyway scripts get longer/larger: dozens of extra log events are fine but thousands are probably not. -->
-	<logger name="org.flywaydb.core.internal.dbsupport.SqlScript" level="debug" />
```
lsmitchell, line 25: No more flyway in pipeline, this is no longer needed.

apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineManagerIT.java
```diff
@@ -29,20 +30,26 @@
...
+import org.mockito.MockitoAnnotations;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /** Integration tests for {@link PipelineManager}, {@link PipelineJobRecordStore}, and friends. */
 public final class PipelineManagerIT {
   private static final Logger LOGGER = LoggerFactory.getLogger(PipelineManagerIT.class);
 
+  /** Mocks the S3 task manager, which is only used in error scenarios. */
+  @Mock S3TaskManager mockS3TaskManager;
```
lsmitchell, line 43: With flyway 5, we can mock this final class now; hurrah!

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -277,7 +302,7 @@ private void handleJobFailure(PipelineJobRecordId jobRecordId, Exception excepti
         jobRecordStore.recordJobFailure(jobRecordId, new PipelineJobFailure(exception));
         jobsEnqueuedHandles.remove(jobRecordId);
       }
-      LOGGER.error("Handle job failure in Pipeline: " + exception.getMessage(), exception);
+      LOGGER.error("Job failure in Pipeline: " + exception.getMessage(), exception);
```
mackec1, line 305: Curious if you tried to log EMERGENCY here; if things go south in the ETL, not even sure why we try to wait/cleanup threads....the thing is hosed.

mackec1 (approved):

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
lsmitchell: Yeah I'm also not sure why them emphasis on the thread cleanup.. I assume it's to try and not leave us half-loaded and in a bad state, but we'd probably either rollback or reload in idempotent mode anyhow. Just trying to keep the general idea of the shutdown process in case it was important (since I have no idea if there was a good reason not to just kill everything the moment the pipeline chokes.)

mjburling (approved):
Very reasonable! I think it's an improvement. LGTM.

PR #1643, opened 2023-03-13
BFD-2476: Execute Partially Adjudicated Smoketest During Deployments

malessi:
**JIRA Ticket:**
[BFD-2746](https://jira.cms.gov/browse/BFD-2746)

As a BFD Engineer I want the Regression Suite to exercise the BFD Server as fully as possible in order to test for regressions in both performance and behavior

---

### What Does This PR Do?

This PR adds the PACA smoke tests to the existing Regression Suite Lambda such that they will run immediately after the existing Regression Suite during deployment and so the outcome of the smoketest will influence the result of the corresponding deployment stage.

These smoke tests differ from the existing Regression Suite in that they use random, non-synthetic data and thus do not have consistent performance characteristics from environment-to-environment or run-to-run; thus, statistics from these smoke tests are _not_ collected as part of our existing Regression Suite performance baseleine. Only the final validation result is considered with respect to the PACA smoketests, and validation fails if there are _any_ failing test cases. These tests will _always_ run with the same static configuration of 10 synthetic users for 10 seconds.

It is expected that a more comprehensive set of PACA endpoint tests will be _merged_ with the existing Regression Suite when BFD-2490 is complete. This is currently blocked until a consistent set of synthetic data is available to test against, as _performance_ regression testing requires such data.

In addition to adding the smoke tests to the Regression Suite Lambda, several code improvements have been made:

1. The entrypoint for stats-related operations, in `bfd_user_base.py`, has been refactored such that the code's behavior is more obvious with respect to failure cases and conditional logic
2. The cases where the Locust process exits with error code `1` have been made much more obvious and have been centralized to `bfd_user_base.py`
3. Multiple type hinting errors and linting errors have been corrected
4. Pylance/Pyright's `strict` type-checking mode has been enabled with some rules explicitly disabled to compromise between safety and practicality (as several of the libraries used do not export proper types, or encourage "best practices" that trigger some rules)
5. Multiple errors with respect to the Regression Suite Lambda Docker image's build process have been corrected:
   1. The `Dockerfile.dockerignore` is now actually respected
   2. Authentication errors with Amazon's Public ECR have been corrected when using `build-push.sh`
   3. Unnecessary image tag pushes have been minimized

This PR has been tested by building a new image and applying it to `test`'s Regression Suite Lambda and verifying that, when run, the Lambda runs both the Regression Suite and PACA smoketests.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist

### DIFF
--- a/apps/utils/locust_tests/common/bfd_user_base.py
+++ b/apps/utils/locust_tests/common/bfd_user_base.py
@@ -9,14 +9,16 @@ from typing import Callable, Dict, List, Optional, Union
 
 from locust import FastHttpUser, events
 from locust.argument_parser import LocustArgumentParser
+from locust.contrib.fasthttp import ResponseContextManager
 from locust.env import Environment
 
 from common import custom_args, data, validation
 from common.locust_utils import is_distributed, is_locust_worker
 from common.stats import stats_compare, stats_writers
-from common.stats.aggregated_stats import StatsCollector
+from common.stats.aggregated_stats import FinalCompareResult, StatsCollector
 from common.stats.stats_config import StatsConfiguration
 from common.url_path import create_url_path
+from common.validation import ValidationResult
 
 _COMPARISONS_METADATA_PATH = None
 """The path to a given stats comparison metadata JSON file for a particular test suite. Should be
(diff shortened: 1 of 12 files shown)


apps/utils/locust_tests/common/bfd_user_base.py
```diff
@@ -161,7 +182,11 @@ def get_by_url(self, url: str, headers: Optional[Dict[str, str]] = None, name: s
             catch_response=True,  # type: ignore -- known Locust argument
         ) as response:
             if response.status_code != 200:
-                response.failure(f"Status Code: {response.status_code}")
+                if isinstance(response, ResponseContextManager):
+                    # pylint: disable=E1121
```
JeremyIR, line 186: Im not as familiar with pylint. What purpose does this comment serve?

malessi: At this point, `response` can be either an instance of `ResponseContextManager` or `FastResponse`. `ResponseContextManager` implements _both_ `failure()` and `failure(exception: str | Exception)` whereas `FastResponse` implements _only_ `failure()`. Therefore, prior to this change, _both_ Pyright/Pylance and Pylint were reporting a possible error here where if `response` was a `FastResponse` calling `response.failure("...")` would throw.

Unfortunately, only Pylance/Pyright seems to understand that line 185 implicitly _coerces_ `response` to `ResponseContextManager` whereas Pylint seems to still believe that `response` could be `FastResponse` if the `isinstance` check is `true`. So, to stop Pylint from reporting false positives, the rule that corresponds with that particular error ([`too-many-function-args`](https://pylint.pycqa.org/en/latest/user_guide/messages/error/too-many-function-args.html)) is disabled using the comment on line 186.

Hopefully that gives some context as to what the comment does.

JeremyIR: Got it

mjburling (dismissed):
Logic appears sound and the container builds _just_ fine with some nice little improvements along the way. LGTM.

mackec1 (approved):
LGTM

JeremyIR (dismissed):
LGTM

PR #1648, opened 2023-03-15
BFD-2329: Final touches to make pipeline gracefully shut down in the event of an error

lsmitchell:
**JIRA Ticket:**
[BFD-2329](https://jira.cms.gov/browse/BFD-2329)

Shuts down the pipeline gracefully in the event of a parse or sql error during loading.

---

### What Does This PR Do?

This fixes two problems which were causing the pipeline application to slowly (72 hour wait) or never shutdown in the event of a field constraint issue or sql error while loading data.

One of the main issues was the use of errorHandlers in RifLoader: The rif loader job is passed a handle to a shutdown method in PipelineApplication to use in the event of an error. The handle was called from threads that process batches of lines in RifLoader. However, invoking this callback from one of the batch sub-threads triggered pipeline to shutdown (using system.exit) which in turn triggers the shutdown mechanism for pipelineManager. PipelineManager waits for all jobs to finishing processing their files before shutting down, however the job's sub-thread that triggered the shutdown in the first place is still invoking the original callback to pipeline application and never finishes, causing a deadlock (as the rif loading job blocks for up to 72 hours waiting for all the sub-threads to complete). Since there is no timeout value for the shutdown loop, it would need to wait the full 72 hour timer on the RifLoader job before the job would time out and the pipeline could shut down. This was solved by removing the error callback, and letting the batch thread simply exit (after logging the error) and complete. This allows us to resolve all the batch threads, and avoid the 72-hour wait to wrap up the job in the event of an error. One last snag is that doing this makes the RifLoader assume the job was successful, so an additional thread-shared volatile boolean acts as a final check to flag the job as failed if any batch thread sets it to true, letting us perform a last check to blow up the job instead of acting as if it completed.

Another issue that was fixed was an incorrect assumption about how the pipeline worked when the graceful shutdown design was implemented. The pipeline works to throw an exception when an unrecoverable error is found; this is meant to bubble up to the pipelineApplication, where an exception handling mechanism shuts everything down using system.exit(). PipelineManger handles the case where a job has failed due to an unhandled exception just fine, but from there was re-throwing the exception hoping it would reach the PipelineApplication to shut everything down. Instead, the exception is lost, as it seems the way that job handlers are threaded does not bubble the exception properly back to the application. This meant in the event of a parse error, the PipelineManager would log and throw the exception correctly, but the final step of shutting the pipeline down would never trigger as the exception never reached PipelineApplication. This was remedied by calling PipelineApplication's shutdown procedure (separated from the exception handler) explicitly when a job failure is handled by PipelineManager.

Major changes: 
- Split PipelineApplication's system.exit call into its own method to make it clearer what we're doing when we explicitly want to shut the pipeline down
- Adjusted the PipelineManager.onFailure to shut the application down explicitly when we catch an unexpected exception, since the exception thrown from the job is in its own thread and never makes it back to the PipelineApplication when thrown
- Adjusted RifLoader to keep track of a thread-shared boolean so it can know to short-circuit the 72-hour wait for batch load timer when we have a SQL error during batch loading (This fixes the application needing to wait for the full 72 hour timeout before shutting down in the event of a sql exception)

Minor changes:
- Reverted some unnecessary changes in the call error handler (PipelineApplication)
- Removed some of the less helpful debug messages about which threads are alive (this wasn't really helping after all)
- Fixed tests to catch the system exit pipeline does during a job failure, as system exit very much breaks junit tests
- Fixed test to account for rifloader change where errors in the load threads indirectly throw a failure exception instead of trying to use the blocking errorHandler

* Check me on if theres a better way to handle the timeout short-circuit than a thread-shared boolean

      * [X] Yes
      * [ ] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [X] Yes
      * [ ] No

This adds the library System Lambda:
https://github.com/stefanbirkner/system-lambda
Which allows us to capture system.exit calls in our tests. Note that this is test-scoped and won't end up in our production released code as such, which hopefully reduces the risk.

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-app/src/test/java/gov/cms/bfd/pipeline/app/PipelineManagerIT.java
lsmitchell, line 1: Primary changes here were just wrapping the couple tests that would cause a fatal shutdown with the System Lambda library catch for system exit, else the JVM exits during the test and they can't run. Also simplified some of the lambdas to be easier to read (thanks intelliJ)

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -354,24 +375,23 @@ public Integer getValue() {
...
           try {
             List<RifRecordLoadResult> processResults =
                 process(recordsBatch, loadedFileId, postgresBatch);
             processResults.forEach(resultHandler::accept);
           } catch (Throwable e) {
-            LOGGER.error("Error caught when processing async batch!");
-            errorHandler.accept(e);
+            LOGGER.error("Error caught when processing async batch!", e);
+            // Cannot use errorHandler here, as it will cause a long wait before shutdown
+            fatalFailure = true;
```
lsmitchell, line 394: Using the errorHandler here was causing a problem. The callback in this subthread caused a shutdown which blocked the subthread from returning until the shutdown was complete. The shutdown was waiting for all jobs to finish, and RifLoader was one of the jobs, which was waiting on its subthreads, and this meant we were deadlocked. The job would sit waiting on the 72 hour timeout on line 330, after which the job would finally give up waiting and we could finish application shutdown, however I doubt anyone actually waited 2 days to watch it gracefully shut down.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -333,6 +339,21 @@ public Integer getValue() {
...
+       * We also can't use a passed-in error handler from up the execution chain or else the pipeline manager shutdown procedure
+       * waits for this thread to shut down, which is waiting on the above 72-hour timeout before it unblocks.
+       *
+       * FUTURE: If we wish to be more discerning about recoverable errors, we could return false here which will
+       * allow the pipeline to continue operating if an error is considered non-fatal (it also currently moves the failed files
+       * to "Done" instead of "Failed", which would need to be fixed.)
+       */
+      if (fatalFailure) {
+        throw new IllegalStateException("Fatal error during rif file processing.");
+      }
```
lsmitchell, line 355: This job needs to definitively return in order to shut down gracefully, so we still can't use the error handler here either. Instead, throw an exception and PipelineManager will trigger the job as failed and shut things down for us.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoaderIT.java
```diff
@@ -245,17 +244,17 @@ public void failOnUpdateBeneficiaryBeforeInsert() {
...
+            IllegalStateException.class,
             () -> {
               loadSample(
                   "SAMPLE_A, bene only, UPDATE",
                   CcwRifLoadTestUtils.getLoadOptions(),
                   editedSample);
             });
 
-    assertTrue(thrown.getMessage().contains("Load errors encountered"));
+    assertTrue(thrown.getMessage().contains("Fatal error during rif file processing"));
```
lsmitchell, line 257: This changed because we're not throwing the exact exception that the subthread hit anymore, we're throwing the custom one when we have a fatalError reported in any subthread. (The original exception in the subthread wasn't particularly valuable, and it's logged anyway)

apps/bfd-pipeline/bfd-pipeline-app/pom.xml
```diff
@@ -115,6 +115,13 @@
 			<scope>runtime</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>com.github.stefanbirkner</groupId>
+			<artifactId>system-lambda</artifactId>
+			<version>${system-lambda-version}</version>
+			<scope>test</scope>
+		</dependency>
```
lsmitchell, line 123: New dependency added for allowing the tests the ability to handle System.Exit() without blowing up the JVM and stopping jUnit altogether. This is test-scoped only, meaning it won't get compiled in our production server code.

Apparently stopping system.exit from killing Junit is quite annoying. Couldnt use mockito either, as the exit does not happen on the same thread as the test, which is a requirement for mocking static methods/classes.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -99,6 +99,15 @@ public final class RifLoader {
...
   private final PipelineApplicationState appState;
 
+  /** The maximum amount of time in hours we will wait for a job to complete loading its batches. */
+  private final int MAX_BATCH_WAIT_TIME_HOURS = 72;
+
+  /**
+   * Keeps track of a fatal error during loading in one of the batch threads so we know to fail the
+   * job (and kill the pipeline).
+   */
+  private static volatile boolean fatalFailure;
```
lsmitchell, line 109: Would be open to if anyone has a better idea than this that doesnt require rearchitecting the application; basically we need a signal from the subthreads to the main thread that we hit some error and the job has failed, and also allows the thread to complete/release. An exception won't work as it doesnt seem to bubble up to the caller (maybe the batch spliterator is eating it or something?) Not sure what other methods there are for talking between threads that arent over-engineered for this rather simple use-case.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
```diff
@@ -611,8 +593,19 @@ public void onFailure(Throwable jobThrowable) {
...
+        // If our job has been interrupted, we are already shutting down, so just ignore it.
+        LOGGER.info("Job interrupted: " + jobRecord.getJobType());
+      } else {
+        /* If the job failed and wasnt cancelled, it threw some other exception mid-stream,
+         * and now we must die as we almost certainly cannot continue with the tainted batch sitting in Incoming.
+         * This will call the proper shutdown procedures and exit gracefully.
+         * FUTURE: We may be able to more fault-tolerant by moving the failed batch
+         * into the Failed folder when this happens and continuing to run
+         */
+        PipelineApplication.shutdown();
```
lsmitchell, line 607: Here we fix a problem where originally this simply logged the error and expected the original exception to bubble back to PipelineApplication for initiating a shutdown, which wasnt happening due to threading. Now we explicitly shut down the pipeline.

mackec1 (approved):
For the pipeline, this looks reasonable; in cases where there are literally thousands of threads, things like a Synchronizer class that holds CountDownLatch can be used to maintain state amongst all threads.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
brianburton: Will the exception be logged somewhere else?

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
mackec1: this could be an AtomicBoolean; allows for atomic operations. It ensures that reads and writes are thread-safe, meaning that multiple threads can access the same boolean without overriding each other's values.

apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/PipelineManager.java
brianburton: Will other jobs have an opportunity to shutdown cleanly?  This looks like it will just terminate the app instantly.

lsmitchell: Yes the original exception is logged before it gets here. This will call PipelineApplicaton which will call System.exit(), which then runs the shutdown routines in PipelineManager (and interrupts any ongoing threads that allow themselves to be interrupted). PipelineApplication will wait for its jobs to either finish (if uninterruptible like RifLoader) or cancel them (if they havent started running or are interruptible.)

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
lsmitchell: I'll look into that; I used volatile here to try and ensure that the value being read was always up-to-date across threads, since its only being written to with one value. But maybe safer just to use AtomicBoolean since the specific way I implemented it isnt immediately obvious.

mackec1 (dismissed):

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
lsmitchell: Adjusted to use AtomicBoolean.

apps/bfd-pipeline/bfd-pipeline-app/pom.xml
lsmitchell: @StewGoin could you review this and thumbs up or down on this dependency?

StewGoin (approved):
Scoping the usage to just tests does mitigate concerns related to the dependency wrt updates, etc... so this is fine to proceed as long as we don't later expand that scope to the running system in Prod/Prod-Sandbox

PR #1658, opened 2023-03-21
BFD-2103: BFD Pipeline SLIs Improvement and Fixes

malessi:
**JIRA Ticket:**
[BFD-2103](https://jira.cms.gov/browse/BFD-2103)

As a BFD Engineer, I want the metrics/SLIs collected for the BFD Pipeline to be accurate and relevant. I want to be able to use these metrics/SLIs to ensure the Pipeline is behaving as expected.

---

### What Does This PR Do?

This PR makes many changes to the existing `bfd-${env}-update-pipeline-slis` Lambda(s) that, as of writing, collect time-based metrics about the BFD Pipeline:

- `time/data-available` -- stores unix timestamps, in UTC, indicating when a particular RIF file for a particular "group"/load was made available in S3 for the BFD Pipeline to load
- `time/data-first-available` -- stores unix timestamps, in UTC, indicating when the _first_ RIF file for a particular group was made available for the BFD Pipeline to load; essentially, samples from this metric represent the moment in time at which a load was _started_ for the BFD Pipeline
- `time/data-loaded` -- stores unix timestamps, in UTC, indicating when a particular RIF file for a particular group was _loaded_ by the BFD Pipeline
- `time/data-fully-loaded` -- stores unix timestamps, in UTC, indicating when the _final_ RIF file for a particular group was loaded by the BFD Pipeline; essentially, samples from this metric represent the moment in time at which a load _finished_
- `time-delta/data-load-time` -- stores _durations_, in seconds, representing the time it took for a particular RIF file in a particular group to go from being _available_ to load having been _loaded_ by the BFD Pipeline; essentially, samples from this metric represent the time it took for a _single RIF file_ to be loaded

Unfortunately, some assumptions made about the BFD Pipeline's behavior, the characteristics of some data loads, and the capabilities of AWS CloudWatch Metrics make the existing set of SLIs inadequate or otherwise entirely inaccurate:

- The Lambda(s) will submit multiple samples, erroneously, for a single group/data load to the `time/data-first-available` and `time/data-fully-loaded` metrics when there should only be one sample per-group per-metric.
- It was assumed that it would be easily possible to calculate the full duration of loads dynamically using the `time/data-first-available` and `time/data-fully-loaded` metrics (by subtracting the latter from the former); unfortunately, AWS CloudWatch Metrics are ill-suited for doing this sort of dynamic calculation, thus making it nearly impossible to create visuals (and alarms) showing historical load durations. This incapability is made even more difficult to workaround given the unintended multiple sample submission behavior mentioned previously

This PR aims to solve the above issues with the following improvements:

- A new metric/SLI is calculated and submitted to AWS CloudWatch Metrics by the Lambda(s):
  - `time-delta/data-full-load-time` -- stores _durations_, in seconds, representing the time it took _in **total**_ for a given BFD Pipeline load to go from being _available_ in S3 to having been _loaded_ by the BFD Pipeline; essentially, samples from this metric represent the time it took to load an entire load/group from start to finish
- The logic checking for the start and end of a given data load has been made more robust and correct, thus the Lambda(s) should no longer submit multiple, erroneous samples to the `time/data-first-available` and `time/data-fully-loaded` metrics
  - The logic for `time/data-first-available` was made more robust by:
    - Using a "sentinel queue", an SQS queue that the Lambda polls for [sentinel](https://en.wikipedia.org/wiki/Sentinel_value) messages. If any messages corresponding to the current group are found, the Lambda assumes that the load has already begun for that group and does not put any samples to the metric; otherwise, the Lambda assumes that the load has just begun and _does_ put data to the metric
    - Setting the [reserved concurrent executions](https://docs.aws.amazon.com/lambda/latest/dg/lambda-concurrency.html#reserved-and-provisioned) of each Lambda to `1`, thus disabling any concurrency for each Lambda. While the above change improved the behavior of the Lambda (as SQS queues are much more immediate than CloudWatch Metrics), Lambdas invoked concurrently close enough together _can_ submit duplicate samples to the metric. Disabling concurrency does not affect the behavior of the Lambda, and I do not expect the number of events the Lambda consumes to ever be in significant enough quantities to negatively affect metric collection
  - The logic for `time/data-fully-loaded` was made more correct by checking for the existence of _all_ non-optional RIF types in the `Done` folder of the Lambda's corresponding ETL S3 bucket _and_ that the `Incoming` folder is empty. Previously, only the second condition was checked which lead to loads that were submitted to the ETL S3 bucket in batches being considered finished as the `Incoming` folder would be temporarily empty when a batch of files were finished loading

These changes were tested by applying them to `test` and running the Lambda repeatedly with a synthetic S3 event notification, ensuring that data was submitted to each metric as expected. These changes have since been undone.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist


ops/terraform/services/pipeline/modules/bfd_pipeline_slis/lambda_src/backoff_retry.py
```diff
@@ -0,0 +1,57 @@
...
+    """Generic function for wrapping another callable (function) that may raise errors and require
+    some form of retry mechanism. Supports passing a list of exceptions/errors for the retry logic
+    to ignore and instead raise to the calling function to handle
+
+    Args:
+        func (Callable[[], T]): The function to retry
+        retries (int, optional): The number of times to retry before raising the error causing the
+        failure. Defaults to PUT_METRIC_DATA_MAX_RETRIES.
+        ignored_exceptions (list[Type[BaseException]] , optional): A list of exceptions to skip
+        iretrying and nstead immediately raise to the calling function. Defaults to [].
```
lsmitchell, line 25:
```suggestion
        retrying and instead immediately raise to the calling function. Defaults to [].
```

lsmitchell (commented):
Looks like some good refactors, but with so many it's a bit hard to understand each block of changes. A few questions.

ops/terraform/services/pipeline/modules/bfd_pipeline_slis/lambda_src/cw_metrics.py
```diff
@@ -0,0 +1,196 @@
...
+from datetime import datetime, timedelta
+from functools import reduce
+from itertools import chain, combinations
+from typing import Any, Tuple, TypeVar
+
+T = TypeVar("T")
+
+
+@dataclass
+class MetricData:
```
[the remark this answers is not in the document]
malessi: Yeah this is just a refactor of helper functions into their own separate files. Unfortunately there were a lot of refactors here and code re-organization because the Lambda was getting too big for a single source file.

ops/terraform/services/pipeline/modules/bfd_pipeline_slis/main.tf
```diff
@@ -33,6 +34,14 @@ resource "aws_lambda_function" "this" {
 
   kms_key_arn = local.kms_key_arn
 
+  # Ensures that only _one_ instance of this Lambda can run at any given time. This stops the
+  # possible duplicate submissions to the first available time metric that could otherwise occur
+  # if two instances of this Lambda are invoked close together. This _does_ take from our total
+  # reserved concurrent executions, so we should investigate an even more robust method of stopping
+  # duplicate submissions
+  # TODO: Investigate other methods of stopping duplicate sample submissions to first available metric
```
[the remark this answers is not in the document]
malessi: We could create a new card for this, although the more I'm thinking about this the more I feel like this isn't much of a problem. Setting reserved concurrency is a completely valid mechanism for restricting concurrent Lambda executions, and since we're using S3 Event Notifications to invoke the Lambda it's [_asynchronously_ invoked](https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html), so AWS will automatically queue events if the Lambda is already running.

I can't think of many other means of stopping concurrent execution without creating more headaches for ourselves. I might just remove this `TODO`.

ops/terraform/services/pipeline/modules/bfd_pipeline_slis/lambda_src/__init__.py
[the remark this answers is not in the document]
malessi: After doing a bit of research on this, the `__init__.py` file _was_ necessary in versions of Python prior to 3.3 in order to signal to Python that a particular directory should be treated as a Python package (and therefore allow for import). 

I don't _think_ it's necessary to keep it, but I feel like it may still be conventional? I know Python supports omitting it, but I can't say if tools like `Pylance` and `Pylint` handle it properly. I know I've had issues with package resolution with `Pylint` _until_ I added an `__init__.py` to the problematic package's directory, so it might still be somewhat necessary. I'm not sure if that particular issue could have been solved another way.

Either way, keeping this here isn't affecting us negatively nor is it yet invalid, so I feel like we should keep it for now to avoid the _potential_ of issues with developer tooling and package management.

lsmitchell (approved):

ops/terraform/services/pipeline/main.tf
```diff
@@ -206,5 +206,6 @@ module "bfd_pipeline_slis" {
   source          = "./modules/bfd_pipeline_slis"
   account_id      = local.account_id
   aws_kms_key_arn = local.kms_key_id
```
billie-rose, line 208: Are `aws_kms_key_arn` and `aws_kms_key_id` both meant to be `local.aws_kms_key_id` ?

ops/terraform/services/pipeline/modules/bfd_pipeline_slis/lambda_src/cw_metrics.py
```diff
@@ -0,0 +1,196 @@
...
+    statistic: str,
+    period: int = 60,
+    start_time: datetime = datetime.utcnow() - timedelta(days=15),
+    end_time: datetime = datetime.utcnow(),
+) -> list[MetricDataResult]:
+    """Wraps the GetMetricData CloudWatch Metrics API operation to allow for easier usage. By
+    default, standard resolution metrics from the current time to 15 days in the past are retrieved
+    from CloudWatch Metrics.
+
+    Args:
```
billie-rose, line 139: Huge nitpick and genuine question - do we need to specify the default values in the docs when it's already specified in the args? Just seems like an extra value to update later if the defaults change

billie-rose: (and sorry! I know this is all refactored code and not part of this PR)

ops/terraform/services/pipeline/modules/bfd_pipeline_slis/lambda_src/backoff_retry.py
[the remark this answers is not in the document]
malessi: I was definitely thinking that--I've written a lot of utility code for some of these Lambdas and feel like we should have some sort of common spot in the repository where we can place it so we're not repeating ourselves as much. I haven't put too much thought as to _where_ we store those libraries and _how_ we would import them, though. Definitely open to suggestions on that!

I do think it's out-of-scope for this work to hoist `backoff_retry.py` out of this Lambda, but I think we should probably put a ticket in the backlog to have someone think about this more deeply at some point.

billie-rose (approved):

ops/terraform/services/pipeline/modules/bfd_pipeline_slis/lambda_src/cw_metrics.py
malessi: That's a great point, and honestly I've fallen into the trap of commenting something, changing the behavior of that something, and then forgetting to update the comment to reflect the new behavior more times that I'd like to admit. I think removing this bit from the function's comment is probably best.

PR #1668, opened 2023-03-30
BFD-2499: Adjust bfd-server-war to use postgres test containers by default instead of HSQL

lsmitchell:
**JIRA Ticket:**
[BFD-2499](https://jira.cms.gov/browse/BFD-2499)

As a BFD contributor, I want to be able to run bfd-server integration tests against a prod-like data store (postgresql) so that I can be confident that the quality communicated by a successful test run is real.

---

### What Does This PR Do?

Tldr: Moves the start and cleanup of the IT test server into code (instead of a shell script) and sets the default database to use with our "IT"s to postgres (using a container). HSQL is still supported as before, per the readme commands. IT tests should also run in your IDE of choice now, if desired!

Note: This adds about 1 minute to the total build by using postgres containers vs hsql, but this imo is a valuable tradeoff for more accurate testing and avoiding HSQL implementation mismatch bugs.

Major:
- Moved the test server setup from the server-start.sh script into code (ServerExecutor in the test code)
   - The test script was not only a maintenance, debug, and understanding headache, but the DB container needed to start before the server, and the server was being launched via maven in the script prior to any code execution. 
   - Rather than add MORE setup to the server-start script, putting everything in code centralizes the entire test setup, lets us share setup code with the rest of the container logic for other modules, and lets us control what order things are setup (to allow the container to be set up before the server is started)
   - This class works similarly to the migrator app's testing code; sets up a command to start the server, runs validations the server can get what files it needs, runs it in a managed process, and cleans up when all tests are run
- Removed the integration step from the maven server script plugins, so they will no longer run when running the ITs or during any other build step. They will now only be run when explicitly calling them with maven commands.
    - I left the scripts and plugins in-place since manually starting/stopping the server uses these plugins, so they can still be called manually when desired (some of our documentation uses these scripts for manual setup/testing, i think)
- Adjusted the startup script to use postgres test containers, in the event its started manually
- Adjusted the "cleanup" script that shuts down the server to use a java shutdown hook which cleans up after the tests are finished running (basically once the JVM is shutting down)
    - This will also catch a ctrl+C, which prevents zombie processes if quitting halfway through a test by any means
-  Sets up a base test class (ServerRequiredTest) for any test which requires the server to run; this will spin up the server, shutdown hook, and datasource for the DB if its not running, or do nothing if its running

Minor:
- Add a maven step to copy a couple required server test keystores from bfd-server/dev/ssl-stores into target/server-work during the build, just to reduce complexity in finding them when booting up the server (less relative path file searches to do)
- Fix a couple minor spots where tests were calling initUnpooledDataSource instead of using the singleton, which could cause desync issues with the datasource
    - Made initUnpooledDatasource private, to encourage the use of `DatabaseTestUtils.get().getUnpooledDataSource()` which will properly either initialize the datasource OR get the existing singleton instead of calling initUnpooledDataSource
- Update README to remove that server tests are still using HSQL
- Moved database cleanup from individual tests into the ServerRequiredTest classes wherever the test was doing it; between each test the database will be truncated to ensure proper data/test isolation
- Copied some code from the pipeline test utils to truncate the DB into ServerTestUtils
    - This unfortunately can't be moved into a shared util easily due to the way it uses JPA class refs that aren't accessible in the common utils location
- Minor fix where RDA tests were inserting a null char instead of a 0 which were breaking the test with postgres 
- Set defaults across the test codebase to use postgres containers instead of HSQL

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -404,6 +404,32 @@
...
+						</goals>
+						<configuration>
+							<outputDirectory>${project.build.directory}/server-work</outputDirectory>
+							<resources>
+								<resource>
+									<directory>${basedir}/../dev/ssl-stores</directory>
+									<includes>
+										<include>server-keystore.jks</include>
+										<include>server-truststore.jks</include>
+									</includes>
```
lsmitchell, line 426: This copies a couple of the files needed for the server to start into target/server-work to facilitate finding them a bit easier; else we need to navigate to outside the module's scope and that's just messy.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -124,23 +124,23 @@ public DataSource dataSource(
       @Value("${" + PROP_DB_CONNECTIONS_MAX + ":-1}") String connectionsMaxText,
       @Value("${" + PROP_DB_SCHEMA_APPLY + ":false}") String schemaApplyText,
       MetricRegistry metricRegistry) {
-    HikariDataSource poolingDataSource;
+    DataSource poolingDataSource;
```
lsmitchell, line 127: The changes here are primarily to allow a non-HSQL datasource to return from createTestDatabase; since psql uses a PGSimpleDataSource instead of a HikariDataSource.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/ServerExecutor.java
```diff
@@ -0,0 +1,285 @@
...
+      String maxHeapArg = System.getProperty("its.bfdServer.jvmargs", "-Xmx4g");
+      String containerImageType =
+          System.getProperty("its.testcontainer.db.image", TEST_CONTAINER_DATABASE_IMAGE_DEFAULT);
+      // FUTURE: Inherit these from system properties? Which of these are valuable to pass?
+      String v2Enabled = "true";
+      String pacEnabled = "true";
+      String pacOldMbiHashEnabled = "true";
+      String pacClaimSourceTypes = "fiss,mcs";
+      String includeFakeDrugCode = "true";
+      String includeFakeOrgName = "true";
```
lsmitchell, line 151: Wondering opinions on which of these are useful to parse from the system props. The server script did read these if set, but not sure if it's valuable to be able to adjust these and break the tests which assume a lot of these are the above values in the process.

One example I found while testing is "pacOldMbiHashEnabled" being set to false fails an RDA test which expects it to be true.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/ServerExecutor.java
```diff
@@ -0,0 +1,285 @@
...
+      args.add(String.format("-DbfdServer.db.url=%s", dbUrl));
+      args.add(String.format("-DbfdServer.db.username=%s", dbUsername));
+      args.add(String.format("-DbfdServer.db.password=%s", dbPassword));
+      args.add(String.format("-DbfdServer.include.fake.drug.code=%s", includeFakeDrugCode));
+      args.add(String.format("-DbfdServer.include.fake.org.name=%s", includeFakeOrgName));
+      args.add(String.format("-Dits.testcontainer.db.image=%s", containerImageType));
+      return args.toArray(new String[0]);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
lsmitchell, line 174: This could be removed, the upper level method already throws IOException and it doesnt really matter if we throw that or this; I'll fix this if there are other adjustments to make.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/RDATestUtils.java
```diff
@@ -284,7 +284,7 @@ private RdaFissClaim fissTestDataB(Mbi mbi) {
             .dcn(FISS_CLAIM_B_DCN)
             .intermediaryNb("99999")
             .hicNo("hicnumbe2")
-            .currStatus('\0')
+            .currStatus('0')
```
lsmitchell, line 287: This caused an issue when running the test with postgres, since it read it as a null value and postgres doesnt allow it for the field type. I talked about it with Brian and we suspect it was a typo anyhow.

apps/bfd-shared-test-utils/src/main/java/gov/cms/bfd/DatabaseTestUtils.java
```diff
@@ -373,7 +384,9 @@ private static DataSource initUnpooledDataSourceForPostgresql(
   private static DataSource initUnpooledDataSourceForTestContainerWithPostgres(
       String username, String password) {
 
-    String testContainerDatabaseImage = System.getProperty("its.testcontainer.db.image", "");
+    String testContainerDatabaseImage =
+        System.getProperty("its.testcontainer.db.image", TEST_CONTAINER_DATABASE_IMAGE_DEFAULT);
+    LOGGER.debug("Starting container, using image {}", testContainerDatabaseImage);
```
lsmitchell, line 389: These comments were helpful while debugging, so leaving these in as debug-level.

cbrunefearless (approved):
LGTM, brought it down on my machine and it works.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/utils/RDATestUtils.java
lsmitchell: Good example of the difference between hsql and postgres and how it could affect a test case (and therefore our ability to catch a bug) though.

mjburling (approved):

PR #1669, opened 2023-03-30
BFD-2471: Introduce Lambda and Jenkins Pipelines to run BFD CCW Pipeline on demand

malessi:
**JIRA Ticket:**
[BFD-2471](https://jira.cms.gov/browse/BFD-2471)

As a BFD Engineer, I want the CCW-variant of the BFD Pipeline to run only where there is new data to load so that resources are not wasted running a large instance that is doing no work

---

### What Does This PR Do?

This PR makes several changes:

- A new Jenkins pipeline, `bfd-job-broker`, has been added to "broker" between our AWS resources and our Jenkins instance, using an SQS queue as the means of control. Previously, it was impossible for us to invoke jobs on our Jenkins instance outside of the instance itself, but with this new pipeline we are able to post to its corresponding SQS indicating what job to run and its parameters. The `bfd-job-broker` pipeline will then poll for these messages, and build the corresponding pipeline/job specified by the message. This new pipeline is used by the Lambda this PR adds to run another Jenkins multibranch pipeline that deploys the `pipeline` Terraservice
  - This pipeline has already been configured to poll for messages and already exists in our Jenkins instance
  - The corresponding infrastructure changes in the `mgmt` environment have be applied and, as of writing, are still applied to the `mgmt` environment
  - This pipeline required a new IAM service user to be created, `bfd-mgmt-jenkins`, as the [AWS SQS Plugin](https://plugins.jenkins.io/aws-sqs-trigger/) used to trigger this pipeline does not support our existing authentication strategies
    - This user's information has been stored securely in SSM, using the `mgmt` module's `base_config` sub-module
- Another, new Jenkins _multibranch_ pipeline, `bfd-deploy-pipeline-terraservice`, has been added to _deploy_ the `pipeline` Terraservice. This new pipeline is used as the mechanism by which the below Lambda starts and stops the CCW pipeline instance, by `apply`ing the Terraservice with the `create_ccw_instance` variable set depending on whether the instance should be running or not
  - While out-of-scope for this PR, this new pipeline essentially duplicates similar `Deploy Pipeline to X` stages in our deployment pipeline. We should consolidate this in the future
  - It is expected that the main deployment pipeline will continue to deploy the `pipeline` Terraservice. The branch from which the main deployment pipeline is deployed from will inform the branch variant of this pipeline that the below Lambda will invoke (so environments remain consistent with their changes)
  - This pipeline supports defining the Pipeline AMI, but the below Lambda does not take advantage of this at the moment. This pipeline will default to no override, which will default the AMI to the lastest built `master` AMI. There is an opportunity to use SSM Parameter Store here to keep AMIs consistent between the deployment pipeline and this pipeline
- A new Lambda, `bfd-${env}-pipeline-manager`, has been added by this PR; its `README` describes its function best:
  > This submodule defines a Lambda and its corresponding infrastructure. This Lambda,
  > `bfd-${env}-pipeline-manager`, "manages" the BFD Pipeline Terraservice; it is invoked when new files
  > are created in the `env`'s corresponding S3 pipeline/ETL bucket at specific paths:
  >
  > - `Synthetic/Incoming/`
  > - `Synthetic/Done/`
  > - `Incoming/`
  > - `Done/`
  >
  > Objects uploaded to or moved to these paths in the ETL bucket will invoke this Lambda, which will
  > then determine if the BFD CCW Pipeline should be started or not. This depends on a few factors:
  >
  > - If the file was uploaded to either of the `Incoming` folders, and there is no _ongoing_ data load
  >   for that file's "group", the `bfd-deploy-pipeline-terraservice` Jenkins pipeline is started using
  >   the `bfd-job-broker` with the arguments set to _create_ the CCW Pipeline instance
  > - Else, if the file was uploaded to either of the `Done` folders, there is no files in the
  >   corresponding `Incoming` folder, all non-optional RIF files are present in the `Done` folder,
  >   _and_ there are no ongoing loads _at all_ (including other groups), the
  >   `bfd-deploy-pipeline-terraservice` Jenkins pipeline is started using the `bfd-job-broker` with the
  >   arguments set to _destroy_ the CCW Pipeline instance
  >
  > In short, this Lambda indirectly invokes a Jenkins pipeline that will either _create_ or _destroy_
  > the CCW Pipeline instance for the current `env`, thus ensuring the CCW Pipeline instance is only
  > running when there is data queued up for load into that `env`. As a side effect of using Terraform
  > to do this management, the `pipeline` Terraservice will be _continuously deployed_ whenever the CCW
  > Pipeline instance needs to start or stop.
  >
  > Note that the branch from which this Lambda/module was deployed from remains consistent, such that
  > whatever branch this module was deployed from will be the branch variant of the
  > `bfd-deploy-pipeline-terraservice` that is invoked. For established environments, this will
  > generally remain `master`, but this allows for consistency when using ephemeral environments that
  > may be based upon non-`master` branches.
  - There is some opportunity for code re-use with respect to this Lambda and the existing `bfd-${env}-pipeline-slis` Lambda, as they share a significant amount of code and have some similar behavior. I believe this work is out-of-scope for this PR, but we should keep this in mind for the future
- The default behavior of the `pipeline` Terraservice always creating the CCW Pipeline instance has been changed to not deploy the CCW Pipeline instance

These changes have been tested by applying the Terraform changes to `test` (where reasonable), and testing the Lambda with synthetic S3 event notifications. The `bfd-job-broker` has been tested by posting messages to its SQS queue and verifying that it builds the target job. The `bfd-deploy-pipeline-terraservice` pipeline has been tested by running the pipeline (with the Terraservice global library modified to only plan) and verifying that the Terraform plan it generates is correct.

The `bfd-job-broker` Jenkins pipeline will be updated to point to `master` when this PR is merged.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist


ops/terraform/services/pipeline/modules/bfd_pipeline_manager/lambda_src/pipeline_manager.py
```diff
@@ -0,0 +1,352 @@
...
+    # "<Synthetic/>/<Incoming/Done>/<ISO date format>/<file name>".
+    if match := re.search(
+        pattern=(
+            rf"^({PipelineLoadType.SYNTHETIC.capitalize()}){{0,1}}/{{0,1}}"
+            rf"({status_group_str})/"
+            rf"([\d\-:TZ]+)/"
+            rf".*({rif_types_group_str}).*$"
+        ),
+        string=decoded_file_key,
+        flags=re.IGNORECASE,
```
billie-rose, line 251: Should we do this if we're also purposefully capitalizing values like load type, and then doing case-sensitive comparisons?

keithdadkins (approved):
Good work and a big step to improve our pipeline.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/lambda_src/pipeline_manager.py
malessi: I don't _think_ it's necessary to use ignore case here, but I think it helps cover any case where a RIF file is uploaded with a mixed case name (i.e. `Bene_...` vs. `bene_....`).

I replaced the `PipelineLoadType.SYNTHETIC.capitalize()` with `PipelineLoadType.SYNTHETIC.value` in the regex pattern as, like you are suggesting, the casing of the pattern doesn't matter. That was a bit of an oversight; there are other places in the code where I explicitly `capitalize()` some string values (that are usually already capitalized) just to make things a little more explicit, but this spot is entirely unnecessary.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/lambda_src/pipeline_manager.py
```diff
@@ -0,0 +1,374 @@
...
+
+@dataclass
+class JenkinsTerraserviceJobParameters:
+    """Represents the parameters for the BFD Pipeline Deploy Terraservice Jenkins Pipeline/job"""
+
+    env: str
+    create_ccw_pipeline_instance: bool
+
+
+def _check_ongoing_load_queue(timeout: int = 1) -> list[OngoingLoadQueueMessage]:
```
billie-rose, line 109: Thoughts on re-naming this something like `_get_ongoing_load_queue_messages` ?

malessi: Agreed, I think your suggested name is more specific and a better fit for what the function actually does.

I'll change it.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+- `Synthetic/Incoming/`
+- `Synthetic/Done/`
+- `Incoming/`
+- `Done/`
+
+Objects uploaded to or moved to these paths in the ETL bucket will invoke this Lambda, which will
+then determine if the BFD CCW Pipeline should be started or not. This depends on a few factors:
+
+- If the file was uploaded to either of the `Incoming` folders, and there are no _ongoing_ or
```
mackec1, line 15: Mitch, are there any safeguards to preclude the Pipeline from firing prematurely; for example, there is an 'implied' prerequisite that pipeline should not start unless/until the manifest(s) is/are available. So CCW provides a manifest per bene/claims file while synthea provides a single manifest file for all data files. In the case of CCW, their 'delivery' contract requires them to provide data files prior to delivering a manifest file for that data.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/lambda_src/pipeline_manager.py
```diff
@@ -0,0 +1,374 @@
...
+import sys
+from dataclasses import asdict, dataclass
+from enum import Enum
+from typing import Any, Optional
+from urllib.parse import quote_plus, unquote
+
+import boto3
+from botocore.config import Config
+
+REGION = os.environ.get("AWS_CURRENT_REGION", "us-east-1")
```
mackec1, line 13: Probably need to support IDEMPOTENT_MODE here???

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/README.md
```diff
@@ -0,0 +1,73 @@
...
+  _queued_ data loads _at all_ (including other groups), the `bfd-deploy-pipeline-terraservice`
+  Jenkins pipeline is started using the `bfd-job-broker` with the arguments set to _create_ the CCW
```
lsmitchell, line 17: Is it possible to have two instances start if two items are added in quick succession/at the same time? For instance, during synthea loads I might move 20+ folders of data into synthetic/incoming for all the future load data; is there a chance the lambda would trigger, queue the creation of the ccw pipeline instance, and before it gets created, see one of the next folders and start another?

It's pretty important we don't ever have two pipelines running for the same environment, so wanted to make sure this was explicitly not possible.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/README.md
```diff
@@ -0,0 +1,73 @@
...
+  Pipeline instance
+- Else, if the file was uploaded to either of the `Done` folders, there are no files in the
+  corresponding `Incoming` folder, all non-optional RIF files are present in the `Done` folder,
+  _and_ there are no ongoing loads _at all_ (including other groups), the
+  `bfd-deploy-pipeline-terraservice` Jenkins pipeline is started using the `bfd-job-broker` with the
+  arguments set to _destroy_ the CCW Pipeline instance
```
lsmitchell, line 23: Just a note that this may not work correctly with synthea data; we will have folders sitting in synthetic/incoming that will not be processed until their week comes due (and we may have up to 2 years worth of these weekly loads sitting and waiting in synthetic/incoming). Depending on the check it may never be able to destroy the instance due to these folders sitting in the incoming folder as expected.

malessi:
After discussing offline with @lsmitchell, I've decided to move this PR back into draft until the following items are implemented:

[ ] When the last file of the last data load is loaded into `Done`, the Lambda should wait for a few minutes to run the Terraservice Jenkins job such that any shutdown tasks the BFD Pipeline needs to run are able to do so
[ ] Due to _future_ Synthetic data loads, the Lambda should evaluate the date of incoming data loads to determine if the Pipeline should be started immediately or if a future run should be scheduled at the point in time where the data of the incoming data load has passed
   * This mimics the current behavior of the Pipeline with respect to Synthetic data loads, as it will wait to load data until its corresponding date has passed
[ ] _Optional_: A runbook should be created outlining the steps to **manually** run the BFD CCW Pipeline _without_ the `bfd-pipeline-manager` Lambda causing out-of-band modifications

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/README.md
malessi: Since this solution doesn't modify the Pipeline itself, we'll be relying on the Pipeline's built-in safeguards to ensure that the Pipeline does not start loading data prematurely. The Lambda _does_ filter based on _what_ file invoked it; for example, if the file uploaded to any of the watched folders in S3 is not a valid RIF file, the Lambda immediately exits.

Basically, you can boil the Lambda down to a mechanism for starting and stopping the Pipeline's _instance_ depending on when we need it available.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/lambda_src/pipeline_manager.py
malessi: This solution doesn't discern between running in non-`IDEMPOTENT` and `IDEMPOTENT` mode, so maybe this is something we should consider for the full implementation? Thanks for bringing this up.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/README.md
malessi: Short answer: no, we cannot ever have two BFD Pipeline instances running at the same time in any given environment. This constraint is enforced by the `pipeline` Terraservice, as only one EC2 instance can exist in our state at any given time for an environment

**Long answer:**

It is _technically_ possible for the `bfd-pipeline-manager` _Lambda_ to run concurrently if S3 Event Notifications are received in quick succession and the locking SQS queue does not update fast enough. This would result in multiple `bfd-job-broker` executions, which in-turn would execute the `bfd-pipeline-terraservice` Jenkins job multiple times. Fortunately, the `bfd-pipeline-terraservice` Jenkins Pipeline uses Jenkins' built-in [Lockable Resources](https://www.jenkins.io/doc/pipeline/steps/lockable-resources/) such that, per-environment, only a single build of the `bfd-pipeline-terraservice` Jenkins Pipeline can be running at any time. Additionally, Terraform itself has a locking mechanism where only one operator/process can modify state at a time, so if the Jenkins lock failed this additional lock would stop concurrent modifications to the `pipeline` Terraservice state.

However, for the same reasons mentioned in the "short answer", only a single BFD CCW Pipeline instance can exist for a given environment at any time due to how its managed within the `pipeline` Terraservice. The only way this constraint could be broken is if the aforementioned locking mechanisms failed and the Terraform state for the `pipeline` Terraservice was somehow corrupted or emptied, thus allowing Terraform to create a new instance even if there was one already running. Of course, that would be a defect with Terraform, and not with this solution itself.

ops/terraform/services/pipeline/modules/bfd_pipeline_manager/README.md
malessi: We discussed this offline, but essentially this can be solved by scheduling runs of `bfd-pipeline-manager` Lambda based upon the folder timestamp of future data loads. This PR does not do this, so it would be susceptible to running indefinitely if many future loads were uploaded to the `Synthetic/Incoming` folder.

malessi:
Closing this PR as the work done here was meant as research for a more complete implementation in [BFD-2537](https://jira.cms.gov/browse/BFD-2537)

PR #1704, opened 2023-04-25
BFD-2549: Introduce EOB performance enhancement for claim types with no claim data

mackec1:
**JIRA Ticket:**
[BFD-2549](https://jira.cms.gov/browse/BFD-2549)

Improve performance of EOBs; in particular, it checks for availability of claims data for a beneficiary prior to invoking a transformer for a given claim type.

### What Does This PR Do?
This PR:

- sets up a database function using a flyway migration (V111__setup_claims_availability_function.sql); function takes in a bene_id and checks all claim types to determine if there is any claims data for that type. If there is data, it sets a bitwise mask denoting data availability. The function returns the bitwise mask accounting for all 8 claim types.
- Add a common utility method that handles invoking the db function, check_claims_mask, and returns a java BitSet object that denotes those claim types that have claims data for the specified bene_id.
- The ExplanationOfBenefitResourceProvider (STU3, R4) will invoke the QueryUtils method, hasClaimsData, which returns a Java BitSet object that enable the resource provider to only process those claim types that will have claims data.

* Verify efficacy of the new flyway migration script.
* Ensure that for ExplanationOfBenefitResourceProvider(s), both STU3 and R4, are optimized to take advantage of the availability of claims data (or not).
* **It should be noted**, that comments/questions were posed as to the efficacy and need for this set of changes; analyzing peering partner queries for ExplanationOfBenefit shows that the **VAST, VAST** majority of requests specify _type=*_ (meaning all claim types). This usage pattern falls prey to a performance hit when there are no claims for a given claim type. This is the problem this PR attempts to mitigate.

It should be noted that several different tacks were developed, allowing for performance comparisons; the code was developed with the following constraints:
- the determination of data for all 8 claims types should **only use a single database round-trip**.
- several SQL solutions were tried/tested; the results were (in descending order of performance, i.e., the most performant is listed first, the least performant is listed last):
  - introduce a database function that handles all 8 claim types and returns a single result row (a mask).
  - introduce 'union' SQL in Java code that that handles all 8 claim types and returns 8 result rows (one per claim type).
  - leave Java code untouched (i.e., no changes to address performance).

All performance testing used the current TEST database using a mixture of claim type requests; the results were:

### WITH db Stored Function returning MASK value + Java BitSet class
       14.68 real         3.06 user         1.17 sys

### JAVA SQL (WITH UNION) + Java BitSet class
       32.17 real        14.92 user         6.25 sys

### Current MASTER
    45.95 real        15.44 user         7.17 sys

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/bfd-data-fda/pom.xml
+++ b/apps/bfd-data-fda/pom.xml
@@ -43,7 +43,7 @@
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-inline</artifactId>
-			<version>${mockito.version}</version>
+			<version>${mockito-inline.version}</version>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/apps/bfd-data-npi/pom.xml
+++ b/apps/bfd-data-npi/pom.xml
@@ -42,7 +42,7 @@
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-inline</artifactId>
-			<version>${mockito.version}</version>
+			<version>${mockito-inline.version}</version>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/apps/bfd-model/bfd-model-dsl-codegen/bfd-model-dsl-codegen-plugin/pom.xml
+++ b/apps/bfd-model/bfd-model-dsl-codegen/bfd-model-dsl-codegen-plugin/pom.xml
@@ -19,9 +19,9 @@
     </description>
 
     <properties>
-        <maven-plugin-api.version>3.9.0</maven-plugin-api.version>
-        <maven-core.version>3.9.0</maven-core.version>
-        <maven-plugin-annotations.version>3.7.1</maven-plugin-annotations.version>
+        <maven-plugin-api.version>3.9.1</maven-plugin-api.version>
+        <maven-core.version>3.9.1</maven-core.version>
+        <maven-plugin-annotations.version>3.8.2</maven-plugin-annotations.version>
         <maven.plugin.version>3.7.1</maven.plugin.version>
     </properties>
 
@@ -110,7 +110,7 @@
             <!-- Used to create mock objects for unit tests. -->
             <groupId>org.mockito</groupId>
             <artifactId>mockito-inline</artifactId>
-            <version>${mockito.version}</version>
+            <version>${mockito-inline.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/apps/bfd-pipeline/bfd-pipeline-ccw-rif/pom.xml
+++ b/apps/bfd-pipeline/bfd-pipeline-ccw-rif/pom.xml
@@ -154,7 +154,7 @@
                         illegal reflection access (JDK9+) in certain circumstances -->
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-inline</artifactId>
-			<version>${mockito.version}</version>
+			<version>${mockito-inline.version}</version>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/pom.xml
@@ -176,7 +176,7 @@
             illegal reflection access (JDK9+) in certain circumstances -->
             <groupId>org.mockito</groupId>
             <artifactId>mockito-inline</artifactId>
-            <version>${mockito.version}</version>
+            <version>${mockito-inline.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaSink.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/RdaSink.java
@@ -60,6 +60,18 @@ public interface RdaSink<TMessage, TClaim> extends AutoCloseable {
   }
 
   /**
+   * Called to determine if a message represents a {@link
+   * gov.cms.mpsm.rda.v1.ChangeType#CHANGE_TYPE_DELETE} change. We filter these messages because
+   * they do not have any meaning within the BFD system.
+   *
+   * @param message message to check
+   * @return true if the message is a delete
+   */
+  default boolean isDeleteMessage(TMessage message) {
+    return false;
+  }
+
+  /**
    * Write the object to the data store and return the number of objects successfully written. The
    * count returned is just the most recent unreported processed count and for asynchronous sinks
    * can reflect values from previously submitted batches.
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
@@ -405,6 +405,8 @@ abstract class AbstractClaimRdaSink<TMessage, TClaim>
                 entityManager.merge(change.getClaim());
                 insertCount += getInsertCount(change.getClaim());
               } else {
+                // We would expect this to have been filtered by the RdaSource so it is safe
+                // to stop processing with an exception here.
                 throw new IllegalArgumentException(
                     "RDA API DELETE changes are not currently supported");
               }
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/FissClaimRdaSink.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/FissClaimRdaSink.java
@@ -10,6 +10,7 @@ import gov.cms.bfd.pipeline.rda.grpc.RdaChange;
 import gov.cms.bfd.pipeline.rda.grpc.source.FissClaimTransformer;
 import gov.cms.bfd.pipeline.sharedutils.PipelineApplicationState;
 import gov.cms.model.dsl.codegen.library.DataTransformer;
+import gov.cms.mpsm.rda.v1.ChangeType;
 import gov.cms.mpsm.rda.v1.FissClaimChange;
 import java.io.IOException;
 import java.util.List;
@@ -58,19 +59,29 @@ public class FissClaimRdaSink extends AbstractClaimRdaSink<FissClaimChange, RdaF
     return !InvalidDcnRegex.matcher(fissClaimChange.getDcn()).find();
   }
 
-  /** {@inheritDoc} */
+  /**
+   * This implementation checks the {@link FissClaimChange#getChangeType}.
+   *
+   * <p>{@inheritDoc}
+   *
+   * @param fissClaimChange message to check
+   * @return true if {@link FissClaimChange#getChangeType} is {@link ChangeType#CHANGE_TYPE_DELETE}.
+   */
+  @Override
+  public boolean isDeleteMessage(FissClaimChange fissClaimChange) {
+    return fissClaimChange.getChangeType() == ChangeType.CHANGE_TYPE_DELETE;
+  }
+
   @Override
   public String getClaimIdForMessage(FissClaimChange object) {
     return object.getClaim().getRdaClaimKey();
   }
 
-  /** {@inheritDoc} */
   @Override
   public long getSequenceNumberForObject(FissClaimChange object) {
     return object.getSeq();
   }
 
-  /** {@inheritDoc} */
   @Nonnull
   @Override
   RdaChange<RdaFissClaim> transformMessageImpl(String apiVersion, FissClaimChange message) {
@@ -79,7 +90,6 @@ public class FissClaimRdaSink extends AbstractClaimRdaSink<FissClaimChange, RdaF
     return change;
   }
 
-  /** {@inheritDoc} */
   @Override
   int getInsertCount(RdaFissClaim claim) {
     return 1 // Add one for the base claim
@@ -89,7 +99,6 @@ public class FissClaimRdaSink extends AbstractClaimRdaSink<FissClaimChange, RdaF
         + claim.getAuditTrail().size();
   }
 
-  /** {@inheritDoc} */
   @Override
   RdaClaimMessageMetaData createMetaData(RdaChange<RdaFissClaim> change) {
     final RdaFissClaim claim = change.getClaim();
@@ -109,7 +118,6 @@ public class FissClaimRdaSink extends AbstractClaimRdaSink<FissClaimChange, RdaF
         .build();
   }
 
-  /** {@inheritDoc} */
   @Override
   MessageError createMessageError(
       String apiVersion, FissClaimChange change, List<DataTransformer.ErrorMessage> errors)
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/McsClaimRdaSink.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/McsClaimRdaSink.java
@@ -11,6 +11,7 @@ import gov.cms.bfd.pipeline.rda.grpc.RdaChange;
 import gov.cms.bfd.pipeline.rda.grpc.source.McsClaimTransformer;
 import gov.cms.bfd.pipeline.sharedutils.PipelineApplicationState;
 import gov.cms.model.dsl.codegen.library.DataTransformer;
+import gov.cms.mpsm.rda.v1.ChangeType;
 import gov.cms.mpsm.rda.v1.McsClaimChange;
 import java.io.IOException;
 import java.util.Comparator;
@@ -40,19 +41,29 @@ public class McsClaimRdaSink extends AbstractClaimRdaSink<McsClaimChange, RdaMcs
         transformer.withMbiCache(transformer.getMbiCache().withDatabaseLookup(transactionManager));
   }
 
-  /** {@inheritDoc} */
+  /**
+   * This implementation checks the {@link McsClaimChange#getChangeType}.
+   *
+   * <p>{@inheritDoc}
+   *
+   * @param mcsClaimChange message to check
+   * @return true if {@link McsClaimChange#getChangeType} is {@link ChangeType#CHANGE_TYPE_DELETE}.
+   */
+  @Override
+  public boolean isDeleteMessage(McsClaimChange mcsClaimChange) {
+    return mcsClaimChange.getChangeType() == ChangeType.CHANGE_TYPE_DELETE;
+  }
+
   @Override
   public String getClaimIdForMessage(McsClaimChange object) {
     return object.getClaim().getIdrClmHdIcn();
   }
 
-  /** {@inheritDoc} */
   @Override
   public long getSequenceNumberForObject(McsClaimChange object) {
     return object.getSeq();
   }
 
-  /** {@inheritDoc} */
   @Nonnull
   @Override
   RdaChange<RdaMcsClaim> transformMessageImpl(String apiVersion, McsClaimChange message) {
@@ -61,7 +72,6 @@ public class McsClaimRdaSink extends AbstractClaimRdaSink<McsClaimChange, RdaMcs
     return change;
   }
 
-  /** {@inheritDoc} */
   @Override
   int getInsertCount(RdaMcsClaim claim) {
     return 1 // Add one for the base claim
@@ -72,7 +82,6 @@ public class McsClaimRdaSink extends AbstractClaimRdaSink<McsClaimChange, RdaMcs
         + claim.getLocations().size();
   }
 
-  /** {@inheritDoc} */
   @Override
   RdaClaimMessageMetaData createMetaData(RdaChange<RdaMcsClaim> change) {
     final RdaMcsClaim claim = change.getClaim();
@@ -96,7 +105,6 @@ public class McsClaimRdaSink extends AbstractClaimRdaSink<McsClaimChange, RdaMcs
         .build();
   }
 
-  /** {@inheritDoc} */
   @Override
   MessageError createMessageError(
       String apiVersion, McsClaimChange change, List<DataTransformer.ErrorMessage> errors)
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/AbstractGrpcRdaSource.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/AbstractGrpcRdaSource.java
@@ -255,13 +255,17 @@ public abstract class AbstractGrpcRdaSource<TMessage, TClaim>
     /** Number of objects that have been received from the RDA API. */
     private final Counter objectsReceived;
     /**
-     * Number of objects that have been successfully stored by the sink. Generally <code>
-     * batches * maxPerBatch</code>
+     * Number of objects that have been successfully stored by the sink. Generally {@code batches *
+     * maxPerBatch}.
      */
     private final Counter objectsStored;
+    /** Number of {@link gov.cms.mpsm.rda.v1.ChangeType#CHANGE_TYPE_DELETE} messages skipped. */
+    private final Counter deleteMessagesSkipped;
+    /** Number of invalid objects skipped. */
+    private final Counter invalidObjectsSkipped;
     /**
-     * Number of batches/transactions used to store the objects. Generally <code>
-     * objectsReceived / maxPerBatch</code>
+     * Number of batches/transactions used to store the objects. Generally {@code objectsReceived /
+     * maxPerBatch}.
      */
     private final Counter batches;
 
@@ -282,6 +286,8 @@ public abstract class AbstractGrpcRdaSource<TMessage, TClaim>
       failures = appMetrics.counter(MetricRegistry.name(base, "failures"));
       objectsReceived = appMetrics.counter(MetricRegistry.name(base, "objects", "received"));
       objectsStored = appMetrics.counter(MetricRegistry.name(base, "objects", "stored"));
+      invalidObjectsSkipped = appMetrics.counter(MetricRegistry.name(base, "skipped", "invalid"));
+      deleteMessagesSkipped = appMetrics.counter(MetricRegistry.name(base, "skipped", "delete"));
       batches = appMetrics.counter(MetricRegistry.name(base, "batches"));
       final String uptimeGaugeName = MetricRegistry.name(base, "uptime");
       uptime = GAUGES.getGaugeForName(appMetrics, uptimeGaugeName);
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/StandardGrpcRdaSource.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/source/StandardGrpcRdaSource.java
@@ -212,13 +212,21 @@ public class StandardGrpcRdaSource<TMessage, TClaim>
               setUptimeToReceiving();
               final TMessage result = responseStream.next();
               metrics.getObjectsReceived().increment();
-              if (sink.isValidMessage(result)) {
+              if (sink.isDeleteMessage(result)) {
+                metrics.getDeleteMessagesSkipped().increment();
+                log.warn(
+                    "skipping DELETE message: claimType={} claimId={} seq={}",
+                    claimType,
+                    sink.getClaimIdForMessage(result),
+                    sink.getSequenceNumberForObject(result));
+              } else if (sink.isValidMessage(result)) {
                 batch.put(sink.getClaimIdForMessage(result), result);
                 if (batch.size() >= maxPerBatch) {
                   processResult.addCount(submitBatchToSink(apiVersion, sink, batch));
                 }
                 lastProcessedTime = clock.millis();
               } else {
+                metrics.getInvalidObjectsSkipped().increment();
                 log.info(
                     "skipping invalid claim: claimType={} claimId={} seq={}",
                     claimType,
--- a/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/FissClaimRdaSinkTest.java
+++ b/apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/FissClaimRdaSinkTest.java
@@ -29,6 +29,7 @@ import gov.cms.bfd.pipeline.rda.grpc.source.FissClaimTransformer;
 import gov.cms.bfd.pipeline.sharedutils.IdHasher;
 import gov.cms.bfd.pipeline.sharedutils.PipelineApplicationState;
 import gov.cms.model.dsl.codegen.library.DataTransformer;
+import gov.cms.mpsm.rda.v1.ChangeType;
 import gov.cms.mpsm.rda.v1.FissClaimChange;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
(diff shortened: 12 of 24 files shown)


apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
```diff
@@ -0,0 +1,74 @@
...
+	V_CARRIER        integer  := 1;    -- public static final int V_CARRIER_HAS_DATA     = (1 << 0);
+	V_INPATIENT      integer  := 2;    -- public static final int V_INPATIENT_HAS_DATA   = (1 << 1);
+	V_OUTPATIENT     integer  := 4;    -- public static final int V_OUTPATIENT_HAS_DATA  = (1 << 2);
+	V_SNF            integer  := 8;    -- public static final int V_SNF_HAS_DATA         = (1 << 3);
+	V_DME            integer  := 16;   -- public static final int V_DME_HAS_DATA         = (1 << 4);
+	V_HHA            integer  := 32;   -- public static final int V_HHA_HAS_DATA         = (1 << 5);
+	V_HOSPICE        integer  := 64;   -- public static final int V_HOSPICE_HAS_DATA     = (1 << 6);
+	V_PART_D         integer  := 128;  -- public static final int V_PART_D_HAS_DATA      = (1 << 7);
+BEGIN
+	SELECT COUNT(BENE_ID) INTO v_cnt FROM carrier_claims WHERE bene_id = v_bene_id limit 1;
```
billie-rose, line 32: Using `COUNT` concerns me as it's prone to doing full sequential table scans and is why the Patient endpoint timed out on occasion. Can we do an EXISTS for these checks?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
```diff
@@ -0,0 +1,74 @@
...
+	IF v_cnt > 0 THEN
+		v_rslt = V_CARRIER;
+	END IF;
+	SELECT COUNT(BENE_ID) INTO v_cnt  FROM inpatient_claims WHERE bene_id = v_bene_id limit 1;
+	IF v_cnt > 0 THEN
+${logic.hsql-only} v_rslt = v_rslt + V_INPATIENT;
+${logic.psql-only} v_rslt = v_rslt # V_INPATIENT;
```
billie-rose, line 39: I could be missing something, but why are we reassigning `v_rslt` to itself for each of these psql conditional statements? Could we do `v_rslt = v_rslt + V_[CLAIM-TYPE]` for both hsql and psql, then at the end cast the final `v_rslt` to a `::bit`? Would eliminate the psql/hsql only logic.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -17,6 +19,47 @@
 
 /** As set of methods to help form JPA queries. */
 public class QueryUtils {
+  /**
+   * Database function that checks all claims for a given beneficiaryId and returns a bitwise mask
+   * value that shows if a given claim type will have any data.
+   */
+  public static final String CHECK_CLAIMS_FOR_DATA =
```
billie-rose, line 26: `CHECK_CLAIMS_FOR_DATA_SQL`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -17,6 +19,47 @@
...
+  private static final int V_SNF_HAS_DATA = (1 << 3);
+  /** bitwise value denoting DME_CLAIMS data for a beneficiary. */
+  private static final int V_DME_HAS_DATA = (1 << 4);
+  /** bitwise value denoting HHA_CLAIMS data for a beneficiary. */
+  private static final int V_HHA_HAS_DATA = (1 << 5);
+  /** bitwise value denoting HOSPICE_CLAIMS data for a beneficiary. */
+  private static final int V_HOSPICE_HAS_DATA = (1 << 6);
+  /** bitwise value denoting PARTD_EVENTS data for a beneficiary. */
+  private static final int V_PART_D_HAS_DATA = (1 << 7);
+
```
billie-rose, line 62: What do you think about doing something like this? It would eliminate the logic inside `QueryUtils` and make it so we don't need to manage so many constant values.

```
public static BitSet hasClaimsData(EntityManager entityManager, long bene_id) {
   boolean carrier_has_data, inpatient_has_data, outpatient_has_data; // etc..
   carrier_has_data = hospice_has_data = outpatient_has_data = false; // etc..
   boolean[] claim_has_data_by_bit_pos = { carrier_has_data, hospice_has_data, outpatient_has_data };

   List<Object> values =
      entityManager
         .createNativeQuery(CHECK_CLAIMS_FOR_DATA)
         .setParameter("beneIdValue", bene_id)
         .getResultList();
   Integer maskVal = (Integer) (values != null && values.size() > 0 ? values.get(0) : 0);
   BitSet rslt = new BitSet(maskVal);
   for (int bit_pos = 0; bit_pos < claim_has_data_by_bit_pos.length; bit_pos++) {
      rslt.set(bit_pos, (maskVal & (1 << bit_pos)) == 1);
   }
   return rslt;
}
```
then...
```
public boolean hasClaimsDataByType(ClaimTypeV2 claimType, BitSet bitset) {
   return switch (claimType) {
      case ClaimTypeV2.Carrier -> bitSet.get(0) == 1;
      case ClaimTypeV2.Inpatient -> bitSet.get(1) == 1;
      ...
      default -> false;
   }
```
both methods residing in EOB endpoint instead of `QueryUtils`

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
mackec1: THat was a quick test to try to create some commonality between psql and hsql;,there is just too much dispersion between sql that the attempt to have more common code was pointless. So the psql is back to the simple 'perform' and the count in hsql is borderline meaningless since there are so few records in out hsql db.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
mackec1: So the using additive ops to emulate a mathematiclal XOR is a bit of a hack; in fact, I had specifically put a comment in the script noting that. While I still prefer the psql support for XOR vs the 'hack', since we only have 8 single-bit masks to worry about, I've just used the additive ops to generate the mask result.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: Done.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: So the Client[12).java is an implementation direction that should never have occurred. In fact, Karl ranted vs. that as a mistake in how the devs implemented this. In fact, those two classes are in line to become a single class if/when we have IOC in which we can inject the *Transformer class.

As for the other suggestion, not sure why we'd want to duplicate code in two EOBs vs a single QueryUtils serving both V1 and V2. As for the for loop, it's possible, but feels brittle in that it becomes more position-based vs. just decomposing the mask.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
```diff
@@ -0,0 +1,106 @@
...
+${logic.hsql-only}	 DECLARE V_HHA            integer DEFAULT 32;
+${logic.hsql-only}	 DECLARE V_HOSPICE        integer DEFAULT 64;
+${logic.hsql-only}	 DECLARE V_PART_D         integer DEFAULT 128;
+
+${logic.hsql-only}	SELECT COUNT(BENE_ID) INTO v_cnt FROM carrier_claims WHERE bene_id = v_bene_id;
+${logic.hsql-only}	IF v_cnt > 0 THEN
+${logic.hsql-only}    SET v_rslt = V_CARRIER;
+${logic.psql-only}  PERFORM 1 FROM carrier_claims WHERE bene_id = v_bene_id limit 1;
+${logic.psql-only}	IF FOUND THEN
+${logic.psql-only}    v_rslt = V_CARRIER;
```
brianburton, line 46: I'm getting cognitive whiplash from seeing the two versions of this function intermingled like this.  For me it would be easier to understand this if the two versions were just one after the other in the file.

Also when displayed in github's review page the indentation is inconsistent.  Maybe it's a tabs vs spaces issue?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -178,4 +221,49 @@ public static boolean isInRange(Instant lastUpdated, DateRangeParam range) {
...
+    BitSet rslt = new BitSet(maskVal);
+    rslt.set(CARRIER_HAS_DATA, (maskVal & V_CARRIER_HAS_DATA) == 1);
+    rslt.set(INPATIENT_HAS_DATA, (maskVal & V_INPATIENT_HAS_DATA) == 1);
+    rslt.set(OUTPATIENT_HAS_DATA, (maskVal & V_OUTPATIENT_HAS_DATA) == 1);
+    rslt.set(SNF_HAS_DATA, (maskVal & V_SNF_HAS_DATA) == 1);
+    rslt.set(DME_HAS_DATA, (maskVal & V_DME_HAS_DATA) == 1);
+    rslt.set(HHA_HAS_DATA, (maskVal & V_HHA_HAS_DATA) == 1);
+    rslt.set(HOSPICE_HAS_DATA, (maskVal & V_HOSPICE_HAS_DATA) == 1);
+    rslt.set(PART_D_HAS_DATA, (maskVal & V_PART_D_HAS_DATA) == 1);
+    return rslt;
```
brianburton, line 267: Since the integer bit mask is being processed here to create a `BitSet` why not create an `EnumSet` instead so that the providers could use a map or iterate over the enum values to decide what to do?  You could possibly even do an intersection of the `claimTypes` and result of this method.  `findByPatient` is begging to be a loop anyway and this could allow that.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -347,12 +348,17 @@ public Bundle findByPatient(
...
+    LOGGER.info(
+        String.format("BitSet for V2 claims, bene_id %d: %s", beneficiaryId, bitSet.toString()));
+
     /*
      * The way our JPA/SQL schema is setup, we have to run a separate search for
      * each claim type, then combine the results. It's not super efficient, but it's
      * also not so inefficient that it's worth fixing.
      */
-    if (claimTypes.contains(ClaimTypeV2.CARRIER)) {
+    if (claimTypes.contains(ClaimTypeV2.CARRIER) && bitSet.get(QueryUtils.CARRIER_HAS_DATA)) {
```
brianburton, line 361: These ifs could be a single loop if we used an `EnumSet`.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -350,75 +351,88 @@ public Bundle findByPatient(
...
+    LOGGER.info(
+        String.format("BitSet for V1 claims, bene_id %d: %s", beneficiaryId, bitSet.toString()));
+
     /*
      * The way our JPA/SQL schema is setup, we have to run a separate search for
      * each claim type, then combine the results. It's not super efficient, but it's
      * also not so inefficient that it's worth fixing.
      */
-    if (claimTypes.contains(ClaimType.CARRIER))
+    if (claimTypes.contains(ClaimType.CARRIER) && bitSet.get(QueryUtils.CARRIER_HAS_DATA)) {
```
brianburton, line 364: `EnumSet`/loop comment here too.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
brianburton: Just noting here that `(maskVal & (1 << bit_pos)) == 1` is incorrect because the bitwise AND does not return 1 or zero.  It returns an integer that includes only those bits which are 1 in both integers.  So the above would need to be something like:
```
int bit = (1 << bit_pos);
boolean bitIsSet =(maskVal & bit) == bit;
```

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
billie-rose: Maybe I just don't understand properly how `${logic.psql-only}` works - it just looks like it's only the `hsql` doing the additive ops and `psql` isn't. Is the `hsql-only` logic applying to both ?

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
billie-rose: NVM, my original comment no longer applies to your updated code.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
billie-rose: NVM, my original comment no longer applies to your updated code.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
billie-rose: Aren't the variables you've declared already position-based though? I don't see how one `LinkedHashMap` is any more or less brittle than 16 variables, but I do understand your concern with code-duplication. I'm not sure if it's better to have dup code or if it's better to have EOB-exclusive functionality in a shared utility class. I guess my preference would be having the duplicated code but I understand that's my personal opinion.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V111__setup_claims_availability_function.sql
mackec1: Reformatted to avoid the whiplash; I always wrestle with minimizing # of lines in a migration vs. just separating per hsql v. psql. Looks better without the whiplash!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: Excellent suggestion; there is follow-up work here to run the EOBs in parallel; will change the BitSet to EnumSet then. Also, cleaned up the mask stuff and added tests to assert mask value setting, etc. Thanks for review and suggestions...always appreciated.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
mackec1: yup. will change to EnumSet when we do the parallel work.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -347,12 +348,17 @@ public Bundle findByPatient(
           paging, eobs, loadedFilterManager.getTransactionTime());
     }
 
+    // See if we have claims data for the beneficiary.
+    BitSet bitSet = QueryUtils.hasClaimsData(entityManager, beneficiaryId);
+    LOGGER.info(
+        String.format("BitSet for V2 claims, bene_id %d: %s", beneficiaryId, bitSet.toString()));
```
lsmitchell, line 354: Should be a debug or somesuch; Doubt we want this clogging the logs for every request.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -350,75 +351,88 @@ public Bundle findByPatient(
       return TransformerUtils.createBundle(paging, eobs, loadedFilterManager.getTransactionTime());
     }
 
+    // See if we have claims data for the beneficiary.
+    BitSet bitSet = QueryUtils.hasClaimsData(entityManager, beneficiaryId);
+    LOGGER.info(
+        String.format("BitSet for V1 claims, bene_id %d: %s", beneficiaryId, bitSet.toString()));
```
lsmitchell, line 357: Similarly, lets make this a debug (or remove it)

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/QueryUtilsTest.java
```diff
@@ -128,6 +132,60 @@ public void testCreateDateRangePredicateBothBounds() {
     verify(builder).and(notNullPredicate, lowerBoundPredicate, upperBoundPredicate);
   }
 
+  /** Test {@link QueryUtils#hasClaimsData} testing claims bitmask values. */
```
lsmitchell, line 135: Can we make this a little more descriptive please? (Describe the test to someone reading this in 6-12 months)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
mackec1: I made them both debug message(s), but reluctantly. To me this had value in that it logged the bene_id AND the mask (i.e., {0, 1, 3, 5, 7} ) that would be applied to the EOB processing. That mask provided forensic info that might be useful in tracking down a problem.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/QueryUtilsTest.java
mackec1: Done.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/QueryUtilsTest.java
```diff
@@ -132,23 +132,43 @@ public void testCreateDateRangePredicateBothBounds() {
     verify(builder).and(notNullPredicate, lowerBoundPredicate, upperBoundPredicate);
   }
 
-  /** Test {@link QueryUtils#hasClaimsData} testing claims bitmask values. */
+  /**
+   * Test {@link QueryUtils#hasClaimsData} asserting various ways an integer mask value can be
+   * created (mathematical XOR or additive integers) and result in the same integer value. The test
+   * further verifies that leveraging operator AND functionality to extract a subset of the bits in
+   * the state, one can arrive at a boolean true/false outcome.@interface.
```
lsmitchell, line 139: Think theres a stray @interface here?

lsmitchell (approved):

lsmitchell (dismissed):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -17,6 +19,47 @@
...
+  public static final int DME_HAS_DATA = 4;
+  /** BitSet index identifier for HHA Claims. */
+  public static final int HHA_HAS_DATA = 5;
+  /** BitSet index identifier for Hospice Claims. */
+  public static final int HOSPICE_HAS_DATA = 6;
+  /** BitSet index identifier for Part D Events. */
+  public static final int PART_D_HAS_DATA = 7;
+
+  /** bitwise value denoting CARRIER_CLAIMS data for a beneficiary. */
+  public static final int V_CARRIER_HAS_DATA = (1 << 0);
```
billie-rose, line 47: Can we use the values defined above in these declarations? E.g.
`public static final int V_CARRIER_HAS_DATA = (1 << CARRIER_HAS_DATA);`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -178,4 +221,59 @@ public static boolean isInRange(Instant lastUpdated, DateRangeParam range) {
...
+   * Query database to determine which claim types have data for the specified beneficiary.
+   *
+   * @param entityManager {@link EntityManager} used to query database.
+   * @param bene_id used to identify the Benficiary to chek claims for.
+   * @return {@link BitSet} denoting which claims have data.
+   */
+  public static BitSet hasClaimsData(EntityManager entityManager, long bene_id) {
+    /*
+     * execute a database function that returns a bitwise mask value that denotes that the given
+     * claim type will have data for the specified beneficiayrId. This represents fast and efficient
```
billie-rose, line 254: `beneficiayrId` -> `bene_id or beneficiaryId`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -178,4 +221,59 @@ public static boolean isInRange(Instant lastUpdated, DateRangeParam range) {
...
     return true;
   }
+
+  /**
+   * Convert an integer value into claims data available BitSet.
+   *
+   * @param maskVal integer value suitable for decomposing into a Claims BitSet.
+   * @return {@link BitSet} denoting which claims have data.
+   */
+  public static BitSet convertClaimsBtimaskValue(int maskVal) {
```
billie-rose, line 231: `convertClaimsBtimaskValue` -> `convertClaimsBitmaskValue`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
mackec1: burling is already talking about a follow-up; basically perform all the claim types to be used in parallel.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/QueryUtilsTest.java
```diff
@@ -128,6 +132,83 @@ public void testCreateDateRangePredicateBothBounds() {
...
+    assertEquals(QueryUtils.V_PART_D_HAS_DATA, (testVal & QueryUtils.V_PART_D_HAS_DATA));
+    assertEquals(QueryUtils.V_HOSPICE_HAS_DATA, (testVal & QueryUtils.V_HOSPICE_HAS_DATA));
+    assertEquals(QueryUtils.V_HHA_HAS_DATA, (testVal & QueryUtils.V_HHA_HAS_DATA));
+    assertEquals(QueryUtils.V_DME_HAS_DATA, (testVal & QueryUtils.V_DME_HAS_DATA));
+    assertEquals(QueryUtils.V_SNF_HAS_DATA, (testVal & QueryUtils.V_SNF_HAS_DATA));
+    assertEquals(QueryUtils.V_OUTPATIENT_HAS_DATA, (testVal & QueryUtils.V_OUTPATIENT_HAS_DATA));
+    assertEquals(QueryUtils.V_INPATIENT_HAS_DATA, (testVal & QueryUtils.V_INPATIENT_HAS_DATA));
+    assertEquals(QueryUtils.V_CARRIER_HAS_DATA, (testVal & QueryUtils.V_CARRIER_HAS_DATA));
+
+    testVal =
```
billie-rose, line 191: I feel as though everything above this line in the test may not be necessary - because the code above is testing the functionality of `&` and `+` for bit manipulation, and I feel like it's safe to assume that neither of those operators in Java are going to change in the future.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: I suppose you could, but that ONLY works for CARRIER_CLAIMS since that has bit 1, so a value of 1. All the others go way beyond their index usage value.

billie-rose:
Just a couple more comments! Thanks for bearing with my nitpicking

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: well yes and no; beneficiaryId is a function of some really old code that was specific to the various ORM models. At one point, Karl wanted to get rid of those Java field names (that he actually created) when we did a big schema migration a while back. We never did that follow-up work, but I tend to use CCW name(s) unless dealing with an ORM model.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: fix; thx for catching that.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/QueryUtilsTest.java
mackec1: True, but I wanted to emphasize that the mask returns the mask value (i.e., not 0, not 1 (unless CARRIER, etc.)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
billie-rose: Ah I was more so suggesting it because of Java naming conventions.

lsmitchell (approved):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -178,4 +221,59 @@ public static boolean isInRange(Instant lastUpdated, DateRangeParam range) {
...
+   * Query database to determine which claim types have data for the specified beneficiary.
+   *
+   * @param entityManager {@link EntityManager} used to query database.
+   * @param bene_id used to identify the Benficiary to chek claims for.
+   * @return {@link BitSet} denoting which claims have data.
+   */
+  public static BitSet hasClaimsData(EntityManager entityManager, long bene_id) {
+    /*
+     * execute a database function that returns a bitwise mask value that denotes that the given
+     * claim type will have data for the specified beneficiayrId. This represents fast and efficient
```
mjburling, line 254: I think @billie-rose is on to something and after a brief discussion... this could be more consistent. After talking to @mackec1, I see where he's coming from on this as far as matching the function. I think after addressing this, we should be in good shape. Sorry for the delay, @mackec1!

```suggestion
  public static BitSet hasClaimsData(EntityManager entityManager, long beneficiaryId) {
    /*
     * execute a database function that returns a bitwise mask value that denotes that the given
     * claim type will have data for the specified beneficiaryId. This represents fast and efficient
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
brianburton: +1 to Billie's suggestion. I don't see why it wouldn't work and it would make the code clearer.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
brianburton: +1 here also.  parameter names are meant to be camel case.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
billie-rose: How do you mean? Isn't it a 1:1 translation? Like if `OUTPATIENT_HAS_DATA = 2`, and `V_OUTPATIENT_HAS_DATA = (1 << 2);`, aren't you just replacing the hard-coded `2` with the variable `OUTPATIENT_HAS_DATA` which is also `2`? And then doing that for all of them? AKA

`V_SNF_HAS_DATA =  (1 << SNF_HAS_DATA);`
`V_DME_HAS_DATA = (1 << DME_HAS_DATA);`

it would be one less hard-coded value in my mind

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
```diff
@@ -178,4 +221,59 @@ public static boolean isInRange(Instant lastUpdated, DateRangeParam range) {
...
+   * Query database to determine which claim types have data for the specified beneficiary.
+   *
+   * @param entityManager {@link EntityManager} used to query database.
+   * @param beneficiaryId used to identify the Benficiary to chek claims for.
+   * @return {@link BitSet} denoting which claims have data.
+   */
+  public static BitSet hasClaimsData(EntityManager entityManager, long beneficiaryId) {
+    /*
+     * execute a database function that returns a bitwise mask value that denotes that the given
+     * claim type will have data for the specified beneficiayrId. This represents fast and efficient
```
mjburling, line 254: Minor typo still persists
```suggestion
     * claim type will have data for the specified beneficiaryId. This represents fast and efficient
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: I don't disagree with either you your thoughts on this; am not going to do it right now as there is a follow-up ticket that will run the EOBs in parallel and I intend to make this bit handling encapsulated in an EnumSet for that. Also, changing the static values to be something like:
rslt.set(CARRIER_HAS_DATA, (maskVal & (1 << CARRIER_HAS_DATA)) != 0);
results in having to replicate that shifting logic in a lot of places which introduces more changes than warranted for code that will change to an EnumSet.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: changed to camel case.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: fixed. thx.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/QueryUtils.java
mackec1: ahh, I see what you were saying; thx, done.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/QueryUtilsTest.java
mackec1: removed some tests.

brianburton (approved):
Nice work!

mjburling (approved):

PR #1712, opened 2023-05-01
BFD-2566: Move linter to pre push hook

cbrunefearless:
**JIRA Ticket:**
[BFD-2566](https://jira.cms.gov/browse/BFD-2566)

User Story:
As a BFD contributor, I want the Java linting step to be moved from the pre-commit to the pre-push hook so that I can avoid the penalty of running the linter on every commit.

AC:
The Java linter executes automatically via pre-push hook
The other functionality of the existing pre-commit hook remains in tact

---

### What Does This PR Do?

Moves the linter to the pre push hook

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


lsmitchell (dismissed):

brianburton (approved):
Looking forward to using this!

mjburling (changes requested):
I don't believe this will be quite as simple as lifting the existing shell function from `pre-commit` and writing it into `pre-push` because it operates off of a `git diff` which _may_ or _may not_ include differences when preparing to push. Some of the existing function applies, but it will need to be triggered when there are relevant differences between the local working copy and whatever is stored in the remote.

cbrunefearless:
> I don't believe this will be quite as simple as lifting the existing shell function from `pre-commit` and writing it into `pre-push` because it operates off of a `git diff` which _may_ or _may not_ include differences when preparing to push. Some of the existing function applies, but it will need to be triggered when there are relevant differences between the local working copy and whatever is stored in the remote.

Im doing a compare on the remote and local of the apps directory to see if we should run the formatter, if there is any changes than we should run the formatter if its in the other folders that are not maven directories, do not run the formatter.

mjburling (approved):

apps/pom.xml
```diff
@@ -586,8 +562,39 @@ checkSecretFilesForPlainText() {
...
+checkJavaFormat() {
+  echo 'Verifying Java source code format...'
+  branch_name="$(git rev-parse --abbrev-ref HEAD 2>&1)"
+
+  git fetch origin 2>&1
+
+  localChanges="$(git diff --name-only "origin/${branch_name}:apps" "${branch_name}:apps" 2>&1)"
+
+  echo "Verifying Java source code format for branch: ${branch_name} ..."
+  if [ "$localChanges" = '' ]; then
```
mackec1, line 579: obviously a matter of preference, but curious...
So if 'no changes detected', why not just exit right there? minimizes 'if' nesting resulting in less indentqation

mackec1 (approved):

apps/pom.xml
cbrunefearless: we could but didnt want to replicate two return 0's

PR #1723, opened 2023-05-04
BFD-2544: PoC for moving transformers into Spring-set instantiated classes for easier testing

lsmitchell:
**JIRA Ticket:**
[BFD-2544](https://jira.cms.gov/browse/BFD-2544)

As a BFD engineer, I want BFD to be implemented in a more canonical, framework-oriented way so that I can take advantage of the features of the framework and more easily reason about and maintain the codebase.

---

### What Does This PR Do?

Makes some adjustments to try and move some of BFD's more static elements into instantiated singletons which are set up by Spring, which makes the component classes easier to mock and test in isolation.

This is just one transformer adjusted with some tests to show the benefits of this strategy. The goal of this would be more isolated tests that would reduce the need to always write an expensive end-to-end test to test various endpoint functionality.

Adjustments:
- BeneficiaryTransformerV2 is now a Spring instantiated singleton bean, configured during startup automatically
    - It's one dependency, metricRegistry, now no longer needs to be passed into each transformation call
    - metricRegistry is injected automatically from the existing metricRegistry bean during the component scan
    - Overall this class profile remains similar to being a static class, as its only instantiated once and kept as a singleton bean; however now it can be more easily isolated/mocked in tests, reducing the need to test the transformer in every resourceProvider test
- Various "R4PatientResourceProviderIT" tests were actually end-to-end tests that could be ITs or unit tests if the transformer was mockable. The following adjustments were made to showcase some efforts that could be made towards this goal:
    - Created R4PatientResourceProviderE2E to contain the actual end to end tests we wish to keep
    - R4PatientResourceProviderIT adjusted to remove all tests which were testing anything beyond the boundary of transformer <> resourceProvider
    - R4PatientResourceProviderTest has additional mocking to support new tests (previously only checked some exception cases, so we not properly mocked)
    - Per the above adjustments, some E2E tests were replaced by other tests (following are tests from the E2E/previous R4PatientResourceProviderIT class):
        - **readExistingPatient**
            - left as an E2E test as a happy path
            - added testReadWhenBeneExistsExpectTransformerCalled (Unit test)
            - added testReadWhenBeneExistsExpectMetricsCalled (Unit test)
        - **readExistingPatientIncludeAddressFieldsTrue**
            - Replaced with testReadWhenHeadersSetExpectHeadersPassedToTransformer (Unit test)
            - The resource provider just passes the headers down, so thats all we really need to test. A test in the controller already tests that the header, once passed, does the right thing in the transformer.
        - **readExistingPatientIncludeIdentifiersFalse / readExistingPatientIncludeIdentifiersBlank**
            - Removed, we no longer use IncludeIdentifiers in v2, so these tests are no longer needed (and setting to false is implicitly tested in other unit tests)
        - **readMissingPatient**
            - Replaced with testReadWhenNoDbResultThrowsException (Unit test)
            - Everything here happens in the resource provider; never needed to be an E2E
        - **searchForExistingPatientByLogicalId**
            - Left this as an E2E happy path test
            - Added testSearchByLogicalIdWhenBeneExistsExpectTransformerCalled (Unit test) which could theoretically just replace it
        - **searchForPatientByLogicalIdWithPaging**
            - Replaced by testSearchByLogicalIdWhenPagingRequestedExpectPageData and testSearchByLogicalIdWhenNoPagingRequestedExpectNoPageData (Unit tests)
        - **searchForMissingPatientByLogicalId**
            - Replaced by testSearchByLogicalIdWhenBeneDoesntExistExpectEmptyBundle (Unit Test)
        - **searchForExistingPatientByMbiHash**
            - Retained the original E2E test since this uses a lot of JPA code
            - Added testSearchByIdentifierIdWhenMbiHashExpectPatient
       - **searchForExistingPatientByMbiHashWithBeneDupes**
           - This is a big honkin test that should/could be multiple tests; left it alone but could be converted with some work
       - **searchForExistingPatientByMbiHashWithPaging**
          - Replaced by testSearchByIdentifierIdWhenPagingRequestedExpectPageData (Unit Test)
      - **searchForExistingPatientByMbiHashHasHistoricMbis**
          - Converted to an integration test with the same name
      - **searchForMissingPatientByMbiHash**
          - Replaced by testSearchByIdentifierIdWhenBeneDoesntExistExpectEmptyBundle
      - **searchForExistingPatientByPartDContractNum**
          - Kept as happy path E2E and added testSearchByCoverageContractWhenContractExistsExpectTransformerCalled (Unit Test)
          - Removed duplicate searchForExistingPatientByPartDContractNumAndYear
      - **searchForPatientByPartDContractNumWithPaging**
          - Replaced by testSearchByCoverageContractWhenPagingRequestedExpectPageData
      - **searchForMissingPatientByPartDContractNum** and **searchForNonExistingPatientByPartDContractNumAndYear**
          - Replaced by testSearchByCoverageContractWhenNoPatientsExpectEmptyBundle (Unit test)
      - **searchForPatientByPartDContractNumWithAnInvalidYear** and **searchForPatientByPartDContractNumWithAInvalidContract**
          - Already covered by existing unit tests, so removed

Performance:
Notes: Run via maven on the command line, building just bfd-server-war; individual test times done with `mvn failsafe:integration-test` and `mvn surefire:test`
Current master:
bfd-server-war total build time: 3 min 46 s
bfd-server-war IT test time: 2 min 16 s
bfd-server-war unit test time: 25.2 s

Post-adjustment:
bfd-server-war total build time: 3 min 37 s
bfd-server-war IT+E2E test time: 2 m 16s
bfd-server-war unit test time: 25.2 s

Speedwise, there isn't much of a difference from this change. While things initially looked promising from IDE test speeds, maven seems much slower and doesnt gain much from this.

This may because intelliJ's timer doesnt start counting until the tests run, and may hide some of the overhead of an extra test class (or other hidden overhead).

Fix includes E2E tests converted into unit tests, cut tests that were redundant due to unit tests, and new unit tests to increase coverage in a couple cases where it was warranted, so this isn't an exact 1:1 replacement, since some tests were completely removed to redundancy and some were added which were additions.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


lsmitchell (commented):
While I'm a bit disappointed the test conversion didn't result in a performance boost, the overall changes still help us test in better isolation and at least did not degrade any performance, as well as aligning us with best practices.

JeremyIR:
Amazing work, Logan! Didnt you say in standup there was 30% reduction of something. Also curious what the mvn failsafe:integration-test and mvn surefire:test results would be in the ec2 instances (Not that you have to go and run those). Also do you have any documentation you could share on how you made these changes.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/R4PatientResourceProviderIT.java
```diff
@@ -2,579 +2,173 @@
...
+    when(requestDetails.getHeader(CommonHeaders.HEADER_NAME_INCLUDE_ADDRESS_FIELDS))
+        .thenReturn("false");
+    // We dont use this anymore on v2, so set it to false since everything should work regardless of
+    // its value
+    when(requestDetails.getHeader(CommonHeaders.HEADER_NAME_INCLUDE_IDENTIFIERS))
+        .thenReturn("false");
+  }
+
+  /** Sets up the default entity manager mocks. */
+  private void mockEntityManager() {
```
mackec1, line 151: nice pattern setup.

mackec1 (approved):
I know this is a POC, and the R4 stuff looks good...shouldn't we (by extension) expand the POC for STU3?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/BeneficiaryTransformerV2.java
```diff
@@ -27,50 +28,52 @@
...
+  /** The Metric registry. */
+  private final MetricRegistry metricRegistry;
+
+  /** Enumerates the options for the currency of an {@link Identifier}. */
+  public enum CurrencyIdentifier {
+    /** Represents a current identifier. */
+    CURRENT,
+    /** Represents a historic identifier. */
+    HISTORIC;
+  }
```
brianburton, line 46: I generally put inner classes and enums at the bottom (though I may violate this from time to time...).  Curious why this was moved from the bottom to the top?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/BeneficiaryTransformerV2.java
```diff
@@ -27,50 +28,52 @@
...
-        metricRegistry
-            .timer(MetricRegistry.name(BeneficiaryTransformerV2.class.getSimpleName(), "transform"))
-            .time();
-    Patient patient = transform(beneficiary, requestHeader, addHistoricalMbis);
-    timer.stop();
-
-    return patient;
+  public BeneficiaryTransformerV2(MetricRegistry metricRegistry) {
+    requireNonNull(metricRegistry);
+    this.metricRegistry = metricRegistry;
```
brianburton, line 59: Great to have this done.  I'm a big fan of dependency injection.

brianburton (approved):
This looks good to me.  Nice to see DI being utilized more.

brianburton:
I'm not a fan of mocking JPA criteria queries.  It's a necessary evil of course (and I've suffered though writing tests of those in the PACA code) but it might be nice to move away from using the query builder and just using JPQL more since that's easier to mock and read.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/BeneficiaryTransformerV2.java
lsmitchell: Combo of style i've seen in the past (and in other classes) and personal preference; not sure what best practice is for this. I'd generally seen enums declared at the top since they are generally not huge constructs (like inner classes may be) and if they're within a class they are probably important to the understanding of that class' functionality. I personally like all vars/enums declared at the top so reading top-to-bottom you have a good idea of all the class level declarations first.

lsmitchell:
> I'm not a fan of mocking JPA criteria queries. It's a necessary evil of course (and I've suffered though writing tests of those in the PACA code) but it might be nice to move away from using the query builder and just using JPQL more since that's easier to mock and read.

Yeah it's hard to discern but the mocks for the JPA were _mostly_ centered around moving past that code as if the query succeeded due to the number of calls to mock to accurately fake a JPA query builder sequence. For anything where we needed to test a more complicated scenario I left the tests in the E2E class so they could just use a real DB and run through the JPA queries for real.

I do agree JPQL would make things easier to read (and primarily, to test), and would be good to move towards if there are no real differences in functionality/performance.

PR #1769, opened 2023-06-01
BFD-2680: program for KMS based secret management

brianburton:
**JIRA Ticket:**
[BFD-2680](https://jira.cms.gov/browse/BFD-2680)

As a BFD engineer I want to be able to encrypt configuration settings with the following features:

* Be able to encrypt only parts of files.
* Be able to use KMS for all encryption.

---

### What Does This PR Do?

Adds a kotlin script that can read and edit text files and encrypt/decrypt specific parts of the file.

You must have kotlin installed and JAVA_HOME set.

**The first time you run the script will take several seconds to compile the script.  After that it should run from the build cache in a second or less.**

The script accepts one required option `--key` with a KMS key ARN.

Two additional options are supported:

* `--endpoint uri` Lets you specify an alternative AWS URI.  This is intended for tesing with localstack. (usually `http://localhost:4566`)
* `--region region` Lets you specify an alternative AWS region.  Defaults to `us-east-1`.

One environment variable is supported:

* `EDITOR` specifies a text editor to use in `edit` mode.  Defaults to `vi`.

To run the script you can just run the `util/cipher/cipher.main.kts` script with appropriate command line arguments.  Run it with no arguments to see a usage message.

```
$ kotlin cipher.main.kts
usage: cipher [options] cat|encrypt|decrypt|edit source [dest]

If dest is omitted the source file will be overwritten.

cat writes a completely decrypted version of source to stdout.

decrypt writes a completely decrypted version of source to dest.

encrypt writes a version of source with all SECURE blocks replaced by CIPHER blocks to dest.
  
edit creates a temporary file with CIPHER blocks converted to SECURE
  blocks and opens it using the editor defined in EDITOR environment variable.
  If the temp file is modified the SECURE blocks are converted back to CIPHER
  blocks the result is written to the dest file.
  Default editor is vi.

Note: Both cat and decrypt remove the secure text delimiters so they are not reversible.

options:
--key arg      : KMS key ARN (REQUIRED)
--endpoint uri : Localhost endpoint for simulated KMS
```

Files are arbitrary text files.  Blocks of sensitive information should be surrounded by `<<SECURE>>` and `<</SECURE>>` to separate them from the rest of the text.  Put the clear text inside the delimiters.

```
my.first.property=plain text
my.first.secret=<<SECURE>>this should be encrypted<</SECURE>>
```

Once you have created a file (or already have one) you can use the `encrypt` command to encrypt it.

```
$ cat sample.properties 
my.first.property=plain text
my.first.secret=<<SECURE>>this should be encrypted<</SECURE>>

$ ./utils/cipher/target/cipher-1.0.0-SNAPSHOT/cipher.sh encrypt sample.properties 
$ cat sample.properties 
my.first.property=plain text
my.first.secret=<<CIPHER>>..lots of ugly base64 stuff...<</CIPHER>>
```

You can view the decrypted version of a file using the `cat` command:

```
$ ./utils/cipher/target/cipher-1.0.0-SNAPSHOT/cipher.sh cat sample.properties 
my.first.property=plain text
my.first.secret=this should be encrypted
```

You can use the `decrypt` command to produce a fully decrypted file:

```
$ KMS_KEY=..whatever...
$ kotlin cipher.main.kts  --key $KMS_KEY decrypt sample.properties decrypted-sample.properties
$ cat decrypted-sample.properties 
my.first.property=plain text
my.first.secret=this should be encrypted
```

The `edit` command doesn't lend itself to samples in markdown.  It works like the `ansible-vault edit` command.  A couple of points about `edit`:

* If you leave the editor without saving the original file will be left unchanged.
* If you leave the editor after saving an empty file the original file will be left unchanged.
* Any secure blocks that have not changed will retain their old encrypted value.

These checks are to provide a way to bail out in the middle as well as to avoid commits with spurious changes.  The re-use of encrypted values is intended to limit the amount of noise in git diffs.

If you want to re-encrypt all `SECURE` blocks in a file just use the `encrypt` command on it again.  Only the `edit` command preserves encrypted values.

This PR introduces a new tool for encrypting configuration settings.  This tool is a kotlin script and uses the following maven dependencies to accomplish its task:

```
@file:DependsOn("org.jetbrains.kotlinx:kotlinx-collections-immutable-jvm:0.3.5")
@file:DependsOn("com.amazonaws:aws-encryption-sdk-java:2.4.0")
@file:DependsOn("software.amazon.awssdk:kms:2.20.74")
@file:DependsOn("com.squareup.okio:okio:3.3.0")
```

* The kotlinx immutable collections library just provides collections used by the script.
* AWS encryption SDK is being used for first time in BFD project.
* AWS KMS SDK is being used for the first time in BFD project.
* okio is just for input/output of bytes by the script.

      * [x] Yes
      * [ ] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [x] Yes
      * [ ] No

* N/A

### Submitter Checklist


apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,462 @@
...
+@file:DependsOn("software.amazon.awssdk:kms:2.20.74")
+@file:DependsOn("com.squareup.okio:okio:3.3.0")
+
+import com.amazonaws.encryptionsdk.AwsCrypto
+import com.amazonaws.encryptionsdk.CryptoResult
+import com.amazonaws.encryptionsdk.kmssdkv2.KmsMasterKey
+import com.amazonaws.encryptionsdk.kmssdkv2.KmsMasterKeyProvider
+import kotlinx.collections.immutable.PersistentMap
+import kotlinx.collections.immutable.persistentMapOf
+import okio.*
```
lsmitchell, line 14: If we're sticking with our java standards in Kotlin, we should avoid star imports. (Not sure if Kotlin handles imports more intelligently where this isn't needed)

apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,462 @@
...
+    var firstIndex = -1L
+    var firstToken = Token.None
+    for (token in Token.values()) {
+        if (token.bytes.size > 0) {
+            val index = this.indexOf(token.bytes)
+            if (index >= 0) {
+                if (firstIndex < 0 || index < firstIndex) {
+                    firstIndex = index
+                    firstToken = token
+//                println("$index $token")
```
lsmitchell, line 121: Should these commented lines be removed?

apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,462 @@
...
+          CIPHER_REGION    Same as --region
+        
+        Secure blocks are created and maintained by humans.
+        
+          ${Span.Secure.start.text}...any text you would like to have encrypted...${Span.Secure.end.text}
+        
+        Cipher blocks are created and maintained by the script.
+        
+          ${Span.Cipher.start.text}...base 64 encrypted data from a secure block...${Span.Cipher.end.text}
+    """.trimIndent()
```
lsmitchell, line 317: Nice usage instructions 👍

lsmitchell (commented):
Looks good overall, appreciate the detailed PR description and usage instructions.

apps/utils/cipher/cipher.main.kts
brianburton: Since this is a script and these are helpful for debugging I left them in rather than have to recreate them later if we need to debug some more.

lsmitchell (approved):

JeremyIR:
This looks great. I know 500 lines isnt necessary long, but is it possible to have a separate file for some of the helper methods? Or is that not feasible. Again not saying that is required, it just might look a bit readable for less stellar engineers like you.

StewGoin (approved):
👍

mjburling (changes requested):
Use of the AWS Encryption SDK is an unexpected but pleasant surprise! I'll have to do a _little_ more homework on the strategic difference between the `KmsClient` alone and this. Your solution seems to be consistent with the complementary [Best Practices Guide](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/best-practices.html). I think the most comforting parts of that guide are as follows (emphasis mine):

> The AWS Encryption SDK is designed to make it easy for you to protect your data using industry standards and best practices. While **many best practices are selected for you in default values**, some practices are optional but recommended whenever it's practical.
> [...]
> Use default values
> The **AWS Encryption SDK designs best practices into its default values**. Whenever possible, use them. For cases where the default is impractical, we provide alternatives, such as algorithm suites without signing. We also provide opportunities to advanced users for customization, such a custom keyrings, master key providers, and cryptographic material managers (CMMs). **Use these advanced alternatives cautiously and have your choices verified by a security engineer whenever possible**.

tl;dr; defaults should yield a secure solution; avoid deviations from the defaults whenever possible; if you adopt any non-default functionality, have a security engineer verify the solution.

I left some modest suggestions/comments here, but this **really** promising!

apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,504 @@
+#!/usr/bin/env kotlin
+
+@file:DependsOn("org.jetbrains.kotlinx:kotlinx-collections-immutable-jvm:0.3.5")
+@file:DependsOn("com.amazonaws:aws-encryption-sdk-java:2.4.0")
+@file:DependsOn("software.amazon.awssdk:kms:2.20.74")
+@file:DependsOn("com.squareup.okio:okio:3.3.0")
```
mjburling, line 6: May as well do latest, but those awsdk point releases increment _pretty_ quickly:
* https://mvnrepository.com/artifact/software.amazon.awssdk/kms
* https://mvnrepository.com/artifact/com.squareup.okio/okio
```suggestion
@file:DependsOn("software.amazon.awssdk:kms:2.20.120")
@file:DependsOn("com.squareup.okio:okio:3.5.0")
```

apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,504 @@
...
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.nio.file.attribute.PosixFilePermissions
+import java.util.*
+import kotlin.system.exitProcess
+
+
+val APP_NAME = "cipher"
+val DEFAULT_REGION = "us-east-1"
+val DEFAULT_EDITOR = "vi"
```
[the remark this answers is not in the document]
brianburton: Hmm.  While I love `emacs` the fact is that `vi` is likely to be installed everywhere even if `emacs` isn't to `vi` seemed the safer default.

apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,504 @@
...
+            decrypted.store(outputFile.sink())
+        }
+
+        "rewind" -> {
+            val original = loadText(inputFile)
+            val rewound = original.rewind(cipher)
+            rewound.store(outputFile.sink())
+        }
+
+        "edit" -> {
```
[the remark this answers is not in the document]
brianburton: Yes, edit could use some attention for this and other reasons.  Maybe this could be added in a follow-on sprint.  I'm sure we'll gain a lot of insights through actual experience with the tool.

brianburton:
> This looks great. I know 500 lines isnt necessary long, but is it possible to have a separate file for some of the helper methods? Or is that not feasible. Again not saying that is required, it just might look a bit readable for less stellar engineers like you.

Scripts are inherently single files.  Splitting into multiple files would complicate installing and using the script on other systems.  If it grows too large we could consider making it a stand alone application.  That would run into further complications though as kotlin makes running scripts easier than running apps.

apps/utils/cipher/cipher.main.kts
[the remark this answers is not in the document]
mjburling: I agree. I don't think it's possible that we will collectively lose sight of usability challenges when we actually start using this, right? With that, I think we can move ahead even without opening a ticket at this time.

apps/utils/cipher/cipher.main.kts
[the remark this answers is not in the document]
mjburling: I agree. This was _mostly_ in jest 😉.

mjburling (approved):
Approving this with **very** limited reservations surrounding usability challenges. The solution works as intended, but there are some known (and likely unknown) user experience issues that we'll defer for now in hopes of getting additional feedback from our users (hey, that's us!).

apps/utils/cipher/cipher.main.kts
mjburling, line 1: Looking for any specific feedback from @malessi prior to acceptance.

apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,505 @@
...
+                val original = loadText(inputFile)
+                val cipherCache = original.extractCipherCache(cipher)
+                val rewound = original.rewind(cipher)
+                rewound.store(tempFile.sink())
+                ProcessBuilder(listOf(editor, tempFile.path))
+                    .redirectInput(ProcessBuilder.Redirect.INHERIT)
+                    .redirectOutput(ProcessBuilder.Redirect.INHERIT)
+                    .redirectError(ProcessBuilder.Redirect.INHERIT)
+                    .start()
+                    .waitFor()
```
malessi, line 471: nit: My `EDITOR` is set to `code -w` ([`-w`/`--wait`](https://code.visualstudio.com/docs/editor/command-line#_core-cli-options) instructs VS Code wait until it's closed before returning to its parent process, necessary for scripts that expect that opening the `EDITOR` will pause execution until it's closed), so this script attempts to open an executable named `code -w`, instead of `code` with the `-w` flag. 

_Definitely_ not a big deal, as I can just specify `EDITOR=nvim kotlin ...` as a workaround, but I figured I'd leave this comment for posterity.

apps/utils/cipher/cipher.main.kts
```diff
@@ -0,0 +1,505 @@
...
+        val encryptResult: CryptoResult<ByteArray, KmsMasterKey> =
+            crypto.encryptData(keyProvider, plainText.toByteArray(), context)
+        return Base64.getUrlEncoder().encodeToString(encryptResult.result).encodeUtf8()
+    }
+
+    // Decrypts the provided base 64 encoded cipher text and returns its equivalent plain text.
+    fun decrypt(cipherText: ByteString): ByteString {
+        val decryptResult = crypto.decryptData(keyProvider, Base64.getUrlDecoder().decode(cipherText.utf8()))
+        return decryptResult.result.toByteString()
+    }
```
malessi, line 199: nit: `decrypt` appears to not work with KMS CMK aliases, instead requiring that the _true_ `key` (not `alias`) ARN is provided. Not a big deal, and definitely not something for this script to handle, but maybe something to explicitly call out in the help text?

Note that `encrypt` _does_ work with aliases. I thought this was surprising, but it might just be a quirk of KMS key aliases that I'm unfamiliar with.

malessi (approved):
This looks great to me! I gave it a thorough test run, merging `test.yaml` and `test.eyaml` into a singular configuration and verifying that this script can encrypt and decrypt it losslessly.

I have a couple _very_ minor comments that I left for posterity's sake, but they don't need to be addressed right now.

apps/utils/cipher/cipher.main.kts
mjburling: > nit: `decrypt` appears to not work with KMS CMK aliases, instead requiring that the _true_ `key` (not `alias`) ARN is provided. Not a big deal, and definitely not something for this script to handle, but maybe something to explicitly call out in the help text?

That's true! I've found working with the KMS API to be a little annoying at times due to the ambiguity surrounding where aliases and IDs can be used vs where Key ARNs are mandated–maybe it's not just me?

Here, we're only using a subset of the encryption-sdk's functionality. Even so, the developers made an intentional decision for us, and we aren't doing any kind of discovery, lookups, or mapping from less specific identifiers (i.e. aliases, key ids, alias arns) to their respective global identifiers of the key arn on decryption. [See this example for a tiny bit more discussion on this](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/java-example-code.html#java-example-strings).

From the help text:
```
[..snipped..]
        Options:
        --key ARN      : KMS key ARN (REQUIRED unless CIPHER_KEY is defined).
        --endpoint URI : Localhost endpoint URI for localstack KMS.
        --region name  : AWS region name.  Defaults to $DEFAULT_REGION.
[..snipped..] 
```

Maybe in a future iteration, we could consider checking input of the `--key` to ensure it's a valid KMS key arn before proceeding 🤔.

apps/utils/cipher/cipher.main.kts
brianburton: OK that's a great feature for the next sprint.  Maybe split the editor string on spaces.  I'd hate to get too elaborate with it.  It could get really elaborate if we wind up dealing with quoted strings etc but simple spaces should be fine.

apps/utils/cipher/cipher.main.kts
brianburton: Another feature to add next sprint!  Maybe we can do an explicit lookup of the ARN from an alias.  Failing that the validity check is an excellent idea.

PR #1784, opened 2023-06-13
BFD-2684: Increase minimum AWS file download part size to 200MB

lsmitchell:
**JIRA Ticket:**
[BFD-2684](https://jira.cms.gov/browse/BFD-2684)

Attempts to fix a bug where AWS downloads get stuck sometimes.

---

### What Does This PR Do?

Increases the async S3 client configuration for maximum file size before splitting the file into smaller parts to 200MB. This will also increase the chunk size of each download piece to 200MB, which should reduce the number of threads needed for a large download.

Also deletes a KMS key addition to the copy request that was causing issues in prod-sbx; basically if the key is found and added to the request the copy fails, because AWS was expecting some specific header that wasnt added correctly. The kms key addition was unneeded, because the copy request if absent the specific kms key will default to the bucket-level encryption, which is what we want anyway. We'll let AWS handle that automatically instead of explicitly adding it.

Also increases the size of the test/prod-sbx pipeline boxes now that we have an on-demand solution for the ccw pipeline creation.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/s3/SharedS3Utilities.java
+++ b/apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/s3/SharedS3Utilities.java
@@ -20,7 +20,6 @@ import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.S3Configuration;
-import software.amazon.awssdk.services.s3.internal.crt.S3CrtAsyncClient;
 import software.amazon.awssdk.services.s3.model.Bucket;
 import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
 import software.amazon.awssdk.services.s3.model.DeleteBucketRequest;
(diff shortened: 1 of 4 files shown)


mackec1 (approved):

lsmitchell:
As a side note locally everything works fine even with the old code; this only seemed to fail on prod-sbx so far for a synthetic load. Suspect this may fix it, but can't be sure without a real test.

billie-rose (dismissed):

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/s3/SharedS3Utilities.java
```diff
@@ -97,7 +96,9 @@ public static S3AsyncClient createS3AsyncClient(Region awsS3Region) {
     if (minioConfig.useMinio) {
       return createS3AsyncMinioClient(awsS3Region, minioConfig);
     }
-    return S3CrtAsyncClient.builder().region(awsS3Region).build();
+    // Split the file download (automatically using the async client) if its >200 MB
+    long mb = 1048576L;
+    return S3AsyncClient.crtBuilder().minimumPartSizeInBytes(200 * mb).region(awsS3Region).build();
```
mjburling, line 101: This seems reasonable. What's the default part size?

I don't think there's any harm in doing this, but it doesn't seem like we're _too_ confident in this change being the solution. I'm building the images based off this PR now and I can deploy the result to `prod-sbx` so we can evaluate it **before** acceptance. What do you think?

lsmitchell: Yup that seems like a good idea. The default previously was 8MB, so a 400MB file would be split into 50 parts. I'm still not entirely sure what the issue is given that things work fine locally, so a test on the box would be prudent.

lsmitchell:
Testing worked with the changes here in prod-sbx; was able to load all data and it was successfully moved to done.

mjburling (approved):

malessi (approved):

PR #1812, opened 2023-06-29
BFD-2720: Fix bug with samhsa matcher class cast in RDA logic

lsmitchell:
**JIRA Ticket:**
[BFD-2720](https://jira.cms.gov/browse/BFD-2720)

After a recent adjustment to the RDA code to use dependency injection, one of the side effects was a mis-cast on some logic that changed as a result of the transformer injection. This fixes that.

---

### What Does This PR Do?

Root cause of this issue was injecting a single transformer based on the resource type into the resource provider. The samhsa matcher REQUIRES the resource to be transformed into a claim, so when resources were passed into the samhsa matcher of a claimResponse resource provider, it used the claimResponse transformer, attempted to cast it to a claim after, and blew up.

This fixes that by moving the samhsa determination logic into the samhsa matcher, and injects the appropriate transformers into THAT class instead of needing the resourceProvider to always have a claim transformer handy.

- Moved hasNoSamhsaData from the Abstract resource provider into the samhsa matcher class
- Inject the claim transformers needed for hasNoSamhsaData into R4ClaimSamhsaMatcher (was already an injected component, but previously had no dependencies)
- Add E2E test to make sure with excludeSamhsa=true the claimResponse endpoint does not blow up
- Add unit tests to ensure hasNoSamhsaData works with claim and claimResponse resources in its new home
- Fix the couple spots where we had instantiated the samhsa matcher to accomadate its new constructor

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/pac/R4ClaimResponseResourceProviderE2E.java
+++ b/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/pac/R4ClaimResponseResourceProviderE2E.java
@@ -1,5 +1,7 @@
 package gov.cms.bfd.server.war.r4.providers.pac;
 
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.rest.client.api.IGenericClient;
 import ca.uhn.fhir.rest.param.DateParam;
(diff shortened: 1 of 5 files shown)


mackec1 (approved):
LGTM; will follow-up on your comment 'original sledgehammer approach of throwing every transformer into the provider' since that is what I am doing in EOB provider(s).

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/pac/R4ClaimResponseResourceProviderE2E.java
```diff
@@ -115,6 +117,33 @@ void shouldGetCorrectClaimResponseResourcesByMbiHash() {
...
+                    Collections.singletonList(new ReferenceParam(RDATestUtils.MBI_OLD_HASH)),
+                    "service-date",
+                    Arrays.asList(new DateParam("gt1970-07-18"), new DateParam("lt1970-07-25")),
+                    "excludeSAMHSA",
+                    Collections.singletonList(new ReferenceParam("true"))))
+            .returnBundle(Bundle.class)
+            .execute();
+
+    // Ensure we got a result and not an exception
+    assertNotNull(claimResult);
```
lsmitchell, line 144: This is a bit rudimentary, and I was attempting to write something better, but the way the test is written uses a json parsing comparitor and apparently some of the expected response does get filtered, so it was going to be a hassle to figure out what the response _should_ be here. For the purposes of this bug I just wanted to make sure we didn't blow up with an E2E test that captures the issue here (and unit tests better test the finer details below)

brianburton (approved):

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/pac/R4ClaimResponseResourceProviderE2E.java
brianburton: I agree it would be good to expand the test comparison later.  But not today.

PR #1835, opened 2023-07-20
BFD-2598: Remove usage of ReflectionUtils

billie-rose:
**JIRA Ticket:**
[BFD-2598](https://jira.cms.gov/browse/BFD-2598)

As a BFD user, I want to know that BFD's continued use of reflection doesn't incur a performance penalty so that I can expect my queries to run as quickly as possible.

---

### What Does This PR Do?

Refactor instances of reflection and hard-coded claims diagnoses/procedures retrieval.
This PR removes `ReflectionUtils.java` and replaces its functionality. It does this by grouping claims diagnosis and procedure fields (via `groupName` inside of the `.yaml` definitions). The auto-generated entity will take any fields with a shared `groupName` and provide an accessor method which returns those properties as a map. All possible fields of diagnoses and procedures can safely be retrieved regardless of whether the claim type has values for it, allowing all claim types to share the same retrieval logic without using reflection.

*Note*: There are a number of changes to `V1` that are more-or-less out of scope for this given story due to the removal of hard-coded accessor methods in favor of the newly auto-generated grouped fields accessor. Still, it felt like these changes were relevant enough to be included and would not hinder the PR's readability or scale.

### Load Testing Results
Each load test was exclusively run against the EOB v2 endpoints:
```
/v2/fhir/ExplanationOfBenefit search by id
/v2/fhir/ExplanationOfBenefit search by id / count=10
/v2/fhir/ExplanationOfBenefit search by id / lastUpdated / includeTaxNumbers
```

The Average, Median, Max, and Min stats refer to the 3 results procured for each endpoint. E.g, if the endpoints' time saved were -5, -10, -15, then the average response time would be -10.

**Negative numbers** represent the time saved (in ms) when **compared** to a load test against master still using Reflection.

**Positive numbers** represent the additional time spent (in ms) or increase in stats when **compared** to a load test against master still using Reflection.

*Run 1* 
| |Median Response Time (ms)|Average Response Time (ms)|Min Response Time (ms)|Max Response Time (ms)|Requests/s|
|-----------|--------------------|---------------------|-----------------|-----------------|----------|
|AVG|-30|-49|1|-3922|-4|-30|-37|-43|-53|-87 |-103|-200|-333|-1767 |-3900 |-3800|
|MEDIAN     |-20                 |-42                  |1                |-4382            |-4        |
|MAX        |-30                 |-56                  |1                |-5017            |-4        |
|MIN        |-40                 |-48                  |0                |-2366            |-4        |

| |50%|66%|75%|80%|90% |95% |98% |99% |99.90%|99.99%|100% |
|---|---|---|---|---|----|----|----|----|------|------|-----|
|AVG|-30|-37|-43|-53|-87 |-103|-200|-333|-1767 |-3900 |-3800|
|MEDIAN     |-20|-30|-30|-40|-70 |-100|-200|-400|-1600 |-4200 |-4300|
|MAX        |-30|-40|-60|-70|-120|-100|-200|-400|-2300 |-4400 |-4700|
|MIN        |-40|-40|-40|-50|-70 |-110|-200|-200|-1400 |-3100 |-2400|

**Negative numbers** represent the time saved (in ms) when **compared** to a load test against master still using Reflection.

**Positive numbers** represent the additional time spent (in ms) or increase in stats when **compared** to a load test against master still using Reflection.
 
*Run 2*
| |Median Response Time (ms)|Average Response Time (ms)|Min Response Time (ms)|Max Response Time (ms)|Requests/s|
|-----------|--------------------|---------------------|-----------------|-----------------|----------|
|AVG        |-17                 |-26                  |0                |-4409            | (+) 5         |
|MEDIAN     |-10                 |-26                  |0                |-4634            | (+) 5         |
|MAX        |-20                 |-29                  |0                |-5396            | (+) 5         |
|MIN        |-20                 |-22                  |0                |-3198            | (+) 4         |

| |50%|66%|75%|80%|90% |95% |98% |99% |99.90%|99.99%|100% |
|-----------|---|---|---|---|----|----|----|----|------|------|-----|
|AVG        |-17|-17|-17|-13|0   |-33 |-67 |-200|-3300 |-4167 |-4400|
|MEDIAN     |-10|-10|-20|-10|0   |-100|0   |-200|-3600 |-4900 |-4700|
|MAX        |-20|-20|-20|-20|0   |0   |-100|-200|-3900 |-4800 |-5300|
|MIN        |-20|-20|-10|-10|0   |0   |-100|-200|-2400 |-2800 |-3200|

### Expanded Tests
In addition to `PoetUtilTest.java` and `GenerateEntitiesFromDslMojoTest.java`, `RifFilesProcessorTest.java` was updated to cover the following test cases:

1. The grouped properties maps contain the correct number of properties; and
1. The grouped properties maps have the expected values; and
1. Expect a `NoSuchMethodException` when calling the last available grouped property + 1
    * This should flag any properties which have been added but not yet classified via `groupName` in the `yaml` file if the size check above passes despite its inclusion
    * IF the size check above it **does** fail, then it's an indicator that this test should be updated and re-verified

* Please pay careful attention to any possible copy-pasta errors between `V1` and `V2`

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist

    4. There isn't too much of a burden on reviewers.
    5. Any problems it causes have a small "blast radius".
    6. It'll be easier to rollback if that becomes necessary.


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DiagnosisUtilV2.java
```diff
@@ -178,86 +177,146 @@ private static void addCodingToCodeableConcept(
...
-   * specific claim type doesn't have the given diagnosis it will just be skipped.
+   * Extracts nearly all diagnosis types from a Claim. If the specific claim type doesn't have the
+   * given diagnosis it will just be skipped.
    *
-   * @param claim the Claim to extract the {@link Diagnosis}es from
+   * @param codes The mapping of diagnosis codes by their property name and respective value
+   * @param codeVersions The mapping of diagnosis code versions by their property name and
+   *     respective value
+   * @param presentOnAdms The mapping of diagnosis "PresentOnAdmissionCode" codes by their property
+   *     name and respective value
```
lsmitchell, line 221: I noticed this is optional and only present on some claim types; is there any more information we could describe here about when this is expected or not?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/SNFClaimTransformerV2.java
```diff
@@ -314,21 +312,18 @@ private ExplanationOfBenefit transformClaim(SNFClaim claimGroup) {
...
-    IntStream.range(FIRST_PROCEDURE, LAST_PROCEDURE + 1)
-        .mapToObj(i -> TransformerUtilsV2.extractCCWProcedure(i, claimGroup))
-        .filter(p -> p.isPresent())
-        .forEach(p -> TransformerUtilsV2.addProcedureCode(eob, p.get()));
+    TransformerUtilsV2.extractCCWProcedures(
+            claimGroup.getProcedureCodes(),
+            claimGroup.getProcedureCodeVersions(),
+            claimGroup.getProcedureDates())
+        .stream()
+        .forEach(p -> TransformerUtilsV2.addProcedureCode(eob, p));
```
lsmitchell, line 326: I like the new code here, cleaner than the way it was done before. (I see it's still done with the counter in the underlying code, so its not totally gone, but at least its abstracted)

lsmitchell (commented):
I like the solution and liking removing reflection as well. I'd certainly err on the side of ensuring that the tests we have actually test the mappings that were changed here, but if testing looks good then this looks like a 👍

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DiagnosisUtilV2.java
billie-rose: From what I gathered it's only present on `InpatientClaims` and I'd guess that's because that's the only claim where it may be relevant to know re:  this explanation in the codebase:

> Diagnosis Present On Admission (POA) Indicator: This field identifies the patient's condition that is present at the time the order for inpatient admission to a general acute care hospital occurs. This indicator is assigned to every principal and secondary diagnosis on an inpatient acute care hospital claim, and the external cause of injury codes. Conditions that develop during an outpatient encounter, including emergency department, observation, or outpatient surgery, are considered as present on admission. The Centers for Medicare and Medicaid Services (CMS) does not require a POA indicator for the external cause of injury code unless it is being reported as another diagnosis. Effective with discharges on or after 10/01/07.

Does that seem right? If so, I can add a quick comment to the method

billie-rose:
> I like the solution and liking removing reflection as well. I'd certainly err on the side of ensuring that the tests we have actually test the mappings that were changed here, but if testing looks good then this looks like a 👍

It does pass all the present tests (including IT tests), but do you think those should be updated to cover more test cases?

apps/bfd-model/bfd-model-dsl-codegen/bfd-model-dsl-codegen-plugin/src/main/java/gov/cms/model/dsl/codegen/plugin/GenerateEntitiesFromDslMojo.java
```diff
@@ -1017,12 +1023,41 @@ TypeSpec createAdditionalFieldsInnerClass(
...
+    return methodSpecs;
+  }
+
   /**
    * Data class holding all of the information required to create accessor methods (setter/getter)
    * for a field.
    */
   @Data
-  @AllArgsConstructor
+  @RequiredArgsConstructor
```
brianburton, line 1060: This class is meant to be immutable.  All fields need to be final.  I think it would be better to add a `@Builder` annotation and change constructor calls to use the builder going forward.  The builder will default any uninitialized object references to null.

apps/bfd-model/bfd-model-dsl-codegen/bfd-model-dsl-codegen-plugin/src/main/java/gov/cms/model/dsl/codegen/plugin/GenerateEntitiesFromDslMojo.java
```diff
@@ -1035,6 +1070,8 @@ static class AccessorSpec {
     private final boolean isNullableColumn;
     /** True if the column is read only. (No setter should be generated.) */
     private final boolean isReadOnly;
+    /** The name of the group of fields to which the field belongs. */
+    private String groupName;
```
brianburton, line 1074: See above.  This one needs to be final like all of the others.

apps/bfd-model/bfd-model-dsl-codegen/bfd-model-dsl-codegen-plugin/src/main/java/gov/cms/model/dsl/codegen/plugin/PoetUtil.java
```diff
@@ -193,6 +201,36 @@ public static MethodSpec createJoinPropertyGetter(
...
+    MethodSpec.Builder methodSpecBuilder =
+        MethodSpec.methodBuilder(fieldToMethodName("get", groupedPropertiesName))
+            .addModifiers(Modifier.PUBLIC)
+            .returns(returnType)
+            .addStatement("$T $L = new $T<>()", returnType, groupedPropertiesName, HashMap.class);
+    propertyNames.stream()
+        .forEach(
+            property ->
+                methodSpecBuilder.addStatement(
+                    "$L.put(\"$L\", $L())",
```
brianburton, line 226: I think it's better to use `$S` for a string literal rather than quoting a `$L`.

https://github.com/square/javapoet#s-for-strings

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DiagnosisUtilV2.java
```diff
@@ -178,86 +177,146 @@ private static void addCodingToCodeableConcept(
...
    * @param label One or more labels to use when mapping the diagnosis.
    * @return a {@link Diagnosis} or {@link Optional#empty()}
    */
   public static Optional<Diagnosis> extractDiagnosis(
-      String substitution, Object claim, Optional<CcwCodebookInterface> ccw, DiagnosisLabel label) {
-
+      String substitution,
+      Map<String, Optional<String>> codes,
+      Map<String, Optional<Character>> codeVersions,
+      Optional<Map<String, Optional<Character>>> presentOnAdms,
```
brianburton, line 194: We might want to consider just passing in a `Map<String, Optional<Character>>` rather than an `Optional<Map<String, Optional<Character>>>` fir `presentOnAdms`.  The callers can just pass `Map.of()` in the cases where they don't have a real map. The  case that does have a map (InpatientClaimTransformerV2) could just pass it in directly.

That would make the code using `presentOnAdms` a lot simpler.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DiagnosisUtilV2.java
```diff
@@ -178,86 +177,146 @@ private static void addCodingToCodeableConcept(
...
-        ReflectionUtils.tryMethod(
-            claim, String.format("getDiagnosis%sPresentOnAdmissionCode", substitution));
-
+        presentOnAdms.isPresent()
+            ? presentOnAdms
+                .get()
+                .getOrDefault(
+                    String.format("diagnosis%sPresentOnAdmissionCode", substitution),
+                    Optional.empty())
+            : Optional.empty();
```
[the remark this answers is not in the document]
brianburton: See note above about passing in the `Map` directly.  Doing so would make this code trivial.

```
Optional<Character> presentOnAdm =
        presentOnAdms.getOrDefault(
                    String.format("diagnosis%sPresentOnAdmissionCode", substitution),
                    Optional.empty());
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerUtils.java
```diff
@@ -2487,85 +2484,157 @@ static void mapEobCommonGroupInpHHAHospiceSNFCoinsurance(
...
+  /**
+   * Extracts all possible diagnosis types from a Claim.
+   *
+   * @param codes The mapping of diagnosis codes by their property name and respective value
+   * @param codeVersions The mapping of diagnosis code versions by their property name and
+   *     respective value
+   * @param presentOnAdms The mapping of diagnosis "PresentOnAdmissionCode" codes by their property
+   *     name and respective value
    * @return the {@link Diagnosis} that can be extracted from the specified {@link InpatientClaim}
    */
```
brianburton, line 2533: Pre-existing issue but this method is called for other claim types too.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/DMEClaimTransformer.java
```diff
@@ -103,35 +102,10 @@ private static ExplanationOfBenefit transformClaim(
...
-            claimGroup.getDiagnosis10CodeVersion(),
-            claimGroup.getDiagnosis11Code(),
-            claimGroup.getDiagnosis11CodeVersion(),
-            claimGroup.getDiagnosis12Code(),
-            claimGroup.getDiagnosis12CodeVersion()))
-      TransformerUtils.addDiagnosisCode(eob, diagnosis);
+    TransformerUtils.extractDiagnoses(
+            claimGroup.getDiagnosisCodes(), claimGroup.getDiagnosisCodeVersions(), Optional.empty())
+        .stream()
+        .forEach(d -> TransformerUtils.addDiagnosisCode(eob, d));
```
brianburton, line 108: Nice!

lsmitchell:
> 

I wouldnt implicitly trust the existing tests cover any code changes; I'd check what we have, if they properly cover whats being changed, and add tests if needed.

lsmitchell (approved):
Thanks for adding additional testing!

PR #1840, opened 2023-07-24
BFD-2768: simplify s3 testing

brianburton:
**JIRA Ticket:**
[BFD-2768](https://jira.cms.gov/browse/BFD-2768)

As a BFD contributor, I want tests for the bfd-pipeline to rely on locally available test doubles so that I can build and test the application without having to configure credentials or create real, AWS S3 resources.

This story involves:

* Replacing minio with localstack S3
* Ensuring all S3 calls in tests use localstack rather than AWS
* Adding code to ensure any attempt at test bucket creation fails if real AWS S3 client is being used

---

### What Does This PR Do?

Overall this PR is intended to improve testability and remove the requirement for GitHub Actions to have AWS access.  With that in mind it does the following:

* Introduces the `AwsClientConfig` and `S3ClientConfig` classes to provide a common abstraction for configuring SSM, SQS, and S3 clients.
* Introduces `S3ClientFactory` interface to allow multiple backend implementations of S3 clients to be used (including mocks) in places where clients need to be created.
* Eliminates `S3MinioConfig` and `MinioTestContainer` from `bfd-pipeline-shared-utils` and removes dependency on testcontainers from that project.
* Introduces `AbstractLocalStackTest` to provide a common way to create tests reliant on a single localstack container.
* Updates various configuration and test classes to use the new features.
* Adds a condition to the AWS credential checks in `LayedConfiguration` and `AppConfiguration` so that they are skipped when using localstack rather than real AWS.
* Removes code to set up AWS permissions on S3 buckets and adds a check to ensure that S3 buckets are only created when using LocalStack.
* Adds additional environment variables to the pipeline to allow the RDA in-process server to be configured to use localstack instead of AWS.
* Removes the `ENV_VAR_KEY_RDA_GRPC_INPROC_SERVER_S3_REGION` setting from the pipeline.  It added complexity and would likely never differ from the default used by the RIF pipeline.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -616,9 +636,12 @@ static CcwRifLoadOptions loadCcwRifLoadOptions(ConfigLoader config, LoadAppOptio
     } else {
       allowedRifFileType = Optional.empty();
     }
-
-    LayeredConfiguration.ensureAwsCredentialsConfiguredCorrectly();
-    ExtractionOptions extractionOptions = new ExtractionOptions(s3BucketName, allowedRifFileType);
+    final S3ClientConfig s3ClientConfig = loadS3ServiceConfig(config);
+    if (s3ClientConfig.isCredentialCheckUseful()) {
+      LayeredConfiguration.ensureAwsCredentialsConfiguredCorrectly();
```
lsmitchell, line 641: Nice; I think this will remove the last remnant of needing our AWS credentials to run the tests.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/s3/AwsS3ClientFactory.java
```diff
@@ -0,0 +1,42 @@
...
+   * Passed to {@link S3CrtAsyncClientBuilder#minimumPartSizeInBytes} to cause large files to be
+   * downloaded in parts of 200 mb.
+   */
+  public static final long MINIMUM_PART_SIZE_FOR_DOWNLOAD = 200L * 1024L * 1024L;
+
+  /** Used to configure the S3 client builders with basic connection settings. */
+  private final S3ClientConfig s3ClientConfig;
+
+  @Override
+  public S3Client createS3Client() {
```
lsmitchell, line 28: External to this work, but we can check the usages of this and see if it'd be beneficial to just use the async client in places at some point in the future; we might get some gains out of the multi part download in other places we download files.

apps/bfd-shared-test-utils/src/main/java/gov/cms/bfd/AbstractLocalStackTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Base class for tests that need to use a shared localstack container. Refer to <a
+ * href="https://java.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers">localstack
+ * docs</a> for explanation of the singleton container pattern.
+ */
+public abstract class AbstractLocalStackTest {
+  /** The localstack image and version to use for integration tests. */
+  public static final DockerImageName LocalStackImageName =
+      DockerImageName.parse("localstack/localstack:2.0.2");
```
lsmitchell, line 14: This might make the localstack image version harder to update; could/should this be read from the pom so it can be updated alongside other versioned items?

apps/bfd-shared-utils/src/main/java/gov/cms/bfd/sharedutils/config/AwsClientConfig.java
```diff
@@ -0,0 +1,107 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.awscore.client.builder.AwsClientBuilder;
+import software.amazon.awssdk.regions.Region;
+
+/**
+ * Value object containing some common settings required to initialize AWS SDK clients. All settings
+ * are optional. Normally construction is via a builder object. The builder class and object have *
```
lsmitchell, line 16:
```suggestion
 * are optional. Normally construction is via a builder object. The builder class and object have
```

apps/bfd-pipeline/bfd-pipeline-shared-test-utils/pom.xml
```diff
@@ -31,6 +31,18 @@
             <artifactId>bfd-pipeline-shared-utils</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>gov.cms.bfd</groupId>
+            <artifactId>bfd-shared-utils</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
lsmitchell, line 38: When my open PR is merged, this should be able to be moved to test scope.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/RdaServerJobIT.java
```diff
@@ -58,8 +63,20 @@ public class RdaServerJobIT extends MinioTestContainer {
...
+  /** A client connected to the localstack container for use in test methods. */
+  private S3Client s3Client;
+
+  /** Populates S3 related fields based on localstack container. */
+  @BeforeEach
+  void initializeS3RelatedFields() {
+    s3ClientConfig = LocalStackS3ClientFactory.createS3ClientConfig(localstack);
+    s3ClientFactory = new AwsS3ClientFactory(s3ClientConfig);
+    s3Client = s3ClientFactory.createS3Client();
+  }
```
lsmitchell, line 79: I think we could move these test variables and setup into the abstract class if all the localStack tests will use the same setup; I did this in the ServerRequiredTest base class for instance.

lsmitchell (commented):
Looks good. Couple suggestions, nothing specifically blocking.

apps/bfd-shared-test-utils/src/main/java/gov/cms/bfd/AbstractLocalStackTest.java
brianburton: Thats a good question.  I did some looking around in the past and did not find a way to get properties like this from the pom at compile time.  We do the same thing in `bfd-shared-test-utils/src/main/java/gov/cms/bfd/DatabaseTestUtils.java`.

An alternative might be to move all of the image names into a single class as constants so they can all be checked and updated in one place.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/RdaServerJobIT.java
brianburton: Not unless we make that class S3 specific.

brianburton: I'll take a look tomorrow.  Maybe I'll introduce an S3 specific subclass.  I'm not a fan of this base class pattern anyway and might revisit it in BFD-2181.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/S3ClientConfig.java
```diff
@@ -0,0 +1,79 @@
...
+public class S3ClientConfig extends AwsClientConfig {
+  /**
+   * Initializes an instance. Any variable can be null. Region defaults to {@link #REGION_DEFAULT}.
+   *
+   * @param region an AWS {@link Region}
+   * @param endpointOverride alternative URI for accessing AWS services (used with localstack)
+   * @param accessKey optional access key
+   * @param secretKey optional secret key
+   */
+  @Builder(builderClassName = "S3Builder", builderMethodName = "s3Builder")
```
mackec1, line 33: Nice construct via Lombok; good to know how this works!

mackec1 (dismissed):
Lots to like about these changes...flexibility, lombok, elegant code! Always learn something in these PRs.

apps/bfd-shared-test-utils/src/main/java/gov/cms/bfd/AbstractLocalStackTest.java
lsmitchell: Neither thing I found was particularly elegant, but there's the http://www.mojohaus.org/properties-maven-plugin/ that lets you write a maven plugin to be read in as a file which this could read from during setup, or there's setting a system var from maven using the surefire plugin: https://stackoverflow.com/questions/3231797/specify-system-property-to-maven-project

If we had the version saved as a maven property like the other versions, one of these two solutions could supply said version for consumption in the application. (There might be a better way that I didn't find, as well)

brianburton: While I applaud the goal I'm not a fan of using a system property for the image name/tag and storing the property value in the pom.  It works fine when running the tests from maven (admittedly our normal case) but if anyone wants to run the tests directly in their IDE without using maven they run into some issues:

When running the test directly in IDEA without maven you need to tell IDEA to set the property directly for the test, otherwise one of two things happen:
* If IDEA does not set the property and the java code has no default value for the property, the test will fail.
* If IDEA does not set the property and the java code has a default value that differs from the value in the pom the test will likely fail if there is some reason why we upgraded the image tag in the first place.

If the code does have a default value then we wind up having to maintain the image tag in two places (pom and java code) rather than one (java code).  Or even three if we include a developer's IDEA configuration.

So I'd prefer to just centralize the version strings in the java code and maybe make a note of it in the pom.

apps/bfd-pipeline/bfd-pipeline-shared-test-utils/pom.xml
brianburton: Why is that?  The `PipelineTestUtils` class requires `DatabaseSchemaManager` to compile.

SInce `bfd-pipeline-shared-test-utils` is a test library it is fine for it to use production code in compile scope.  It's only the other way around that's problematic.

It turns out this dependency declaration is redundant in this pom so I'll remove it.

lsmitchell: Oops you're right I read it as the test version but this is the normal one. Ignore!

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/test/java/gov/cms/bfd/pipeline/rda/grpc/RdaServerJobIT.java
brianburton: I have Added AbstractLocalStackS3Test and removes LocalStackS3ClientFactory.

lsmitchell (approved):
👍

JeremyIR:
Looks fine. DefaultS3TransferManager is deprecated in further version of AWS SDK, hence why it wasnt upgraded in this PI. Maybe in a future ticket we can explore swapping that out and ensuring via s3 tests that it wont have negative consequences.

JeremyIR (approved):

PR #1854, opened 2023-08-01
BFD-2729: Improve Regression Suite with PAC Data

malessi:
**JIRA Ticket:**
[BFD-2729](https://jira.cms.gov/browse/BFD-2729)

As a BFD release engineer, I want the regression suite to include PAC endpoints so that I can trust a successful regression suite execution yields a healthy deployment.

---

### What Does This PR Do?

This PR updates the V2 Regression Suite to include test cases that exercise the `Claim` and `ClaimResponse` endpoints using static, synthetic data. Consequently, the Regression Suite Lambda has been updated such that it will only run the V2 Regression Suite. Additionally, all test suites except `high_volume_suite` (Load Test Suite) have been updated to pull data without exhausting the source list (by randomly sampling or by cycling through all elements), and several type errors have been fixed.

This PR has been verified by:

- Running `v2/regression_suite.py` locally against `test`, verifying that there were no failures
- Running `v2/regression_suite.py` locally against `test` with local statistics collection and comparison enabled, verifying that statistics for the new test cases are _stable_ and _consistent_ and that the comparison passes everytime
- Running `high_volume_suite.py` locally against `test`, verifying that it works as expected and there are no failures
- Running `high_volume_suite.py` locally against `test` with both included and excluded tags, verifying that only the tasks that should be ran were run

A new Regression Suite Docker image will be built once this PR is accepted and merged into `master`.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist

### DIFF
--- a/.github/workflows/ci-java.yml
+++ b/.github/workflows/ci-java.yml
@@ -1,10 +1,7 @@
 name: 'CI - Java'
-on: pull_request
-
-permissions:
-  id-token: write   # This is required for requesting the AWS IAM OIDC JWT
-  contents: read    # This is required for actions/checkout
-  packages: write
+on:
+  pull_request:
+  merge_group:
 
 env:
   # workflow file matchers - workflow jobs will only run if matching files are found
@@ -78,12 +75,6 @@ jobs:
     needs: workflow
     if: needs.workflow.outputs.files
     steps:
-      - name: 'Configure AWS credentials'
-        uses: aws-actions/configure-aws-credentials@v2
-        with:
-          role-to-assume: ${{ secrets.GHA_AWS_IAM_ROLE_ARN }}
-          role-session-name: ci-java
-          aws-region: us-east-1
       - name: 'Checkout repo'
         uses: actions/checkout@v3
       - name: 'Setup JDK'
@@ -109,15 +100,15 @@ jobs:
           EOF
 
       - name: 'Run Maven Build on FDA Drug Utility'
-        run: mvn --batch-mode -Dmaven.build.cache.enabled=false install
+        run: mvn --threads 1C --batch-mode -Dmaven.build.cache.enabled=false install
         working-directory: ./apps/bfd-data-fda
 
       - name: 'Run Maven Build on NPI Org Utility'
-        run: mvn --batch-mode -Dmaven.build.cache.enabled=false install
+        run: mvn --threads 1C --batch-mode -Dmaven.build.cache.enabled=false install
         working-directory: ./apps/bfd-data-npi
 
       - name: 'Run Maven Build'
-        run: mvn --quiet --batch-mode -Dmaven.build.cache.enabled=false verify
+        run: mvn --threads 1C --quiet --batch-mode -Dmaven.build.cache.enabled=false verify
         working-directory: ./apps
 
       - name: Login to GitHub Container Registry
--- a/.github/workflows/ci-runbook.yml
+++ b/.github/workflows/ci-runbook.yml
@@ -3,6 +3,7 @@ on:
   pull_request:
     paths:
       - docs/runbooks/*
+  merge_group:
 
 jobs:
   build-runbook-readme:
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -4,6 +4,7 @@ on:
     paths:
       - .github/workflows/ci-rust.yml
       - apps/utils/db-query-checker/*
+  merge_group:
 
 jobs:
 
--- a/.github/workflows/ci-terraform.yml
+++ b/.github/workflows/ci-terraform.yml
@@ -1,5 +1,7 @@
 name: 'CI - Terraform'
-on: pull_request
+on:
+  pull_request:
+  merge_group:
 
 env:
   # workflow file matchers - workflow jobs will only run if matching files are found
--- a/apps/pom.xml
+++ b/apps/pom.xml
@@ -110,7 +110,7 @@
         <snakeyaml.version>2.0</snakeyaml.version>
 
         <jakarta.xml.bind.api.version>2.3.2</jakarta.xml.bind.api.version>
-        <org.glassfish.jaxb.runtime.version>2.3.1</org.glassfish.jaxb.runtime.version>
+        <org.glassfish.jaxb.runtime.version>2.3.8</org.glassfish.jaxb.runtime.version>
         <jakarta.annotation.api.version>1.3.5</jakarta.annotation.api.version>
 
         <logback.jackson.version>0.1.5</logback.jackson.version>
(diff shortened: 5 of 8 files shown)


apps/utils/locust_tests/common/data.py
```diff
@@ -3,16 +3,34 @@
...
+    def __call__(self, uri: str) -> CopyableEnumerable[Any]:
+        ...
+
+
+@runtime_checkable
+class SamplingLoadFunction(Protocol):
+    def __call__(
+        self, uri: str, table_sample_pct: Optional[float] = None
+    ) -> CopyableEnumerable[Any]:
+        ...
```
malessi, line 22: These [`Protocol`]() classes are ["Callback Protocols"](https://mypy.readthedocs.io/en/latest/protocols.html#callback-protocols) and serve as _much_ stronger type guarantees for the type of the function that's provided as the data loading function. These `Protocol`s are used to ensure only the following functions pass static type analysis when passed into `load_from_parsed_opts()`/`load_from_uri()`:

```
def load_function(uri: str) -> CopyableEnumerable[Any]: ...
def load_function(uri: str, table_sample_pct: Optional[float] = None) -> CopyableEnumerable[Any] ...
```

Anything else will be rejected by the type checker.

Additionally, they are both `@runtime_checkable`, as it is necessary to check _which_ type of `LoadFunction` the given `Callable` is during runtime. Without this attribute, the below `LoadFunction` `Union` doesn't work.

apps/utils/locust_tests/common/data.py
```diff
@@ -3,16 +3,34 @@
...
+
+
+LoadFunction = Union[NonSamplingLoadFunction, SamplingLoadFunction]
```
malessi, line 25: These [`Protocol`](https://mypy.readthedocs.io/en/latest/protocols.html) classes are ["Callback Protocols"](https://mypy.readthedocs.io/en/latest/protocols.html#callback-protocols) and serve as _much_ stronger type guarantees for the type of the function that's provided as the data loading function. These `Protocol`s are used to ensure only the following functions pass static type analysis when passed into `load_from_parsed_opts()`/`load_from_uri()`:

```
def load_function(uri: str) -> CopyableEnumerable[Any]: ...
def load_function(uri: str, table_sample_pct: Optional[float] = None) -> CopyableEnumerable[Any] ...
```

Anything else will be rejected by the type checker. Fortunately, _all_ of the corresponding "load functions" in `db.py` match this method signature, and so count as `LoadFunction`s.

Additionally, they are both `@runtime_checkable`, as it is necessary to check _which_ type of `LoadFunction` the given `Callable` is during runtime. Without this attribute, the below `LoadFunction` `Union` doesn't work.

malessi:
Pulling this out of draft. Thanks @brianburton for discovering the slow MBIs query in `prod` and _really_ quickly providing a modified query that uses an index!

apps/utils/locust_tests/common/bfd_user_base.py
```diff
@@ -164,7 +164,7 @@ def ssl_context_factory(self) -> ssl.SSLContext:
             context.load_verify_locations(cafile=self.server_public_key)
         return context
 
-    def get_by_url(self, url: str, headers: Optional[Dict[str, str]] = None, name: str = ""):
+    def get_by_url(self, url: str, headers: Optional[Mapping[str, str]] = None, name: str = ""):
```
malessi, line 167: [`Mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) indicates that the argument is an immutable/read-only key-to-value collection.

It's used here to replace `dict` type hints for arguments that are never mutated. This replacement also fixed a few type errors.

apps/utils/locust_tests/v1/regression_suite.py
```diff
@@ -77,17 +78,17 @@ class RegressionV1User(BFDUserBase):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.bene_ids = master_bene_ids.copy()
-        self.contract_data = master_contract_data.copy()
-        self.hashed_mbis = master_hashed_mbis.copy()
+        self.bene_ids = itertools.cycle(master_bene_ids.copy())
```
malessi, line 81: [`itertools.cycle`](https://docs.python.org/3/library/itertools.html#itertools.cycle) returns an iterator on an iterable that, once reaching the end of the iterable, loops back around to the first element in the given iterable. Thus, data is never exhausted when running tests for an arbitrary amount of time.

This is _especially_ important for PAC data, as the number of synthetic MBIs is _very_ small in each environment. We would exhaust the list of MBIs within _seconds_ without this new behavior.

apps/utils/locust_tests/v1/regression_suite.py
```diff
@@ -77,17 +78,17 @@ class RegressionV1User(BFDUserBase):
...
+        self.hashed_mbis = itertools.cycle(master_hashed_mbis.copy())
 
     @tag("coverage", "coverage_test_id_count")
     @task
     def coverage_test_id_count(self):
         """Coverage search by ID, Paginated"""
         self.run_task_by_parameters(
             base_path="/v1/fhir/Coverage",
-            params={"beneficiary": self.bene_ids.pop(), "_count": "10"},
+            params={"beneficiary": next(self.bene_ids), "_count": "10"},
```
malessi, line 91: [`next(...)`](https://docs.python.org/3/library/functions.html?highlight=next#next) simply returns the next element from the iterator created by `itertools.cycle(...)` above.

apps/utils/locust_tests/high_volume_suite.py
```diff
@@ -60,24 +71,45 @@ def _(environment: Environment, **kwargs):
...
     @tag("eob_test_id_count_type_pde_v1", "v1")
     @task
     def eob_test_id_count_type_pde_v1(self):
         """Explanation of Benefit search by ID, type PDE, paginated"""
-        self.run_task_by_parameters(
+        self.user.run_task_by_parameters(
             base_path="/v1/fhir/ExplanationOfBenefit",
             params={
-                "patient":  self.bene_ids.pop(),
+                "patient": random.choice(self.user.bene_ids),
```
malessi, line 112: [`random.choice(...)`](https://docs.python.org/3/library/functions.html?highlight=next#next) is used to pick a random element from the enumerable. This ensures the source data collection is not mutated, nor exhausted, and that data is picked at random; therefore, this is equivalent to using `random.shuffle()` without needing to be concerned with data exhaustion.

malessi (commented):
Some additional context that may be helpful for reviewers:

apps/utils/locust_tests/high_volume_suite.py
lsmitchell: If i'm understanding this right, I think we shouldn't be applying the new random choice pattern to the high volume suite. For the regression suite we don't care about reusing data, but the high volume suite does, we want to avoid calling anything twice to avoid cached hits influencing the performance result numbers.

lsmitchell (commented):
Thanks for adding the context review comments! Helped a lot in understanding things quicker.

apps/utils/locust_tests/high_volume_suite.py
malessi: I'm not sure how big of an impact that would have on the High Volume Suite's characteristics, but you are right; using `random.choice()` invites the possibility that a single `HighVolumeUser` _could_ re-use data that had already been used. And, the High Volume Suite doesn't use the limited synthetic datasets, so thinking about it now I don't think we should be concerned about data exhaustion at all.

I've reverted this change to the High Volume Suite. It now `shuffle`s each `list` of data, and data is `pop`ped from each `list` when it's used, effectively making it impossible for any `HighVolumeUser` (individually) to reuse data. This does mean that a `HighVolumeUser` could run out of data, but solving that is out-of-scope of this PR anyways.

lsmitchell (approved):

mjburling (approved):

PR #1885, opened 2023-08-16
BFD-2842: colima compatibility and images in pom

brianburton:
**JIRA Ticket:**
[BFD-2842](https://jira.cms.gov/browse/BFD-2842)

As a BFD engineer I want tests to be easily configured and to support test containers on docker platforms other than docker desktop.

---

Running tests under colima (instead of Docker Desktop) failed for a variety of reasons.  This PR makes code changes necessary to allow the tests to pass in that environment.

* Localstack 2.0.2 was not compatible so we now use 2.2.0
* Localstack FIFO queues are broken either in 2.2.0 or when running in colima.  They aren't really necessary for the migrator so SqsDao now uses normal queues.  This simplifies sending messages but complicates receiving them since order isn't guaranteed so tests now account for any possible ordering.
* Tests were failing due to a race condition caused by the postgresql container initializing more slowly under colima.  We now use `waitFor` to cause tests to wait for the container to begin listening to its port before proceeding.

In addition the image to use for localstack is now defined in the root `pom.xml` file.  The class that uses it has a default value in case the property is not defined when a test runs (e.g. in IDEA).  Comments were added to the pom for both image properties to identify where these defaults are defined.

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-db-migrator/src/test/java/gov/cms/bfd/migrator/app/SqsDaoIT.java
```diff
@@ -29,24 +31,27 @@ void setUp() {
...
+    String queueName = "my-test-queue";
+    String queueUri = dao.createQueue(queueName);
     String message1 = "this is a first message";
     String message2 = "this is a second message";
-    dao.sendMessage(queueUri, messageGroupId, "1", message1);
-    dao.sendMessage(queueUri, messageGroupId, "2", message2);
-    assertEquals(Optional.of(message1), dao.nextMessage(queueUri));
-    assertEquals(Optional.of(message2), dao.nextMessage(queueUri));
+    dao.sendMessage(queueUri, message1);
+    dao.sendMessage(queueUri, message2);
```
mackec1, line 48: Could we really try to muck with things by sending message 2 first here?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/ServerExecutor.java
mjburling, line 1: 👍 Nice little _campsite_ fix here, too. Thanks!

apps/bfd-db-migrator/src/main/java/gov/cms/bfd/migrator/app/SqsDao.java
mjburling, line 1: OK! These changes should yield a little less work for putting the native signaling to work, too.

mjburling (approved):
✨

mackec1 (approved):

PR #1901, opened 2023-08-24
BFD-2856 : BFD Server ASG Network Scaling Policies are causing erroneous capacity oscillation

billie-rose:
**JIRA Ticket:**
[BFD-2856](https://jira.cms.gov/browse/BFD-2856)

The new network-traffic-based Scaling Policies introduced in [BFD-2804](https://jira.cms.gov/browse/BFD-2804) are designed to be more aggressive during bursts of network traffic, opting to scale-out quickly and to defer scale-in until network traffic has subsided for five minutes. The intention when designing these scaling policies was for there to be discrete "stages" of scaling for a given range of network traffic:

3 instances when network traffic is less than 100MB/min
6 instances when network traffic is greater than 100MB/min but less than or equal to 200MB/min
9 instances when network traffic is greater than 200MB/min but less than or equal to 400MB/min
12 instances when network traffic is greater than 400MB/min
Given the above, the ideal would be that after a minute of network traffic holding at a given range above, the BFD Server ASG would scale to the corresponding number of instances. Then, the BFD Server ASG would not scale back in until the network traffic holds at a lower range for 5 minutes (i.e. at minute 1, network is 250MB/min so scale-out to 9, then minutes 2 - 6 network is ~120MB/min so scale-in to 6).

Unfortunately, this is not the case.

**Details**
The BFD Server ASG Scaling Policies are causing capacity oscillation during scaling events, as the scale-out and scale-in evaluation periods differ and both scale-out and scale-in are unaware of the current capacity of the ASG. It is possible that a large spike of network traffic causes scale-out to a large number of instances (12, 9, etc.), then this network traffic subsides slightly for several minutes thus prompting scale-in to occur and remove a small number of instances. However, because the network traffic did not decrease such that it fell beneath the lowest scale-out threshold, and the scale-out policy is designed such that it adds instances to the ASG when exceeding said threshold, the scale-out policy will immediately and erroneously add instances back to the ASG. This causes an "oscillation" effect during a scaling event where the scale-in correctly removes a small number of instances, and then the scale-out policy incorrectly requests new instances immediately after.

---

### What Does This PR Do?

Update the BFD Server ASG scale-out  and scale-in policy such that they set Desired Capacity (DC) to the proper value given a range of network traffic.

- Update the ASG scale-out policy so that it can set DC to a value greater than the current DC, but it cannot decrease the DC.
- Implement a CloudWatch Alarm that uses Metric Math to calculate a value representing a scalar multiplier of additional capacity.
- Configure the alarm such that missing data is treated as OK and the metric value alarmed on by this Alarm is sparse as values are only generated when capacity should be scaled-out (the current DC is less than the wanted DC), so the Alarm will only enter ALARM if the DC should be set to a value larger than it already is.
- Update the ASG scale-in policy to set (not remove from) the DC to a static value
- Set the CloudWatch Alarm associated with the ASG scale-in policy to evaluate on 10 consecutive periods of 60 seconds (10 minute evaluation period) rather than 5.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


malessi (approved):

ops/terraform/services/server/modules/bfd_server_asg/main.tf
```diff
@@ -246,33 +251,34 @@ resource "aws_autoscaling_policy" "filtered_networkin_low_scaling" {
...
 
 resource "aws_cloudwatch_metric_alarm" "filtered_networkin_high" {
   alarm_name          = "bfd-${var.role}-${local.env}-networkin-high"
-  comparison_operator = "GreaterThanThreshold"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
   datapoints_to_alarm = 1
   evaluation_periods  = 1
-  threshold           = 100 * 1000000 # 100 megabytes
+  threshold           = 1
   treat_missing_data  = "ignore"
```
malessi, line 276:
```suggestion
  treat_missing_data  = "notBreaching"
```

Missed this one! The Alarm's metric is sparse, and if there is no value for said metric during the Alarm's evaluation period that means that there should be no change in scaling capacity. So, the Alarm's state should be set to `OK`.

malessi (changes requested):

mjburling (approved):

malessi (approved):

PR #1923, opened 2023-09-07
BFD-2849: Adjust case where startIndex >= maxResults to throw 400 instead of 500

lsmitchell:
**JIRA Ticket:**
[BFD-2849](https://jira.cms.gov/browse/BFD-2849)

Calling the EOB (and probably other) endpoints with a startIndex greater than the number of entries in the response returns a 500; the response should be a 400, since this is a user input error and can be corrected by adjusting the request.

---

### What Does This PR Do?

For every endpoint (via adjusting the bundling logic when pagination is request in the TransformerUtils) we will throw an InvalidRequestException with a useful error message when the `startIndex >= totalResources.size()` which will cause a 400 at the response layer in HAPI-FHIR.

While many restAssured tests were added to cover this case for each endpoint (including PAC endpoints), 99% of the tests are just copy/pasted from the newly added EOB tests with minor alterations for the endpoint, number of results expected, and version (between v1/v2), so much of this is the same code. 

For each endpoint, only tests that were related to the bug were added. One exception is v1 EOB, where all the tests that existed for v2 worked so figured I might as well copy them all over.

For v1 restAssured E2Es, I just copied exactly what was in the V2 tests, and changed it to v1. (Only exception being the EOB tests where the paths for taxNum were different and a couple asserts had to have their path adjusted).

Major Adjustments:
- Adjusts EOB resource provider in both v1 and v2 to not swallow InvalidRequestExceptions when creating the bundle, since this would prevent throwing a 400 during the bundling
- Adds logic to throw an InvalidRequestException while creating bundles for all endpoints (including PAC endpoints) if the startIndex >= max number of resources returned
- Adds RestAssured test classes for every endpoint, both v1 and v2
   - Tests added for each endpoint were to cover this bugfix only, but add a nice framework for future test additions

Minor Adjustments:
- Cleanup in OffsetLinkBuilder to move the validations of the field sizes into a method instead of being spread between constructor, getters, and elsewhere. This lines up with how fields were validated in PatientLinkBuilder.
    - Fixed a couple tests as a result of the validation changes, since some validation checks were done in getters before
- Fixes some spots near code I was changing which were using string concatenation to make strings instead of String.Valueof
- Adds some common restAssured test methods to ServerTestUtils so they can be shared

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

Signoff from the various peering partners, noted in this confluence document:
https://confluence.cms.gov/pages/viewpage.action?pageId=896217997

### Submitter Checklist

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/OffsetLinkBuilder.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/OffsetLinkBuilder.java
@@ -48,14 +48,27 @@ public final class OffsetLinkBuilder implements LinkBuilder {
    */
   public OffsetLinkBuilder(RequestDetails requestDetails, String resource) {
     this.pageSize = parseIntegerParameters(requestDetails, Constants.PARAM_COUNT);
-    startIndex = parseIntegerParameters(requestDetails, "startIndex");
+    this.startIndex = parseIntegerParameters(requestDetails, "startIndex");
+    this.serverBase = requestDetails.getServerBaseForRequest();
+    this.resource = resource;
+    this.requestDetails = requestDetails;
+    validate();
+  }
+
+  /**
+   * Verifies that the link parameters, if not empty, are legal values. If the values are not legal,
+   * throws an {@link InvalidRequestException} to describe the error and return an http 400 to the
+   * caller.
+   */
+  private void validate() {
+    if (pageSize.isPresent() && pageSize.get() < 0) {
+      throw new InvalidRequestException(
+          String.format("Value for pageSize cannot be negative: %s", pageSize.get()));
+    }
     if (startIndex.isPresent() && startIndex.get() < 0) {
       throw new InvalidRequestException(
           String.format("Value for startIndex cannot be negative: %d", startIndex.get()));
     }
-    this.serverBase = requestDetails.getServerBaseForRequest();
-    this.resource = resource;
-    this.requestDetails = requestDetails;
   }
 
   /**
@@ -95,12 +108,7 @@ public final class OffsetLinkBuilder implements LinkBuilder {
   @Override
   public int getPageSize() {
     if (!isPagingRequested()) throw new BadCodeMonkeyException();
-    if (!pageSize.isPresent()) return 10;
-    if (pageSize.get() < 0) {
-      throw new InvalidRequestException(
-          String.format("Value for pageSize cannot be negative: %s", pageSize.get()));
-    }
-    return pageSize.get();
+    return pageSize.orElse(10);
   }
 
   /** {@inheritDoc} */
@@ -117,14 +125,7 @@ public final class OffsetLinkBuilder implements LinkBuilder {
    */
   public int getStartIndex() {
     if (!isPagingRequested()) throw new BadCodeMonkeyException();
-    if (startIndex.isPresent()) {
-      if (startIndex.get() < 0) {
-        throw new InvalidRequestException(
-            String.format("Value for startIndex cannot be negative: %s", startIndex.get()));
-      }
-      return startIndex.get();
-    }
-    return 0;
+    return startIndex.orElse(0);
   }
 
   /**
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
@@ -46,7 +46,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
 import javax.persistence.PersistenceContext;
@@ -335,16 +334,17 @@ public final class R4ExplanationOfBenefitResourceProvider extends AbstractResour
     boolean filterSamhsa = Boolean.parseBoolean(excludeSamhsa);
     CanonicalOperation operation = new CanonicalOperation(CanonicalOperation.Endpoint.V2_EOB);
     operation.setOption("by", "patient");
-    operation.setOption("IncludeTaxNumbers", "" + includeTaxNumbers);
+    operation.setOption("IncludeTaxNumbers", String.valueOf(includeTaxNumbers));
     operation.setOption(
         "types",
         (claimTypesRequested.size() == ClaimType.values().length)
             ? "*"
             : claimTypesRequested.stream()
                 .sorted(Comparator.comparing(ClaimType::name))
-                .collect(Collectors.toList())
+                .toList()
                 .toString());
-    operation.setOption("pageSize", paging.isPagingRequested() ? "" + paging.getPageSize() : "*");
+    operation.setOption(
+        "pageSize", paging.isPagingRequested() ? String.valueOf(paging.getPageSize()) : "*");
     operation.setOption(
         "_lastUpdated", Boolean.toString(lastUpdated != null && !lastUpdated.isEmpty()));
     operation.setOption(
@@ -377,6 +377,9 @@ public final class R4ExplanationOfBenefitResourceProvider extends AbstractResour
                 Optional.ofNullable(serviceDate),
                 filterSamhsa,
                 includeTaxNumbers);
+      } catch (InvalidRequestException e) {
+        // If we're throwing a 400, pass it back up
+        throw e;
       } catch (Exception e) {
         LOGGER.error(e.getMessage(), e);
       }
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
@@ -1765,9 +1765,9 @@ public final class TransformerUtilsV2 {
        * TODO: Above bug should be fixed as of 1/1/20; re-investigate this
        */
       int endIndex = Math.min(paging.getStartIndex() + paging.getPageSize(), resources.size());
-      resourcesSubList = resources.subList(paging.getStartIndex(), endIndex);
       // Throw a 400 if startIndex >= results, since we cant sublist with these values
       validateStartIndexSize(paging.getStartIndex(), resources.size());
+      resourcesSubList = resources.subList(paging.getStartIndex(), endIndex);
       bundle = TransformerUtilsV2.addResourcesToBundle(bundle, resourcesSubList);
       paging.setTotal(resources.size()).addLinks(bundle);
       // Add number of paginated resources to MDC logs
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
@@ -1766,6 +1766,8 @@ public final class TransformerUtilsV2 {
        */
       int endIndex = Math.min(paging.getStartIndex() + paging.getPageSize(), resources.size());
       resourcesSubList = resources.subList(paging.getStartIndex(), endIndex);
+      // Throw a 400 if startIndex >= results, since we cant sublist with these values
+      validateStartIndexSize(paging.getStartIndex(), resources.size());
       bundle = TransformerUtilsV2.addResourcesToBundle(bundle, resourcesSubList);
       paging.setTotal(resources.size()).addLinks(bundle);
       // Add number of paginated resources to MDC logs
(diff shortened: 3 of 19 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/OffsetLinkBuilder.java
```diff
@@ -48,14 +48,27 @@ public final class OffsetLinkBuilder implements LinkBuilder {
    */
   public OffsetLinkBuilder(RequestDetails requestDetails, String resource) {
     this.pageSize = parseIntegerParameters(requestDetails, Constants.PARAM_COUNT);
-    startIndex = parseIntegerParameters(requestDetails, "startIndex");
+    this.startIndex = parseIntegerParameters(requestDetails, "startIndex");
+    this.serverBase = requestDetails.getServerBaseForRequest();
+    this.resource = resource;
+    this.requestDetails = requestDetails;
+    validate();
```
lsmitchell, line 55: This is some cleanup to make this class do all its validations upfront, like PatientLinkBuilder does. Made more sense to me than waiting until we call the parameter to bother checking if its valid. (Not technically related to the fix, but I was in this code poking around for answers)

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -377,6 +377,9 @@ public Bundle findByPatient(
                 Optional.ofNullable(serviceDate),
                 filterSamhsa,
                 includeTaxNumbers);
+      } catch (InvalidRequestException e) {
+        // If we're throwing a 400, pass it back up
+        throw e;
```
lsmitchell, line 382: This is important; swallowing all exceptions from here with the below Exception block was stopping me from throwing 400s and caused a lot of headaches figuring out why I wasnt getting the expected 400 once I added the code fix. Here we carve out an exception for our 400 exception so it can bubble up properly.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
```diff
@@ -1764,6 +1764,14 @@ public static Bundle createBundle(
        * TODO: Above bug should be fixed as of 1/1/20; re-investigate this
        */
       int endIndex = Math.min(paging.getStartIndex() + paging.getPageSize(), resources.size());
+      // Throw a 400 if startIndex >= results, since we cant sublist with these values
+      if (paging.getStartIndex() >= resources.size()) {
+        throw new InvalidRequestException(
+            String.format(
+                "Value for startIndex (%d) must be less than than result size (%d)",
+                paging.getStartIndex(), resources.size()));
+      }
```
lsmitchell, line 1773: This is the crux of the fix that affects all endpoints' bundling; everything calls this method to do bundling (except PAC, which does it in the abstract resource provider directly)

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/ServerRequiredTest.java
```diff
@@ -45,6 +45,9 @@ public class ServerRequiredTest {
   @GuardedBy("class synchronized")
   protected static RequestSpecification requestAuth;
 
+  /** The Server test utils instance, for convenience and brevity. */
+  protected ServerTestUtils testUtils = ServerTestUtils.get();
```
lsmitchell, line 49: Added this for some convenience, instead of needing a lengthy ServerTestUtils.get() everywhere its used.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/ServerTestUtils.java
```diff
@@ -399,6 +405,103 @@ public static List<Object> parseData(List<StaticRifResource> sampleResources) {
     return recordsParsed;
   }
 
+  /**
```
lsmitchell, line 408: Moved these methods out of the EOB restAssured test, since all the new RestAssured tests needed them too.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/OffsetLinkBuilderTest.java
```diff
@@ -92,12 +92,11 @@ public void testNegativePageSizeExpectException() {
     String[] value = {"-1"};
     params.put(Constants.PARAM_COUNT, value);
     when(requestDetails.getParameters()).thenReturn(params);
-    OffsetLinkBuilder linkBuilder = new OffsetLinkBuilder(requestDetails, "");
 
     assertThrows(
         InvalidRequestException.class,
         () -> {
-          linkBuilder.getPageSize();
+          new OffsetLinkBuilder(requestDetails, "");
```
lsmitchell, line 99: Minor fix since we're validating in the constructor instead of the getter now.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageE2E.java
```diff
@@ -0,0 +1,101 @@
...
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+
+import gov.cms.bfd.server.war.ServerRequiredTest;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/** Endpoint end-to-end test for the V2 Coverage endpoint. */
+public class CoverageE2E extends ServerRequiredTest {
```
lsmitchell, line 12: The tests in this class are shared for all the others; once you've reviewed one of the test classes, the others are copy/pastes with the names, endpoint, and number of expected results changed.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitE2E.java
```diff
@@ -0,0 +1,929 @@
...
+import java.util.Date;
+import java.util.List;
+import org.apache.commons.lang3.tuple.ImmutableTriple;
+import org.apache.commons.lang3.tuple.Triple;
+import org.hl7.fhir.r4.model.Patient;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+/** Endpoint end-to-end test for the V1 explanation of benefits endpoint. */
```
lsmitchell, line 40: This is a full copy of the V2 restAssured test for EOB, save for testEobByPatientIdWithIncludeTaxNumbersTrueExpectTaxNumbers since the tax number path is different in v1, so it had its paths adjusted. (And of course adjusting the v2 -> v1)

lsmitchell (commented):
Adding some context to ease reviews.

lsmitchell:
Noted in the PR, but reposting here for visibility, this should NOT be merged until we have signoffs from the peering partners that it will not need corresponding changes on their side.

tracked here: https://confluence.cms.gov/pages/viewpage.action?pageId=896217997

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
brianburton: Can a client of the API know before making a request what the `resources.size()` will be?  If not is this really a client error?  Or just an edge case where we should return an empty response?

lsmitchell: I considered this a bit when designing the fix, whether to just do some default behavior like returning no results.

1. They wont necessarily know ahead of time if they are making their first call, but an empty response isnt as clear that there were results and with a fixed input they would return, and I'd rather indicate there was a potential issue with the nonsensical input.
2. Generally startIndex is used with a query where the result amount _should_ be known by having made a call first, otherwise why are they indexing into the results?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/OffsetLinkBuilder.java
mackec1: makes sense!

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
```diff
@@ -49,14 +43,14 @@
 public class ExplanationOfBenefitE2E extends ServerRequiredTest {
```
mackec1, line 43: Isn't this a slight deviation of class naming? We now have two ExplanationOfBenefitE2E, just in diff packages.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/OffsetLinkBuilder.java
brianburton: Where would this `IllegalRequestException` be handled if it's thrown in the constructor?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
mackec1: yea I wouldn't worry about it...in fact it is preferable from having the 'full blown name', especially if we have a V3 coming down the road.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
brianburton, line 1773: Might be a good candidate for a utility method.  Seeing exactly the same lines of code in 3 different classes.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -335,16 +334,17 @@ public Bundle findByPatient(
     boolean filterSamhsa = Boolean.parseBoolean(excludeSamhsa);
     CanonicalOperation operation = new CanonicalOperation(CanonicalOperation.Endpoint.V2_EOB);
     operation.setOption("by", "patient");
-    operation.setOption("IncludeTaxNumbers", "" + includeTaxNumbers);
+    operation.setOption("IncludeTaxNumbers", String.valueOf(includeTaxNumbers));
```
brianburton, line 337: I like this change.  Was never a fan of the `empty string +` pattern.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/ServerTestUtils.java
```diff
@@ -399,6 +405,103 @@ public static List<Object> parseData(List<StaticRifResource> sampleResources) {
...
+   * @return the pagination link
+   */
+  public String getPaginationLink(Response response, String paginationLinkName) {
+
+    List<Map<String, ?>> links = response.jsonPath().getList("link");
+    return links.stream()
+        .filter(m -> m.get("relation").equals(paginationLinkName))
+        .findFirst()
+        .orElseThrow()
+        .get("url")
```
brianburton, line 501: I like the use of `orElseThrow()`to avoid calling two methods named `get` from two different classes here.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
lsmitchell: I think per the engineering sync chat I'll leave these, and I'll propose a story to nix the prefix/postfix R4/V2/V1/stu3 on the other classes as a separate effort.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/OffsetLinkBuilder.java
lsmitchell: It gets caught by HAPI-FHIR and turned into a 400. It's HAPI-FHIR's exception, and the only way to get it to throw 400s which is kinda gross since it ties us to the implementation, but every non-HAPI exception creates a 500 in their internal code.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
```diff
@@ -49,14 +43,14 @@
...
 
   /** The base eob endpoint. */
-  public static String EOB_ENDPOINT;
+  public static String eobEndpoint;
```
brianburton, line 46: Why was this ever public?

lsmitchell: Good point, it can be protected.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/TransformerUtilsV2.java
lsmitchell: I'll make one for the TransformerUtilsV2 since it gets shared between here and PAC code, but two usages are in the separate transformerUtils versions which dont have a good place to share code between them. Maybe in future work to pull out shared code we can reuse the new util method I'll make in both v1 and v2 locations and put it in some commonTransformerUtils class or something.

brianburton (approved):
Looks good to me!

mackec1 (approved):

lsmitchell:
Got sign off from all peering partners that this would not require any changes on their side, so this is ready to merge.

brianburton (approved):

brandoncruz3 (approved):

PR #1942, opened 2023-09-19
BFD-2862: BFD Server ASG scale-in policy is evaluating too soon after scale-out causing premature scale-in

malessi:
**JIRA Ticket:**
[BFD-2862](https://jira.cms.gov/browse/BFD-2862)

The BFD Server scale-in policy will immediately scale-in instances when NetworkIn drops beneath a particular scale-in threshold range, unless prior to scale-in the BFD Server ASG scaled-out to 12 instances. The expected behavior is that scale-in does not occur until 10 minutes after scale-out, but the current scale-in policy allows for near immediate scale-in after scale-out.

---

### What Does This PR Do?

This PR updates the scaling "configuration" such that:
- scale-in should no longer be able to immediately, erroneously activate and scale-in the Server after a scale-out.
- the period of time that must pass before scale-in can occur has been increased to 15 minutes from 10 minutes to ensure the Server remains scaled-out during our typical "bursty" bulk loads. 
- the [instance warmup](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-default-instance-warmup.html?icmpid=docs_ec2as_help_panel) has been removed entirely, as [BFD-2863](https://jira.cms.gov/browse/BFD-2863) introduced Lifecycle Hooks that are only completed once a given instance is warmed-up, so an arbitrary instance warmup is unnecessary.

This PR has been validated by:

- `terraform apply`ing the changes to the `test` environment's `server`, verifiying that the changes apply successfully and as expected
- manually setting the desired capacity of the `test` environment's `server` ASG to capacities greater than 3 (thus simulating scale-out), verifying that the ASG scales-in after 15 minutes to the appropriate number of instances

#### Context

The original scale-in implementation used a single CloudWatch Alarm with a single, corresponding scale-in policy attached to the ASG. The scale-in alarm would ALARM if, over a _consecutive_ period of 10 minutes, the `NetworkIn` was beneath 400 MB/min. Therefore, if at any point the `NetworkIn` exceeded 400 MB/min, the alarm would transition to `OK`. After such a transition, the alarm would only ever transition back to `ALARM` after 10 consecutive minutes of network traffic less than 400 MB/min.

If the scale-in policy's corresponding alarm is _ever_ in the `ALARM` state, that means the scale-in policy is active and its thresholds are evaluated to determine which scaling action to take. Because 400 MB/min was the _end_ of the range of scaling stages configured for the server and network traffic only ever exceeds 400 MB/min when we get larger loads, the scale-in alarm was practically _always_ in `ALARM`. This meant that the scale-in policy could almost always _evaluate_. So, there was often a scenario where the Server would get enough traffic to scale-out but because the traffic did not quite exceed 400 MB/min, the scale-in alarm never stopped `ALARM`ing and the scale-in policy would be able to evaluate immediately afterwards. This would usually result in scaling oscillation, where the Server would scale at one minute and then immediately scale-in the next minute.

This has been resolved by splitting the scale-in to `n - 1` scale-in policies and `n - 1` scale-in alarms, where `n` is the number of scaling stages that are defined within code. We have 4 stages, so this implementation will create 3 distinct scale-in alarms and corresponding policies. Each scale-in policy and alarm evaluates a particular stage of scaling (ranges of network traffic). By moving the threshold evaluation to CloudWatch Metric Math, only returning a value if the number of instances exceed the scale-in policy's desired capacity target, and configuring each alarm to treat missing data as `OK`, the alarms will now transition to `OK` whenever a scaling event occurs. This "breaks" the consecutive periods of evaluation, forcing each alarm to re-evaluate after scale-out over the _full_ 15 minute period before transitioning to a new state, thus ensuring the scale-in policies cannot evaluate immediately after scale-out.

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist

### DIFF
--- a/ops/terraform/services/server/modules/bfd_server_asg/main.tf
+++ b/ops/terraform/services/server/modules/bfd_server_asg/main.tf
@@ -5,11 +5,16 @@ locals {
   rds_reader_endpoint = data.external.rds.result["CustomEndpoint"] == "" ? data.external.rds.result["ReaderEndpoint"] : data.external.rds.result["CustomEndpoint"]
 
   additional_tags = { Layer = var.layer, role = var.role }
-  scaleout_asg_capacities = [
-    { capacity = length(var.env_config.azs) * 2, metric_lower_bound = 1 * var.scaling_networkin_interval_mb, metric_upper_bound = 2 * var.scaling_networkin_interval_mb },
-    { capacity = length(var.env_config.azs) * 3, metric_lower_bound = 2 * var.scaling_networkin_interval_mb, metric_upper_bound = 4 * var.scaling_networkin_interval_mb },
-    { capacity = length(var.env_config.azs) * 4, metric_lower_bound = 4 * var.scaling_networkin_interval_mb, metric_upper_bound = null }
+
+  scaling_capacity_step = length(var.env_config.azs)
+  scaling_stages = [
+    { begin = 0, end = 1 * var.scaling_networkin_interval_mb, desired_capacity = local.scaling_capacity_step * 1 },
+    { begin = 1 * var.scaling_networkin_interval_mb, end = 2 * var.scaling_networkin_interval_mb, desired_capacity = local.scaling_capacity_step * 2 },
+    { begin = 2 * var.scaling_networkin_interval_mb, end = 4 * var.scaling_networkin_interval_mb, desired_capacity = local.scaling_capacity_step * 3 },
+    { begin = 4 * var.scaling_networkin_interval_mb, end = null, desired_capacity = local.scaling_capacity_step * 4 },
   ]
+  scalein_config  = slice(local.scaling_stages, 0, length(local.scaling_stages) - 1)
+  scaleout_config = slice(local.scaling_stages, 1, length(local.scaling_stages))
 
   on_launch_lifecycle_hook_name = "bfd-${local.env}-${var.role}-on-launch"
 }
@@ -138,7 +143,7 @@ resource "aws_autoscaling_group" "main" {
   min_elb_capacity          = var.lb_config == null ? null : var.asg_config.min
   wait_for_capacity_timeout = var.lb_config == null ? null : "20m"
 
-  health_check_grace_period = 600 # Temporary, will be lowered when/if lifecycle hooks are implemented
+  health_check_grace_period = 600                                   # Temporary, will be lowered when/if lifecycle hooks are implemented
   health_check_type         = var.lb_config == null ? "EC2" : "ELB" # Failures of ELB healthchecks are asg failures
   vpc_zone_identifier       = data.aws_subnet.app_subnets[*].id
   load_balancers            = var.lb_config == null ? [] : [var.lb_config.name]
@@ -197,13 +202,15 @@ resource "aws_autoscaling_group" "main" {
 #
 
 resource "aws_cloudwatch_metric_alarm" "filtered_networkin_low" {
-  alarm_name          = "bfd-${var.role}-${local.env}-networkin-low"
-  comparison_operator = "LessThanThreshold"
+  for_each = { for v in local.scalein_config : "${v.desired_capacity}-instances" => v }
+
+  alarm_name          = "bfd-${var.role}-${local.env}-networkin-low-${each.key}"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
   datapoints_to_alarm = 10
   evaluation_periods  = 10
-  threshold           = 400 * 1000000 # 400 megabytes
-  treat_missing_data  = "ignore"
-  alarm_actions       = [aws_autoscaling_policy.filtered_networkin_low_scaling.arn]
+  threshold           = 1
+  treat_missing_data  = "notBreaching"
+  alarm_actions       = [aws_autoscaling_policy.filtered_networkin_low_scaling[each.key].arn]
 
   metric_query {
     id          = "m1"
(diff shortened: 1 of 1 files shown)


brandoncruz3 (dismissed):

ops/terraform/services/server/modules/bfd_server_asg/main.tf
```diff
@@ -5,11 +5,38 @@ locals {
...
+    # low network traffic. This is intended to keep the Server scaled-out as traffic is often bursty
+    # over a given period rather than sustained at a given threshold
+    scale_in = {
+      eval_period                  = 1 * 60
+      datapoints_to_alarm          = 15
+      consecutive_periods_to_alarm = 15
+    }
+  }
+  scaling_stages = [
+    # Between 0 MB/min and 100 MB/min we want to be at 3 instances
```
billie-rose, line 29: Huge nitpick and not important, but I would have these comments say something like "between 0 MB/min and 1x the scaling interval MB/min..." so we don't have to update these comments/have misleading comments should that interval change w/o updating the description

ops/terraform/services/server/modules/bfd_server_asg/main.tf
```diff
@@ -230,65 +259,71 @@ resource "aws_cloudwatch_metric_alarm" "filtered_networkin_low" {
...
+  }
+
+  metric_query {
+    id    = "e2"
+    label = "Set to ${each.value.desired_capacity} capacity units"
+    expression = "IF(${join(" && ", compact([
+      each.value.begin > 0 ? "networkin > ${each.value.begin}" : null,
+      each.value.end != null ? "networkin <= ${each.value.end}" : null,
+      "m3 > ${each.value.desired_capacity}"
+    ]))}, 1)"
```
billie-rose, line 296: Dumb question but should this be `1 * ${each.value.desired_capacity}`..?

ops/terraform/services/server/modules/bfd_server_asg/main.tf
billie-rose: also - one of the first "gotcha" thoughts I had was to make understand which values between `begin` and `end` were exclusive/inclusive - can we make a comment for that here?

ops/terraform/services/server/modules/bfd_server_asg/main.tf
```diff
@@ -230,65 +259,71 @@ resource "aws_cloudwatch_metric_alarm" "filtered_networkin_low" {
...
+    id          = "networkin"
     label       = "FilteredNetworkIn"
+    return_data = false
+  }
+
+  metric_query {
+    id    = "e2"
+    label = "Set to ${each.value.desired_capacity} capacity units"
+    expression = "IF(${join(" && ", compact([
+      each.value.begin > 0 ? "networkin > ${each.value.begin}" : null,
```
billie-rose, line 293: Not sure if it's even possible, but if it is, should we have `begin` be inclusive and `end` exclusive? Just in case `networkin` is ever 0?

billie-rose:
Thanks for all the work you've put into this :) just a handful of comments.

ops/terraform/services/server/modules/bfd_server_asg/main.tf
malessi: It's definitely possible, but would require the corresponding scale-out stages be swapped as well. If the ranges don't match between the two, we end up with inconsistent behavior if `NetworkIn` is ever _exactly_ on the boundary of a given range (i.e. 100 MB/min, 200 MB/min, etc.).

For this particular case, the ternary ensures that if the alarm's stage has a range with a `begin` of 0, the generated Metric Math looks like this:

```
IF(networkin <= {END} && m3 > {DESIRED_CAPACITY}, 1)
```
so that _any_ `networkin` beneath `end` will return data. Obviously, this is different from:
```
IF(networkin > 0 && networkin <= {END} && m3 > {DESIRED_CAPACITY}, 1)
```
which would result in a `networkin` of 0 returning no data (thus "breaking" consecutive periods of evaluation and forcing the alarm to start re-evaluating from the start of a new 15 minute period).

We should be covered by the above, but I have no problem inverting the range inclusiveness to make `begin` inclusive and `end` exclusive if that would be more understandable. We would be able to remove this ternary, for example.

ops/terraform/services/server/modules/bfd_server_asg/main.tf
malessi: Assuming you're asking about line 296: this `1` could really be any value, assuming the value broke the alarm's threshold. The value returned by the `IF()`, in this case `1`, can be thought of a boolean value indicating whether the ASG _could_ be scaled-down at that given period. If there are 15 consecutive periods of this boolean value being "true" (`1`), then the alarm enters `ALARM` and the scale-in policy can evaluate.

Does that make sense? I can see how the label for this expression is a little vague, so I can definitely reword it to make the above more clear. Definitely not a dumb question, either -- AWS's scaling mechanisms aren't really designed for this particular scaling configuration so this is coloring outside the lines a bit.

ops/terraform/services/server/modules/bfd_server_asg/main.tf
malessi: I agree. I'll reword these comments to indicate the which bounds are inclusive/exclusive. I'll also make the values more generic.

ops/terraform/services/server/modules/bfd_server_asg/main.tf
billie-rose: Gotcha! Thank you for elaborating - makes total sense.

ops/terraform/services/server/modules/bfd_server_asg/main.tf
billie-rose: I think all is well with what you have here :) No need to go and change things up when they're perfectly functional as-is

billie-rose (approved):

mjburling (approved):

brandoncruz3 (approved):

PR #1955, opened 2023-09-26
BFD-2928: Convert remaining EOB V2 E2E into RestAssured

lsmitchell:
**JIRA Ticket:**
[BFD-2928](https://jira.cms.gov/browse/BFD-2928)

As a BFD Engineer I would like the existing resource provider E2E tests to use restAssured so that we are testing our REST endpoints as customers use them.

---

### What Does This PR Do?

A majority of the tests inside of `R4ExplanationOfBenefitResourceProviderE2E.java` were converted into RestAssured tests and placed into ExplanationOfBenefitE2E as part of the hackathon work. This PR converts remaining straggler tests and test cases not captured in that initial work, with the intention of completely replacing R4ExplanationOfBenefitResourceProviderE2E with RestAssured versions of the tests.

This story scope is specifically the v2 versions of the EOB E2E.

- Fixes the (v2) ExplanationOfBenefitE2E samhsa tests to cover all claim types as they were covered in the original E2E
    - Moves the samhsa adding logic into the common server utility class
- Adds tests to cover all claim types for the read endpoint, and extracts common logic to make the tests DRY
- Adjust the pagination test to use an odd pagination value to cover a test case from the original E2E
    - Test case is where we have an odd pagination number (and an even number of results) we make sure the last 2 items dont cause an out-of-bounds exception
- Adds missing restAssured test for pagination if only startIndex is present
- Added samhsa test to ensure default behavior for when samhsa flag is not explicitly added
- Added some test utility method adjustments to cover all claim types instead of just carrier
- Deleted R4ExplanationOfBenefitResourceProviderE2E since all test cases are now covered in ExplanationOfBenefitE2E

One omission is the field comparison tests that existed in R4ExplanationOfBenefitResourceProviderE2E. These tests copy the same case as EndpointJsonResponseComparatorV2IT, as they just load the SampleA data and run it through the resource provider+transformer, then compare the output to the original loaded data. The EndpointJson tests do the same thing (but with the final json output) so these tests are superfluous.

* Check if theres anything I missed about the field-checking tests removed from the deleted E2E where they are not just covered by the endpointJson tests

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
```diff
@@ -53,42 +59,87 @@ public void setupTest() {
...
   }
 
   /** Verifies that the EOB returns a 400 when called with a negative count parameter. */
   @Test
   public void testEobByPatientIdWithNegativeCountExpect400() {
 
     String patientId = testUtils.getPatientId(testUtils.loadSampleAData());
-    // It's also possible to add these to the call below, but I liked the request string as
-    // one would write it when calling the API
+    // Its possible to add these via the API, but
```
mackec1, line 142: seems a little wonky....", but" what?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
```diff
@@ -926,44 +1051,89 @@ public void searchEobByPatientIdWithNoResultsAndPaginationRequestedExpect200() {
...
-    return loadedRecords.stream()
-        .filter(r -> r instanceof OutpatientClaim)
-        .map(OutpatientClaim.class::cast)
-        .findFirst()
-        .get();
+  private String getClaimIdFor(List<Object> loadedRecords, ClaimType claimType) {
+    return switch (claimType) {
+      case CARRIER:
+        CarrierClaim carrier = getClaim(loadedRecords, CarrierClaim.class);
+        yield String.valueOf(carrier.getClaimId());
```
mackec1, line 1081: NICE! I like this use of yield.

mackec1 (dismissed):
one dangling comment

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
lsmitchell: Oops, had a fix for this and never pushed it. Fixed!

brandoncruz3 (approved):

mackec1 (approved):

PR #1965, opened 2023-10-02
BFD-2929: Convert V1 EOB E2E tests to RestAssured

lsmitchell:
**JIRA Ticket:**
[BFD-2929](https://jira.cms.gov/browse/BFD-2929)

As a BFD Engineer I would like the existing resource provider E2E tests to use restAssured so that we are testing our REST endpoints as customers use them.

---

### What Does This PR Do?

- Moves the RestAssured tests for EOB v2 into a common class to be extended by the V1 and V2 EOB E2Es, since 99% of the test logic in RestAssured is the same between versions
- Provides a V1 and V2 class for inheriting the tests, as well as stub methods that call the super for ease of calling these tests from an IDE (Calling a single test when inheriting them from the super class doesnt work from IDE, so we need a hook method in the appropriate versioned test class)
- Moves test cases that existed in v1 but not v2 into the common shared test pool

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist


apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/ExplanationOfBenefitE2EBase.java
```diff
@@ -0,0 +1,1115 @@
+package gov.cms.bfd.server.war;
```
lsmitchell, line 1: This is 90% copying the V2 tests into a common location, and 10% creating the few test cases that existed in the v1 suite but not v2 suite. This class allows for all test cases to be shared across both versions, ultimately increasing our coverage for both.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
lsmitchell, line 1: Changes consist of adjusting methods to call their common version, adding the test cases added from V1. The one method that isnt common is due to path differences between v1 and v2.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/stu3/providers/ExplanationOfBenefitE2E.java
lsmitchell, line 1: Changes consist of adjusting methods to call their common version. The one method that isnt common is due to path differences between v1 and v2.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/ExplanationOfBenefitE2E.java
```diff
@@ -65,7 +76,7 @@ public void setupTest() {
    */
   @Test
   public void testReadEobCarrierIdExpectClmIdInResponse() {
-    verifySuccessfulResponseAndClaimIdFor(ClaimType.CARRIER);
+    super.testReadEobCarrierIdExpectClmIdInResponse();
   }
```
brianburton, line 80: I'd prefer to see these removed if they are not necessary.  They add some confusion and extra maintenance burden and another thing a maintainer would think was necessary to do when adding another test to the base class.

lsmitchell: Yeah I didn't love it. It seems using the Structure tab you mention we can run individual inherited tests without issue, so I'm ok to annotate in the base class and reduce the duplication. I did add some light instructions on how to run the individual tests in-IDE in the class doc, since it wasn't clear to me at the time.

lsmitchell:
Moved the test annotations to the base test and removed the duplicated test methods out of the child classes.

brandoncruz3 (approved):

brianburton (approved):
Nice work.

mackec1 (approved):

PR #1974, opened 2023-10-03
BFD-2507: Fix bug relating to Athena query failures being ignored

aadedejifearless:
**JIRA Ticket:**
[BFD-2507](https://jira.cms.gov/browse/BFD-2507)

- Description
Our Locust tests do not treat Athena queries failing as actual failures, so when this case happens we see no failures in our deployment pipeline.
- Background
A relatively recent string of failures in the Regression Suite caused by upgrades to Athena would have been caught faster if the aggregated stats framework treated FAILED Athena queries as true failures instead of ignoring them outright (treating FAILED queries as an absence of possible stats to compare against). [BFD-2456](https://jira.cms.gov/browse/BFD-2456) resolved this particular issue, but future issues could occur that would result in ignored FAILED Athena queries.
---

### What Does This PR Do?

- Updates the `stats_compare.py` to send the failure to the pipeline. 

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/utils/locust_tests/common/stats/stats_compare.py
+++ b/apps/utils/locust_tests/common/stats/stats_compare.py
@@ -119,7 +119,7 @@ def do_stats_comparison(
             str(stats_config.stats_store.value),
             str(ex),
         )
-        return FinalCompareResult.NOT_APPLICABLE
+        return FinalCompareResult.FAILED
 
     if not previous_stats:
         logger.warning(


apps/utils/locust_tests/common/stats/stats_compare.py
```diff
@@ -119,7 +119,7 @@ def do_stats_comparison(
             str(stats_config.stats_store.value),
             str(ex),
         )
-        return FinalCompareResult.NOT_APPLICABLE
+        return FinalCompareResult.FAILED
```
billie-rose, line 122: Just wanted to double-check that there aren't any scenarios where it would be OK for the stats to not load? @malessi ?

malessi: I took a look at this with Ade and I'm _pretty_ sure the only times `RuntimeError`s are raised when we're running an Athena query is (a `RuntimeError` is never explicitly raised by the `StatsFileLoader`):

1. When `query_result` is `None` in `__get_stats_from_query(...)` (lines 204 - 207 in `stats_loaders.py`):

   ```python
   def __get_stats_from_query(self, query: str) -> List[AggregatedStats]:
       query_result = self.__run_query(query)
       if not query_result:
           raise RuntimeError("Athena query result was empty or query failed")
   ```

   `query_result` is never empty/`None` _even when the query retuns no results_ (I tested this locally by running a Regression Suite test configured to compare against a tag that doesn't exist).

   So, if it were empty/`None` something is definitely wrong.

2. When the `status` of the running Athena query becomes `FAILED` or `CANCELLED` (lines 258 - 261 in `stats_loaders.py`):

   ```python
   if status == "SUCCEEDED":
       break
   elif status == "FAILED" or status == "CANCELLED":
       raise RuntimeError(f"Query failed to complete -- status returned was {status}")
   ```

   This is the main case we want to signal failure for, as the query `status` only ever becomes `FAILED` if something is wrong with the query itself. 
   
   Had we already been signaling failure (by setting the compare result to `FAILED` instead of `NOT_APPLICABLE`) in this case we would have immediately caught the issues that occurred a few months back.

lsmitchell (approved):

malessi (approved):

apps/utils/locust_tests/common/stats/stats_compare.py
aadedejifearless: @billie-rose are you good with this?

billie-rose (approved):

PR #2015, opened 2023-10-25
BFD-2933: Convert AccessJsonE2E to restAssured

lsmitchell:
**JIRA Ticket:**
[BFD-2933](https://jira.cms.gov/browse/BFD-2933)

As a BFD Engineer I would like existing IT tests to use restAssured so that we are testing our REST endpoints as customers use them.

---

### What Does This PR Do?

- Coverts the AccessJsonIT to AccessJsonE2E and converts the test to use RestAssured
  - Fixes the test to wipe access.json before running, since before it was not and the test results were based on what (if anything) had run before it. The test previously did not pass when run solo.
  - Removes two headers which don't get set as part of the test as written: Content-Location (which I don't know if/how we use this today, but we dont explicitly set or use it) and Accept-Charset (which has been deprecated and advised to not be used since 2012, and we dont explicitly set or check for this in code)
- Fixed the assert to keep track of all failed headers and do a single assert, which makes the output a lot more useful to read as well as checking for all headers at once instead of the first failed one

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

### DIFF
--- a/apps/bfd-server/bfd-server-shared-utils/src/main/java/gov/cms/bfd/server/sharedutils/BfdMDC.java
+++ b/apps/bfd-server/bfd-server-shared-utils/src/main/java/gov/cms/bfd/server/sharedutils/BfdMDC.java
@@ -131,10 +131,6 @@ public class BfdMDC {
   public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT =
       computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept");
 
-  /** MDC key for the http request header accept charset. */
-  public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT_CHARSET =
-      computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept-Charset");
-
   /** MDC key for the http request header accept encoding. */
   public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT_ENCODING =
       computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept-Encoding");
--- a/apps/bfd-server/bfd-server-shared-utils/src/main/java/gov/cms/bfd/server/sharedutils/BfdMDC.java
+++ b/apps/bfd-server/bfd-server-shared-utils/src/main/java/gov/cms/bfd/server/sharedutils/BfdMDC.java
@@ -131,6 +131,10 @@ public class BfdMDC {
   public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT =
       computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept");
 
+  /** MDC key for the http request header accept charset. */
+  public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT_CHARSET =
+      computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept-Charset");
+
   /** MDC key for the http request header accept encoding. */
   public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT_ENCODING =
       computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept-Encoding");
--- a/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
+++ b/apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
@@ -1,18 +1,28 @@
 package gov.cms.bfd.server.war;
 
 import static io.restassured.RestAssured.given;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import gov.cms.bfd.model.rif.entities.Beneficiary;
 import gov.cms.bfd.server.sharedutils.BfdMDC;
 import gov.cms.bfd.server.war.commons.TransformerConstants;
+import java.io.BufferedWriter;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
 import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Verifies that access.json is written to as expected. */
 public class AccessJsonE2E extends ServerRequiredTest {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(AccessJsonE2E.class);
+
   /** Verifies that access.json is written to within BFD-server-war via API call. */
   @Test
   public void verifyAccessJson() throws IOException {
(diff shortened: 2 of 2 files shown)


mackec1 (dismissed):
Nice...continue to see benefits of RestAssured!

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
```diff
@@ -0,0 +1,123 @@
...
+    Path accessLogJson =
+        ServerTestUtils.getWarProjectDirectory()
+            .resolve("target")
+            .resolve("server-work")
+            .resolve("access.json");
+
+    // Empty the access json to avoid pollution from other tests
+    try (BufferedWriter writer = Files.newBufferedWriter(accessLogJson)) {
+      writer.write("");
+      writer.flush();
```
brianburton, line 48: This would do the same thing in one line:

`Files.writeString(accessLogJson, "")`

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
```diff
@@ -0,0 +1,123 @@
...
+            BfdMDC.RESOURCES_RETURNED);
+
+    List<String> missingHeader = new ArrayList<>();
+    for (String header : headersToCheck) {
+      if (!content.contains(header)) {
+        missingHeader.add(header);
+      }
+    }
+    // Check we have no missing headers in a way that makes a useful assertion error if not
+    assertEquals(new ArrayList<>(), missingHeader, "Missing expected headers in access.json!");
```
brianburton, line 121: Loving this!  Big improvement over the old way.

brandoncruz3 (dismissed):

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
```diff
@@ -0,0 +1,123 @@
...
+        .expect()
+        .statusCode(200)
+        .when()
+        .get(requestString);
+
+    // Verify that the access JSON is working, as expected.
+    try {
+      TimeUnit.MILLISECONDS.sleep(1000); // Needed to resolve a race condition
+    } catch (InterruptedException ignored) {
+    }
```
mjburling, line 63: This is a curious few lines. I see _this_ implementation requires that we not only maintain the solution, but we double the value. It wasn't _too_ offensive when we originally accepted this as part of #1557, but I do recall pairing on this and asking if it could be removed... I **think** it was only kept because we couldn't get GitHub Actions to behave without it at the time. Is that where you saw failures or were you seeing these failures locally, too?

In general, can we say that we're more tolerant of sleeps like this in tests?

mjburling (changes requested):
This is super tidy, @lsmitchell. Well done!

Unfortunately, we need to better understand what impacts log content/format changes may have. Maybe we can just leave those improvements out for the moment?

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
lsmitchell: I didn't see any issues removing this when I was testing, but it existed before and I didn't want the test failing in a week cause I didn't locally see the issue a few times. If you're thinking the original cause is fixed, I can nix it, since I don't like thread sleeps either.

apps/bfd-server/bfd-server-shared-utils/src/main/java/gov/cms/bfd/server/sharedutils/BfdMDC.java
```diff
@@ -131,10 +131,6 @@ public class BfdMDC {
   public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT =
       computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept");
 
-  /** MDC key for the http request header accept charset. */
-  public static final String HTTP_ACCESS_REQUEST_HEADER_ACCEPT_CHARSET =
-      computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, HEADER_PREFIX, "Accept-Charset");
-
```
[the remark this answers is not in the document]
lsmitchell: I can leave it in, although it wasn't actually being used anywhere outside this one test. 

Separate from this effort, we may look at if this value is used or anyone cares about it; according to all documentation I can find (like this [protocol rfc](https://www.rfc-editor.org/rfc/rfc9110.html#name-accept-charset) amongst others) accept-charset has been deprecated for almost a decade, and its generally accepted practice to not use/accept/return it.

brianburton (dismissed):

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
lsmitchell: Removed it, as long as github actions passes for this PR, I'll assume it'll be fine. (If not, can add it back if we see an issue)

mjburling (dismissed):
Very tidy!

I've scheduled a follow-up for the interactions between access.json and the glue schema in [BFD-3022](https://jira.cms.gov/browse/BFD-3022).

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
```diff
@@ -0,0 +1,115 @@
...
+            BfdMDC.HTTP_ACCESS_REQUEST_HEADER_HOST_ENCODING,
+            BfdMDC.HTTP_ACCESS_REQUEST_HEADER_USER_AGENT,
+            BfdMDC.HTTP_ACCESS_REQUEST_HTTP_METHOD,
+            BfdMDC.HTTP_ACCESS_REQUEST_OPERATION,
+            BfdMDC.HTTP_ACCESS_REQUEST_QUERY_STR,
+            BfdMDC.HTTP_ACCESS_REQUEST_TYPE,
+            BfdMDC.HTTP_ACCESS_REQUEST_URI,
+            BfdMDC.HTTP_ACCESS_REQUEST_URL,
+            BfdMDC.HTTP_ACCESS_RESPONSE_DURATION_PER_KB,
+            BfdMDC.HTTP_ACCESS_RESPONSE_DURATION_MILLISECONDS,
```
aadedejifearless, line 94: @lsmitchell I think there's a duplication here.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/AccessJsonE2E.java
```diff
@@ -0,0 +1,115 @@
...
+            BfdMDC.DATABASE_QUERY_TYPE,
+            BfdMDC.DATABASE_QUERY_BATCH_SIZE,
+            BfdMDC.DATABASE_QUERY_SIZE,
+            BfdMDC.DATABASE_QUERY_MILLI,
+            BfdMDC.DATABASE_QUERY_SUCCESS,
+            BfdMDC.DATABASE_QUERY_SOURCE_NAME,
+            BfdMDC.HAPI_RESPONSE_TIMESTAMP_MILLI,
+            BfdMDC.HAPI_POST_PROCESS_TIMESTAMP_MILLI,
+            BfdMDC.HAPI_PRE_HANDLE_TIMESTAMP_MILLI,
+            BfdMDC.HAPI_PRE_HANDLE_TIMESTAMP_MILLI,
```
aadedejifearless, line 77: Same here

aadedejifearless (approved):

mjburling (approved):

brandoncruz3 (approved):

PR #2047, opened 2023-11-07
BFD-2887: Run server-load in docker-host

billie-rose:
**JIRA Ticket:**
[BFD-2887](https://jira.cms.gov/browse/BFD-2887)

During an incident on 9/8/23 we were able to make a temporary fix to work around the issues we were having with our builds. We determined that the best long term solution for this problem would be replatforming.

---

### What Does This PR Do?

Migrate the server-load process from an AMI to a Docker image.
This PR eliminates the server-load AMI build process and instead builds a server-load-controller docker image which is then launched onto the docker-host AMI when the server-load job is run.

Note that the `requirements.txt` was re-generated using `pip-compile` and required due to a conflict with `awscli` and the previous version of `boto3`.

The server-load node build has been renamed as part of this PR to better describe its role and relationship with the server-load controller.

      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No
      * [ ] Yes
      * [X] No

      * [ ] Yes
      * [X] No

* N/A

### Submitter Checklist


ops/jenkins/bfd-build-server-load-controller/build-push.sh
```diff
@@ -0,0 +1,42 @@
...
+# Push image to ECR
+docker push "$IMAGE_NAME_NODE" --all-tags
+
+# Put the image tag in SSM
+aws ssm put-parameter \
+  --name "$SSM_IMAGE_TAG_NODE" \
+  --value "${IMAGE_NAME_NODE}:${DOCKER_TAG}" \
+  --overwrite \
+  --type String \
+  --region "$AWS_REGION"
```
brandoncruz3, line 41: Can we add our default tags to any parameters we are putting into SSM via CLI? This will help note what params are not in IaC and are created via CLI. Thanks!

brandoncruz3 (changes requested):
Would like to have that `put-parameter` updated to include default tags. Other than that, I think this is a great PR. Thanks

lsmitchell (dismissed):
Looks simplified and improved to me 👍

ops/jenkins/bfd-build-server-load-controller/build-push.sh
billie-rose: Sorry, I've read this comment a few times and I'm not sure what you mean?

brandoncruz3: No worries. I was wondering if you could add our default tags to the put-parameter call on lines L36-L41. Something like 
```
aws ssm put-parameter \
  --name "$SSM_IMAGE_TAG_NODE" \
  --value "${IMAGE_NAME_NODE}:${DOCKER_TAG}" \
  --overwrite \
  --type String \
  --tags Key=Environment,Value=mgmt Key=stack,Value=mgmt Key=Terraform,Value=False \
  Key=application,Value=bfd Key=business,Value=oeda \
  --region "$AWS_REGION"
```

brandoncruz3: Of course, whatever tags you think are appropriate would suffice here. Thanks!

billie-rose: Ahh, I see. Good call

billie-rose: I ended up doing this OOB since you cannot overwrite SSM params while also passing in tags, and writing code to handle this conditional each time the image is built seems extraneous to me...thoughts?

brandoncruz3 (approved):

lsmitchell (approved):

malessi (approved):
Nothing brings me more joy than seeing an Ansible role disappear.

This looks great, and _quite_ a bit easier to follow than the existing solution. Thanks Billie!

PR #2156, opened 2024-01-22
BFD-3013: Pipeline SLIs Lambda fails to submit metrics when RIFs load too quickly

malessi:
**JIRA Ticket:**
[BFD-3013](https://jira.cms.gov/browse/BFD-3013)

The Pipeline SLIs Lambda, `bfd-${env}-update-pipeline-slis`, will fail to submit Pipeline metrics to CloudWatch Metrics when a RIF file loads too quickly. Specifically, it will fail to submit metrics to the following CloudWatch Metrics:

- `time/data-loaded`
- `time/data-fully-loaded`
  - If, and only if, the RIF file that loaded too quickly was the final RIF loaded
  - If the Lambda fails to submit metrics to this CloudWatch Metric, the corresponding data load will be considered incomplete indefinitely, and corresponding Pipeline SLO Alarms will alarm erroneously

---

### What Does This PR Do?

This PR:

- Updates the Pipeline SLIs Lambda to use [DynamoDB Tables](https://aws.amazon.com/dynamodb/) to track when a particular RIF and data load is made available for duration-to-load metric calculations, instead of querying CloudWatch Metrics
  - These Tables are configured to use ["On-Demand Capacity"](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html#HowItWorks.OnDemand). Considering the _very_ small average size of each item that is stored in these Tables (about 120 bytes per-item), and the _relative_ infrequency of storage and retrieval, using "On-Demand Capacity" should be _significantly_ more cost effective than using the cheapest provisioned configuration. Rough calculations (using [On-Demand pricing information](https://aws.amazon.com/dynamodb/pricing/on-demand/)) indicates that each Table should cost almost nothing per-month
  - All items put to these Tables have explicit [TTLs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html) such that they expire 7 days after their corresponding RIF or data load was made available in S3. This ensures the Tables don't become any larger than they need to be, as CloudWatch Metrics is the long-term data storage for this situation
  - CloudWatch Metrics does not guarantee that data for a given metric is available to be retrieved within 10 minutes of the data's submission. So, if a RIF finished within 10 minutes (which was often the case for `DME`), load duration metrics would not be submitted for the `DME` RIF and the Lambda would error-out
- Updates the Pipeline SLIs Lambda to continue if it is unable to calculate time deltas (or durations to load) for individual RIFs or the load itself
  - This resolves the issue where, if one of these metrics could not be calculated, the Pipeline SLIs Lambda would exit immediately and _not_ calculate "fully loaded" metrics that are used for alerting. This would cause the load to erroenously be considered unfinished indefinitely by the alerting mechanisms 
- Updates the Pipeline SLIs Lambda to use [Powertools for AWS Lambda](https://docs.powertools.aws.dev/lambda/python/latest/), a Library developed by AWS that improves the ergonomics of developing Lambdas in Python
  - The Lambda has been updated to use the Lambda Layer provided by Powertools
- Updates the precondition checks using Powertools to be more robust and simpler
- Removes all explicit `Exception` handling in the Lambda's handler, allowing the Lambda Runtime to receive those `Exception`s. Now, failing invocations are now properly represented in the CloudWatch Metrics for the Lambda
  - This would, in normal circumstances, cause AWS Lambda to retry the Lambda up to two times. This Lambda has side effects, and we don't want those side effects to occur more than once if the Lambda fails halfway through. So, the Lambda's [asynchronous invocation retry count](https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html) has been reduced to 0.
    - This doesn't necessarily cover non-asynchronous invocations. Fortunately, this Lambda is _always_ invoked asynchronously, but if we wanted to be _real_ sure that side effects only occur once we could look into applying Powertools' [idempotency utilities](https://docs.powertools.aws.dev/lambda/python/latest/utilities/idempotency/) to the functions with side effects. This would allow the Lambda to retry as many times as necessary _without_ concern for additional, unnecessary side effects
  - This is a new paradigm, as _most_ of our bespoke Lambdas swallow all `Exception`s in order to avoid AWS Lambda's default retry behavior (as most of our Lambdas have side effects). Provided that this new solution proves to work correctly, **we should investigate updating our other Lambdas to _not_ swallow `Exception`s**. We would then have more accurate metrics on our Lambdas
- Adds two unit test suites (this Lambda has had many problems in the past, thus necessitating more thorough, automated testing):
  - A test suite for the Pipeline SLIs Lambda's handler testing various valid and invalid scenarios
    - This unit test suite _only_ tests the `handler()` function, and is non-comprehensive. It _is_ sufficient to ensure that the Lambda's `handler()` is attempting to submit the right metrics for "happy path" cases, and that _most_ error scenarios are tested. But, not all code paths are tested completely
  - A test suite testing new helper functions related to DynamoDB
- `dev-requirements` have been added to aid developers working on this particular Lambda in creating a Python environment with the appropriate type hinting for `boto3`

This PR has been verified by:

- Running the new unit test suites, verifying that all tests pass
- `terraform apply`ing the updated Lambda to `test`, verifying that it `apply`s successfully and that there were no unexpected changes
- Invoking the aforementioned Lambda with a mock event indicating a RIF was made available in `Incoming`, verifying that the Lambda posted the correct Metrics, that the correct DynamoDB Tables were updated, and that there were no errors
- Invoking the aforementioned Lambda with a mock event indicating a RIF was made available in `Done`, verifying that the Lambda posted the correct Metrics and that there were no errors
- Invoking the aforementioned Lambda with an _invalid_ mock event, verifying that the underlying `Exception` was logged along with its stack trace and that the Lambda was _not_ retried

    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

    - [ ] Yes
    - [x] No

- N/A

### Submitter Checklist


ops/terraform/services/pipeline/modules/bfd_pipeline_slis/lambda_src/update_pipeline_slis.py
malessi, line 1: This file has been refactored extensively, i.e.:

- all `print(...)` usages have been replaced with appropriate `Logger` functions
- multiple cases of unnecessary nesting have been refactored by inverting the corresponding `if` and using guard pattern, where appropriate
   - this refactoring is largely the reason why the diff is so large
- logging statements that were originally in `try-raise` alongside the `backoff_retry` operation they were logging were extracted out of the `try-raise` blocks so that it's more obvious exceptions are being swallowed for the `backoff_retry`
- obviously, formatting was ran _after_ the above changes which further exacerbates the problem

malessi: When I say the _logic_ hasn't changed, I mean that the Lambda still roughly does the same thing it's done since it was written:

1. Lambda is invoked from an S3 event from a file landing in either the `Incoming` or `Done` folders of the Pipeline bucket
2. Lambda inspects the event:

   1. If the event indicates the file was a valid RIF:

      1. Metrics with the value of the event's time, as a unix timestamp, are posted to the `time/data-available` _if_ the event indicates the RIF was added to `Incoming` _or_ `time/data-loaded` _if_ the event indicates the RIF was added to `Done`
      2. Then, if the RIF was added to `Incoming`:
         1. A state store (once CloudWatch Metrics, now DynamoDB) is checked for any events generated by previous invocations of this Lambda indicating when/if the current data load (represented by its ISO-8601 timestamp) had already been "made available" (as in, the load this RIF is a part of has already started)
            1. If an event is returned for the current data load, the Lambda exits as there's nothing more to do
            2. Else, if there was no such event, this RIF is the first RIF of a new data load and the Lambda then:
               1. Generates a new event indicating when this new data load first appeared, and submits it to the state store (DynamoDB)
               2. Generates `time/data-first-available` metrics with the value of the event's time as a unix timestamp and submits those metrics to CloudWatch Metrics
               3. The Lambda then exits
      3. Or, if the RIF was added to `Done`:
         1. A state store (once CloudWatch Metrics, now DynamoDB) is checked for any events generated by previous invocations of this Lambda indicating when/if the current RIF had already been "made available"
            1. If an event is returned indicating when the RIF was first made available:
               1. A "time delta" is computed by taking the delta of time between the RIF's `Incoming` S3 event and its current `Done` event -- this is the time it took to load the RIF
               2. This "time delta", or duration-to-load, is converted to seconds and submitted to a CloudWatch Metric named `time-delta/data-load-time`
            2. Else, the Lambda logs a warning and continues
         2. Then, the Lambda executes the following checks (which, in short, check if the load is complete), and only continues if _both_ pass:
            1. The Lambda looks at the `Done` folder for the current load and checks to see if _all_ **required** RIFs exist within the `Done` folder. If they do, this check passes, and if they don't (indicating the load _isn't_ finished) it fails
            2. The Lambda looks at the `Incoming` folder for the current load and checks if it's empty. If it is this check passes, and otherwise fails
         3. Assuming the aforementioned checks pass:
            1. The Lambda takes the event time from its invocation time, converts it to a unix timestamp, and posts it to CloudWatch Metrics under the `time/data-fully-loaded` Metric
            1. The state store (in DynamoDB now) is once again checked for any events submitted by previous invocations of the Lambda for events corresponding to when the RIF's data load was first made available
               1. If an event is returned indicating when the _RIF_ was first made available:
                  1. A "time delta" is computed by taking the delta of time between the current `Done` S3 event and the event returned from the state store
                  2. This "time delta", or duration-to-load the _RIF_, is converted to seconds and submitted to a CloudWatch Metric named `time-delta/data-load-time`
               2. If no event is returned (this is _technically_ invalid, but _could_ happen), a warning is logged and the Lambda continues
            2. The state store is checked one final time for any events corresponding to when the _data load_ was first made available
               1. If an event is returned indicating when the _data load_ was first made available:
                  1. A "time delta" is computed by taking the delta of time between the current `Done` S3 event and the event returned from the state store
                  2. This "time delta", or duration-to-load the _entire data load_, is converted to seconds and submitted to a CloudWatch Metric named `time-delta/data-full-load-time`
               2. If no event is returned (this is _technically_ invalid, but _could_ happen), a warning is logged and the Lambda continues

   2. Else, the Lambda exits

Overall, the major change to _function_ is the change from querying CloudWatch Metrics when generating duration-to-load metrics to querying DynamoDB. As mentioned in the summary, this is because CloudWatch Metrics do not _guarantee_ Metrics are available to be retrieved within 10 minutes of their submission. So, fast loading RIFs (like `DME`) would often return no data for when they were first made available using CloudWatch Metrics.

mjburling (approved):

billie-rose (approved):

PR #2301, opened 2024-04-23
jimfuqian/BFD3364 Docker compose based dev local bfd server setup

JFU-NAVA-PBC:
**JIRA Ticket:**
[BFD-3364](https://jira.cms.gov/browse/BFD-3364)

As a BFD developer, I want an easy way to fire up a local BFD server and perform development work with it in order to simply the Developer experience.

Preferably, A docker compose based solution that only need one CLI:

e.g. docker compose up bfd

Benefits:

- Native OS agnostic, same CLI on all OS machines
- No need to install java, maven, etc. on native OS
- Ready to debug after 2-3 CLIs

---

### What Does This PR Do?

Added docker artifacts docker compose yaml and docker file etc. to setup a local dev bfd server and etc.

Check out the PR and try out the docker compose per instructions and examples in:

./apps/readme-dev.md

And get bfd up anbd running and use your IDEs to do the dev tasks.

Note,

Tested out on Mac, Linux (Ubuntu), Windows (WSL2) 

      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No
      * [ ] Yes
      * [x] No

      * [ ] Yes
      * [x] No

* N/A

### Submitter Checklist

* [ ] The data dictionary has been updated with any field mapping changes, if any were made.
* [ ] All new and modified code is appropriately commented, such that the what and why of its design would be reasonably clear to engineers, preferably ones unfamiliar with the project.
* [ ] Any deviations from the other policies in the [DASG Engineering Standards](https://github.com/CMSgov/cms-oeda-dasg/blob/master/policies/engineering_standards.md) are specifically called out in this PR, above.


apps/Dockerfile.local
```diff
@@ -0,0 +1,61 @@
+ARG JAVA_VERSION=21
+ARG MAVEN_VERSION=3.9.6
+
+FROM maven:${MAVEN_VERSION}-amazoncorretto-${JAVA_VERSION}
```
aschey-forpeople, line 4: Would it be possible to hook into the configuration we have for jib instead of creating a separate Dockerfile here?

aschey-forpeople: Yeah okay, looks like jib is mostly just used to create an image from the existing build process. I thought it was a bit more flexible. Instead of having one `Dockerfile.local`, would it make sense to have a separate one for each application and docker-compose could orchestrate them? That would make it a little clearer how the different services interact with each other and you could view the log output separately per service.

JFU-NAVA-PBC: > Yeah okay, looks like jib is mostly just used to create an image from the existing build process. I thought it was a bit more flexible. Instead of having one `Dockerfile.local`, would it make sense to have a separate one for each application and docker-compose could orchestrate them? That would make it a little clearer how the different services interact with each other and you could view the log output separately per service.

Could you elaborate a bit on "application" here referring to? are they just the bfd modules under ./apps/?
if we want to filter out the logging events from each bfd components, we probably have to go into the logged records contents e.g. to grep a json record's certain property...?

For services (e.g. API end points exposed on port 6500), there are only 2 services defined in the docker compose, bfd, and db.

PR #2391, opened 2024-07-29
BFD-3550: Add cross-reference ID and cross-reference switch to CCW pipeline

aschey-forpeople:
**JIRA Ticket:**
BFD-3550

### What Does This PR Do?

- Adds the `xref_grp_id` and `xref_sw` columns to the database and the CCW pipeline
- For now, we will need to support both files with and without these columns, so the test changes here are mostly additive to assure we can still support both scenarios. Once we only need to support files with the xref columns included, we can see about consolidating these extra tests and test files.
- One bug fix is included here which is to check if a CSV field contains a header before trying to get the value - this prevents an exception from being thrown
- Some test setup logic was based on the assumption that the CSV header and the order of the fields in the YAML RIF schemas should match. Places that make this assumption have been fixed so this is no longer the case.
- The `withHeader` method for creating CSVs has been throwing deprecation warnings for a while. I replaced all instances of these with the builder pattern that they now want you to use.

### Validation

Created an ephemeral environment from test, ran database migrations, and re-ran an old pipeline load that contained a beneficiary file.


apps/bfd-model/bfd-model-rif-samples/src/test/java/gov/cms/bfd/model/rif/samples/SampleDataColumnsTest.java
```diff
@@ -42,8 +44,8 @@ public void logEnumColumns() {
    * gov.cms.bfd.model.rif.samples.StaticRifResourceGroup#SAMPLE_A} data file headers.
    */
   @Test
-  public void verifySampleAColumns() {
-    verifyColumns(StaticRifResourceGroup.SAMPLE_A);
+  public void verifySampleBColumns() {
+    verifyColumns(StaticRifResourceGroup.SAMPLE_B);
```
aschey-forpeople, line 48: I left the sample A files alone for now so we don't mess with the existing tests. Sample B contains the new xref columns which we need to use for this test since we're checking that all columns in the file are in the schema.

apps/bfd-model/bfd-model-rif-samples/src/test/java/gov/cms/bfd/model/rif/samples/SampleDataColumnsTest.java
```diff
@@ -56,20 +58,17 @@ private void verifyColumns(StaticRifResourceGroup sampleGroup) {
       Enum<?>[] columnsInEnum = getColumnsInEnum(sampleFile.getRifFileType());
 
       // Use a CSVParser to parse the header out of the sample file.
-      CSVFormat parserFormat = CSVFormat.DEFAULT.withDelimiter('|');
+      CSVFormat parserFormat = CSVFormat.DEFAULT.builder().setDelimiter('|').build();
```
aschey-forpeople, line 61: All the changes to `.builder()` are to remove deprecation warnings on `withDelimiter`

apps/bfd-model/bfd-model-rif-samples/src/test/java/gov/cms/bfd/model/rif/samples/SampleDataColumnsTest.java
```diff
@@ -86,18 +85,18 @@ private void verifyColumns(StaticRifResourceGroup sampleGroup) {
...
-            columnNameFromSample,
-            columnNameFromEnum,
-            String.format(
-                "Unable to match column '%d' from sample data for '%s'.\nSample Columns: %s\nEnum Columns:   %s\n",
-                col,
-                sampleFile.name(),
-                toHeaderFormat(columnsInSample, c -> c),
-                toHeaderFormat(columnsInEnum, c -> c.name())));
+
+      HashSet<String> sampleSet = new HashSet<>(Arrays.asList(columnsInSample));
```
aschey-forpeople, line 89: Changes here are to prevent the order from mattering. Instead of checking that the columns have the same index, we only care that all columns are included.

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -493,7 +493,7 @@ private void tweakIfBeneficiary(
        * state as a BeneficiaryHistory record. (Note: this has to be done AFTER the secret hashes
        * have been updated, as per above.
        */
-      updateBeneficaryHistory(
+      updateBeneficiaryHistory(
```
aschey-forpeople, line 496: This is just fixing a typo

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoader.java
```diff
@@ -869,6 +871,19 @@ static boolean isBeneficiaryHistoryEqual(
         newBeneficiaryRecord.getMbiEffectiveDate(), oldBeneficiaryRecord.getMbiEffectiveDate())) {
       return false;
     }
+
+    if (!Objects.equals(
```
aschey-forpeople, line 875: This will instruct the pipeline to add a new history record if the xref id or switch has changed

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/test/java/gov/cms/bfd/pipeline/ccw/rif/load/RifLoaderIT.java
```diff
@@ -1149,14 +1215,10 @@ private RifFile editSampleRecords(
...
-      // Build a CSVFormat with the specific header needed for the RIF file type.
-      String[] csvHeader =
-          Stream.concat(
-                  Stream.of("DML_IND"),
-                  Arrays.stream(rifFileEvent.getFile().getFileType().getColumns())
-                      .map(e -> e.name()))
-              .toArray(String[]::new);
-      CSVFormat csvFormat = RifParsingUtils.CSV_FORMAT.withHeader(csvHeader);
+      // Get the header from the original file
+      List<String> csvHeader = RifParsingUtils.createCsvParser(inputFile).getHeaderNames();
```
aschey-forpeople, line 1219: Instead of trying to create a new header with everything in the schema, we instead pull the header from the original file since the order could be different.

apps/bfd-model/bfd-model-rif/src/main/resources/db/scripts/clear_ccw_pipeline_load.sql
```diff
@@ -0,0 +1,128 @@
+/*
```
aschey-forpeople, line 1: Created a runbook that explains how to use this script for re-running a pipeline load: https://github.com/CMSgov/beneficiary-fhir-data/wiki/How-to-clean%E2%80%90up-and-reload-test-data-from-a-CCW-pipeline-load

aschey-forpeople:
Closing since this effort has been abandoned.

PR #2393, opened 2024-08-01
BFD-3508: Introduce Automatic Scaling of RDS Reader Nodes

malessi:
**JIRA Ticket:**
BFD-3508

### What Does This PR Do?

**NOTE: These changes modify the names of some Terraform resource definitions in `common`, specifically existing application autoscaling resources and the `aws_rds_cluster_instance.nodes` resource. We will need to move these resources to their new names prior to deployment in all environments**

This PR implements RDS application autoscaling which allows us to automatically scale-in and scale-out read replicas in our Aurora clusters.

More specifically:

- Application autoscaling targets and policies have been defined in `common/app-autoscaling.tf` to enable CPU-based automatic scaling of RDS reader nodes. Corresponding SSM parameters have been introduced to control the target CPU utilization (set to 75% average across all nodes in all environments), the scale-out and scale-in cooldowns (300 seconds in all environments), and the min and max capacity (differs between environments, see changes to environment `.yaml`s)
- The Terraform that manages the nodes in the Aurora cluster has been refactored such that Terraform only creates the _writer_ node. If reader nodes are desired in a given environment, the `/bfd/*/common/nonsensitive/rds_min_reader_nodes` SSM parameter should be set to the minimum number of desired reader nodes and the application autoscaling policies will then create the readers automatically after the writer node is available
- The writer node in each Aurora cluster is now named `<CLUSTER_ID>-writer-node`. There is no means by which we can control the name of the read replicas created by application autoscaling, so there is no reason to continue with our existing naming scheme
- Existing scheduled scaling policies/actions in `test` have been extended to also cover ephemeral environments so that ephemeral environments will also scale-in after hours and scale-out during work hours. A new variable override has been added to `common`, `disable_rds_scheduling_override`, that allows the operator to disable this scheduled scaling behavior if there is a need
- A corresponding parameter that controls the aforementioned override has been added to the `bfd-deploy-apps` Jenkins pipeline for use with ephemeral environments
- A `destroy`-time [`local-exec` provisioner](https://developer.hashicorp.com/terraform/language/resources/provisioners/local-exec) has been added to the `aws_rds_cluster.aurora_cluster` resource that runs a newly-added bash script: `ops/terraform/services/common/scripts/destroy-autoscaled-nodes.sh`. This is necessary as Terraform does not manage autoscaled nodes and Aurora clusters cannot be deleted if they are not empty, so attempting to `terraform destroy` an ephemeral environment with autoscaled nodes would otherwise be impossible. This new script will enumerate all autoscaled reader nodes within the cluster, mark them for deletion, and wait indefinitely for them to be deleted before Terraform attempts to destroy the Aurora cluster
- Some Terraform resources have been renamed to better reflect their purpose
- Unused `prod` endpoints have been removed entirely (the "beta reader" and "readers" endpoints, specifically)

- Scrutinize the `ops/terraform/services/common/scripts/destroy-autoscaled-nodes.sh` for any possible logic error. I've added as many safeguards as I could possibly think of, but an earlier version of this script caused our most recent incident
- Does 75% CPU across all nodes make sense as a target? Does this need to be higher/lower? Do the values for the cooldowns also make sense?
- Does it make sense to do scheduled scale-in/scale-out in ephemeral environments?

- Adds any new software dependencies
- Modifies any security controls
- Adds new transmission or storage of data
- Any other changes that could possibly affect security?

### Validation

- `terraform apply`ing these changes to the `3508-test` ephemeral environment, _verifying that_:
  - It `apply`s successfully
  - No unexpected resources were created
  - Application autoscaling policies automatically create the minimum number of reader nodes specified by configuration
- Deploying `server` _immediately after_ `common` in `3508-test`, _verifying that_:
  - The BFD Server instances pass their healthchecks and are able to connect to the database, even if the autoscaled reader node(s) haven't yet scaled out
  - (Assuming BFD-3507 is in effect) the BFD Server automatically connects to the autoscaled reader node once it is available and all connections are drained from the writer
- Waiting until 7 PM ET after deployment of `3508-test` with an autoscaled reader node in the Aurora cluster, _verifying that_ the autoscaled node is scaled-in due to off hours scale-in scheduled actions
  - Correspondingly, the work hours scale out scheduled action successfully _scaled-out_ reader nodes
- `terraform destroy`ing `common` in `3508-test`, _verifying that_:
  - The `local-exec` provisioner on the cluster successfully deleted the autoscaled reader nodes without affecting other nodes in other clusters
  - All resources are destroyed successfully without error

### DIFF
--- a/ops/terraform/services/base/values/prod.yaml
+++ b/ops/terraform/services/base/values/prod.yaml
@@ -108,8 +108,12 @@
 /bfd/${env}/common/nonsensitive/rds_master_username: bfduser
 /bfd/${env}/common/nonsensitive/rds_cluster_identifier: bfd-prod-aurora-cluster
 /bfd/${env}/common/nonsensitive/rds_instance_class: db.r6i.12xlarge
-/bfd/${env}/common/nonsensitive/rds_instance_count: "3"
 /bfd/${env}/common/nonsensitive/rds_security_group: bfd-prod-aurora-cluster
+/bfd/${env}/common/nonsensitive/rds_min_reader_nodes: 1
+/bfd/${env}/common/nonsensitive/rds_max_reader_nodes: 3
+/bfd/${env}/common/nonsensitive/rds_scaling_cpu_target: 75
+/bfd/${env}/common/nonsensitive/rds_scale_in_cooldown: 300
+/bfd/${env}/common/nonsensitive/rds_scale_out_cooldown: 300
 /bfd/${env}/common/nonsensitive/vpc_name: bfd-prod-vpc
 /bfd/${env}/common/nonsensitive/vpn_security_group: bfd-prod-vpn-private
 /bfd/${env}/migrator/nonsensitive/ref_dir: /opt/bfd-db-migrator
--- a/ops/terraform/services/base/values/test.yaml
+++ b/ops/terraform/services/base/values/test.yaml
@@ -111,8 +111,12 @@
 /bfd/${env}/common/nonsensitive/rds_master_username: bfduser
 /bfd/${env}/common/nonsensitive/rds_cluster_identifier: bfd-test-aurora-cluster
 /bfd/${env}/common/nonsensitive/rds_instance_class: db.r6i.2xlarge
-/bfd/${env}/common/nonsensitive/rds_instance_count: "1"
 /bfd/${env}/common/nonsensitive/rds_security_group: bfd-test-aurora-cluster
+/bfd/${env}/common/nonsensitive/rds_min_reader_nodes: 0
+/bfd/${env}/common/nonsensitive/rds_max_reader_nodes: 0
+/bfd/${env}/common/nonsensitive/rds_scaling_cpu_target: 75
+/bfd/${env}/common/nonsensitive/rds_scale_in_cooldown: 300
+/bfd/${env}/common/nonsensitive/rds_scale_out_cooldown: 300
 /bfd/${env}/common/nonsensitive/vpc_name: bfd-test-vpc
 /bfd/${env}/common/nonsensitive/vpn_security_group: bfd-test-vpn-private
 /bfd/${env}/migrator/nonsensitive/ref_dir: /opt/bfd-db-migrator
--- a/ops/terraform/services/common/app-autoscaling.tf
+++ b/ops/terraform/services/common/app-autoscaling.tf
@@ -1,6 +1,6 @@
-resource "aws_appautoscaling_target" "replicas" {
-  # only applicable for test environment
-  count = local.env == "test" ? 1 : 0
+resource "aws_appautoscaling_target" "dynamic_replicas" {
+  # only applicable for test environment or ephemeral environments
+  count = local.env == "test" || local.is_ephemeral_env ? 1 : 0
 
   service_namespace  = "rds"
   scalable_dimension = "rds:cluster:ReadReplicaCount"
@@ -9,8 +9,8 @@ resource "aws_appautoscaling_target" "replicas" {
   # NOTE: These are the initial, additional capacity settings for the target. Scaling events by schedule and
   # policy directly update these values on the app autoscaling target. As a result, changes to either
   # min_capacity and max_capacity fields are ignored by the lifecycle meta argument below.
-  min_capacity = 0
-  max_capacity = 0
+  min_capacity = local.rds_min_reader_nodes
+  max_capacity = local.rds_max_reader_nodes
 
   lifecycle {
     ignore_changes = [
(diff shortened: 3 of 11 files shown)


ops/terraform/services/base/values/prod.yaml
```diff
@@ -108,8 +108,12 @@
 /bfd/${env}/common/nonsensitive/rds_master_username: bfduser
 /bfd/${env}/common/nonsensitive/rds_cluster_identifier: bfd-prod-aurora-cluster
 /bfd/${env}/common/nonsensitive/rds_instance_class: db.r6i.12xlarge
-/bfd/${env}/common/nonsensitive/rds_instance_count: "3"
 /bfd/${env}/common/nonsensitive/rds_security_group: bfd-prod-aurora-cluster
+/bfd/${env}/common/nonsensitive/rds_min_reader_nodes: 1
+/bfd/${env}/common/nonsensitive/rds_max_reader_nodes: 3
+/bfd/${env}/common/nonsensitive/rds_scaling_cpu_target: 75
+/bfd/${env}/common/nonsensitive/rds_scale_in_cooldown: 300
+/bfd/${env}/common/nonsensitive/rds_scale_out_cooldown: 300
```
malessi, line 116: Curious what others' opinions are on these values. These are the _default_ values for CPU-based application autoscaling, and they work _fine_ when testing against an ephemeral database cluster. But, I'm not sure if these values are _too_ conservative or not.

I certainly think we want our "scale in cooldown" to remain high just the databases don't scale-in unnecessarily, but perhaps we could do with a lower value for the "scale out cooldown"? As it stands, it takes around 15 minutes between each scale-out event, which is a _bit_ long.

Thoughts?

ops/terraform/services/base/values/test.yaml
```diff
@@ -111,8 +111,12 @@
 /bfd/${env}/common/nonsensitive/rds_master_username: bfduser
 /bfd/${env}/common/nonsensitive/rds_cluster_identifier: bfd-test-aurora-cluster
 /bfd/${env}/common/nonsensitive/rds_instance_class: db.r6i.2xlarge
-/bfd/${env}/common/nonsensitive/rds_instance_count: "1"
 /bfd/${env}/common/nonsensitive/rds_security_group: bfd-test-aurora-cluster
+/bfd/${env}/common/nonsensitive/rds_min_reader_nodes: 0
+/bfd/${env}/common/nonsensitive/rds_max_reader_nodes: 0
```
malessi, line 116: I _think_ the existing scheduled scaling in `test` is _supposed_ to give us a single reader node, but scheduled action has been suspended on the `test` cluster so we've been running with a single writer for some time. Given that fact, I've disabled reader nodes in `test`. Should we instead have a single reader node (we set the min and max to `1`)?

ops/terraform/services/base/values/prod.yaml
mjburling: I used a 50% threshold as part of the analysis earlier this year to approximate the potential savings in this strategy. I think you could reduce that threshold to 50% and we can go ahead an evaluate this in time.

ops/terraform/services/base/values/test.yaml
mjburling: I think this is fine. With the pre-purchase agreement going into effect in June, without **this** work... I went ahead and disabled the scale-out/scale-in schedules.

ops/terraform/services/common/scripts/destroy-autoscaled-nodes.sh
```diff
@@ -0,0 +1,93 @@
...
+set -euo pipefail
+
+# Be _extra_ careful about the cluster ID given by the $DB_CLUSTER_ID environment variable.
+# Specifically, remove all whitespace (which is invalid for a cluster ID) and ensure that the
+# resulting, trimmed ID is non-empty before even making any deletion attempts as the AWS CLI will
+# treat an empty cluster ID as equivalent to "give me all clusters" when calling
+# describe-db-clusters
+trimmed_cluster_id="$(echo "$DB_CLUSTER_ID" | tr -d '[:space:]')"
+if [[ -z $trimmed_cluster_id ]]; then
+  echo "DB_CLUSTER_ID must not be an empty string or whitespace"
```
aschey-forpeople, line 16: As discussed, we should try to add some check to make sure this is an ephemeral environment

ops/terraform/services/base/values/prod.yaml
malessi: Agreed. I think we can be more aggressive in when we scale out because we're simply not scaling _at all_ just yet. I'll change it to 50% in all environments.

malessi: Done in commit [0f7a9b3](https://github.com/CMSgov/beneficiary-fhir-data/pull/2393/commits/0f7a9b358f4817da07c8c94ba35024a5744a2c52).

ops/terraform/services/common/scripts/destroy-autoscaled-nodes.sh
malessi: Agreed. I'll add some explicit checks on the environment here to ensure that we can only run this script within an ephemeral environment.

malessi: Done in [b2c98bc](https://github.com/CMSgov/beneficiary-fhir-data/pull/2393/commits/b2c98bc5a9dd23feb8a62270ebfd604af436f4e8) and [8d870a5](https://github.com/CMSgov/beneficiary-fhir-data/pull/2393/commits/8d870a5aecccc4af1f55b64ef826a5014863421d).

ops/terraform/services/common/app-autoscaling.tf
```diff
@@ -1,6 +1,15 @@
-resource "aws_appautoscaling_target" "replicas" {
-  # only applicable for test environment
-  count = local.env == "test" ? 1 : 0
+locals {
+  enable_rds_scheduled_scaling = !var.disable_rds_scheduling_override && (local.env == "test" || local.is_ephemeral_env)
+  replicas_scaling_target      = local.enable_rds_scheduled_scaling ? one(aws_appautoscaling_target.dynamic_replicas) : one(aws_appautoscaling_target.static_replicas)
+}
+
+resource "aws_appautoscaling_target" "dynamic_replicas" {
```
malessi, line 6: Note that we _do_ need two variants of this `aws_appautoscaling_target` resource specifically because `lifecycle` meta-arguments only support [literal values](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#literal-values-only). So, we can't do something like:

```terraform
lifecycle {
  ignore_changes = local.enable_rds_scheduled_scaling ? [
    min_capacity,
    max_capacity
  ] : []
}
```

This is because the ternary cannot be evaluated at the time where the `lifecycle` meta-argument is evaluated, so Terrraform doesn't allow it. Instead, we have two variants of this resource: `dynamic_replicas` for when scheduled scaling of RDS nodes is enabled and `static_replicas` for when it isn't. Which variant is chosen and used by dependent resources depends on the `local.enable_rds_scheduled_scaling` (see the definition for `local.replicas_scaling_target`).

timothy-ellis-ky:
there's a lot of good stuff in here, @malessi.
I haven't made it all the way through to submit a legitimate review, but I did want to toss this out for consideration: in looking for an auto-scaling read replica example, I came across this public module source that does not leverage any helper scripts - https://github.com/terraform-aws-modules/terraform-aws-rds-aurora - not sure if you have seen it.

ops/terraform/services/common/scripts/destroy-autoscaled-nodes.sh
```diff
@@ -0,0 +1,111 @@
...
+  exit 1
+fi
+
+# jq expression returns the first element in the returned array as a precaution against a scenario
+# where multiple clusters are erroneously returned. Otherwise, the nodes in all clusters would be
+# marked for deletion. Additionally, we select for only the nodes with names containing
+# "autoscaling" to avoid possibly deleting the writer node which is managed by Terraform directly
+nodes_in_cluster="$(
+  aws rds describe-db-clusters --db-cluster-identifier "$trimmed_cluster_id" |
```
timothy-ellis-ky, line 43: I realize we have a number of defaults that likely cover this item, but would it be prudent to include "--output json" on all of the CLI commands as a safety precaution?

timothy-ellis-ky (commented):
see commentary

ops/terraform/services/common/scripts/destroy-autoscaled-nodes.sh
```diff
@@ -0,0 +1,111 @@
...
+      ((deleting_nodes += 1))
+    fi
+  done
+
+  if ((deleting_nodes == 0)); then
+    echo "All autoscaling nodes in $trimmed_cluster_id have been deleted"
+    exit
+  fi
+
+  echo "$deleting_nodes autoscaling node(s) have not yet been deleted. Sleeping for 5 seconds before retrying"
```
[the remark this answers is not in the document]
malessi: The provisioner isn't tied to Terraform's update interval, so once it's invoked Terraform logs any output immediately and doesn't interfere with the execution of the script. In short, this logic is entirely unaffected by Terraform's internal check interval.

I could certainly increase the interval to 10 seconds, but I think the only thing that'd accomplish is to just increase the time (by a few seconds) it takes to `terraform destroy` ephemeral environments with autoscaled nodes. I'm fine with that, but it would not be necessary for the `provisioner` to work properly.

ops/terraform/services/common/scripts/destroy-autoscaled-nodes.sh
malessi: This is definitely reasonable; being explicit is always a good idea. I'll make this change.

Related: I do think there are quite a few places where the AWS CLI is invoked and `--output json` is assumed without being explicitly specified, so maybe this is part of a wider discussion of whether we should trust that `--output json` is always the default wherever we run the AWS CLI? I think it's reasonable for us to continue to assume that, but maybe going forward we should always specify the output format explicitly.

malessi: Done in [e5f63dc](https://github.com/CMSgov/beneficiary-fhir-data/pull/2393/commits/e5f63dc9945afd6cd523ada185885a45a3e009b8).

ops/terraform/services/common/app-autoscaling.tf
```diff
@@ -20,37 +29,73 @@ resource "aws_appautoscaling_target" "replicas" {
...
+resource "aws_appautoscaling_scheduled_action" "work_hours_scale_out" {
+  depends_on = [aws_rds_cluster_instance.writer]
+
+  count = local.enable_rds_scheduled_scaling ? 1 : 0
+
+  name               = "bfd-${local.env}-work-hours-scale-out"
+  service_namespace  = local.replicas_scaling_target.service_namespace
+  resource_id        = local.replicas_scaling_target.resource_id
+  scalable_dimension = local.replicas_scaling_target.scalable_dimension
   schedule           = "cron(00 07 ? * MON-FRI *)"
```
malessi, line 75: @aschey-forpeople brought up a good point during demo: for contributors working outside of ET there may be cases where this `cron` schedule (and the corresponding scale-in `cron`) _do_ overlap with 9-5 working hours. Should we consider modifying these schedules to make them a _bit_ more flexible? Perhaps something like 6 AM ET for scale-out (this scheduled action) and 8 PM ET for scale-in?

malessi: On the other hand, I think in the (probably _very_) exceptional case where testing needs to happen against an ephemeral cluster _with_ autoscaling enabled _outside_ of the existing 7 AM - 7 PM ET window could be served by simply enabling the `disable_rds_scheduling_override`.

aschey-forpeople: Yeah, that's fine. Considering this won't be something we need to mess with frequently, redeploying `common` with that parameter seems perfectly acceptable.

malessi:
> there's a lot of good stuff in here, @malessi. I haven't made it all the way through to submit a legitimate review, but I did want to toss this out for consideration: in looking for an auto-scaling read replica example, I came across this public module source that does not leverage any helper scripts - https://github.com/terraform-aws-modules/terraform-aws-rds-aurora - not sure if you have seen it.

Historically we've avoided using such modules for a few reasons:

1. The Terraform service/module we're writing is _not_ complex enough to merit introducing a third-party module; essentially, introducing the module would be more work than simply using the `aws` provider's base `resource`s
2. The service/module is doing something special that isn't well served by the third-party module that aims to cover the more general case, so we'd have to otherwise define base-`aws` provider `resource` anyways
3. The third-party module abstracts away some "complexity" that we actually _want_ to be very explicit about

I believe 2 and 3 are most relevant for `common` here; we want to be very explicit about our cluster definition, and it appears we would need to still define _most_ (if not all) of `app-autoscaling.tf` ourselves due to the special scheduled scaling behavior in `test` and ephemeral environments.

I could totally be wrong here, though. I haven't spent enough time researching to really definitively say whether or not we'd be better served deferring to third-party modules in our Terraform code. Honestly, I'm also a fan of being explicit when it comes to infrastructure as code, even at the cost of increased verbosity.

timothy-ellis-ky:
no arguments on implementing the aforementioned third-party module, simply noting that it offers options that may not require the scale-down script.

timothy-ellis-ky (approved):

PR #2395, opened 2024-08-01
BFD-3531: Updates coverage resource with more fields for c4dic implementation

meliGuzman:
**JIRA Ticket:**
BFD-3531

### What Does This PR Do?

- Adds extensions for the color information
- Changes the identifier field to point to MBI instead of bene_id
- If the C4DIC profile is specified in a coverage request, we separates Part X plans into different Coverage resources. If no profile is specified we default to current behavior of the C4BB profile
- Sets the status field to "active"
- Adds the contact info under the telecom field
- Sets the subscriber field the same as the beneficiary field
- Updated logic for coverage ID depending on whether or not C4DIC is enabled
- Fixed some tests in CoverageTransformerV2 that were using a problematic class variable

- Run BFD server locally and run these curl requests to verify that you can see the fields populated on the coverage and contained organization resource as instructed below

### Validation

I ran the BFD server locally and ran these curl requests to verify that I can see the fields populated on the coverage and contained organization resources

1. Feature Flag ENABLED & C4DIC profile specified --> should see 4 coverage resources with the C4DIC fields basically duplicated

curl --silent --insecure --cert $BFD_PATH/apps/bfd-server/dev/ssl-stores/client-unsecured.pem "https://localhost:6500/v2/fhir/Coverage?beneficiary=567834&_profile=http://hl7.org/fhir/us/insurance-card/StructureDefinition/C4DIC-Coverage&_format=json"

2. Feature Flag ENABLED & NO profile specified --> should see only 4 coverage resources for the Part X plans. The C4DIC coverage fields will not be shown.

curl --silent --insecure --cert $BFD_PATH/apps/bfd-server/dev/ssl-stores/client-unsecured.pem "https://localhost:6500/v2/fhir/Coverage?beneficiary=567834&_format=json"

3. Feature Flag ENABLED & C4BB profile specified --> should see only 4 coverage resources for the Part X plans. The C4DIC coverage fields will not be shown. Same as `#2` (above).

curl --silent --insecure --cert $BFD_PATH/apps/bfd-server/dev/ssl-stores/client-unsecured.pem "https://localhost:6500/v2/fhir/Coverage?beneficiary=567834&_profile=http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-Coverage&_format=json"


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -122,11 +125,33 @@ private Coverage transformC4Dic(Beneficiary beneficiary) {
...
     // update Coverage.meta.lastUpdated
     TransformerUtilsV2.setLastUpdated(coverage, beneficiary.getLastUpdated());
 
+    TransformerUtilsV2.setPeriodStart(
+        coverage.getPeriod(), beneficiary.getMedicareCoverageStartDate());
+    coverage.setStatus(CoverageStatus.ACTIVE);
+    coverage.setSubscriber(TransformerUtilsV2.referencePatient(beneficiary));
+
+    List<Extension> colorExtensions = new ArrayList<>();
+    TransformerUtilsV2.addExtension(
```
aschey-forpeople, line 141: Taking a look at the example JSON, theirs has a few differences: https://hl7.org/fhir/us/insurance-card/STU1.1/Coverage-Example-Coverage1.json.html. Might be worth taking a look and seeing if ours should have the same structure.

```json
{
      "extension": [
        {
          "url": "http://hl7.org/fhir/us/insurance-card/StructureDefinition/C4DIC-BackgroundColor-extension",
          "valueCodeableConcept": {
            "coding": [
              {
                "system": "http://terminology.hl7.org/CodeSystem/IECColourManagement",
                "code": "#00bfff"
              }
            ]
          }
        },
        {
          "url": "http://hl7.org/fhir/us/insurance-card/StructureDefinition/C4DIC-ForegroundColor-extension",
          "valueCodeableConcept": {
            "coding": [
              {
                "system": "http://terminology.hl7.org/CodeSystem/IECColourManagement",
                "code": "#000000"
              }
            ]
          }
        },
        {
          "url": "http://hl7.org/fhir/us/insurance-card/StructureDefinition/C4DIC-HighlightColor-extension",
          "valueCodeableConcept": {
            "coding": [
              {
                "system": "http://terminology.hl7.org/CodeSystem/IECColourManagement",
                "code": "#ceebf5"
              }
            ]
          }
        }
      ],
      "url": "http://hl7.org/fhir/us/insurance-card/StructureDefinition/C4DIC-ColorPalette-extension"
    },
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -892,12 +947,50 @@ private void addC4DicIdentifier(Coverage coverage, Beneficiary beneficiary) {
...
+          Optional.empty());
+    }
+    if (beneficiary.getPartBCoverageStartDate().isPresent()) {
+      createCoverageClass(
+          coverage,
+          CoverageClass.PLAN,
+          TransformerConstants.COVERAGE_PLAN_PART_B,
+          Optional.empty());
+    }
+    if (beneficiary.getHmoCoverageCount().isPresent()
```
aschey-forpeople, line 977: Good find 👍. I was afraid we'd have to check each of the HMO monthly indicator columns.

aschey-forpeople: One potential question though - it looks like this field is updated gradually as the year goes on. If you look for a bene with reference year = 2024 and HMO coverage count = 8, there are no results, so it looks like this field is a reflection of the current year up until the previous month. If we query this in January, will we get a value of 0 or 1?

aschey-forpeople: This might just be an edge case we have to live with, but it would be good to know more about the behavior here.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -868,10 +893,40 @@ private void addC4DicPayor(Coverage coverage) {
...
+
+    ContactPoint phoneContact =
+        TransformerUtilsV2.createContactPoint(
+            ContactPoint.ContactPointSystem.PHONE,
+            TransformerConstants.C4DIC_MEDICARE_SERVICE_PHONE_NUMBER,
+            null);
+
+    ContactPoint emailContact =
+        TransformerUtilsV2.createContactPoint(
+            ContactPoint.ContactPointSystem.EMAIL, TransformerConstants.C4DIC_MEDICARE_EMAIL, null);
```
aschey-forpeople, line 924: As far as I can tell, there isn't actually a publicly listed email address to contact on medicare.gov. Maybe it's better to use `ContactPointSystem.URL` here instead?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
meliGuzman: I like that two level hierarchy better! so something like `extension : [ {extension : [color1, color2, etc]}, {future subfield extension}]` I'll add that change in

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
meliGuzman: You're right! Good catch!

dondevun (approved):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -884,18 +952,99 @@ private void addC4DicIdentifier(Coverage coverage, Beneficiary beneficiary) {
...
+   * Determines if the beneficiary is subscribed to Part C. Due to a lag with ingestion, we check if
+   * last_updated > current month, we check if the previous month is present.
+   *
+   * @param beneficiary the value for {@link Beneficiary}
+   * @param month The month that the beneficiary record was last updated within the database.
+   * @return true/false if beneficiary has Part C
+   */
+  private boolean beneficiaryHasPartC(Beneficiary beneficiary, int month) {
+    return switch (month) {
+      case 1 -> beneficiary.getPartCContractNumberJanId().isPresent()
```
aschey-forpeople, line 1022: We may want to check prod and see if this is for sure the case, but since we're already checking against the last updated month here, I think we can omit the check for the previous month.

My initial idea was to check the current month and the last updated month, but I think just checking the last updated month should be sufficient unless there's a case where we get a load on the first week of the month that doesn't provide the contract IDs for some reason.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -884,18 +952,99 @@ private void addC4DicIdentifier(Coverage coverage, Beneficiary beneficiary) {
...
+    }
+    if (beneficiary.getPartBCoverageStartDate().isPresent()) {
+      createCoverageClass(
+          coverage,
+          CoverageClass.PLAN,
+          TransformerConstants.COVERAGE_PLAN_PART_B,
+          Optional.empty());
+    }
+    Optional<Instant> lastUpdatedOpt = beneficiary.getLastUpdated();
+    if (lastUpdatedOpt.isPresent()) {
```
aschey-forpeople, line 993: Is it possible for `last_updated` to be null if it's the first time the bene is loaded in? If so, we may want to add a fallback here to check the current month if we don't have a `last_updated` time.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
meliGuzman: @aschey-forpeople Hmm that's what i was thinking about as well but wasn't sure. If you/team think it's safe enough then I can remove the check on the previous month

aschey-forpeople: I think the way it's written here would cause it to say they still have coverage for the current month if we've already ingested a load that says they lost coverage, so I think this way is more likely to cause incorrect values.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2Test.java
```diff
@@ -108,7 +109,7 @@ public void setup() {
 
     Calendar calen = Calendar.getInstance();
     calen.set(2021, 3, 17);
-    beneficiary.setLastUpdated(calen.getTime().toInstant());
+    beneficiary.setLastUpdated(calen.getTime().toInstant()); //
```
aschey-forpeople, line 112:
```suggestion
    beneficiary.setLastUpdated(calen.getTime().toInstant());
```
Stray trailing comment here

aschey-forpeople (dismissed):

aschey-forpeople (dismissed):

MahiFentaye (dismissed):

alex-dzeda (changes requested):
It looks like this would fail profile validation on C4DIC due to the existence of 4 classes w/ Plan. It looks like the future of differentiating medicare types is going to be via the coverage.type field (see https://hl7.org/fhir/us/core/STU7/StructureDefinition-us-core-coverage.html ctrl + f medicare). Addressing the former is more critical than preparing for the future; so it could be that an A + B card has a plan w/ "Medicare Fee For Service," whereas C + D -> "Medicare Advantage w/ Drug" (or something more descriptive based upon PTC_PLAN_TYPE_CD elements?)

A general rule that will need to be enforced is that if the user has coverages that require different identifiers, they'll need a coverage resource for each type (eg parts A + B can be combined, but D will be a separate Coverage resource, unfortunately. I think a part C bene may luck into having a single insurance card if the member numbers end up being the same between C and D).

meliGuzman:
Hey @alex-dzeda and @aschey-forpeople, I addressed Alex's feedback and followed our decision from this thread! https://cmsgov.slack.com/archives/C010WDXAZFZ/p1723738336818769?thread_ts=1723662533.423649&cid=C010WDXAZFZ If the C4DIC profile is specified in a coverage request, we separates Part X plans into different Coverage resources. If no profile is specified we default to current behavior of the C4BB profile. Lmk if I missed anything!

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -86,77 +89,83 @@ public Coverage transform(MedicareSegment medicareSegment, Beneficiary beneficia
...
     Timer.Context timer = createTimerContext("part_a");
     Coverage coverage = new Coverage();
-    Profile profile = Profile.C4BB;
+    Profile profile = (enabledProfiles.contains(Profile.C4DIC)) ? Profile.C4DIC : Profile.C4BB;
 
     addProfile(coverage, profile);
-
-    coverage.setId(CommonTransformerUtils.buildCoverageId(MedicareSegment.PART_A, beneficiary));
+    if (profile.equals(Profile.C4DIC)) {
+      coverage.setId(
```
aschey-forpeople, line 153: I think we can reduce a bit of this if we keep the single `buildCoverageId` method and pass the profile in there to do the check.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -86,77 +89,83 @@ public Coverage transform(MedicareSegment medicareSegment, Beneficiary beneficia
...
+   * @param enabledProfiles the CARIN {@link Profile} to use
    * @return {@link MedicareSegment#PART_A} {@link Coverage} resource for the specified {@link
    *     Beneficiary}
    */
-  private Coverage transformPartA(Beneficiary beneficiary) {
+  private Coverage transformPartA(Beneficiary beneficiary, Set<Profile> enabledProfiles) {
     Timer.Context timer = createTimerContext("part_a");
     Coverage coverage = new Coverage();
-    Profile profile = Profile.C4BB;
+    Profile profile = (enabledProfiles.contains(Profile.C4DIC)) ? Profile.C4DIC : Profile.C4BB;
```
aschey-forpeople, line 149: Since we're treating this as an either/or, I think we just need to pass in a single profile rather than a `Set<Profile>`.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
meliGuzman: @aschey-forpeople Hmm but in that method it gets the url prefix from the MedicareSegment so `return new IdDt(
        COVERAGE_SIMPLE_CLASSNAME,
        String.format("%s-%d", medicareSegment.getUrlPrefix(), beneficiaryId));`

meliGuzman: Changed!

alex-dzeda (changes requested):
Please update the metadata endpoint to provide documentation on appropriate values for supporting _profile on the Coverage endpoint. Currently, there is no "documentation" mentioned, and I'm thinking it may be appropriate to note the acceptable values here, and have the display dependent upon feature flag, as well?
image

aschey-forpeople (approved):

PR #2402, opened 2024-08-16
BFD-3575: Support proper retry of BFD Server operations throwing FailoverSuccessExceptions

malessi:
**JIRA Ticket:**
BFD-3575

### What Does This PR Do?

This PR aims to resolve the intermittent `500`s caused by the AWS JDBC Wrapper throwing `FailoverSuccessSQLException`s upon successful failover during RDS scale-in/scale-out. These exceptions are intended to signal to the application that the operation should be retried, and so there is no mechanism within the JDBC Wrapper itself for automatically handling them. 

Instead, this PR introduces the [`spring-retry`](https://github.com/spring-projects/spring-retry) library to the `bfd-server-war` module which enables automatic retrying via the `@Retryable` annotation on methods within Spring Beans. All `@Search` and `@Read` methods within all `*Provider` classes in `bfd-server-war` have been annotated with a new composed annotation, `@RetryOnRDSFailover`, that composes the `@Retryable` annotation with some default values configured to automatically retry the method if any `FailoverSQLException`s are encountered.

- **Adds any new software dependencies**
  - This PR adds the `spring-retry` and `spring-aspects` libraries to `bfd-server-war`
  - This PR adds the `spring-test` library at the `test` scope to `bfd-server-war`

### Validation

- Running the new `SpringRetryUtilsTest` unit tests and the `RetryOnRDSFailoverIT` Spring integration tests, _verifying that_ they pass
- Modifying the `CommonQueries::findBeneficiary` method to throw a `RuntimeException` wrapping a `FailoverSuccessSqlException` when it is **first invoked**, then running the `PatientE2EBase::testReadWhenNoHistoricalMbisExpect200` E2E test, _verifying that_:
  - The corresponding, calling `PatientResourceProvider::read` method is successfully retried (verified via debugging)
  - The test passes after the retry
- Modifying the `CommonQueries::findBeneficiary` method to throw a `FailoverSuccessSqlException` when it is **first invoked**, then running the `PatientE2EBase::testReadWhenNoHistoricalMbisExpect200` E2E test, _verifying that_:
  - The corresponding, calling `PatientResourceProvider::read` method is successfully retried (verified via debugging)
  - The test passes after the retry
- Building a BFD Server AMI with these changes, deploying the BFD Server changes to ephemeral environment `3575-test` with the RDS cluster configured to scale from 1 to 3 instances, and then running a significant load test locally against the `3575-test` ELB configured with Cross-AZ load balancing, _verifying that_
  - No `500`s were returned during any scale-in or scale-out events, even when deleting two reader nodes during a period of significant load
  - There were no performance regressions 
  - **Conversely, running the same set of tests against a BFD Server without the proper retry policies in place resulted in 10 `500`s, in total, caused by RDS scale-in and scale-out**

### DIFF
--- a/apps/bfd-server/bfd-server-war/pom.xml
+++ b/apps/bfd-server/bfd-server-war/pom.xml
@@ -14,17 +14,6 @@
         The primary WAR for the BFD Server project. Provides a FHIR API.
     </description>
 
-    <repositories>
-        <repository>
-            <id>spring-releases</id>
-            <name>Spring Releases</name>
-            <url>https://repo.spring.io/release</url>
-            <snapshots>
-                <enabled>false</enabled>
-            </snapshots>
-        </repository>
-    </repositories>
-
     <dependencies>
         <dependency>
             <!-- BFD Shared Utilities include a wrapper around MDC for logging. -->
--- a/apps/bfd-server/bfd-server-war/pom.xml
+++ b/apps/bfd-server/bfd-server-war/pom.xml
@@ -261,6 +261,13 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <!-- We need to be able to retry our Provider entrypoint methods upon
+            failover of a node in our Aurora Clusters-->
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <!-- Mockito utilities for JUnit / Jupiter. -->
             <groupId>org.mockito</groupId>
             <artifactId>mockito-junit-jupiter</artifactId>
--- a/apps/bfd-server/bfd-server-war/pom.xml
+++ b/apps/bfd-server/bfd-server-war/pom.xml
@@ -152,6 +152,7 @@
             <artifactId>spring-data-jpa</artifactId>
         </dependency>
         <dependency>
+            <!-- Dependency for spring-retry-->
             <groupId>org.springframework</groupId>
             <artifactId>spring-aspects</artifactId>
         </dependency>
--- a/apps/bfd-server/bfd-server-war/pom.xml
+++ b/apps/bfd-server/bfd-server-war/pom.xml
@@ -14,6 +14,17 @@
         The primary WAR for the BFD Server project. Provides a FHIR API.
     </description>
 
+    <repositories>
+        <repository>
+            <id>spring-releases</id>
+            <name>Spring Releases</name>
+            <url>https://repo.spring.io/release</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
+    </repositories>
+
     <dependencies>
         <dependency>
             <!-- BFD Shared Utilities include a wrapper around MDC for logging. -->
@@ -150,6 +161,17 @@
                 mostly just need its Hibernate JPA glue. -->
             <groupId>org.springframework.data</groupId>
             <artifactId>spring-data-jpa</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-aspects</artifactId>
+        </dependency>
+        <dependency>
+            <!-- We need to be able to retry our Provider entrypoint methods upon
+            failover of a node in our Aurora Clusters-->
+            <groupId>org.springframework.retry</groupId>
+            <artifactId>spring-retry</artifactId>
+            <version>2.0.7</version>
         </dependency>
 
         <dependency>
--- a/apps/bfd-server/bfd-server-war/pom.xml
+++ b/apps/bfd-server/bfd-server-war/pom.xml
@@ -171,7 +171,7 @@
             failover of a node in our Aurora Clusters-->
             <groupId>org.springframework.retry</groupId>
             <artifactId>spring-retry</artifactId>
-            <version>2.0.7</version>
+            <version>${spring-retry.version}</version>
         </dependency>
 
         <dependency>
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
@@ -59,12 +59,14 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.orm.jpa.JpaTransactionManager;
 import org.springframework.orm.jpa.LocalContainerEntityManagerFactoryBean;
 import org.springframework.orm.jpa.support.PersistenceAnnotationBeanPostProcessor;
+import org.springframework.retry.annotation.EnableRetry;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 /** The main Spring {@link Configuration} for the Blue Button API Backend application. */
 @Configuration
 @ComponentScan(basePackageClasses = {ServerInitializer.class})
 @EnableScheduling
+@EnableRetry(proxyTargetClass = true)
 public class SpringConfiguration extends BaseConfiguration {
   /**
    * The {@link String } property that is used to hold drug code file name that is used for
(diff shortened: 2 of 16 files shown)


apps/bfd-server/bfd-server-war/pom.xml
```diff
@@ -14,6 +14,17 @@
...
+        <repository>
+            <id>spring-releases</id>
+            <name>Spring Releases</name>
+            <url>https://repo.spring.io/release</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
+    </repositories>
+
```
malessi, line 27: I honestly don't know what to do here because the `spring-retry` library is _supposed_ to be available on Maven Central, but it seems that [just isn't the case](https://repo.maven.apache.org/maven2/springframework/). However, the Spring Release repository does seem to have the latest release version of `spring-retry` available.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SpringConfiguration.java
```diff
@@ -59,12 +59,14 @@
...
 import org.springframework.orm.jpa.LocalContainerEntityManagerFactoryBean;
 import org.springframework.orm.jpa.support.PersistenceAnnotationBeanPostProcessor;
+import org.springframework.retry.annotation.EnableRetry;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 /** The main Spring {@link Configuration} for the Blue Button API Backend application. */
 @Configuration
 @ComponentScan(basePackageClasses = {ServerInitializer.class})
 @EnableScheduling
+@EnableRetry(proxyTargetClass = true)
```
malessi, line 69: As far as I understand it, if we were to use interfaces for each provider class rather than the concrete implementation we wouldn't need to specify `proxyTargetClass` here (the `Proxy` will explicitly subclass the class with any `@Retryable`-decorated methods).

The side effect of doing it this way is that all provider classes need to be non-`final` so that the generated `Proxy`s _can_ subclass them properly.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/RetryOnRdsFailover.java
```diff
@@ -0,0 +1,31 @@
...
+ * FailoverSQLException} is not possible for all provider operations. Instead, a SpEL expression is
+ * used to get the root cause of the thrown {@link Exception} and see if it is an instance of {@link
+ * FailoverSQLException} (note that the root cause of a {@link FailoverSQLException} is itself).
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Retryable(
+    exceptionExpression =
+        "T(org.apache.commons.lang3.exception.ExceptionUtils).getRootCause(#root) instanceof T(software.amazon.jdbc.plugin.failover.FailoverSQLException)",
```
malessi, line 28: This `exceptionExpression` uses [SpEL](https://docs.spring.io/spring-framework/reference/core/expressions.html) (Spring Expression Language). Essentially, this string will be evaluated as an expression at runtime for each `Exception` that `@RetryOnRdsFailover`-decorated methods throw. `#root`, in this context, is the `Exception` that was thrown, so this expression is essentially equivalent to:

```java
// Assume #root is ex
return ExceptionUtils.getRootCause(ex) instanceof FailoverSQLException
```

We must specify the fully-qualified path to each type/static class (using the `T(...)` syntax) because we cannot guarantee that these types are imported when this expression is evaluated. I do realize that this is not entirely ideal, and I could extract this expression out into a separate `SpringRetryUtils` static utility class or something similar if we're uncomfortable with this line.

aschey-forpeople: I'm assuming `@Retryable(retryFor = FailoverSqlException.class)` doesn't work because FailoverSqlException is the inner exception rather than the outer one?

aschey-forpeople: Would be nice if you could have it call a method instead of this stringly-typed business, but if this is the only choice, I think it's fine to keep it inline.

malessi: > I'm assuming `@Retryable(retryFor = FailoverSqlException.class)` doesn't work because FailoverSqlException is the inner exception rather than the outer one?

Yep. There's some downstream methods called by the providers that throw, essentially: `new RuntimeException(new JDBCException(new FailoverSuccessSQLException)))`. Even if they don't wrap it in an outer `RuntimeException`, the resulting exception will almost always be a `JDBCException` wrapping a `FailoverSuccessSQLException`. So we can't rely on `retryFor` here because it only looks at the type of the outermost `Exception` for matching.

> Would be nice if you could have it call a method instead of this stringly-typed business, but if this is the only choice, I think it's fine to keep it inline.

I thought about it some more and I think I'm leaning more towards a `SpringRetryUtils` static class with a static method called something like `shouldRetryFailover(...)` that implements the same logic as this SpEL expression. We'd still need to use the canonical class name (i.e. `gov.cms.bfd.server.war.commons.SpringRetryUtils`) in the SpEL that calls it, though, something like:

```
exceptionExpression = "T(gov.cms.bfd.server.war.commons.SpringRetryUtils).shouldRetryFailover(#root)"
```

_But_, the `shouldRetryFailover` method could be unit tested and we could be more confident that the package path of `SpringRetryUtils` won't change under us without warning.

I'm going to refactor this following the above.

aschey-forpeople: Nice, I think that's a worthy improvement

malessi: This has been done in [4e291f3](https://github.com/CMSgov/beneficiary-fhir-data/pull/2402/commits/4e291f3b1f6b2d3c9d4a18011209cb8132266e86)

malessi:
@sb-benohe Added you as this adds two libraries for use at runtime, `spring-retry` and `spring-aspects`, and one for testing, `spring-test`.

dondevun (approved):

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/commons/RetryOnRDSFailoverIT.java
```diff
@@ -0,0 +1,190 @@
...
+     *     Exception}s to throw prior to executing actual method logic
+     * @param guid random GUID that is used as a key to identify a given test-case run for tracking
+     *     retry counts
+     * @return {@code true}, always
+     * @throws Exception whichever {@link Exception} that is at the top of the provided list of
+     *     {@link Exception}s, if not empty
+     */
+    @RetryOnRDSFailover(backoff = @Backoff(delay = 1))
+    boolean retryableOperation(List<Exception> exceptions, String guid) throws Exception {
+      RETRY_COUNTS_PER_GUID.put(guid, RETRY_COUNTS_PER_GUID.getOrDefault(guid, -1) + 1);
```
malessi, line 77: For anyone curious: I opted to do the counting here manually rather than rely on something like a [`Mockito.spy`](https://site.mockito.org/javadoc/current/org/mockito/Spy.html) to verify call counts of `RetryableOperationBean#retryableOperation` because it seems that `spring-retry` either doesn't call the `@Retryable`-annotated method on retries or `spy` doesn't play nice with the proxied `RetryableOperationsBean` class. Either way, attempting to verifying retry attempts using Mockito's call counting (with `verify()` and `times()`) always returns that the method was called once, even though it _should_ return that it was called multiple times.

So, tracking the number of retry attempts manually was honestly the easiest approach I could think of. A `ConcurrentHashMap` is used in case these tests are ever run in parallel, also.

dondevun (approved):

aschey-forpeople (approved):
Looks good, nice work on the tests!

PR #2405, opened 2024-08-20
BFD-3544: Update-BFD-Test-DB-connectivity-scripts to add add username and password options

MahiFentaye:
**JIRA Ticket:**
BFD-3544

### What Does This PR Do?

add `username` and `password` options to connect to test db from local

### Validation


apps/utils/scripts/run-bfd-server
```diff
@@ -111,13 +115,26 @@ Option defaults:
 EOF
 }
 
-while getopts "d:i:kl:xzZh" option ; do
+# Prompt for test database values
+prompt_for_test_db() {
+   read -p "Host: " db_host
+   read -p "Port: " db_port
+   read -p "Username: " "username"
+   read -sp "Password: " "password"
```
dondevun, line 123: Did you test this? RDS tokens appear to be too long for the password input. I think a better approach would be to take the DB parameters in using Bash command line arguments.

apps/utils/scripts/run-bfd-server
```diff
@@ -85,11 +85,14 @@ debug_enabled=false
 debug_suspend=n
 db_host=localhost
 db_port=5432
+username=
+password=
```
aschey-forpeople, line 89: I think we can default these to the local ones just like the db_host has a default here
```suggestion
username=bfd
password=InsecureLocalDev
```

apps/utils/scripts/run-bfd-server
```diff
@@ -248,9 +258,16 @@ setvar paths/files/truststore "$bin_dir/ssl/${truststore_file}"
...
-setvar db/username "bfd"
-setvar db/password "InsecureLocalDev"
+# Conditional database URL setting
+if $host_set && [[ -n "$username" ]] && [[ -n "$password" ]]; then
+  setvar db/url "jdbc:postgresql://${db_host}:${db_port}/fhirdb?logServerErrorDetail=false"
+    setvar db/username $username
+    setvar db/password $password
+else
+ setvar db/url "jdbc:postgresql://${db_host}:${db_port}/bfd?logServerErrorDetail=false"
+   setvar db/username "bfd"
```
aschey-forpeople, line 268: If we add the default as I mentioned above then I think we can move these out of the conditional

MahiFentaye: the only reason I didn't set the values as default is because it made sense to check for null incase a user only set the host but not the username and password

aschey-forpeople: Yeah, I'm wondering if we can just fix it so that the database name doesn't have to be different for local development.

aschey-forpeople: Actually I think we can. We just need to change the database name in `run-db-migrator` and then we can always connect to `fhirdb` here instead of `bfd`. And then we shouldn't need to do any additional checks.

MahiFentaye: okay, looking into this

dondevun: @aschey-forpeople @MahiFentaye That's a little bit out of scope for this ticket. Let's create a new one, as I think it will be a little bit more work than just changing one line.

aschey: Actually I think it is one line if I'm not missing something. The migrator script hardcodes the db name, we just need to change it there.

apps/utils/scripts/run-db-migrator
```diff
@@ -238,7 +238,7 @@ fi
 
 # Set up the configuration variables using the setvar function.
 export BFD_ENV_NAME=laptop
-setvar db/url "jdbc:postgresql://${db_host}:${db_port}/bfd"
+setvar db/url "jdbc:postgresql://${db_host}:${db_port}/fhirdb"
```
MahiFentaye, line 241: the wiki instructions need to be updated to use `fhirbd` for [Testing Local Environment Setup](https://github.com/CMSgov/beneficiary-fhir-data/wiki/Testing-Your-Local-Environment-Setup) once this is approved

aschey-forpeople: Looks good! I think we just need to make the same change for `run-bfd-pipeline` and then we should be set.

aschey-forpeople (approved):

dondevun (approved):

PR #2432, opened 2024-09-13
BFD-3517: Duplicate GET Locust test suites with POST

MahiFentaye:
**JIRA Ticket:**
BFD-3517

### What Does This PR Do?
BFD-3517_duplicate POST Locust test suites
adds `post_by_url` method that takes a `body` parameter
when `body` is present, it will make a `POST` request instead of a `GET` 

There is a copy of each test suit to a new POST version

### Validation

This has been tested against test db
locust -f v2/regression_suite_post.py --users 10 --spawn-rate 10 --spawned-runtime 30s
locust -f v1/regression_suite_post.py --users 10 --spawn-rate 10 --spawned-runtime 30s
locust -f high_volume_suite_post.py --users 10 --spawn-rate 10 --spawned-runtime 30s


dondevun:
@MahiFentaye Please use a little more verbosity with the PR template.

dondevun:
@MahiFentaye I think this looks good. Have you tested it?

MahiFentaye:
yes, I have tested @dondevun

dondevun (approved):
Looks Good. Please update the PR template Validation section with your testing steps.

aschey-forpeople (changes requested):
Whoops, just found a typo

apps/utils/locust_tests/v2/partially_adjudicated_post.py
```diff
@@ -0,0 +1,124 @@
...
+        self.__post(
+            "ClaimResponse/_search",
+            "claimResponseServiceDateLastUpdated",
+            self.SERVICE_DATE_LAST_UPDATED,
+        )
+
+    @tag("claim", "exclude-samsa")
+    @task
+    def get_claim_response_with_exclude_samsa(self):
+        self.__post("ClaimRespons/_search", "excludeSAMSA", True)
```
aschey-forpeople, line 124: Typo here ^

aschey-forpeople: I would expect these tests to be failing with the wrong endpoint name here. Is that not the case?

apps/utils/locust_tests/v2/partially_adjudicated_post.py
```diff
@@ -0,0 +1,124 @@
...
+    @tag("claim", "service-date", "last-updated")
+    @task
+    def get_claim_with_service_date_and_last_updated(self):
+        """Get single Claim with last updated and service date"""
+        self.__post("Claim/_search", "claimServiceDateLastUpdated", self.SERVICE_DATE_LAST_UPDATED)
+
+    @tag("claim", "exclude-samsa")
+    @task
+    def get_claim_with_exclude_samsa(self):
+        self.__post("Claim/_search", "excludeSAMSA", True)
```
mjburling, line 91: Looks like a typo (or propagated typo) here, too.

aschey-forpeople:
I think we also need to update `app.py` so that it invokes the new tests.

apps/utils/locust_tests/v2/partially_adjudicated_post.py
aschey-forpeople: It looks like that's actually an issue with the existing code lol. So we haven't been testing with `excludeSAMHSA` set properly at all... fun.

MahiFentaye: Good catch. yes, this two excludeSAMHSA tests don't pass for both GET and POST. It was easily missed because it doesn't tell you a test failed either, You have to look at the log above and you will see errors related to it.
I do see issues with hitting both `Claim` and `ClaimResponse` endpoints even after correcting the `excludeSAMHSA` typo. I will start a thread

MahiFentaye: This test files were old and didn't need a post version since they aren't  being used anymore, The new test files are deleted now. Thanks to Austin!

apps/utils/locust_tests/v1/regression_suite_post.py
```diff
@@ -0,0 +1,189 @@
...
+    @tag("patient", "patient_test_id")
+    @task
+    def patient_test_id(self):
+        """Patient search by ID"""
+        self.run_task_by_parameters(
+            base_path="/v1/fhir/Patient/_search",
+            body={
+                "_id": next(self.bene_ids),
+            },
+            name="/v1/fhir/Patient/id"
```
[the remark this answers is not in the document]
MahiFentaye: updated

dondevun (approved):

apps/utils/locust_tests/high_volume_suite_post.py
```diff
@@ -0,0 +1,485 @@
+"""High Volume Load test suite for BFD Server endpoints."""
```
aschey-forpeople, line 1: Do we need to update one of the other files to call this test suite or is that going to be done later?

malessi: I think we can do that later when we remove all `GET` suites. We did need to modify the infrastructure that runs the Regression Suite since it gates the deployment and we want to verify the `POST` endpoints as part of it whereas the High Volume Suite (HVS) is ran quite rarely and _usually_ only against ephemeral environments. We can probably hold off on modifying the HVS's supporting infrastructure to run the `POST`-variant until after the `GET` endpoints are removed.

aschey-forpeople: Sure, sounds good 👍

aschey-forpeople (approved):

PR #2452, opened 2024-09-30
BFD-3658: Create new SAMHSA list

alex-dzeda:
**JIRA Ticket:**
BFD-3658

### What Does This PR Do?

This PR re-creates the existing SAMHSA list, with date-definite start and end dates for each code. It does not meet the criteria for BFD-3659, as further work is ongoing to determine the needs behind some of the historical codes. The formatting also differs slightly from the ticket. Given that there have been no updates to the ICD-9/ICD-10 lists since at least 2015, I believe it's more prudent to store a diff, rather than utilizing 1/5/24 as the starting point.

This file will be cached / loaded into memory / into a database at runtime and referred to while data is ingestion. With the intent of this being used to find a union with a continuous timeframe, is a hardcoded date, or some other constant preferred for start and end dates? 

### Validation

The count for the "systems" and the codes provided match the existing list, with the addition of two historical entries for MS-DRG 522 + 523.


apps/bfd-server/bfd-server-war/src/main/resources/security_labels.csv
```diff
@@ -0,0 +1,678 @@
...
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0009,1970-01-01,ACTIVE
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0010,1970-01-01,ACTIVE
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0011,1970-01-01,ACTIVE
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0012,1970-01-01,ACTIVE
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0013,1970-01-01,ACTIVE
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0014,1970-01-01,ACTIVE
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0015,1970-01-01,ACTIVE
+https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets,H0050,1970-01-01,ACTIVE
+http://www.ama-assn.org/go/cpt,99408,1970-01-01,ACTIVE
+http://www.ama-assn.org/go/cpt,99409,1970-01-01,ACTIVE
```
aschey-forpeople, line 16: Not a big deal, but for ease of reading through this, maybe it's better to have all the codes grouped by system?

alex-dzeda:
Updated to adopt yaml. Of note, only a few codes have start dates that are not 1970; I tracked the changes over time in the individual CCLF IP list, and codes added AFTER they were issued are marked as beginning at 1970, in order to prevent overshares that would otherwise occur.

dondevun (approved):

PR #2457, opened 2024-10-04
BFD-3688: Log nonsensitive MBI identifiers to MDC for Claim and ClaimResponse requests

malessi:
**JIRA Ticket:**
BFD-3688

### What Does This PR Do?

This PR introduces additional logging to the `Claim` and `ClaimResponse` endpoints that logs the `mbi_id` and `mbi_hash` of the MBI associated with the FISS or MCS claim to the MDC. Since the MDC log informs BFD Insights, these fields are now available to query within Athena as `mdc_mbi_hash` and `mdc_mbi_id`. Correspondingly, the Glue Table and Parquet schemas have been updated to include these new columns.

Additionally, E2E tests were added verifying that these identifiers are logged to the `access.json` MDC log for any valid request against `Claim` or `ClaimResponse`.

**Note that the Glue Crawlers in `prod-sbx` and `prod` must be run after this is merged before `mdc_mbi_hash` and `mdc_mbi_id` can be queried upon. I will do this upon merge, out-of-band.**

- ~~Adds any new software dependencies~~

### Validation

- Running all automated tests **including the new E2E tests**, _verifying that_ they pass
- Deploying all changes to `test` and running the Regression Suite against the Server in `test`, _verifying that_:
  - `mbi_id` and `mbi_hash` appear in valid `Claim` and `ClaimResponse` access.json logs 
  - `mbi_id` and `mbi_hash` are queryable in BFD Insights data using Athena

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
@@ -57,6 +57,7 @@ import org.hl7.fhir.r4.model.Bundle;
 import org.hl7.fhir.r4.model.ClaimResponse;
 import org.hl7.fhir.r4.model.IdType;
 import org.hl7.fhir.r4.model.Resource;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Allows for generic processing of resource using common logic. Claims and ClaimResponses have the
@@ -285,6 +286,26 @@ public abstract class AbstractR4ResourceProvider<T extends IBaseResource>
   }
 
   /**
+   * Returns the given {@link RdaFissClaim} or {@link RdaMcsClaim} entity's MBI ID given the
+   * resource type.
+   *
+   * @param <T> generic type extending {@link IBaseResource}
+   * @param claimIdType the {@link ResourceTypeV2} indicating what the claim's type is (either fiss
+   *     or mcs)
+   * @param claimEntity the claim entity itself; either a {@link RdaFissClaim} or {@link
+   *     RdaMcsClaim}
+   * @return the MBI ID associated with the given claim entity, or <code>null</code> if not found
+   */
+  private <T extends IBaseResource> @Nullable String getClaimEntityMbiId(
+      ResourceTypeV2<T, ?> claimIdType, Object claimEntity) {
+    return switch (claimIdType.getTypeLabel()) {
+      case "fiss" -> ((RdaFissClaim) claimEntity).getMbiRecord().getMbiId().toString();
+      case "mcs" -> ((RdaMcsClaim) claimEntity).getMbiRecord().getMbiId().toString();
+      default -> null;
+    };
+  }
+
+  /**
    * Implementation specific claim type parsing.
    *
    * @param typeText String to parse representing the claim type.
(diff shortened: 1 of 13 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
```diff
@@ -282,6 +289,39 @@ private T transformEntity(
...
+   *     RdaMcsClaim}
+   * @return the {@link Mbi} associated with the given claim entity, or <code>null</code> if not
+   *     found
+   */
+  private <T extends IBaseResource> @Nullable Mbi getClaimEntityMbi(
+      ResourceTypeV2<T, ?> claimIdType, Object claimEntity) {
+    return switch (claimIdType.getTypeLabel()) {
+      case "fiss" -> ((RdaFissClaim) claimEntity).getMbiRecord();
+      case "mcs" -> ((RdaMcsClaim) claimEntity).getMbiRecord();
+      default -> null;
```
aschey-forpeople, line 309: I would probably throw an exception here since this case should never get hit.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
```diff
@@ -449,24 +485,40 @@ Bundle createBundleFor(
...
+            // There may be multiple resource types which will each result in a list of claim
+            // entities. So, to ensure that we have a flat list of entities to their type, we use
+            // flatMap
+            .flatMap(
+                type ->
+                    claimDao
+                        .findAllByMbiAttribute(
+                            type, mbi, bundleOptions.isHashed, lastUpdated, serviceDate)
+                        .stream()
+                        .map(e -> new ImmutablePair<>(e, type)))
```
aschey-forpeople, line 499: I don't feel _too_ strongly about this, but considering the degree of nesting here, maybe this is worth breaking into a helper function?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
```diff
@@ -55,6 +58,7 @@
 import org.hl7.fhir.r4.model.ClaimResponse;
 import org.hl7.fhir.r4.model.IdType;
 import org.hl7.fhir.r4.model.Resource;
+import org.jetbrains.annotations.Nullable;
```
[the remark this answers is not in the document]
malessi: Good catch. I fixed this in [01dcdfa](https://github.com/CMSgov/beneficiary-fhir-data/pull/2457/commits/01dcdfa04195a6baecaa4e753fd69cf05d8f423d).

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
malessi: Agreed. I've gone ahead and done this in [64a0dbd](https://github.com/CMSgov/beneficiary-fhir-data/pull/2457/commits/64a0dbd164a5e7226697d34dd271b975eadb8351).

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
malessi: I'm hesitant about refactoring this, but I definitely could be convinced otherwise. Thinking about it, the resulting helper method would have a rather complicated argument list (all arguments to `findAllByMbiAttribute`) and return type (`Stream<ImmutablePair<ResourceTypeV2...`). Additionally, although this is far less relevant, I can't think of any other spot in the code where this hypothetical helper method would be useful other than to replace lines 495 - 499 in this file.

I feel like this would result in more code for not too much gain. I'm no fan of unnecessary nesting, but I feel this is readable enough and anyone with a good grasp of `Stream`s should be able to grok the intent fairly quickly (take a list of `ResourceTypeV2`s, for each find list of claim entities of that type, flat map each type to each entity in the aforementioned list by pairing them using `ImmutablePair`).

dondevun (approved):

PR #2465, opened 2024-10-17
BFD-3644 Update C4DIC Implementation - Plan ID + Group ID

MahiFentaye:
**JIRA Ticket:**
BFD-3644
Plan ID updated to a combination of Contract ID + PBP ID for part C and part D C4DIC coverage resources
The change gets the value of the current month and gets Contract ID and PBP for the current month and concatenates the to as Plan ID

Part C 
old
```
{
"type": {
"coding": [
{
"system": "http://terminology.hl7.org/CodeSystem/coverage-class",
"code": "plan",
"display": "Plan"
}
]
},
"value": "Part C"
}
```
new 

```
{
"type": {
"coding": [
{
"system": "http://terminology.hl7.org/CodeSystem/coverage-class",
"code": "plan",
"display": "Plan"
}
]
},
"value": "Y0002-800"
}
```

Part D
old: 
```
{
"type": {
"coding": [
{
"system": "http://terminology.hl7.org/CodeSystem/coverage-class",
"code": "plan",
"display": "Plan"
}
]
},
"value": "Part D"
}
```
new:
```
{
"type": {
"coding": [
{
"system": "http://terminology.hl7.org/CodeSystem/coverage-class",
"code": "plan",
"display": "Plan"
}
]
},
"value": "Z0007-800"
}
```

### Validation


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -1098,4 +1099,133 @@ Timer.Context createTimerContext(String partId) {
...
+   * concatenates Part C Contract And Pbp.
+   *
+   * @param beneficiary the value for {@link Beneficiary}
+   * @return concatenated Contract And Pbp
+   */
+  private String concatenatePartCContractAndPbp(Beneficiary beneficiary) {
+    int month = Calendar.getInstance().get(Calendar.MONTH) + 1;
+    String contractId = getPartCContractNumber(beneficiary, month).orElse(null);
+    String pbpId = getPartCPbpNumber(beneficiary, month).orElse(null);
+    return contractId + "-" + pbpId;
```
dondevun, line 1113: Since contractId and pbpId can technically be null, you should have a null check here. Same in the concatenatePartDContractAndPbp() method.

MahiFentaye: @alex-dzeda should we display `Part C` or `Part D` like before in case of null?

dondevun (approved):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -342,6 +338,15 @@ private Coverage transformPartC(Beneficiary beneficiary, Profile profile) {
       transformEntitlementDualEligibility(coverage, beneficiary);
     }
 
+    String ContractAndPbpID = concatenatePartCContractAndPbp(beneficiary);
```
aschey-forpeople, line 341: It looks like the changes here will affect every response, but we only want to make these changes for the C4DIC profile (`profile.equals(Profile.C4DIC))`)

aschey-forpeople (changes requested):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -442,6 +441,14 @@ private Coverage transformPartD(Beneficiary beneficiary, Profile profile) {
       transformEntitlementDualEligibility(coverage, beneficiary);
     }
 
+    String ContractAndPbpID = concatenatePartDContractAndPbp(beneficiary);
```
aschey-forpeople, line 444: Variable here should be camelcase
```suggestion
    String contractAndPbpID = concatenatePartDContractAndPbp(beneficiary);
```

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -1098,4 +1105,138 @@ Timer.Context createTimerContext(String partId) {
...
+
+  /**
+   * concatenates Part D Contract And Pbp.
+   *
+   * @param beneficiary the value for {@link Beneficiary}
+   * @return concatenated Contract And Pbp
+   */
+  private String concatenatePartDContractAndPbp(Beneficiary beneficiary) {
+
+    int month = Calendar.getInstance().get(Calendar.MONTH);
```
aschey-forpeople, line 1133: Might be nice to make a helper function for this rather than duplicating the logic for part C and part D

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
MahiFentaye: Good Catch

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -342,6 +345,24 @@ private Coverage transformPartC(Beneficiary beneficiary, Profile profile) {
...
+            getPartCContractNumber(beneficiary), getPartCPbpNumber(beneficiary));
+
+    if (profile.equals(Profile.C4DIC)) {
+      // Hide resource Part C when Contract ID or PBP ID is null
+      if (contractAndPbpID == null) {
+        return null;
+      }
+      createCoverageClass(coverage, CoverageClass.PLAN, contractAndPbpID, Optional.empty());
+    } else {
+      createCoverageClass(
```
aschey-forpeople, line 359: Looks like the group is still missing here.

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/coverageSearchByPatientId.json
```diff
@@ -908,15 +908,6 @@
         }
       } ],
       "class" : [ {
-        "type" : {
```
aschey-forpeople, line 911: It's a good idea to look at these files to ensure nothing is being changed accidentally. We shouldn't be seeing these values removed here.

aschey-forpeople (changes requested):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
MahiFentaye: good catch

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/coverageSearchByPatientId.json
MahiFentaye: yeah, this was part of removing the Group. adding that back for C4DIC should fix this as well

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4CoverageResourceProvider.java
```diff
@@ -213,6 +213,11 @@ public Coverage read(@IdParam IdType coverageId) {
 
     Coverage coverage =
         coverageTransformer.transform(coverageIdSegment.get(), beneficiaryEntity, profileUsed);
+
+    if (coverage == null || coverage.isEmpty()) {
```
aschey-forpeople, line 217: Do we need the check for `isEmpty()` here? I'm not entirely sure what would cause that condition. Maybe it's safer to only check null here?

MahiFentaye: I am not sure what would cause that either but I thought it would better to cover it. Doesn't it just add extra layer of safety if that ever happens?

aschey-forpeople: It looks like that method returns true if all of the properties are empty. I don't think we want to throw an exception in that case.

MahiFentaye: updated

aschey-forpeople (approved):

dondevun (approved):

PR #2466, opened 2024-10-17
BFD-3603: Change careTeam[].qualification to use taxonomy

dondevun:
**JIRA Ticket:**
BFD-3603

### What Does This PR Do?

This PR changes careTeam[].qualification to use taxonomy, instead of prvdr_spclty. The npiOrgLookup module has been refactored to include Practitioner NPIs, instead of just Organization NPIs, and the taxonomy mappings have been downloaded and put into the resources folder to use as a source during lookup.

### Validation

Tested against golden bene for all claim types.


apps/bfd-data-npi/src/main/java/gov/cms/bfd/data/npi/utility/DataUtilityCommons.java
```diff
@@ -262,6 +266,36 @@ private static void buildNPIResource(
...
+   * @throws IOException exception thrown.
+   */
+  public static Map<String, String> processTaxonomyDescriptions() throws IOException {
+    ClassLoader classLoader = DataUtilityCommons.class.getClassLoader();
+    File taxonomyFile =
+        new File(
+            Objects.requireNonNull(classLoader.getResource("nucc_taxonomy_240.csv")).getFile());
+    Map<String, String> taxonomyMapping = new HashMap<>();
+    try (FileInputStream is = new FileInputStream(taxonomyFile);
+        BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
```
aschey-forpeople, line 282: Would it be easier to use a CSV reader here?

dondevun: Maybe, but if we did that, then we should probably refactor the rest of the code in this module, as well, to use the CSV reader -- this follows the same pattern that already exists to read the NPI file. I'm not sure if introducing a library and refactoring all of this code is the best course of action, but I'm open to a conversation about it.

MahiFentaye (dismissed):

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/NPIOrgLookupIT.java
```diff
@@ -15,7 +15,7 @@ public class NPIOrgLookupIT {
   public void VerifyAValidNPIOrg() throws IOException {
     NPIOrgLookup npiOrgLookup = NPIOrgLookup.createNpiOrgLookup();
     Optional<String> orgDisplay = npiOrgLookup.retrieveNPIOrgDisplay(Optional.of("1497758544"));
-    assertEquals("CUMBERLAND COUNTY HOSPITAL SYSTEM", orgDisplay.get());
+    assertEquals("CUMBERLAND COUNTY HOSPITAL SYSTEM, INC", orgDisplay.get());
```
dondevun, line 18: Adding the CSV reader uncovered a bug where org names with commas were not displaying correctly.

apps/bfd-data-npi/src/main/java/gov/cms/bfd/data/npi/utility/DataUtilityCommons.java
dondevun: Good call on the CSV reader. CSV parsing bugs were uncovered.

aschey-forpeople: Awesome, thanks for looking at that.

apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
```diff
@@ -1050,13 +1050,6 @@
             "code" : "performing",
             "display" : "Performing provider"
           } ]
-        },
-        "qualification" : {
```
aschey-forpeople, line 1054: Are all of these supposed to be removed now? Might be nice to update these tests to include the new fields.

dondevun:
The new NPI enrichment story will refactor a lot of this code, so I think I'm going to abandon this, and continue the work in the new NPI enrichment story.

apps/bfd-data-npi/src/main/java/gov/cms/bfd/data/npi/utility/DataUtilityCommons.java
```diff
@@ -271,45 +315,42 @@ private static void buildNPIResource(
...
+        String taxonomyCode = csvRecord.get(TAXONOMY_CODE_FIELD);
         // entity type code 2 is organization
-        if (!Strings.isNullOrEmpty(entityTypeCode) && Integer.parseInt(entityTypeCode) == 2) {
-          out.write(npi + "\t" + orgName);
-          out.newLine();
+        if (!Strings.isNullOrEmpty(entityTypeCode)) {
+          if (Integer.parseInt(entityTypeCode) == 2 && !Strings.isNullOrEmpty(orgName)) {
+            out.write(npi + "\t" + orgName);
+            out.newLine();
+          } else if (!Strings.isNullOrEmpty(taxonomyCode)) {
```
dondevun, line 350: This should resolve the empty rows in the constructed file.

MahiFentaye (dismissed):

apps/bfd-data-npi/src/main/java/gov/cms/bfd/data/npi/utility/DataUtilityCommons.java
aschey-forpeople: What's the reason that some of the org names are empty?

aschey-forpeople: Makes sense, thanks!

apps/bfd-data-npi/src/main/java/gov/cms/bfd/data/npi/utility/DataUtilityCommons.java
aschey-forpeople: Actually, I don't think this is always the case. I added this line
```java
 if (Integer.parseInt(entityTypeCode) == 2 && Strings.isNullOrEmpty(orgName)) {
            System.out.println("empty org");
 }
```
and it got a few hits before I stopped it.

apps/bfd-data-npi/src/main/java/gov/cms/bfd/data/npi/utility/DataUtilityCommons.java
aschey-forpeople: Actually, I don't think this is always the case. I added this line
```java
 if (Integer.parseInt(entityTypeCode) == 2 && Strings.isNullOrEmpty(orgName)) {
            System.out.println("empty org");
 }
```
and it got a few hits before I stopped it.

MahiFentaye (approved):

PR #2470, opened 2024-10-24
BFD-3716: Filter any MBI being logged to MDC

aschey-forpeople:
**JIRA Ticket:**
BFD-3716

### What Does This PR Do?

The current MBI filter in the MDC logger has a few limitations:
- Only looks for strings that start with `us-mbi|` which is not always the case
- Doesn't handle url-encoded values
- Only looks at the querystring
- Only looks at the Patient endpoint

This PR makes the following changes:
- Looks at the complete request parts for any MBIs (querystring, headers, and URL)
- Doesn't rely on there being a `us-mbi|` prefix - this _could_ produce false positives in theory, but I can't think of any conditions that would cause this. Even if it does, it's better than a false negative.
- Removes any filtering on a specific endpoint
- Adds test cases for the filter logic

- Are there any other spots that we should filter on?
- Is the regex filter sufficient?

This does have security implications.

### Validation

Unit tests. Also ran locally and verified the masking was being applied in the live server output.

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
@@ -162,6 +162,9 @@ public class RequestResponsePopulateMdcFilter extends OncePerRequestFilter {
    * @return modified input
    */
   private String replaceAllMbis(String input) {
+    if (input == null) {
+      return input;
+    }
     Matcher matcher = MBI_REGEX.matcher(input);
     while (matcher.find()) {
       input = input.replace(matcher.group(), OBFUSCATED_MBI);
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
@@ -61,7 +61,7 @@ public class RequestResponsePopulateMdcFilter extends OncePerRequestFilter {
 
       // See the following link for more info on the MBI format
       // https://www.cms.gov/medicare/new-medicare-card/understanding-the-mbi-with-format.pdf
-      Pattern.compile("(?:\\d\\p{Alpha}\\p{Alnum}){2}\\d\\p{Alpha}{2}\\d{2}");
+      Pattern.compile("\\d\\p{Alpha}\\p{Alnum}\\d-?\\p{Alpha}\\p{Alnum}\\d-?\\p{Alpha}{2}\\d{2}");
 
   /** {@inheritDoc} */
   @Override
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
@@ -52,15 +52,16 @@ public class RequestResponsePopulateMdcFilter extends OncePerRequestFilter {
   private static final String RESPONSE_PREFIX = "response";
 
   /** obfuscated replacement for MBI. */
-  private static final String OBFUSCATED_MBI = "us-mbi|*********";
-
-  /** Regex Pattern to check for an /Patient request end-point. */
-  private static final Pattern PATIENT_REQUEST_REGEX = Pattern.compile("fhir/Patient");
+  private static final String OBFUSCATED_MBI = "*********";
 
   /** Regex Pattern to check for an MBI. */
   private static final Pattern MBI_REGEX =
-      Pattern.compile(
-          "us-mbi\\|[1-9][^SLOIBZsloibz0-9][^SLOIBZsloibz][0-9]-?[^SLOIBZsloibz0-9][^SLOIBZsloibz][0-9]-?[^SLOIBZsloibz0-9][^SLOIBZsloibz0-9][0-9][0-9]");
+      // Real MBIs disallow certain characters, but we'll be conservative here for simplicity's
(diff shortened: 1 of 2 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
```diff
@@ -113,65 +114,61 @@ private void handleRequest(HttpServletRequest request) {
...
-    CanonicalOperation operation = new CanonicalOperation(CanonicalOperation.Endpoint.OTHER);
+    // Record the basic request components.
+    CanonicalOperation operation =
+        new CanonicalOperation(CanonicalOperation.Endpoint.matchByHttpUri(request));
+    BfdMDC.put(
+        BfdMDC.computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, "http_method"), request.getMethod());
+    String url = replaceAllMbis(request.getRequestURL().toString());
+    BfdMDC.put(BfdMDC.computeMDCKey(MDC_PREFIX, REQUEST_PREFIX, "url"), url);
 
-    if (request instanceof HttpServletRequest) {
```
aschey-forpeople, line 119: I removed this check because it serves no purpose - the argument is already an `HttpServletRequest` so this check serves no purpose. Must've been left over from a previous refactor or something.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
```diff
@@ -267,7 +264,7 @@ private static String getClientSslPrincipalDistinguishedName(HttpServletRequest
   private static X509Certificate getClientCertificate(HttpServletRequest request) {
     X509Certificate[] certs =
         (X509Certificate[]) request.getAttribute("jakarta.servlet.request.X509Certificate");
-    if (certs == null || certs.length <= 0) {
+    if (certs == null || certs.length == 0) {
```
aschey-forpeople, line 267: Unrelated change here, but it irks me to see a `<= 0` check on a value that can only be positive.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilterTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Mock HttpServletRequest servletRequest;
+  @Mock HttpServletResponse servletResponse;
+  @Mock FilterChain filterChain;
+
+  /** Set up tests with a mocked MDC adapter and make sure the MDC context is clear. */
+  @BeforeEach
+  public void beforeEach() {
+    mdcMock.clear();
+    BfdMDC.setMDCAdapter(mdcMock);
```
aschey-forpeople, line 39: Unfortunately the MDC adapter is a static variable which interferes with other tests when setting the mock value here. I'll have to think about the best workaround for this.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
```diff
@@ -52,15 +52,16 @@ public class RequestResponsePopulateMdcFilter extends OncePerRequestFilter {
...
   /** Regex Pattern to check for an MBI. */
   private static final Pattern MBI_REGEX =
-      Pattern.compile(
-          "us-mbi\\|[1-9][^SLOIBZsloibz0-9][^SLOIBZsloibz][0-9]-?[^SLOIBZsloibz0-9][^SLOIBZsloibz][0-9]-?[^SLOIBZsloibz0-9][^SLOIBZsloibz0-9][0-9][0-9]");
+      // Real MBIs disallow certain characters, but we'll be conservative here for simplicity's
+      // sake, and so we can easily test against synthetic MBIs.
+
+      // See the following link for more info on the MBI format
+      // https://www.cms.gov/medicare/new-medicare-card/understanding-the-mbi-with-format.pdf
+      Pattern.compile("\\d\\p{Alpha}\\p{Alnum}\\d-?\\p{Alpha}\\p{Alnum}\\d-?\\p{Alpha}{2}\\d{2}");
```
aschey-forpeople, line 64: Although we don't support MBIs formatted with dashes, the original regex did check for it, and per the document linked above, it is a valid format. So it's plausible to think someone may attempt to construct a query with a dash-formatted MBI.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilterTest.java
```diff
@@ -0,0 +1,180 @@
...
+    BfdMDC.setMDCAdapter(previousAdapter);
+  }
+
+  /***
+   * Test arguments.
+   * @return arguments
+   */
+  public static Stream<Arguments> loggingTest() {
+    return Stream.of(
+        arguments("nonsensitive", "nonsensitive"),
```
aschey-forpeople, line 71: If anyone has ideas for additional test cases to add here, I'd be happy to include them.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/RequestResponsePopulateMdcFilter.java
mjburling: Casting a wider net for potential matches here makes a lot of sense. The scope is already limited by what should or _could_ appear as part of the request. And I agree–it doesn't much imagination to consider how an errant, dash-delimited MBI might be introduced here.

dondevun (approved):

dondevun:
I didn't see any issues jump out at me for this; I think it is probably ok.

malessi (approved):

PR #2471, opened 2024-10-29
BFD-3706: Refactor NPI Enrichment

dondevun:
**JIRA Ticket:**
BFD-3706

### What Does This PR Do?

This code refactors the NPI enrichment to a cleaner, more object oriented approach. 

The TSV that is stored on disk and in memory at runtime has been changed to use JSON. When the `retrieveNPiOrgDisplay()` method is called, the JSON is mapped to a POJO (instead of a TSV string, as exists right now). This is a much cleaner approach, which allows all of the information that we could possible need from an NPI enrichment to be stored in a single, easy to understand object. This also allows us to easily add more fields as needed, without the need for much refactoring.

### Validation


apps/bfd-data-npi/src/main/java/gov/cms/bfd/data/npi/utility/DataUtilityCommons.java
```diff
@@ -271,47 +333,59 @@ private static void buildNPIResource(
...
+                .taxonomyDisplay(taxonomyMap.get(taxonomyCode))
+                .providerFirstName(providerFirstName)
+                .providerMiddleName(providerMiddleName)
+                .providerLastName(providerLastName)
+                .providerNamePrefix(providerPrefix)
+                .providerNameSuffix(providerSuffix)
+                .providerCredential(providerCredential)
+                .build();
+        String json = objectMapper.writeValueAsString(npiData);
+        dos.write((npi + "\t" + json).getBytes());
```
aschey-forpeople, line 385: Since we're in refactoring mode here, I think we way want to also use a CSV writer here instead of hardcoding the tab character. Since we know this NPI data is poorly validated, it's not implausible to think that maybe someone could put a tab character in their org name or something.

PR #2473, opened 2024-10-30
Update CoverageTransformerV2.java

MahiFentaye:
**JIRA Ticket:**
BFD-3703

### What Does This PR Do?

Include Medicaid Eligibility and Entitlement reason extensions for C4DIC POC

### Validation


dondevun:
@MahiFentaye It looks like you need to run `mvn clean install -DgenerateTestData=true` to regenerate the test data.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -690,6 +718,74 @@ private void transformEntitlementDualEligibility(Coverage coverage, Beneficiary
...
 
+  /**
+   * current monthly Medicare-Medicaid dual eligibility code extensions to the provided {@link
+   * Coverage} resource.
+   *
+   * @param coverage the {@link Coverage} to generate
+   * @param beneficiary the {@link Beneficiary} to generate Coverage for
+   * @param currentMonth the value of current month in integer
+   */
+  private void currentMonthTransformEntitlementDualEligibility(
```
dondevun, line 729: This isn't a big deal, so feel free to ignore, but shouldn't the verb come first in the method name?

dondevun (approved):

dondevun (approved):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -274,9 +280,17 @@ private Coverage transformPartB(Beneficiary beneficiary, Profile profile) {
     if (beneficiary.getBeneEnrollmentReferenceYear().isPresent()) {
       addCoverageDecimalExtension(
           coverage, CcwCodebookVariable.RFRNC_YR, beneficiary.getBeneEnrollmentReferenceYear());
-      // Monthly Medicare-Medicaid dual eligibility codes
-      transformEntitlementDualEligibility(coverage, beneficiary);
 
+      if (profile.equals(Profile.C4DIC)) {
```
aschey-forpeople, line 284: All of these conditional checks are getting a bit unwieldy. I'm starting to wonder if C4DIC warrants having it's own transform methods. I don't want to add too much scope to what you already have here, but maybe it's something to consider in a subsequent PR.

MahiFentaye: haha . . . I actually love this comment. I did think of refactoring the methods multiple time just to avoid that . . . just seemed way out of the scope the current story

aschey-forpeople: Yep, it's tricky because the resources are still _mostly_ similar. So we'd have to figure out the best way to reduce duplication, likely by adding more groups of helper methods.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -449,8 +472,10 @@ private Coverage transformPartD(Beneficiary beneficiary, Profile profile) {
       addCoverageDecimalExtension(
           coverage, CcwCodebookVariable.RFRNC_YR, beneficiary.getBeneEnrollmentReferenceYear());
 
-      // Monthly Medicare-Medicaid dual eligibility codes
-      transformEntitlementDualEligibility(coverage, beneficiary);
+      if (!isProfileC4DIC) {
```
aschey-forpeople, line 475: Could this go in the `else` clause of the subsequent conditional to mirror what you have for the part C transformer? That will eliminate one extra conditional.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
```diff
@@ -208,8 +208,14 @@ private Coverage transformPartA(Beneficiary beneficiary, Profile profile) {
     if (beneficiary.getBeneEnrollmentReferenceYear().isPresent()) {
       addCoverageDecimalExtension(
           coverage, CcwCodebookVariable.RFRNC_YR, beneficiary.getBeneEnrollmentReferenceYear());
-      // Monthly Medicare-Medicaid dual eligibility codes
-      transformEntitlementDualEligibility(coverage, beneficiary);
+
+      if (profile.equals(Profile.C4DIC)) {
```
aschey-forpeople, line 212: Maybe make an `isProfileC4IDC` variable here to so it matches the other method?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
MahiFentaye: I tried that but this needs to happen under the following condition
` if (beneficiary.getBeneEnrollmentReferenceYear().isPresent()) ` which in my test cases C4DIC data wasn't satisfying

dondevun (dismissed):

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/CoverageTransformerV2.java
aschey-forpeople: What's the reason for that check? Should the corresponding C4DIC logic also be inside that check?

MahiFentaye: so I was testing a case that I kinda manipulated in a way to view hidden values but looks like this okay with the actual test data. I have tested both the normal and C4DIC cases and this should be okay

dondevun (approved):

aschey-forpeople (dismissed):

aschey-forpeople (approved):

PR #2480, opened 2024-11-05
BFD-3732: Create migration script for SAMHSA 2.0 tables

dondevun:
**JIRA Ticket:**
BFD-3732

### What Does This PR Do?

Adds the necessary tables for the SAMHSA 2.0 filter.

### Validation

Migrator successfully tested on ephemeral database.


apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V20241101111900986__create_samhsa_2_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+    CONSTRAINT fk_inpatient_tags_clm_id_to_inpatient_claims_clm_id
+        FOREIGN KEY (clm_id)
+            REFERENCES ccw.inpatient_claims(clm_id)
+    );
+CREATE INDEX IF NOT EXISTS inpatient_tags_clm_id_idx
+    ON ccw.inpatient_tags(clm_id);
+
+CREATE TABLE IF NOT EXISTS rda.mcs_tags (
+    tag_id BIGINT,
+    clm_id CHARACTER VARYING(15) NOT NULL,
```
MahiFentaye, line 97: any reason `clm_id` data type is `BIGINT` for ccw schema and CHARACTER VARYING(15) for rda?

dondevun: Yes, it has to do with the foreign keys. mcs_claims and fiss_claims use `CHARACTER VARYING` for the claim id type, which necessitates the corresponding key in the tags tables being the same.

MahiFentaye (dismissed):

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V20241101111900986__create_samhsa_2_tables.sql
```diff
@@ -0,0 +1,119 @@
+DROP TYPE IF EXISTS tag_code CASCADE;
+CREATE TYPE tag_code AS ENUM('R', '42CFRPart2');
+
+CREATE TABLE IF NOT EXISTS ccw.hha_tags (
+    tag_id BIGINT,
+    clm_id BIGINT NOT NULL,
+    code TAG_CODE,
```
aschey-forpeople, line 7: Do we need a `NOT NULL` constraint on the `code` column as well?

dondevun: Yes, probably should.

apps/bfd-model/bfd-model-rif/src/main/resources/db/migration/V20241101111900986__create_samhsa_2_tables.sql
```diff
@@ -0,0 +1,119 @@
+DROP TYPE IF EXISTS tag_code CASCADE;
```
aschey-forpeople, line 1: Hmm, I know a simple "create if not exists" statement doesn't exist for types, but I'm not sure how I feel about the potential for accidentally dropping all these tables if this script is erroneously ran. Stack Overflow suggests this might work?

```sql
DO $$
BEGIN
    IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'tag_code') THEN
        CREATE TYPE tag_code AS ENUM('R', '42CFRPart2');
    END IF;
END$$;
```

dondevun: Let me try it.

aschey-forpeople (dismissed):
This looks good, but just FYI you can add the constraints inline and the indexes without an explicit name so you don't have to worry about typos in the constraint and index names.

aschey-forpeople:
I think we may want to point these PRs to a feature branch? Probably don't want to be merging these into `master` yet.

dondevun:
> I think we may want to point these PRs to a feature branch? Probably don't want to be merging these into `master` yet.

Yes, that is probably a good idea. As an aside, this particular PR has been absorbed by the RDA ingestion PR (which is why it's in draft). I think I will create a feature branch, and merge the SAMHSA PRs into that as they get approved.

PR #2492, opened 2024-11-14
BFD-3672: Populate tags for claim / EOB on ingestion from RDA

dondevun:
**JIRA Ticket:**
BFD-3672

### What Does This PR Do?

This is the first stage in the creation of the SAMHSA filter 2.0. This creates the tags for the RDA claims, and creates a new class, SamhsaUtil. This class will allow for the evaluation of the claims for SAMHSA codes as they get saved to the database, and persists the associated SAMHSA tags, if necessary.

### Validation

Tested by running the Pipeline locally. Additionally, Integration tests and Unit tests have been created.


apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -0,0 +1,392 @@
...
+    return entries.isEmpty() ? Optional.empty() : Optional.of(entries);
+  }
+
+  /**
+   * Check if a given code is a SAMHSA code.
+   *
+   * @param code the code to check.
+   * @return If the code is SAMHSA, returns the SAMHSA entry. Otherwise, an empty optional.
+   */
+  public Optional<SamhsaEntry> isSamhsaCode(Optional<String> code) {
```
aschey-forpeople, line 361: Maybe this should be called `getSamhsaCode`? `isSamhsaCode` implies that it returns a boolean.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
```diff
@@ -408,6 +410,7 @@ private void mergeBatch(long maxSeq, Collection<RdaChange<TClaim>> changes) {
                 var metaData = createMetaData(change);
                 entityManager.merge(metaData);
                 entityManager.merge(change.getClaim());
+                samhsaUtil.processClaim(change.getClaim(), entityManager);
```
aschey-forpeople, line 413: I'm being slightly paranoid here, but we may want to consider doing these inserts in an explicit transaction. Theoretically, if a request comes in right when we're inserting the data, they could get the new claims after they're inserted, but before the SAMHSA tags are created, resulting in a leak. If we do it in a transaction, they shouldn't be able to see the new data until after all the inserts have completed.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -0,0 +1,392 @@
...
+   *
+   * @param mcsClaim The claim to check.
+   * @return a list of TagDetail objects, one for each SAMHSA code found in the claim.
+   */
+  private Optional<List<TagDetails>> getPossibleMcsSamhsaFields(RdaMcsClaim mcsClaim) {
+    List<TagDetails> entries = new ArrayList<>();
+    LocalDate serviceDate =
+        mcsClaim.getIdrHdrFromDateOfSvc() == null
+            ? LocalDate.parse("1970-01-01")
+            : mcsClaim.getIdrHdrFromDateOfSvc();
```
aschey-forpeople, line 194: Might be worth checking with @alex-dzeda to make sure we're using the right dates. I recall there was some nuance to this when we were discussing this previously.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -0,0 +1,392 @@
...
+        isSamhsaCode(Optional.ofNullable(fissClaim.getAdmitDiagCode())),
+        "fiss_claims",
+        "admit_diag_code",
+        null,
+        entries,
+        serviceDate,
+        throughDate);
+    for (RdaFissRevenueLine revenueLine : fissClaim.getRevenueLines()) {
+      buildDetails(
+          isSamhsaCode(Optional.ofNullable(revenueLine.getApcHcpcsApc())),
```
aschey-forpeople, line 300: Might be worth a comment here stating that this column should never contain SAMHSA data, but we're checking it in case someone misinterprets this field.

apps/bfd-pipeline/bfd-pipeline-rda-grpc/src/main/java/gov/cms/bfd/pipeline/rda/grpc/sink/direct/AbstractClaimRdaSink.java
dondevun: After a little bit of research, it appears that `executeProcedure` does run in a transaction:

```
Runs a transaction that does not produce a result (usually an insert or update). Ensures that the transaction is properly committed if the procedure completes normally or rolled back if the procedure throws an exception. Internally uses the adapter method ThrowingConsumer.executeAsFunction and discards the null result.
```

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
alex-dzeda: The safest option is to imagine a time period (eg start and end date). If any of those dates is after an applicable start date, I'd consider it to be within range.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -0,0 +1,394 @@
...
+      String column,
+      Integer lineNum,
+      List<TagDetails> detailsList,
+      LocalDate serviceDate,
+      LocalDate throughDate) {
+    if (entry.isPresent()) {
+      try {
+        LocalDate startDate = LocalDate.parse(entry.get().getStartDate());
+        LocalDate endDate =
+            entry.get().getEndDate().equals("Active")
```
aschey-forpeople, line 252: Maybe make this a case-insensitive comparison just to be safe?

dondevun: Good call

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -0,0 +1,396 @@
...
+      Integer lineNum,
+      List<TagDetails> detailsList,
+      LocalDate serviceDate,
+      LocalDate throughDate) {
+    if (entry.isPresent()) {
+      try {
+        LocalDate startDate = LocalDate.parse(entry.get().getStartDate());
+        LocalDate endDate =
+            entry.get().getEndDate().equalsIgnoreCase("Active")
+                ? LocalDate.now()
```
aschey-forpeople, line 253: I'm not sure if defaulting to `now()` is the safest option for an active code since it would cause issues if the claim through date is in the future for some reason. I'd probably use `LocalDate.MAX` here.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -0,0 +1,396 @@
...
+                : LocalDate.parse(entry.get().getEndDate());
+
+        // if the throughDate is not between the start and end date,
+        // and the serviceDate is not between the start and end date,
+        // then the claim falls outside the date range of the SAMHSA code.
+        if ((throughDate.isBefore(startDate) || throughDate.isAfter(endDate))
```
aschey-forpeople, line 259: This looks correct, but it's a bit hard to follow. Maybe it would help with readability to make an `isBetween` helper function?

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
dondevun: Absolutely.

aschey-forpeople (approved):
Thanks for taking this on, this looks good!

PR #2497, opened 2024-11-19
BFD-3663: Populate tags for claim / EOB on ingestion from CCW

dondevun:
**JIRA Ticket:**
BFD-3663

### What Does This PR Do?

Populate SAMHSA tags for claims on ingestion from CCW. 

### Validation

Unit tests and integration tests have been created.


apps/bfd-model/bfd-model-rif/src/main/java/gov/cms/bfd/model/rif/samhsa/CcwTagKey.java
```diff
@@ -0,0 +1,33 @@
+package gov.cms.bfd.model.rif.samhsa;
+
+import java.io.Serializable;
+import java.util.Objects;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+/** The composite key for RDA tags. */
```
aschey-forpeople, line 9:
```suggestion
/** The composite key for CCW tags. */
```

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -153,6 +206,697 @@ private List<McsTag> checkAndProcessMcsClaim(RdaMcsClaim mcsClaim) {
...
+    return Collections.emptyList();
+  }
+
+  /**
+   * Constructs a list of TagDetail objects for a SNF claim.
+   *
+   * @param claim The claim to check.
+   * @return a list of TagDetail objects, one for each SAMHSA code found in the claim.
+   */
+  private Optional<List<TagDetails>> getPossibleSnfFields(SNFClaim claim) {
```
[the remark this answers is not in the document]
dondevun: Yeah, I thought about this a bit... The problem is that there's just enough differences to make any attempt at abstraction cumbersome. Let me think about it some more.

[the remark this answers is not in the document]
dondevun: Did some heavy refactoring, managed to combine most of the logic by using a yaml file that contains the methods for each class, and invoking them using reflection.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/resources/claim_samhsa_methods.yaml
```diff
@@ -0,0 +1,706 @@
+- claimClass: CarrierClaim
```
aschey-forpeople, line 1: Clever solution! My only concern is that we're giving up a lot of compile-time safety by invoking all of this dynamically. Alternatively, we could have an adapter class for each claim type - `HHASamhsaAdapter`, `InpatientSamhsaAdapter`, etc. that would wrap each claim object and implement some interface or wrap some base class that we could operate on using common methods just like what you've done in these yaml files. It would be a bit more boilerplate than using reflection, but I think it would be a bit less prone to accidental breakage. Thoughts?

dondevun: If your concern about compile-time safety is that the methods may not exist, then the integration tests should alleviate that -- Every method in the yaml file is exercised. Or did you have a different concern?

The adapter solution would work, but I think it would be a bit more tightly coupled than the yaml solution. What I mean is that if, in the future, we add new SAMHSA fields, we would need only add two lines into the YAML file (column and method).

aschey-forpeople: If we leave this code as is and never change it, then yes, the tests will catch it. My main concern with these kinds of things is that it's more difficult to refactor and extend safely in the future. I think the concept here makes sense - with the adapter pattern we can still use `getFields()` as you have here and we should only need to add 1-2 lines of code to add the new field there.

dondevun: Been thinking about this, and I've come to agree with what you have been trying to get at: passing the methods along from the adapter with a `Supplier` object may be a better solution for the sake of maintainability Reflection would make the code a lot more compact, but:
1. Too easy to make a mistake, and
2. I'm not convinced that the execution time of invoking a method with reflection is advisable due to the volume of claims that we process at a time.

MahiFentaye (approved):

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -96,19 +127,56 @@ public void createSamhsaMap() throws IOException {
...
-  public <TClaim> void processClaim(TClaim claim, EntityManager entityManager) {
-    switch (claim) {
-      case RdaFissClaim fissClaim -> {
-        Optional<List<FissTag>> tags = Optional.of(checkAndProcessFissClaim(fissClaim));
-        persistTags(tags, entityManager);
-      }
-      case RdaMcsClaim mcsClaim -> {
-        Optional<List<McsTag>> tags = Optional.of(checkAndProcessMcsClaim(mcsClaim));
-        persistTags(tags, entityManager);
+  public <TClaim> boolean processClaim(TClaim claim, EntityManager entityManager) {
```
aschey-forpeople, line 132: Do we need the return value here? Doesn't seem like it's used right now.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -153,6 +225,210 @@ private List<McsTag> checkAndProcessMcsClaim(RdaMcsClaim mcsClaim) {
     return Collections.emptyList();
   }
 
+  /**
+   * Checks for SAMHSA codes in an HHA claim and constructs the tags.
+   *
+   * @param claim The claim to check.
+   * @return A list of tag entities to persist.
+   */
+  private List<HhaTag> checkAndProcessHhaClaim(HHAClaim claim)
```
aschey-forpeople, line 234: I think this logic can operate on an input of `SamhsaAdapterBase` so it doesn't need to be duplicated for each claim type.

dondevun: Maybe? Each tag type would still need to be instantiated individually, so I'm not sure how much value there would be to moving this to the adapter. Can you elaborate on what you're thinking?

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
dondevun: It will be used in the backfill code, to keep track of how many tags were created.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
aschey-forpeople: Yeah, we would probably need a method in the adapter called `buildTag` or something in the adapter that instantiates the tag object from the claim ID, code, and details, but if we did that, we wouldn't need a separate `checkAndProcessXXXClaim` method for each claim type in here and we could get rid of the big switch statement on line 135. I haven't tried it myself, so I could be missing something, but I think that would work?

aschey-forpeople: I'm hoping the adapter can simplify the logic in this file as much as possible so we can eliminate all of the switch statements and confine any duplicated code to the adapters. It won't prevent all the annoying boilerplate, but it at least prevents the need for big switch statements and keeps the main logic separate which I think helps with readability.

dondevun: We would still need to know which adapter to instantiate; we may be able to move the switch logic to a factory class, but I don't think it will be possible to eliminate it completely.

aschey-forpeople: Yeah that's true. It's just one line to instantiate the adapter though so that's not the end of the world.

dondevun: Refactored based on this conversation.

PR #2506, opened 2024-12-03
Bfd 3664

dondevun:
**JIRA Ticket:**
BFD-3664

### What Does This PR Do?

Replace me.

### Validation


apps/bfd-pipeline/bfd-pipeline-app/src/main/java/gov/cms/bfd/pipeline/app/AppConfiguration.java
```diff
@@ -484,6 +508,26 @@ static ConfigLoader createConfigLoaderForTesting(Map<String, String> getenv) {
...
+   *
+   * @param config Config loader
+   * @param ccwPipelineEnabled True if the CCW pipeline is enabled in this instance.
+   * @return {@link BackfillConfigOptions}
+   */
+  public static BackfillConfigOptions loadBackfillConfigOptions(
+      ConfigLoader config, boolean ccwPipelineEnabled) {
+    boolean enabled = config.booleanOption(SSM_PATH_SAMHSA_BACKFILL_ENABLED).orElse(false);
+    // We don't want to run if we're on a CCW Pipeline instance
+    if (!enabled || ccwPipelineEnabled) {
```
aschey-forpeople, line 522: It is a bit confusing that this runs on the RDA pipeline, but I understand the reasoning. Ideally this could run as its own pipeline, but that would increase the complexity a fair bit. I think this is fine for now and we can revisit once we're running in ECS.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
```diff
@@ -0,0 +1,268 @@
...
+ */
+public abstract class AbstractSamhsaBackfill {
+  /** The Logger. */
+  private static final Logger LOGGER = LoggerFactory.getLogger(AbstractSamhsaBackfill.class);
+
+  /**
+   * Query to retrieve a list of claims objects, ignoring claims that already have SAMHSA tags. Will
+   * start at a given claim id, and limit the results to a given limit.
+   */
+  protected String QUERY_WITH_STARTING_CLAIM =
```
aschey-forpeople, line 30: I think my main concern with this is that we could potentially miss claims if new data is being updated while this is running since PAC claim IDs are not monotonic and older claims could be updated after we've already processed them. If we can be sure this runs in under a day or two, perhaps we can just pause the other pipelines while this runs.

aschey-forpeople: The other option would be to key off of `last_updated`, but that would require adding another index to all of the claim tables.

dondevun: I thought this would be a problem at first, as well, but then I realized that any new data that comes in will be processed by the RDA/CCW pipeline (which makes a call to SamhsaUtil before the claims are written).

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
```diff
@@ -0,0 +1,268 @@
...
+  private void saveProgress(
+      String table, Optional<String> lastClaimId, EntityManager entityManager) {
+    Query query = Objects.requireNonNull(entityManager).createNativeQuery(UPSERT_PROGRESS_QUERY);
+    query.setParameter("tableName", table);
+    query.setParameter("lastClaim", lastClaimId.get());
+    query.executeUpdate();
+  }
+
+  /**
+   * Rims a query to get the last claim id processed.
```
dondevun, line 252:
```suggestion
   * executes a query to get the last claim id processed.
```

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
aschey-forpeople: Ah yeah, very true. Should be fine then.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
```diff
@@ -0,0 +1,278 @@
...
+            startingClaim.orElse(""),
+            "tagTableName",
+            tableEntry.getTagTable(),
+            "limit",
+            String.valueOf(limit));
+    StringSubstitutor strSub = new StringSubstitutor(params);
+    String queryStr =
+        strSub.replace(
+            startingClaim.isPresent() ? QUERY_WITH_STARTING_CLAIM : QUERY_WITH_NO_STARTING_CLAIM);
+    return entityManager.createNativeQuery(queryStr, tableEntry.getClaimClass());
```
aschey-forpeople, line 141: Would it be possible to use `createQuery` instead of `createNativeQuery`? That will at least perform some type checking to ensure the type is assignable to the query result. I believe that would remove the need for `select *` as well.

ops/terraform/services/base/values/test.yaml
```diff
@@ -152,6 +152,8 @@
 /bfd/${env}/pipeline/nonsensitive/rda/cleanup/run_size: UNDEFINED
 /bfd/${env}/pipeline/nonsensitive/rda/cleanup/transaction_size: UNDEFINED
 /bfd/${env}/pipeline/nonsensitive/rda/instance_type: m6a.large
+/bfd/${env}/pipeline/nonsensitive/rda/samhsa/backfill/enabled: "false"
+/bfd/${env}/pipeline/nonsensitive/rda/samhsa/backfill/batch_size: 15000
```
aschey-forpeople, line 156: Might want to test with a larger batch size here to see if it helps. 100,000 or so shouldn't be a problem.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
dondevun: Yeah, we can do that. The check for existing tags may be a bit tricky, but I'll see what I can do.

dondevun: I think these changes should work fine. Testing performance now.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/QueryConstants.java
```diff
@@ -0,0 +1,43 @@
...
+      "SELECT stmt_cov_from_date, stmt_cov_to_date from ${claimTable} where ${claimField} = :claimId ";
+
+  /** Gets the date ranges for a claim in a line item table from its parent table. */
+  public static final String GET_CLAIM_DATES_MCS =
+      "SELECT idr_hdr_from_date_of_svc, idr_hdr_to_date_of_svc from ${claimTable} where ${claimField} = :claimId ";
+
+  /** Query to perform upsert on the backfill progress table for a given claim table. */
+  public static final String UPSERT_PROGRESS_QUERY =
+      """
+      INSERT INTO ccw.samhsa_backfill_progress
```
aschey-forpeople, line 31: I suppose we have to remember to reset the progress table each time we want to run a new backfill job. I was thinking we could tie it to a job ID or something that can increment each time, but that's probably not worth the complexity.

aschey-forpeople (approved):

MahiFentaye (approved):

PR #2509, opened 2024-12-10
BFD-3789: Minimal alignment of server-load infrastructure with BFD-2588

timothy-ellis-ky:
**JIRA Ticket:**
BFD-3789

### What Does This PR Do?

Aligns server-load infrastructure with ASG improvements implemented with BFD-2588

* missing resource updates 

* [X] I have considered the above security implications as it relates to this PR.  

### Validation

Draft until identifying steps to verify all changes

### DIFF
--- a/ops/terraform/services/server/server-load/data-sources.tf
+++ b/ops/terraform/services/server/server-load/data-sources.tf
@@ -66,9 +66,18 @@ data "aws_launch_template" "template" {
   name = "bfd-${local.env}-fhir"
 }
 
-data "aws_autoscaling_group" "asg" {
-  name = "${data.aws_launch_template.template.name}-${data.aws_launch_template.template.latest_version}"
+# BFD-3789
+# data "aws_autoscaling_group" "asg" {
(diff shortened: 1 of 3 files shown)


malessi (changes requested):
Thanks for taking a crack at this. I _do_ think the `server-load` Terraservice (and corresponding Python) needs some serious care, _especially_ after BFD-2558, but I feel that the changes required to get it up to shape are more extensive than what is covered here.

Rather than spending time doing that, I think you would be better served by running a locally distributed Locust load test using the Regression Suite tests using the `apps/utils/locust_tests/run_locally_distributed.py` Python script. I've personally used it to achieve 100 simulated users at up to 1000 req/s without overwhelming my local machine. I suspect that level of load would be more than sufficient for evaluating the performance of `pgAudit`.

I think BFD-3789 can be closed for the time being until fixing the Load Tests as a whole is higher priority.

timothy-ellis-ky:
Deferring this work until future reassessment

PR #2518, opened 2025-01-08
BFD-3126 + BFD-3129: Pipeline manages its own scale-in + remove s3 move logic

aschey-forpeople:
**JIRA Ticket:**
BFD-3126

### What Does This PR Do?

**Note:** BFD-3129 had to be combined with this one because some of the changes here require that the manifest files remain in the Incoming folder until the load is completed. This change was required in this PR in order for the integration tests to pass.

- Updates the CCW pipeline to modify the ASG capacity from 1 to 0 once it has determined there is no more work to do
  - Only happens once we've received a `ManifestList.done` file
  - On receiving this file, we ensure all manifest files are done processing and every manifest in the manifest list is accounted for
  - Also check that every non-synthetic data load has a manifest list file
  - Termination signal is propagated through the pipeline's signal handlers to ensure all cleanup logic is executed
  - New permissions are given to the pipeline so it can modify its ASG appropriately
- Removes scale-in logic from the scheduler lambda since it's no longer needed
- Adds logic to the scheduler to treat the new manifest list as a valid manifest file
  - This isn't super important, but it does allow the pipeline to be scaled out when we receive a manifest list while the pipeline is not already scaled out. This shouldn't happen in normal circumstances, but it is useful for testing. 
- Adds a new script for converting a synthetic data load to a non-synthetic data format
  - Useful for testing different scenarios using synthetic data in the test environment
- Removes all S3 move logic

Note that the manifest list file will not be moved to the "done" folder since we're removing the move logic soon anyway.

### Validation

Unit/integration tests, manually tested both synthetic and non-synthetic loads in an ephemeral environment.


apps/utils/synthetic-load-converter/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.poetry]
```
aschey-forpeople, line 1: I decided to go rogue and use poetry here. Overall I think it's much easier to manage than messing with `pip-compile`.

.github/pull_request_template.md
```diff
@@ -52,5 +52,5 @@ Please indicate if this PR does any of the following:
 
 ### Validation
 
-Have you fully verified and tested these changes? Is the acceptance criteria met? Please provide reproducible testing instructions, code snippets, or screenshots as applicable.
+**Have you fully verified and tested these changes? Is the acceptance criteria met? Please provide reproducible testing instructions, code snippets, or screenshots as applicable.**
```
aschey-forpeople, line 55: Decided to bold this bit so it's easier to differentiate the question from the response

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadJob.java
```diff
@@ -77,6 +79,12 @@ public final class CcwRifLoadJob implements PipelineJob {
    */
   public static final Duration MAX_MANIFEST_AGE = Duration.ofDays(60);
 
+  /**
+   * The date when the final manifest list was implemented. Pipeline runs before this date won't
+   * have a manifest list.
+   */
+  private static final Instant FINAL_MANIFEST_CUTOFF = Instant.parse("2024-12-12T00:00:00Z");
```
dondevun, line 86: This has a magic number feel, to me. Not sure how else to implement, though.

aschey-forpeople: We can remove it if/when we get rid of the old manifest loads, but it's necessary for now because manifest loads older than that don't have a manifest list file.

aschey-forpeople: There's similar logic elsewhere to handle older records before some feature was implemented.

dondevun (approved):

MahiFentaye (dismissed):

dondevun (approved):

apps/utils/scripts/run-bfd-server
```diff
@@ -181,12 +181,15 @@ function setvar() {
...
-    echo export $env_name
+    env_name=$(echo "BFD_${ssm_name}" | tr '/' '_' | tr '[a-z]' '[A-Z]')
+    echo export "$env_name"
     eval "export ${env_name}=\"${value}\""
   else
     full_ssm_name="${SSM_PARAMETER_PATH}/${ssm_name}"
-    echo put-parameter $full_ssm_name
+    # some values may need to be escaped to be used as an env var
+    # but should be unescaped in SSM
+    value=$(echo "${value}" | tr -d '\')
```
aschey-forpeople, line 191: Unrelated, had to make this change so that variables with quotes can be used in SSM and environment variables

apps/bfd-pipeline/bfd-pipeline-ccw-rif/src/main/java/gov/cms/bfd/pipeline/ccw/rif/CcwRifLoadJob.java
```diff
@@ -421,6 +394,43 @@ public void close() throws Exception {
...
+        .filter(id -> !id.manifestId().isFutureManifest())
+        .map(id -> id.manifestId().getTimestamp())
+        .collect(Collectors.toSet());
+  }
+
+  /**
+   * Returns the flattened list of manifests from the manifest lists.
+   *
+   * @param finalManifestLists final manifest lists from the 'Incoming' folder
+   */
```
malessi, line 415:
```suggestion
   * @param finalManifestLists final manifest lists from the 'Incoming' folder
   * @return set of manifests from all final manifest lists in 'Incoming'
   */
```

Looks like this javadoc is missing a `@return`.

malessi (approved):

PR #2534, opened 2025-01-29
BFD-3565: Simplify BFD Server error alerting infrastructure

malessi:
**JIRA Ticket:**
BFD-3565

### What Does This PR Do?

**Due to resource name changes, the `bfd_server_error_alerts` module will need to be destroyed and then re-`apply`'d in all environments. This has been done in `test`, and will be done out-of-band in all other environments once this PR is merged.**

This PR greatly simplifies the BFD Server's `500` errors alerting solution and infrastructure. The "scheduler Lambda" has been removed outright along with its supporting Terrraform. The "alerter Lambda"'s Terraform has been simplified and its architecture updated to `arm64` for (likely negligible) reduced cost. See diagram below for more details.

For context, there was originally some fear that running Log Insights queries every 5 minutes to check for errors on our logs would be prohibitively expensive, so we introduced some complexity to the alerting infrastructure to ensure that we were only running those queries when there were errors. Sometime after that a change was made that inadvertently disabled that safeguard such that the Log Insights queries (specifically, the "alerter Lambda") began executing, in every environment, every 5 minutes regardless of whether there were errors. We have since noticed no real increase in CloudWatch spend. So, we realized that this complexity was unnecessary and could be removed, resulting in this PR.

#### New Infrastructure Diagram

bfd_server_error_alerts_diagram

### Validation

* `terraform apply`ing these changes to `server` in `test`, _verifying that_:
  * Changes are `apply`'d successfully with no errors
  * All resources are created as expected
* Waiting 5 minutes for the `rate()` schedule to execute the alerter Lambda, _verifying that_ the Lambda executes as expected and there are no errors


.github/scripts/pre-commit.sh
```diff
@@ -62,7 +62,7 @@ runShellCheckForCommitFiles() {
 
       # Skip binary formats
       case "$extension" in
-        "zip" | "p12" | "pfx" | "cer" | "pem")
+        "zip" | "p12" | "pfx" | "cer" | "pem" | "png" | "jpg")
```
malessi, line 65: Trying to commit the changes to the diagram PNG resulted in an `invalid byte sequence` error from `sed`, so I opted to exclude some image types from pre-commit hooks entirely.

aschey-forpeople: Sounds like something that someone putting MBIs in a JPEG would say...

aschey-forpeople (approved):

.github/scripts/pre-commit.sh
malessi: Clearly what we _actually_ need is some form of pre-commit checks for image steganography 😉

MahiFentaye (approved):

PR #2542, opened 2025-02-05
BFD-3848: Transient SFTP connection errors are affecting BFD EFT Outbound in test

malessi:
**JIRA Ticket:**
BFD-3848

### What Does This PR Do?

This PR updates the maximum Lambda retries to 3 instead of disabling retries outright. Initially, the fear was that this Lambda was not idempotent, but after some investigation we determined that it is actually idempotent and the Lambda as a whole can be retried. Accordingly, the CloudWatch Alarms alerting on failures for this Lambda has been updated to alert after 3 failures in a 10 minute period so that the on-call is not alerted if the Lambda is able to recover automatically.

### Validation

* `terraform apply` `eft` in `test`, _verifying that_ the `apply` is successful and all resources are updated

### DIFF
--- a/ops/terraform/services/eft/main.tf
+++ b/ops/terraform/services/eft/main.tf
@@ -316,8 +316,8 @@ resource "aws_lambda_function_event_invoke_config" "sftp_outbound_transfer" {
   function_name          = one(aws_lambda_function.sftp_outbound_transfer[*].function_name)
   maximum_retry_attempts = 2
 
-  # On failure we want failing events to land into a DLQ such that responding engineers can analyze
-  # the event and retry, if necessary
+  # If the Lambda exhausts all of its retry attempts, we want failing events to land into a DLQ such
+  # that responding engineers can analyze the event and retry, if necessary
   destination_config {
     on_failure {
       destination = one(aws_sqs_queue.sftp_outbound_transfer_dlq[*].arn)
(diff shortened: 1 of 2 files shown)


ops/terraform/services/eft/main.tf
```diff
@@ -313,12 +313,8 @@ resource "aws_sqs_queue" "sftp_outbound_transfer_dlq" {
 resource "aws_lambda_function_event_invoke_config" "sftp_outbound_transfer" {
   count = length(local.eft_partners_with_outbound_enabled) > 0 ? 1 : 0
 
-  function_name = one(aws_lambda_function.sftp_outbound_transfer[*].function_name)
-  # This Lambda is invoked by SNS, which invokes the Lambda asynchronously. By default, AWS Lambda
-  # retries failing Functions twice before dropping the event, but because this Lambda has side
-  # effects we don't want to retry if it fails. Instead, we will drop failing events into a DLQ for
```
aschey-forpeople, line 319: Will the DLQ still be used if the retries are exhausted?

aschey-forpeople (dismissed):

ops/terraform/services/eft/main.tf
malessi: Yep, it will be. I can reword this comment to make it explicit that the DLQ will contain the message if retries are exhausted instead of outright removing it.

malessi: I've reworded a comment below this one to indicate that the DLQ will contain the failing event after all retries are exhausted in [14d2539](https://github.com/CMSgov/beneficiary-fhir-data/pull/2542/commits/14d2539db326cb3c8350f651065a4fe220eb7d3f).

PR #2548, opened 2025-02-12
BFD-3854: IDR pipeline POC

aschey-forpeople:
**JIRA Ticket:**
BFD-3854

### What Does This PR Do?

Creates the initial POC for the IDR pipeline. The Snowflake connector is not fully set up yet since we don't have access.
In the meantime, we can mock the IDR schema directly in Postgres for testing purposes. I've currently included a small subset of the data and simplified queries so we can quickly make changes without dealing with the large number of columns that we will eventually need to handle.

The current approach used here is to stream data out of IDR in batches and load them into Postgres using `COPY`. It's possible that this approach will change once we are able to do performance testing with real data.

There is a simple integration test included to verify the data load is working on a basic level.

### Validation

Tests included.


apps/bfd-pipeline/bfd-pipeline-idr/model.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import BaseModel
+
+
+class IdrBeneficiary(BaseModel):
```
aschey-forpeople, line 4: pydantic's model classes make it easy to work with typed data coming out of the database. It does introduce some overhead due to runtime type checking though, so we may want to tweak this in the future if the performance hit is large.

apps/bfd-pipeline/bfd-pipeline-idr/extractor.py
```diff
@@ -0,0 +1,58 @@
...
+            batch: list[T] = cur.fetchmany(self.batch_size)
+            while len(batch) > 0:
+                yield batch
+                batch = cur.fetchmany(self.batch_size)
+
+
+class SnowflakeExtractor(Extractor):
+    def __init__(self, batch_size: int):
+        super().__init__()
+        self.conn = snowflake.connector.connect(user="", password="", account="")
```
aschey-forpeople, line 41: This is untested for now because we don't have Snowflake access. This will be used for prod data and the `PostgresExtractor` above can be used for test data.

malessi (approved):

mjburling (approved):

PR #2557, opened 2025-02-19
BFD-3886: Fix logging, permissions in MGMT Action

TimRotenSB:
**JIRA Ticket:**
BFD-3886

### What Does This PR Do?

This pull request corrects a bug found in BFD-3873 regarding the github actions role permissions. Additionally, it renames the log groups to mirror the updated github action name and adjusts a quicksight contributor to their proper role (over-permissioning occured during the prior PR by mistake)

- Validate the changes do what we're claiming they do
- Validate terraform infrastructure plans found under the associated log groups
- Verify formatting complies with team and terraform standards

### Validation


malessi (approved):

ops/terraform/env/mgmt/cloudwatch.tf
```diff
@@ -1,5 +1,5 @@
 # CloudWatch Log Group for the "CI - Update OPS Infrastructure" Terraform plan/apply logs
-resource "aws_cloudwatch_log_group" "gha_ci_ops_infra" {
-  name       = "/bfd/${local.env}/gha/ci-ops-infra"
+resource "aws_cloudwatch_log_group" "cd-terraform-mgmt-deploy" {
+  name       = "/bfd/${local.env}/gha/cd-terraform-mgmt-deploy"
```
mjburling, line 3: Do we intend on performing a _little_ bit of surgery to ensure we don't destroy the logs affiliated with the existing log group?

mjburling (commented):
This is a lot cleaner/clearer than the initial attempt.

@dondevun Maybe along with the #2554, should the wording on the PR... [the older, more complex template](https://github.com/CMSgov/beneficiary-fhir-data/blob/3e5d7be10bb8f657b55d2fccfc0a16d21ee16d4e/.github/pull_request_template.md#what-needs-to-be-merged-and-deployed-before-this-pr) used to include a section that dealt with e.g. out-of-band application of terraform and/or one-time, manual manipulation of the terraform state. I don't want to return to the old template, but maybe there's still something missing from the new template?

ops/terraform/env/mgmt/cloudwatch.tf
TimRotenSB: Research says that the correct way to do this would be to use a lambda with cloudwatch log permissions that reads each log stream and streams it to the new log group. 

What would be your recommendation? I see a few potentials:

1. The option above.. Would require we write the lambda to consume the existing logs and the ingestion time on the logs would change. 
2. Leave the old log group in place and just make a completely new one labeled like the one above. (only downside being actually remembering where and what the old log group is for long term)

How much value do we see from keeping these? What's been the history of going back 1 or 2 months for these?

mjburling: > What would be your recommendation @mjburling [..]

I would suggest that we simply remove the old log group from state and proceed. The days of the AWS account that this impacts today are numbered.

> How much value do we see from keeping these? What's been the history of going back 1 or 2 months for these?

I don't believe these are critical, but I trust that you'd agree that maintaining _some_ history of terraform's automated interactions between GitHub and the BFD AWS account retains some value for security/audit purposes...

mjburling (approved):

malessi (approved):

PR #2561, opened 2025-02-20
BFD-3875 updating claim query to also get the Security Tags, adding a shadow flag to test Samhsa 2.0 changes with ConsentInterceptorSimulation

MahiFentaye:
**JIRA Ticket:**
BFD-3875

adding a shadow flag to test Samhsa 2.0 changes with ConsentInterceptorSimulation

### Validation

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
+++ b/apps/bfd-server/bfd-server-war/src/test/resources/endpoint-responses/v2/eobByPatientIdAll.json
@@ -16,12 +16,7 @@
       "id" : "outpatient-1234567890",
       "meta" : {
         "lastUpdated" : "IGNORED_FIELD",
-        "profile" : [ "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-ExplanationOfBenefit-Outpatient-Institutional" ],
-        "security" : [ {
-          "system" : "http://terminology.hl7.org/CodeSystem/v3-Confidentiality",
-          "code" : "N",
-          "display" : "Normal"
-        } ]
+        "profile" : [ "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-ExplanationOfBenefit-Outpatient-Institutional" ]
       },
       "contained" : [ {
         "resourceType" : "Organization",
@@ -903,12 +898,7 @@
       "id" : "dme-2188888888",
       "meta" : {
         "lastUpdated" : "IGNORED_FIELD",
-        "profile" : [ "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-ExplanationOfBenefit-Inpatient-Institutional" ],
-        "security" : [ {
-          "system" : "http://terminology.hl7.org/CodeSystem/v3-Confidentiality",
-          "code" : "N",
-          "display" : "Normal"
-        } ]
+        "profile" : [ "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-ExplanationOfBenefit-Inpatient-Institutional" ]
       },
       "contained" : [ {
         "resourceType" : "Observation",
@@ -1639,12 +1629,7 @@
       "id" : "hha-2925555555",
       "meta" : {
         "lastUpdated" : "IGNORED_FIELD",
-        "profile" : [ "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-ExplanationOfBenefit-Professional-NonClinician" ],
-        "security" : [ {
-          "system" : "http://terminology.hl7.org/CodeSystem/v3-Confidentiality",
-          "code" : "N",
-          "display" : "Normal"
-        } ]
+        "profile" : [ "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-ExplanationOfBenefit-Professional-NonClinician" ]
       },
       "contained" : [ {
         "resourceType" : "Organization",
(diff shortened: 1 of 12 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/SecurityTagManager.java
```diff
@@ -72,50 +72,55 @@ public Set<String> queryTagsForClaim(String claimId, Class<?> tagClass) {
    * @return SecurityLevel
    */
   public List<Coding> getClaimSecurityLevel(String claimId, Class<?> tagClass) {
+    if (!samhsaV2Enabled) {
+      return new ArrayList<>();
+    }
+
+    List<String> securityTags = queryTagsForClaim(claimId, tagClass).stream().toList();
```
aschey-forpeople, line 79: This isn't strictly related to this PR, but I just noticed that this is going to be doing a separate DB query for each claim. That might be okay since the query is simple, but some EOB resources will have a lot of claims which means a lot of roundtrips to the database. We may want to think about loading these in a single query.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
```diff
@@ -470,6 +477,13 @@ public Bundle findByPatient(
                 getResourceTypes(), mbiString, lastUpdated, serviceDate, paging, bundleOptions);
       }
 
+      // This is evaluating the new Samhsa2.0 data scrubbing compared to the old one
+      // a difference will be logged as error
+      IBaseResource resourceCopy = bundleResource.copy();
+      Bundle v2SamhsaScrubbedResource =
+          (Bundle) v2SamhsaConsentSimulation.simulateScrubbing(requestDetails, resourceCopy);
+      v2SamhsaConsentSimulation.logMissingClaimIds(v2SamhsaScrubbedResource, bundleResource);
+
```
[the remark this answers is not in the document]
aschey-forpeople: Thinking about this some more... rather than trying to compare the resources after they've been modified, it might be easier/more performant to just compare the outputs of the functions that say whether the claim has samhsa data. For the old code, that would be `samhsaMatcher.test(claim)` and `shouldRedactResource(resource)` for the new one. It doesn't exercise the entire flow, but I think we can be reasonably sure the rest of the code works as long as the logic that determines if SAMHSA data is present works correctly.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimDao.java
```diff
@@ -135,6 +146,132 @@ public <T> List<T> findAllByMbiAttribute(
...
+    List<ClaimWithSecurityTagsV2<T>> claimEntitiesWithTags = new ArrayList<>();
+    Long claimId = null;
+    String mcsFissClaimId = null;
+    String query = null;
+
+    if (claimEntities != null) {
+      // Iterate over each claim and get its security tags
+      for (T claimEntity : claimEntities) {
+        // Identify claim type and get the claimId
+        switch (claimEntity) {
```
aschey-forpeople, line 215: This is still issuing a separate query for each claim. What we probably want to do is something like `SELECT t.code FROM table where t.claim in (:claimIds)` which will give a list of all codes in a single query.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsV2.java
```diff
@@ -0,0 +1,22 @@
+package gov.cms.bfd.server.war.r4.providers.pac.common;
+
+import java.util.Set;
+import lombok.Getter;
+
+/** DAO of ClaimWithSecurityTagsV2. */
+@Getter
+public class ClaimWithSecurityTagsV2<T> {
```
aschey-forpeople, line 8: I don't think we need the version suffix here. We can use the same class for V1 and V2.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsV2.java
```diff
@@ -0,0 +1,22 @@
...
+  private T claimEntity;
+  private Set<String> securityTags;
+
+  /**
+   * The tag ClaimWithSecurityTagsV2.
+   *
+   * @param claimEntity claimEntity
+   * @param securityTags securityTags
+   */
+  public ClaimWithSecurityTagsV2(T claimEntity, Set<String> securityTags) {
```
aschey-forpeople, line 18: Can use `@AllArgsConstructor` instead of defining this manually

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PatientClaimsEobTaskTransformerV2.java
```diff
@@ -368,15 +403,108 @@ private <T> List<T> findClaimTypeByPatient() {
...
+
+    List<Object[]> results = new ArrayList<>();
+
+    if (tagTable != null) {
+      String query =
+          String.format("SELECT t.claim, t.code FROM %s t WHERE t.claim IN :claimIds", tagTable);
+
+      results =
+          entityManager
+              .createQuery(query, Object[].class)
```
aschey-forpeople, line 463: I think you can use `String[]` instead of `Object[]` here

PR #2575, opened 2025-03-05
BFD-3921: Cached Truststore/Keystore for Server Container Workloads

malessi:
**JIRA Ticket:**
BFD-3921

### What Does This PR Do?

This PR introduces resources (in `certstores.tf`) and external scripts (in `scripts/`) to generate the BFD Server's truststore and keystore and store them in a new, per-environment S3 Bucket, `bfd-<env>-server-certstores`. These certstores will then be pulled by the containerized BFD Server workload so that it does not need to spend time on startup generating them.

These certstores will not be generated _unless_ the component certificates have changed _or_ the store in S3 was modified out-of-band, even if the certstores do not exist on the machine locally. This ensures deployments do not spend time generating these files when it is not necessary.

### Validation

* `terraform apply`ing these changes to `test`'s `server`, _verifying that_:
  * The certstores are created, as expected
  * Repeated `apply`s with no changes to the component certificates do not result in any changes to related resources
  * Out-of-band modifications to the certstores in S3 result in their regeneration
  * Removing the `generated/truststore.pfx` file without modifying the corresponding S3 object or component certificates results in no change to its resources
* Using `keytool` to inspect the truststore and keystore, _verifying that_ both stores are valid and include the expected certificate data


ops/terraform/services/server/certstores.tf
```diff
@@ -0,0 +1,143 @@
...
+  keystore_s3_key     = "keystore.pfx"
+  keystore_base64     = local.sensitive_service_config["server_keystore_base64"]
+}
+
+data "aws_ssm_parameters_by_path" "client_certificates" {
+  path = "/bfd/${local.env}/${local.service}/nonsensitive/client_certificates"
+}
+
+resource "aws_s3_bucket" "certstores" {
+  bucket        = "bfd-${local.env}-${local.service}-certstores"
```
mjburling, line 27: What do you think about using `bucket_prefix` instead of `bucket` in this instance? Probably ought to be what we do going forward. I don't think adopting that here would require any coordination/orchestration beyond this module... ought to be safe.

mjburling (changes requested):
I've left an in-line comment on `bucket_prefix` for consideration. Beyond that, maybe we consider adjusting the configuration for terraform docs to include publishing the usage of modules and regenerating?

ops/terraform/services/server/certstores.tf
malessi: Yeah, that's totally fine with me. I'll make the change.

malessi: I've made this change in [2e59e2c](https://github.com/CMSgov/beneficiary-fhir-data/pull/2575/commits/2e59e2cfb191d19562c72001588888de4813fb0c), and have re-`apply`'d `server` to `test` making sure that it `apply`s just fine.

mjburling (approved):

aschey-forpeople (approved):

PR #2576, opened 2025-03-05
BFD-3875: Refactor Security Tag Query and Introduce SAMHSA V2 Shadow Flag

MahiFentaye:
**JIRA Ticket:**
BFD-3875

### What Does This PR Do?

This pull request refactors the process of retrieving security tags for claims. Instead of fetching them individually for each claim, we now retrieve all security tags at once and associate them with the claims in a new `ClaimWithSecurityTags` class. This change improves efficiency by reducing database calls.

Additionally, we are introducing a Shadow flag `samhsaV2Shadow`, which is currently set to `true`. When enabled, the Shadow flag will ensure that even if the `Samhsa2.0` flag is off, the system will still compare whether a claim contains restricted SAMHSA data. It will perform this comparison between the old check and the new one, and log the claim IDs if the results different.

### Validation


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/ClaimResponseTypeV2.java
```diff
@@ -33,7 +41,85 @@ public final class ClaimResponseTypeV2<TEntity>
           RdaMcsClaim.class,
           RdaMcsClaim.Fields.mbiRecord,
           RdaMcsClaim.Fields.idrClmHdIcn,
-          List.of(RdaMcsClaim.Fields.idrHdrFromDateOfSvc, RdaMcsClaim.Fields.idrHdrToDateOfSvc));
+          List.of(RdaMcsClaim.Fields.idrHdrFromDateOfSvc, RdaMcsClaim.Fields.idrHdrToDateOfSvc),
+          "McsTag");
+
+  /** Instance for Carrier claims. */
+  public static final ClaimResponseTypeV2<CarrierClaim> CARRIER =
```
aschey-forpeople, line 48: I don't think you need these. This file is just for PAC data (FISS/MCS).

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/ClaimTypeV2.java
```diff
@@ -32,7 +40,85 @@ public final class ClaimTypeV2<TEntity> extends AbstractResourceTypeV2<Claim, TE
           RdaMcsClaim.class,
           RdaMcsClaim.Fields.mbiRecord,
           RdaMcsClaim.Fields.idrClmHdIcn,
-          List.of(RdaMcsClaim.Fields.idrHdrFromDateOfSvc, RdaMcsClaim.Fields.idrHdrToDateOfSvc));
+          List.of(RdaMcsClaim.Fields.idrHdrFromDateOfSvc, RdaMcsClaim.Fields.idrHdrToDateOfSvc),
+          "McsTag");
+
+  /** Instance for Carrier claims. */
+  public static final ClaimTypeV2<CarrierClaim> CARRIER =
```
aschey-forpeople, line 47: Same here, shouldn't need these for PAC

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,183 @@
...
+
+    List<Object[]> results = new ArrayList<>();
+
+    if (tagTable != null) {
+      String query =
+          String.format("SELECT t.claim, t.code FROM %s t WHERE t.claim IN :claimIds", tagTable);
+
+      results =
+          entityManager
+              .createQuery(query, Object[].class)
```
aschey-forpeople, line 80: I think you can use `String[]` instead of `Object[]` here

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Provides common logic for performing DB interactions. */
+public class ClaimWithSecurityTagsDao {
```
aschey-forpeople, line 17: This is under `r4/providers/pac` but there shouldn't be anything v2 or PAC-specific in here. I think this can go under `common`.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  /**
+   * Builds a mapping from claim IDs to their security tags.
+   *
+   * @param claimEntities the claim entity
+   * @param claimIds The list of claim IDs
+   * @param entityIdAttribute the claim type
+   */
+  public void collectClaimIds(
```
aschey-forpeople, line 133: I think it might make the intent clearer if this returns the set of claim IDs instead of modifying the parameter that's passed in.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,183 @@
...
+
+        // Get the value of the entityIdField
+        Object claimIdValue = entityIdField.get(claimEntity);
+
+        // If a valid claim ID is found, add it to the claimIds list
+        if (claimIdValue != null) {
+          claimIds.add(claimIdValue.toString());
+        }
+      } catch (NoSuchFieldException e) {
+        LOGGER.debug("Field '{}' not found for claim entity: {}", entityIdAttribute, claimEntity);
```
aschey-forpeople, line 150: Would we expect this error condition ever happen? Perhaps `debug` isn't severe enough of a log level.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,183 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(ClaimWithSecurityTagsDao.class);
+
+  /**
+   * Builds a mapping from claim IDs to their security tags.
+   *
+   * @param claimEntities the claim entity
+   * @param claimIds The list of claim IDs
+   * @param claimType the claim type
+   */
+  public void collectClaimIds(
```
aschey-forpeople, line 28: This method doesn't perform any database queries, so I think it may not belong in this class.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/PatientClaimsEobTaskTransformer.java
```diff
@@ -199,7 +213,15 @@ public void setIncludeTaxNumbers(boolean includeTaxNumbers) {
...
     LOGGER.debug("TransformPatientClaimsToEobTaskpwd.call() started for {}", id);
     try {
-      eobs.addAll(transformToEobs(findClaimTypeByPatient()));
+      List<ClaimWithSecurityTags<T>> claims = findClaimTypeByPatient();
+      eobs.addAll(transformToEobs(claims));
+
+      for (ExplanationOfBenefit eob : eobs) {
+        boolean samhsaMatcherTest = samhsaMatcher.test(eob);
+        // Log missing claim for samhsa V2 Shadow check
+        v2SamhsaConsentSimulation.logMissingClaim(claims, samhsaMatcherTest);
```
aschey-forpeople, line 222: I think this isn't quite going to work the way it's written here. It's comparing one EOB object against all of the claims when it should be comparing each claim one by one. I think you may want to modify `transformToEobs` and then do the check per claim in that method.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DMEClaimTransformerV2.java
```diff
@@ -64,38 +69,47 @@ final class DMEClaimTransformerV2 implements ClaimTransformerInterfaceV2 {
...
    */
   @Trace
   @Override
-  public ExplanationOfBenefit transform(Object claim, boolean includeTaxNumber) {
+  public ExplanationOfBenefit transform(Object claimEntity, boolean includeTaxNumber) {
+
+    Object claim = claimEntity;
+    List<Coding> securityTags = new ArrayList<>();
+
+    if (claimEntity instanceof ClaimWithSecurityTags<?> claimWithSecurityTags) {
```
aschey-forpeople, line 102: Do we still need this conditional?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DMEClaimTransformerV2.java
```diff
@@ -117,7 +131,9 @@ private ExplanationOfBenefit transformClaim(
     // Required values not directly mapped
     eob.getMeta().addProfile(Profile.C4BB.getVersionedEobInpatientUrl());
 
-    eob.getMeta().setSecurity(securityTags);
+    if (samhsaV2Enabled) {
```
aschey-forpeople, line 134: Is this check needed? I thought it would just return an empty array if samhsa v2 is not enabled.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PatientClaimsEobTaskTransformerV2.java
```diff
@@ -342,6 +366,42 @@ private <T> List<T> findClaimTypeByPatient() {
...
+
+    if (!claimIds.isEmpty()) {
+      // Make ONE query to get all claim-tag relationships
+      Map<String, Set<String>> claimIdToTagsMap =
+          claimWithSecurityTagsDao.buildClaimIdToTagsMap(
+              claimType.getEntityTagType(), claimIds, entityManager);
+
+      // Process all claims using the map from the single query
+      claimEntities.stream()
+          .forEach(
```
aschey-forpeople, line 388: Can use `map` instead of `forEach` here.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimDao.java
```diff
@@ -88,12 +96,13 @@ public <T> T getEntityById(ResourceTypeV2<?, T> resourceType, String id) {
    * @return A list of entities of type T retrieved matching the given parameters.
    */
   @Trace
-  public <T> List<T> findAllByMbiAttribute(
+  public <T> List<ClaimWithSecurityTags<T>> findAllByMbiAttribute(
```
aschey-forpeople, line 99: The other methods to fetch claims also need to return `ClaimWithSecurityTags`. There's another method in here `getEntityById` that needs to be modified.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsDao.java
aschey-forpeople: I think we should probably log an error and re-throw

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/DMEClaimTransformerV2.java
aschey-forpeople: Ah that's right, makes sense.

apps/bfd-server/bfd-server-war/src/test/java/gov/cms/bfd/server/war/r4/providers/CarrierClaimTransformerV2Test.java
```diff
@@ -136,7 +136,8 @@ public void before() throws IOException {
             metricRegistry,
             drugCodeDisplayLookup,
             NPIOrgLookup.createTestNpiOrgLookup(),
-            securityTagManager);
+            securityTagManager,
```
aschey-forpeople, line 139: I see the existing tests got updated which is good. Do we have any tests for the new functionality?

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,114 @@
...
+   * @param entityIdAttribute the claim type
+   * @return set of ClaimIds
+   */
+  public Set<String> collectClaimIds(List<Object> claimEntities, String entityIdAttribute) {
+    Set<String> claimIds = new HashSet<>();
+    for (Object claimEntity : claimEntities) {
+
+      try {
+        Field entityIdField = claimEntity.getClass().getDeclaredField(entityIdAttribute);
+        entityIdField.setAccessible(true);
```
aschey-forpeople, line 73: Do we need this line? This field should already be public I believe.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Builds a mapping from claim IDs to their security tags.
+   *
+   * @param claimEntities the claim entity
+   * @param entityIdAttribute the claim type
+   * @return set of ClaimIds
+   */
+  public Set<String> collectClaimIds(List<Object> claimEntities, String entityIdAttribute) {
```
aschey-forpeople, line 67: I think `collectClaimIds` can call `extractClaimId` for each element instead of re-implementing the logic here.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,114 @@
...
+      results =
+          entityManager
+              .createQuery(query, String[].class)
+              .setParameter("claimIds", claimIds)
+              .getResultList();
+    }
+
+    // Build the map from results
+    Map<String, Set<String>> claimIdToTagsMap = new HashMap<>();
+    for (Object[] result : results) {
```
aschey-forpeople, line 50: This is already a `String[]` so you can use `String[]` here and avoid calling `toString` below.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/SamhsaV2InterceptorShadow.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * logs Missing Claim Ids.
+   *
+   * @param entity the resource being processed
+   * @param hasSamhsaData the hasNoSamhsaData boolean
+   */
+  public void logMissingClaim(Object entity, boolean hasSamhsaData) {
+    if (samhsaV2Shadow && entity instanceof ClaimWithSecurityTags<?> claimWithSecurityTags) {
```
aschey-forpeople, line 56: I think we can use `ClaimWithSecurityTags` as the parameter here and avoid this conditional.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
```diff
@@ -524,6 +530,13 @@ Bundle createBundleFor(
 
     List<T> resources =
         entitiesWithType.stream()
+            .map(
+                pair -> {
+                  // Log if missing claim for samhsa V2 Shadow check before filtering
+                  samhsaV2InterceptorShadow.logMissingClaim(
```
aschey-forpeople, line 536: Can probably do this check in the call to `filter` below to avoid calling `hasNoSamhsaData` twice.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/common/ClaimWithSecurityTagsDao.java
aschey-forpeople: This method isn't called by the method that does the actual query so it doesn't seem like it needs to be here. `SecurityTagManager` is probably a better place.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Provides common logic for performing DB interactions. */
+public class ClaimWithSecurityTagsDao {
```
aschey-forpeople, line 16: Can probably call this `SecurityTagsDao` so it's a bit less verbose.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/ClaimWithSecurityTagsDao.java
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Builds a mapping from claim IDs to their security tags.
+   *
+   * @param tagTable The table containing security tags
+   * @param claimIds The list of claim IDs
+   * @param entityManager the entityManager
+   * @return A map from claim ID to a list of security tags
+   */
+  public Map<String, Set<String>> buildClaimIdToTagsMap(
+      String tagTable, Set<String> claimIds, EntityManager entityManager) {
```
aschey-forpeople, line 29: `entityManager` should probably be injected in the constructor like it is in `ClaimDao`. This class should probably be initialized in the same manner as `ClaimDao` for consistency.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/R4ExplanationOfBenefitResourceProvider.java
```diff
@@ -239,9 +241,11 @@ metricRegistry, getClass().getSimpleName(), "query", "eob_by_id")) {
             "eob_by_id", eobByIdQueryNanoSeconds, claimEntity == null ? 0 : 1);
       }
     }
-
+    Set<String> securityTags = new HashSet<>();
+    ClaimWithSecurityTags claimWithSecurityTags =
```
aschey-forpeople, line 245: This isn't actually loading the security tags for the claim.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/PatientClaimsEobTaskTransformerV2.java
```diff
@@ -23,16 +28,21 @@
...
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
+import org.apache.poi.ss.formula.functions.T;
```
aschey-forpeople, line 41: Guessing this was imported accidentally

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/ClaimType.java
```diff
@@ -41,54 +41,62 @@ public enum ClaimType {
       CarrierClaim_.claimId,
       CarrierClaim_.beneficiaryId,
       (entity) -> ((CarrierClaim) entity).getDateThrough(),
+      "CarrierTag",
```
aschey-forpeople, line 44: To reduce the risk of typos, we should use the class name here: `CarrierTag.class.getName()`

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/V1SamhsaConsentInterceptor.java
```diff
@@ -37,48 +39,53 @@ public ConsentOutcome willSeeResource(
...
+
+    // Process the resource if it is a Bundle
+    if (theResource instanceof Bundle bundle) {
+      processBundle(bundle);
+    }
+
+    return ConsentOutcome.PROCEED;
+  }
+
+  boolean applySamhsaFiltering(RequestDetails theRequestDetails) {
```
aschey-forpeople, line 55: This should probably be called `shouldFilterSamhsa` since it's not actually applying any filters.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/McsClaimResponseTransformerV2.java
```diff
@@ -105,14 +114,17 @@ public McsClaimResponseTransformerV2(
    */
   @Trace
   public ClaimResponse transform(Object claimEntity, boolean includeTaxNumbers) {
-    if (!(claimEntity instanceof RdaMcsClaim)) {
+    ClaimWithSecurityTags<?> claimWithSecurityTags = (ClaimWithSecurityTags<?>) claimEntity;
```
aschey-forpeople, line 117: I think we should change the interface that defines this method to accept `ClaimWithSecurityTags<T> claimEntity` instead of `Object claimEntity`. That will provide a bit more type safety.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/r4/providers/pac/AbstractR4ResourceProvider.java
```diff
@@ -238,7 +259,9 @@ public T read(@IdParam IdType claimId, RequestDetails requestDetails) {
       throw new ResourceNotFoundException(claimId);
     }
 
-    Mbi claimEntityMbi = getClaimEntityMbi(claimIdObj.getRight(), claimEntity);
+    Mbi claimEntityMbi =
+        getClaimEntityMbi(
+            claimIdObj.getRight(), ((ClaimWithSecurityTags<?>) claimEntity).getClaimEntity());
```
aschey-forpeople, line 264: If we change the variable definition to `ClaimWithSecurityTags<?> claimEntity` instead of `Object claimEntity`, we don't need to cast here.

apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/SecurityTagManager.java
```diff
@@ -141,4 +104,61 @@ public List<org.hl7.fhir.dstu3.model.Coding> getClaimSecurityLevelDstu3(
...
+      case CarrierClaim carrierClaim -> String.valueOf(carrierClaim.getClaimId());
+      case DMEClaim dmeClaim -> String.valueOf(dmeClaim.getClaimId());
+      case HHAClaim hhaClaim -> String.valueOf(hhaClaim.getClaimId());
+      case HospiceClaim hospiceClaim -> String.valueOf(hospiceClaim.getClaimId());
+      case InpatientClaim inpatientClaim -> String.valueOf(inpatientClaim.getClaimId());
+      case OutpatientClaim outpatientClaim -> String.valueOf(outpatientClaim.getClaimId());
+      case SNFClaim snfClaim -> String.valueOf(snfClaim.getClaimId());
+      case PartDEvent partDEvent -> String.valueOf(partDEvent.getEventId());
+
+      default -> null; // Return null for unsupported claim types
```
aschey-forpeople, line 161: I think we should throw a `BadCodeMonkeyException` here since this case should never be hit. If we do make some mistake here and return null, causing the security tags to be skipped, we could have a security leak.

PR #2583, opened 2025-03-11
BFD-3890: Define Initial ECS Cluster

mjburling:
**JIRA Ticket:**
BFD-3890

### What Does This PR Do?

This PR introduces a basic ECS Cluster definition under a `cluster` terraservice and makes use of SSM hierarchy lookups added here to the `bfd-terraservice` module.

* Are we keeping things simple?

### Validation

Resources defined in this new terraservice are available in the `test` environment.


malessi (approved):

ops/terraform/services/cluster/main.tf
```diff
@@ -0,0 +1,74 @@
...
+  #       This requires adjustments to the `mgmt` module
+  #       Address in BFD-3945
+  configuration {
+    execute_command_configuration {
+      kms_key_id = local.kms_data_key_arn
+      logging    = "OVERRIDE"
+
+      log_configuration {
+        cloud_watch_log_group_name = aws_cloudwatch_log_group.this.name
+      }
```
TimRotenSB, line 56: Missing `tags` argument - additionally with tags, may want to add an ephemeral toggle within the tags for guardduty container runtime monitoring which enables itself based on the cluster being tagged (although this might not be enabled at an account level yet we could tag it ahead of time for when this IS turned on) 

Documentation for reference: https://aws.amazon.com/blogs/security/using-amazon-guardduty-ecs-runtime-monitoring-with-fargate-and-amazon-ec2/

Tag to include is: `GuardDutyManaged` `True/False`

TimRotenSB (approved):
Looks good for a jumping off point - Only noticed the missing tag in the cluster level resource.

aschey-forpeople (dismissed):

ops/terraform/services/cluster/main.tf
malessi: FYI, as far as `tags`: [`tags` is automatically derived](https://www.hashicorp.com/en/blog/default-tags-in-the-terraform-aws-provider#:~:text=Auto%20Scaling%20Groups.-,using%20default%20tags,-You%20can%20set) from the module-level `default_tags` specified in the `aws` `provider` block in `terraform.tf`; specifically, all resources that support default tags--including `aws_ecs_cluster`--will get their tags from `local.default_tags`. Take a look at the Cluster in the Console, you'll see it has the expected set of tags already.

mjburling: Indeed. I've added the additional, resource-level tag for `GuardDutyManaged:true` for grins on the `ecs_cluster` definition as of 6005a96e0. We'll more fully address this in BFD-3945.

aschey-forpeople (approved):

malessi (approved):

PR #2632, opened 2025-04-16
BFD-4019: Bug in NDC lookup

dondevun:
… normalize drug code before lookup.

**JIRA Ticket:**
BFD-4019

### What Does This PR Do?

This PR fixes two issues:
1. Drug code display values will be removed if the drug code cannot be looked up, instead of showing the placeholder.
2. Drug codes are normalized from the format `000000000` to `00000-0000` before the lookup.

### Validation

Verified against golden beneficiary.

### DIFF
--- a/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/CommonTransformerUtils.java
+++ b/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/CommonTransformerUtils.java
@@ -808,11 +808,22 @@ public final class CommonTransformerUtils {
    * @return placeholder for the drug code.
    */
   public static String buildReplaceDrugCode(Optional<String> drugCode) {
-    if (drugCode.isPresent()) {
-      return String.format("replaceDrugCode[%s]", drugCode.get());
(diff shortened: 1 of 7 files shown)


apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/commons/CommonTransformerUtils.java
```diff
@@ -808,13 +808,24 @@ public static Optional<NPIData> buildReplaceTaxonomy(Optional<String> npi) {
...
     }
   }
 
+  /**
+   * Normalizes the drug code to match the database values. *
+   *
+   * @param drugCode The drug code to normalize.
+   * @return The normalized drug code.
+   */
+  public static String normalizeDrugCode(String drugCode) {
```
aschey-forpeople, line 824: Would it be cleaner to normalize the other way? i.e. remove any hyphens from the string instead, that way we're not relying on substring indexes.

aschey-forpeople: Okay, fair enough. I'm fine with this for now.

aschey-forpeople (approved):

PR #2636, opened 2025-04-18
BFD-4020: Refactor backfill to add details column

dondevun:
…ion of details column.

**JIRA Ticket:**
BFD-4020

### What Does This PR Do?

Adds a details column to the backfill tags. Additionally, this will make sure that the column that the code comes from matches the system that the code belongs to before creating the tag. This will eliminate the need for any checking on the server, as no tags should be created for claims that don't have legitimate SAMHSA codes.

### Validation

### DIFF
--- a/ops/terraform/services/base/values/ephemeral.yaml
+++ b/ops/terraform/services/base/values/ephemeral.yaml
@@ -47,7 +47,7 @@
 /bfd/${env}/server/nonsensitive/pac/claim_source_types: fiss,mcs
 /bfd/${env}/server/nonsensitive/c4dic/enabled: "false"
 /bfd/${env}/server/nonsensitive/samhsa_v2/enabled: "false"
-/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "true"
+/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "false"
 /bfd/${env}/server/nonsensitive/lb_is_public: false
 /bfd/${env}/server/nonsensitive/launch_template_volume_iops: 3000
 /bfd/${env}/server/nonsensitive/launch_template_volume_size_gb: 60
@@ -79,3 +79,6 @@
 /bfd/${env}/server/nonsensitive/ecs/resources/memory: 8192
 /bfd/${env}/pipeline/nonsensitive/rda/npi_fda_load_job/batch_size: 100000
 /bfd/${env}/pipeline/nonsensitive/rda/npi_fda_load_job/enabled: true
+/bfd/${env}/pipeline/nonsensitive/rda/samhsa/backfill/enabled: "true"
+/bfd/${env}/pipeline/nonsensitive/rda/samhsa/backfill/batch_size: 100000
+/bfd/${env}/pipeline/nonsensitive/rda/samhsa/backfill/log_interval: 900
--- a/ops/terraform/services/base/values/prod-sbx.yaml
+++ b/ops/terraform/services/base/values/prod-sbx.yaml
@@ -143,7 +143,7 @@
 /bfd/${env}/server/nonsensitive/pac/claim_source_types: fiss,mcs
 /bfd/${env}/server/nonsensitive/c4dic/enabled: "false"
 /bfd/${env}/server/nonsensitive/samhsa_v2/enabled: "false"
-/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "true"
+/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "false"
 /bfd/${env}/server/nonsensitive/new_relic/app_name: BFD Server prod-sbx
 /bfd/${env}/server/nonsensitive/heathcheck/request_timeout_seconds: "15"
 /bfd/${env}/server/nonsensitive/heathcheck/request_backoff_seconds: "1"
--- a/ops/terraform/services/base/values/prod.yaml
+++ b/ops/terraform/services/base/values/prod.yaml
@@ -193,7 +193,7 @@
 /bfd/${env}/server/nonsensitive/pac/claim_source_types: fiss,mcs
 /bfd/${env}/server/nonsensitive/c4dic/enabled: "false"
 /bfd/${env}/server/nonsensitive/samhsa_v2/enabled: "false"
-/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "true"
+/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "false"
 /bfd/${env}/server/nonsensitive/new_relic/app_name: BFD Server prod
 /bfd/${env}/server/nonsensitive/heathcheck/request_timeout_seconds: "15"
 /bfd/${env}/server/nonsensitive/heathcheck/request_backoff_seconds: "1"
--- a/ops/terraform/services/base/values/test.yaml
+++ b/ops/terraform/services/base/values/test.yaml
@@ -193,7 +193,7 @@
 /bfd/${env}/server/nonsensitive/pac/claim_source_types: fiss,mcs
 /bfd/${env}/server/nonsensitive/c4dic/enabled: "true"
 /bfd/${env}/server/nonsensitive/samhsa_v2/enabled: "false"
-/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "true"
+/bfd/${env}/server/nonsensitive/samhsa_v2/shadow: "false"
 /bfd/${env}/server/nonsensitive/heathcheck/request_timeout_seconds: "15"
 /bfd/${env}/server/nonsensitive/heathcheck/request_backoff_seconds: "1"
 /bfd/${env}/server/nonsensitive/heathcheck/testing_bene_id: "-88888888888888"


apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -104,14 +106,27 @@ public class SamhsaUtil {
...
+  private static Map<String, List<SamhsaEntry>> samhsaMap = new HashMap<>();
 
   /** Instance of this class. Will be a singleton. */
   private static SamhsaUtil samhsaUtil;
 
   /** The file from which SAMHSA entries are pulled. Must be in the resources folder. */
   private static final String SAMHSA_LIST_RESOURCE = "security_labels.yml";
 
+  private static final String[] PRCDR_SYSTEMS = {
+    "http://www.cms.gov/Medicare/Coding/ICD10", "http://www.cms.gov/Medicare/Coding/ICD9"
```
aschey-forpeople, line 118: I think we already have these URLs defined in constants elsewhere. Might be good to reference those here instead of defining them again.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -179,59 +194,139 @@ private Object[] getClaimDates(
...
+    } else {
+      datesObject = getClaimDates(claimId, tableEntry, entityManager);
+      // Put the dates in a map, in case any other line items with the same id have samhsa
+      // codes.
+      datesMap.put(claimId.toString(), datesObject);
+    }
+    return datesObject;
+  }
+
+  private static boolean isInvalidClaimDate(Object[] datesObject, SamhsaEntry entry) {
```
aschey-forpeople, line 277: It looks like some of this logic is duplicated between here and `buildDetails`. Might be nice to make it a bit more reusable.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
```diff
@@ -163,6 +194,39 @@ public Long call() {
...
+   * @return The List of column positions.
+   */
+  public static List<Integer> getColumnPositions(
+      COLUMN_TYPE columnType,
+      Map<String, COLUMN_TYPE> queryColumns,
+      boolean uniqueColumnType,
+      boolean required) {
+    List<Integer> results = new ArrayList<>();
+    int i = 0;
+    for (Map.Entry<String, COLUMN_TYPE> entry : queryColumns.entrySet()) {
```
aschey-forpeople, line 213: If you want to keep using the functional style, this could be done with a combination of `IntStream` and `filter`, but this is fine too.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
```diff
@@ -74,6 +81,30 @@ public abstract class AbstractSamhsaBackfill implements Callable {
...
+
+  @Getter @Setter int fromDatePosition;
+  @Getter @Setter int toDatePosition;
+  @Getter @Setter int claimIdPosition;
+  @Getter @Setter List<Integer> samhsaColumnPositions;
+  @Getter @Setter int lineNumPosition;
+  @Getter @Setter String[] columnNames;
+
+  /** The column type for a column. */
+  public enum COLUMN_TYPE {
```
aschey-forpeople, line 95: I think the convention is to use `PascalCase` for the enum definition.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
```diff
@@ -177,45 +241,69 @@ public Long call() {
...
    * @param table The table.
    * @param claimField The claim id field.
    * @param columns The columns to check.
    * @return The query string for a particular table.
    */
-  protected static String buildQueryStringTemplate(
-      String table, String claimField, String... columns) {
+  protected String buildQueryStringTemplate(String table, String claimField, String... columns) {
     String concatColumns = String.join(", ", columns);
+    List<String> samhsaColumnOrder = new ArrayList<>();
```
aschey-forpeople, line 302: Is all this logic around column order due to the query returning the results as a list? Maybe we can return the results as a map instead? That way we can reference the values by key instead of index. https://stackoverflow.com/questions/4371384/can-a-jpa-query-return-results-as-a-java-map

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
aschey-forpeople: Sure, fair enough

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -179,61 +196,132 @@ private Object[] getClaimDates(
...
+    LocalDate startDate = LocalDate.parse(entry.getStartDate());
+    LocalDate endDate =
+        entry.getEndDate().equalsIgnoreCase("Active")
+            ? LocalDate.MAX
+            : LocalDate.parse(entry.getEndDate());
+    getStartEndDateForCode result = new getStartEndDateForCode(startDate, endDate);
+    return result;
+  }
+
+  private record getStartEndDateForCode(LocalDate startDate, LocalDate endDate) {}
```
aschey-forpeople, line 288: Is the convention for record classes to use camel case instead of pascal case? Usually I've seen them use pascal case like normal class types.

aschey-forpeople: I'd maybe call this `CodeDateRange` or something rather than start a class name with a verb.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/AbstractSamhsaBackfill.java
```diff
@@ -176,45 +185,47 @@ public Long call() {
...
     Optional<Object[]> dates = Optional.empty();
     // Line item tables pull the active dates with a separate query, while parent tables use the
     // original query.
     if (!tableEntry.getLineItem()) {
-      dates = Optional.of(new Object[] {claim[1], claim[2]});
-      codesOffset = 3;
+      try {
+        dates =
+            Optional.of(
+                new Object[] {
```
aschey-forpeople, line 197: I know it's not strictly related to this PR, but do we need to pass `Object[]` around everywhere for the claim dates? Would it be possible to query them as date objects instead?

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -179,61 +194,132 @@ private Object[] getClaimDates(
...
+    return result;
+  }
+
+  private record CodeDateRange(LocalDate startDate, LocalDate endDate) {}
+
+  private static SamhsaEntry getEntryForCode(String columnName, String code) {
+    List<String> columnSystems = getSystemsForColumn(columnName);
+    List<SamhsaEntry> entryList = new ArrayList<>();
+    // Get all the entries for the systems that this code may belong to
+    columnSystems.forEach(system -> entryList.addAll(samhsaMap.get(system)));
```
aschey-forpeople, line 292: You can do a `flatMap` here instead of a `forEach` with `addAll`:

```java
 columnSystems.stream()
    .flatMap(c -> samhsaMap.get(c).stream())
    .filter(e -> e.getCode().equals(code))
    .findFirst()
    .orElse(null);
```

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/SamhsaUtil.java
```diff
@@ -165,75 +180,145 @@ public static void createSamhsaMap() throws IOException {
...
     Map<String, String> params =
         Map.of("claimTable", tableEntry.getParentTable(), "claimField", tableEntry.getClaimField());
     StringSubstitutor strSub = new StringSubstitutor(params);
     String queryStr = strSub.replace(tableEntry.getDatesQuery());
-    Query query = entityManager.createNativeQuery(queryStr);
+    Query query = entityManager.createNativeQuery(queryStr, Tuple.class);
     query.setParameter("claimId", claimId);
-    return (Object[]) query.getSingleResult();
+    Tuple result = (Tuple) query.getSingleResult();
+    return new Date[] {(Date) result.get(0), (Date) result.get(1)};
```
aschey-forpeople, line 191: I think we usually use `LocalDate`. Is there a reason we used `Date` here instead? Just want to make sure there's no funny timezone business happening.

aschey-forpeople: Yeah, I just want to make sure there's no timezone issues with this. There _shouldn't_ be since we set the JVM process to use UTC, but it's hard to be sure just by looking at it.

aschey-forpeople: JPA does support automatic conversion to `LocalDate` (we're using it with the CCW pipeline), but that may not work with native queries.

apps/bfd-pipeline/bfd-pipeline-shared-utils/src/main/java/gov/cms/bfd/pipeline/sharedutils/samhsa/backfill/RDASamhsaBackfill.java
```diff
@@ -13,123 +15,114 @@ public class RDASamhsaBackfill extends AbstractSamhsaBackfill {
...
+
   /** Columns for fiss revenue lines. */
-  private static String[] FISS_REVENUE_LINES_SAMHSA_COLUMNS =
-      new String[] {"apc_hcpcs_apc", "hcpc_cd"};
+  private String[] FISS_REVENUE_LINES_SAMHSA_COLUMNS =
+      new String[] {RDA_POSITION, "apc_hcpcs_apc", "hcpc_cd"};
 
   /** Columns for fiss_diagnosis_codes. */
-  private static String[] FISS_DIAGNOSIS_SAMHSA_COLUMNS = new String[] {"diag_cd2"};
+  private String[] FISS_DIAGNOSIS_SAMHSA_COLUMNS = new String[] {RDA_POSITION, "diag_cd2"};
```
aschey-forpeople, line 48: What's the purpose of including `RDA_POSITION` here? There's no SAMHSA data in that column so it seems a little odd.

aschey-forpeople (approved):

PR #2637, opened 2025-04-18
BFD-4015: V3 Patient Read endpoint MVP

aschey-forpeople:
**JIRA Ticket:**
BFD-4015

### What Does This PR Do?

- Adds the initial Read endpoint for the Patient resource
  - FHIR mappings are based on the current contents of `patient.map`
  - Edge cases around beneficiaries with multiple MBIs and bene_sks have not been thoroughly tested yet - will be handled in a later story
- Creates the setup for integration tests
  - Integration tests use the HAPI FHIR client to verify the endpoints function correctly
  - Snapshot testing is used to validate that responses match the expectation
  - `testcontainers` is used to test against an ephemeral Postgres instance
  - The database setup is still in a rough draft, for now I've just copied over the sql script used to create the BFD database
- Updates the IDR pipeline with some missing info
  - Adds the beneficiary MBI history table and death date
  - Marks all fields as non-nullable and introduces normalization to handle any null values coming in

New dependencies are being added

### Validation

Integration tests


apps/pom.xml
```diff
@@ -99,6 +100,7 @@
 
         <hibernate.version>6.4.4.Final</hibernate.version>
         <junit.version>5.12.0</junit.version>
+        <junit.platform.version>1.12.0</junit.platform.version>
```
aschey-forpeople, line 103: This is a transitive dependency, but for some reason `bfd-server-ng` was resolving to an older (incompatible) version, so I had to force it to use the latest.

sb-benohe (approved):
New dependencies look good, SNYK Scans are currently clean for highs and criticals, Approved

apps/bfd-pipeline/bfd-pipeline-idr/bfd.sql
```diff
@@ -9,27 +9,28 @@ CREATE TABLE idr.beneficiary(
...
-    bene_line_1_adr VARCHAR(45),
-    bene_line_2_adr VARCHAR(45),
-    bene_line_3_adr VARCHAR(40),
-    bene_line_4_adr VARCHAR(40),
-    bene_line_5_adr VARCHAR(40),
-    bene_line_6_adr VARCHAR(40),
-    cntct_lang_cd VARCHAR(3),
-    idr_trans_efctv_ts TIMESTAMPTZ,
-    idr_trans_obslt_ts TIMESTAMPTZ,
+    bene_mbi_id VARCHAR(11) NOT NULL,
```
alex-dzeda, line 12: Let's add IDR_UPDT_TS to all of these, also removal of SSN.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/PatientIT.snap
```diff
@@ -0,0 +1,57 @@
+gov.cms.bfd.server.ng.PatientIT.patientRead=[
+{
```
aschey-forpeople, line 2: @alex-dzeda here's the sample output from the tests.

apps/bfd-pipeline/bfd-pipeline-idr/bfd.sql
```diff
@@ -39,15 +40,26 @@ CREATE TABLE idr.beneficiary_history(
...
+    bene_ssn_num VARCHAR(9) NOT NULL,
+    idr_trans_efctv_ts TIMESTAMPTZ NOT NULL,
+    idr_trans_obslt_ts TIMESTAMPTZ NOT NULL,
     bfd_created_ts TIMESTAMPTZ NOT NULL,
     bfd_updated_ts TIMESTAMPTZ NOT NULL,
     PRIMARY KEY(bene_sk, idr_trans_efctv_ts)
 );
 
+CREATE TABLE idr.beneficiary_mbi_history (
```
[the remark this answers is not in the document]
aschey-forpeople: Sure, I can rename it. What's the reason for storing `IDR_UPDT_TS`? I'm using the time we ingest the data for `lastUpdated` since that would be the time that the data is actually available to our users.

apps/bfd-pipeline/bfd-pipeline-idr/bfd.sql
aschey-forpeople: Will we need SSN in the future? If we don't add it now, we'll have to go backfill the data later.

apps/bfd-pipeline/bfd-pipeline-idr/mock-idr.sql
```diff
@@ -41,6 +42,16 @@ CREATE TABLE cms_vdm_view_mdcr_prd.v2_mdcr_bene_hstry (
     PRIMARY KEY(bene_sk, idr_trans_efctv_ts)
 );
 
+CREATE TABLE cms_vdm_view_mdcr_prd.v2_mdcr_bene_mbi_id (
```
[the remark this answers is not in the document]
aschey-forpeople: I have a copy of `bfd.sql` for the API project because it needs to be part of the java classpath. In the future, we'll remove that file and just run the pipeline to load in the test data. So yes, but it's a temporary workaround.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/model/Identity.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Transforms the identity record to a {@link Patient.PatientLinkComponent} if the bene_sk is
+   * different from the current beneficiary's bene_sk.
+   *
+   * @param currentPatient current version of the beneficiary
+   * @return patient link
+   */
+  public Optional<Patient.PatientLinkComponent> toFhirLink(Patient currentPatient) {
```
alex-dzeda, line 92: If current bene SK != BENE_XREF_EFCTV_SK, then current BENE_SK is replaced by the BENE_XREF_EFCTV_SK. 
Conversely, when they're the same, yet other BENE_SKs exist, the BENE_XREF_EFCTV_SK replaces those BENE_SKs.

apps/bfd-pipeline/bfd-pipeline-idr/bfd.sql
alex-dzeda: We might, but not in the next year or so, at least. I think the possibility is low enough that the backfill is fine if it is ever needed.

apps/bfd-pipeline/bfd-pipeline-idr/bfd.sql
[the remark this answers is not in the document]
alex-dzeda: It's not critical, but it is a nice way to stamp out an exact version of a resource.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/model/Identity.java
aschey-forpeople: Changed the logic here. May not be 100% correct yet, but we need to incorporate the changed for kill credit anyway, so that will be done separately.

apps/bfd-pipeline/bfd-pipeline-idr/bfd.sql
aschey-forpeople: Fair enough. Made both of these changes.

apps/bfd-pipeline/bfd-pipeline-idr/bfd.sql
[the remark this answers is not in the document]
aschey-forpeople: Sure, doesn't hurt. Added it.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/model/Identity.java
aschey-forpeople: One question - if the current patient record is an older one (BENE_SK != BENE_XREF_EFCTV_SK), do we add links  to any bene records besides the "replaced by" record that points to BENE_XREF_EFCTV_SK?

alex-dzeda: nope - the more bare bones the better. If >2 BENE_SKs merge into a different BENE_XREF_EFCTV_SK, they'd both only point to the BENE_XREF_EFCTV_SK (replacedBy). We _could_ add the seealso to those, but given that we're going to be merging claims on the current SK, I don't think there's a benefit, while there's a risk of people just getting confused, even though traversing that link would still point to the current one in the subsequent link.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/model/Identity.java
```diff
@@ -86,18 +90,32 @@ public Optional<Identifier> toFhirIdentifier() {
...
-      return Optional.empty();
+  public Optional<Patient.PatientLinkComponent> toFhirLink(Patient patientToCompare) {
+    var compareSk = patientToCompare.getId();
+    var beneSkStr = beneSk.toString();
+    var xrefSkStr = xrefSk.toString();
+    var beneSkMatches = beneSkStr.equals(compareSk);
+    var currentIsXref = xrefSkStr.equals(beneSkStr);
+    var compareIsXref = xrefSkStr.equals(compareSk);
+
+    if (currentIsXref && !beneSkMatches) {
```
alex-dzeda, line 104: Can you add a comment above to enumerate more detail on which SK is being referred to? It's fuzzy if one just looks at the method. The naming of these makes me think it might be flipped. eg if current bene_sk (eg Patient/<bene_sk>) == bene_xref_efctv_sk

aschey-forpeople: Yeah, I'm having a hard time trying to come up with variable names that make sense. Essentially, what happens is that we pull all of the bene_sk/mbi pairs in a separate query, then this method gets called on each result from that. So it will iterate through the bene_sk/mbi pair that matches the requested patient record as well as any historical/newer values.

I renamed a few things and added comments. Let me know if that helps.

alex-dzeda (approved):

PR #2654, opened 2025-04-29
BFD-4045: Add standalone file to generate/update existing BB2 User Accounts to v3.

alex-dzeda:
…ent collisions

**JIRA Ticket:**
BFD-4045

### What Does This PR Do?

This PR adds another generation script to populate a set of synthetic MBIs + beneficiaries corresponding to the existing BB2 demo accounts (BBUser00000 - BBUser10000), to reduce the burden of BB2 in setting up a v3 sandbox. The characteristics file used is also a part of this, to ensure it is able to be replicated. A conscious decision was made to mostly copy and edit the beneficiary generator, rather than merge them into a single file, for efficiency's sake. 
Of note, the golden beneficiary is included at the very end. The golden bene in V3 will now include the edge cases / merge events. It has multiple MBIs, and multiple BENE_SKs. These functionalities are critical improvements in V3, because it means more people will have access to their data, and provide a specific point of reference if folks want to test out these scenarios as they build for V3.

Do we want to put the extra effort into merging these files? We can if necessary, but I actually kind of like the option to maintain the BB2 accounts separately. 

All of the MBIs produced are synthetic and not valid. Among other reasons, the 2nd digit _only_ allows characters that would be invalid in a real MBI.

### Validation

The script is runnable by typing python BB2-Patient-Generator.py in the requisite directory. the outputs will be in the same directory, with the file names beginning with "SYNTHETIC" 
All of the BB2 user accounts have an "S" in the 2nd digit of the MBI. In both generation files, I removed "S" as an option for the second letter, ensuring that we will not have collisions between the BB2 accounts and any further generated accounts.


apps/bfd-model/bfd-model-idr/sample-data/generator/BB2-Patient-Generator.py
```diff
@@ -0,0 +1,151 @@
...
+fake = Faker()
+
+used_bene_sk = [] #We have this + used_mbi in case we want to expand this and ensure no collisions (just populate w/ existing list)
+used_mbi = []
+bene_table = []
+bene_hstry_table = []
+mbi_table = {}
+address_options = []
+bb_accounts = []
+with open('beneficiary-components/addresses.csv', 'r') as file:
```
aschey-forpeople, line 19: Would it be possible to move some of the logic like this and the MBI generator to a shared utils file that we can import from both the BB2 generator and the other patient generator?

alex-dzeda: I think the address loader, gen_mbi, and gen_bene_sk are easily moved into a shared file, anything else you think is critical? The rest diverges a bit.

aschey-forpeople: That sounds good to me. Anything that diverges can be kept separate.

sb-benohe (approved):

sb-benohe:
all synthetic data

aschey-forpeople (approved):

PR #2687, opened 2025-05-23
BFD-3862: Reconcile Terraform with Multi-Account Structure

malessi:
**JIRA Ticket:**
BFD-3862

### What Does This PR Do?

This PR introduces support for multi-account _OpenTofu_ Greenfield Terraservices whilst still supporting the legacy account. The ergonomics of working within the Greenfield Terraservices are _largely_ the same, although there are a few compromises made here due to limitations in OpenTofu's early evaluation support and OpenTofu Workspaces requiring state in order to function.

In more detail, this PR:

- Switches _every_ Greenfield Terraservice in both `ops/services` and `ops/platform` to use latest OpenTofu `1.10.0-beta1` instead of Terraform `1.5.0`
  - This is necessary as Terraform does not have Early Evaluation whereas OpenTofu does. We rely on using Early Evaluation to dynamically choose the OpenTofu state Bucket, KMS Key, and other `backend`-related configuration
- Introduces a new `root.tofu.tf` for both `ops/services` and `ops/platform` that dynamically configures the `s3` `backend` for every Greenfield Terraservice using early evaluation with some `tofu init`/`tofu workspace select`-only `variable`s (see below)
- Replaces every Greenfield Terraservice's `terraform.tf` with a `tofu.tf` symlink that links to the appropriate `root.tofu.tf` mentioned above
  - For environment-specific Terraservices (`ops/services`) the `root.tofu.tf` dynamically selects the appropriate S3 Bucket and S3 Key based upon the `local.service` defined in each service's `main.tf` along with the `terraform.workspace` OR `var.parent_env`
  - For platform Terraservices (`ops/platform`) the `root.tofu.tf` dynamically selects the appropriate S3 Bucket and S3 Key based upon the `local.service` defined in each service's `main.tf` along with the `terraform.workspace` OR `var.account_type` (since platform services create common resources in each account)

##### Legacy

Nothing's changed! Since `var.greenfield` defaults to `false`, `tofu init` and `tofu workspace select` do not require specifying the `var.account_type`/`var.parent_env` for _any_ Terraservice.

##### Greenfield

The ergonomics of working within the Greenfield Terraservices _in the Greenfield accounts_ are **slightly** different compared to the legacy account. This is due to now needing to support multiple accounts, along with a different state structure where each _parent environment_ now has its own state within its own Bucket instead of a single `bfd-tf-state` Bucket.

**Switching environments (environment-specific) or accounts (platform services)**

_NOTE: This isn't quite how working in the Greenfield account is now since `var.greenfield` exists; however, this is the intended UX when we do migrate fully to Greenfield and the legacy accounts are no longer in consideration (`var.greenfield` is removed)_

Since we now separate our resources across multiple accounts, the workflow for environment-specific Terraservices will look something like:

```bash
kion s # select the appropriate prod/non-prod account
tofu init -var=parent_env=test -reconfigure # -reconfigure is REQUIRED; -migrate-state is potentially dangerous!
tofu workspace select -var=parent_env=test test # workspace could also be ephemeral, e.g. 1234-test, just not something like ####-prod
tofu plan # no more need for parent_env!
```

Similarly, platform services will look like:

```bash
kion s # select the appropriate prod/non-prod account
tofu init -var=account_type=non-prod -reconfigure # -reconfigure is REQUIRED; -migrate-state is potentially dangerous!
tofu workspace select -var=account_type=non-prod non-prod
tofu plan # no more need for account_type!
```

- ~~Adds any new software dependencies~~

- [x] I have created tests to sufficiently ensure the reliability of my code, if applicable. If this is a modification to an existing piece of code, I have audited the associated tests to ensure everything works as expected.

### Validation

- `tofu apply`ing in Greenfield non-prod _and_ in Legacy with several platform/environment-specific Terraservices, _verifying that_:
  - Everything works as expected provided that the correct invocation is used to `init` or `workspace select`
  - Trying to `workspace select` an invalid account type or environment (i.e. `var.parent_env` is `test` and workspace is `prod`) results in an error


aschey-forpeople (approved):

mjburling (approved):
This scratches the itch, I think. Fortunately, it's not even _that_ clever. Let's give it a shot. Thanks!

PR #2751, opened 2025-07-15
BFD-4071: Patient Identities filtering out kill credit

brick-green:
**JIRA Ticket:**
BFD-4071

### What Does This PR Do?

Replace me.

### Validation


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/BeneficiaryRepository.java
```diff
@@ -84,6 +84,48 @@ SELECT new PatientIdentity(ROW_NUMBER() OVER (ORDER BY abi.beneSk) rowId, abi.be
...
+   *
+   * @param beneSk bene surrogate key
+   * @param beneXrefSk computed bene surrogate key
+   * @return list of patient identities representing all active identities connected to the bene
+   *     record
+   */
+  public List<PatientIdentity> getValidBeneficiaryIdentities(long beneSk, long beneXrefSk) {
+    return entityManager
+        .createQuery(
+            """
```
aschey-forpeople, line 99: Mind de-indenting this a bit?

aschey-forpeople: Also maybe put the select column list on separate lines so it isn't quite so long

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/BeneficiaryRepository.java
```diff
@@ -84,6 +84,48 @@ SELECT new PatientIdentity(ROW_NUMBER() OVER (ORDER BY abi.beneSk) rowId, abi.be
...
+                                FROM BeneficiaryXref bx
+                                WHERE bx.beneSk = bene.beneSk
+                                  AND bx.beneXrefSk = bene.xrefSk
+                                  AND bx.beneKillCred = '1'
+                            )
+                            AND NOT EXISTS (
+                                SELECT 1
+                                FROM BeneficiaryXref bx_init
+                                WHERE bx_init.beneSk = :beneSk
+                                  AND bx_init.beneXrefSk = :beneXrefSk
```
aschey-forpeople, line 118: Does this clause catch anything additional? `:beneXrefSk` and `bene.xrefSk` should always be the same.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/model/BeneficiaryXref.java
```diff
@@ -0,0 +1,40 @@
...
+  private String beneHicnNum;
+
+  @Column(name = "bene_kill_cred_cd")
+  private String beneKillCred;
+
+  @Column(name = "idr_updt_ts")
+  private ZonedDateTime idrUpdateTimestamp;
+
+  @Column(name = "src_rec_ctre_ts")
+  private ZonedDateTime srcRecCtreTimestamp;
```
aschey-forpeople, line 33: A lot of these columns were added to the database because we need them for the unique key, but we don't need to map them in the model if they're not actually referenced in the code

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/BeneficiaryRepository.java
```diff
@@ -84,6 +84,48 @@ SELECT new PatientIdentity(ROW_NUMBER() OVER (ORDER BY abi.beneSk) rowId, abi.be
...
+  /**
+   * Queries for current and historical MBIs and BENE_SKs, along with their start/end dates.
+   * Beneficiary records with kill credit switch set to "1" or overshare mbi are filtered out
+   *
+   * @param beneSk bene surrogate key
+   * @param beneXrefSk computed bene surrogate key
+   * @return list of patient identities representing all active identities connected to the bene
+   *     record
+   */
+  public List<PatientIdentity> getValidBeneficiaryIdentities(long beneSk, long beneXrefSk) {
```
aschey-forpeople, line 96: I think we can get rid of the old `getPatientIdentities` method since it's unused now.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/BeneficiaryRepository.java
brick-green: This is meant to test if the relationship between the input `:beneSk` and `:beneXrefSk` is valid. If it's invalid, then nothing is supposed to be return. I got that direction from this line in the 5/13 update:

`
Ensure that the BENE_SK -> BENE_XREF_EFCTV_SK relationship does not have a kill credit switch of 1. If it's a valid relationship, pull MBIs + BENE_SKs from BENE and BENE_HSTRY on those BENE_SKs.
`

I read that as saying nothing should be returned in the event that the kill credit switch = 1. What do you think?

aschey-forpeople (approved):
Thanks for persisting through much pain with this one

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/PatientReadIT.java
```diff
@@ -33,7 +33,21 @@ void patientReadValidString() {
 
   @Test
   void patientReadMergedBene() {
-    var patient = patientRead().withId("181968400").execute();
+    var patient = patientRead().withId("792872340").execute();
```
aschey-forpeople, line 36: Sorry, last minute thought - it might be nice to have an extra assert or two on these that just assert the number of links or identifiers. The snapshots are good for generic happy-path tests, but since these tests are specifically for merged benes, having a few more asserts like that will make it easier to see what we're trying to check here.

brick-green: Yeah good call for sure. Added some

PR #2778, opened 2025-08-13
BFD-4198: Filter SAMHSA information on V3 calls

np-complete0:
**JIRA Ticket:**
BFD-4198

### What Does This PR Do?
Implements SAMHSA filter for V3.

### Validation


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +64,67 @@ public Bundle searchByBene(
...
+      var sys = SecurityLabels.getSecurityLabelKeys();
+
+      filteredEobs =
+          eobs.stream()
+              .filter(
+                  claim ->
+                      claim.getClaimProcedures().stream()
+                          .noneMatch(
+                              e ->
+                                  sys.stream()
```
aschey-forpeople, line 87: That's a lot of nesting. Can we make some helper functions for this so it's a bit easier to read?

apps/bfd-shared-utils/src/main/java/gov/cms/bfd/sharedutils/SecurityLabels.java
```diff
@@ -0,0 +1,89 @@
...
+
+    try {
+      List<Map<String, Object>> yamlData = yaml.load(inputStream);
+
+      for (Map<String, Object> entry : yamlData) {
+        String system = (String) entry.get("system");
+        resultMap.computeIfAbsent(system, k -> new ArrayList<>()).add(entry);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
```
aschey-forpeople, line 51: This should use a proper logger and it should probably re-throw the error. We want the application to fail to start if this doesn't work since this is critical functionality.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +64,67 @@ public Bundle searchByBene(
     var eobs =
         claimRepository.findByBeneXrefSk(
             beneXrefSk.get(), serviceDate, lastUpdated, count, startIndex, sourceIds);
+
+    return getFhirEobs(eobs, true);
+  }
+
+  Bundle getFhirEobs(List<Claim> eobs, boolean filterSamhsa) {
+
+    List<Claim> filteredEobs = eobs;
```
aschey-forpeople, line 73: We should add some test cases for these.

aschey-forpeople, line 73: We try to use `var` to keep variable declarations simple.

```suggestion
    var filteredEobs = eobs;
```

apps/bfd-shared-utils/src/main/java/gov/cms/bfd/sharedutils/SecurityLabels.java
```diff
@@ -0,0 +1,95 @@
...
+ * <pre>
+ * - system: http://hl7.org/fhir/sid/icd-10-cm
+ *   code: F19.159
+ *   startDate: 1970-01-01
+ *   endDate: ACTIVE
+ * </pre>
+ *
+ * @version 1.0
+ */
+public class SecurityLabels {
```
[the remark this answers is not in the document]
np-complete0: ok

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimProcedure.java
```diff
@@ -10,12 +10,16 @@
...
+import lombok.Getter;
+import lombok.Setter;
 import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.Coding;
 import org.hl7.fhir.r4.model.ExplanationOfBenefit;
 
 /** Procedure and diagnosis table. */
 @Entity
+@Getter
+@Setter
```
aschey-forpeople, line 22: Do we need `@Setter` here? These fields should be read-only once they're deserialized from the database.

np-complete0: I can do this only:

image

aschey-forpeople: We tend to prefer E2E/integration tests instead of unit tests on this project. If we create an E2E test using SAMHSA data, then we can verify that the correct columns are being checked, the YAML file is being loaded correctly, the filter is removing the correct items, the JSON response has the correct meta tags (once that's added, etc). It'll make testing the complete flow easier and then we don't need to add any testing-specific code to the main application logic. You can take a look at the other tests for examples.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +119,45 @@ public Bundle searchById(
...
+    String dictCode =
+        entry.get("code") != null
+            ? entry.get("code").toString().trim().replace(".", "").toLowerCase()
+            : "";
+    return dictCode;
+  }
+
+  // Returns true if the given claim contains any procedure that matches a SAMHSA
+  // security label code from the provided dictionary and keys.
+  private boolean claimHasSamhsa(
```
aschey-forpeople, line 146: The logic here needs to check the code system in the SAMHSA map to ensure it's comparing against the matching ICD version and code type.

aschey-forpeople: They do, but only for the columns that match the code type. For example, `clm_prcdr_cd` and `clm_dgns_cd` in `v2_mdcr_clm_prod` are ICD codes, so they should only check against the ICD procedure and diagnosis code lists, respectively. There can be overlaps between the code systems that can cause false positives otherwise.

np-complete0: so in the `clm_prcdr_cd` and `clm_dgns_cd` only systems should be:
```
"http://hl7.org/fhir/sid/icd-9-cm",
"http://hl7.org/fhir/sid/icd-10-cm",
"http://www.cms.gov/Medicare/Coding/ICD9",
"http://www.cms.gov/Medicare/Coding/ICD10"
```
?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +70,41 @@ public Bundle searchByBene(
...
+
+      return emptyBundle;
+    }
+
+    var fhirEobs = filteredEobs.stream().map(Claim::toFhir).toList();
+
     return FhirUtil.bundleOrDefault(
-        eobs.stream().map(Claim::toFhir), claimRepository::claimLastUpdated);
+        fhirEobs.stream().map(eob -> (Resource) eob), claimRepository::claimLastUpdated);
   }
```
brick-green, line 108: This `getFhirEobs` method mixes together logic. IMO it would be easier to read if some of it was split out into separate methods. Perhaps one for `filterSamhsaClaims()` then one for `createEmptyBundle()`?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +70,41 @@ public Bundle searchByBene(
...
+                  claim -> {
+                    return !claimHasSamhsa(claim, dict, sys);
+                  })
+              .toList();
+    }
+
+    if (filteredEobs.isEmpty()) {
+      Bundle emptyBundle = new Bundle();
+      emptyBundle.setType(Bundle.BundleType.COLLECTION);
+      emptyBundle.setTypeElement(null);
```
brick-green, line 99: Does `emptyBundle.setTypeElement(null);` not overwrite the `emptyBundle.setType(Bundle.BundleType.COLLECTION);`?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -0,0 +1,69 @@
...
+ * test database seeded by the integration test pipeline, and then asserts that requesting that
+ * claim via the FHIR EOB endpoint returns no results (i.e. it was filtered).
+ */
+public class EobSamhsaFilterIT extends IntegrationTestBase {
+
+  private final FhirContext context = FhirContext.forR4(); // Create a FhirContext for R4
+  IGenericClient client = context.newRestfulGenericClient(getServerUrl());
+
+  protected long testClaimId = 1071939711295L;
+  private String codeToCheck = "F11921"; // Opioid use
```
aschey-forpeople, line 28: Is this code actually present in the test data? I don't see any new data being added to the csvs.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
brick-green, line 1: Can you test for the empty bundle logic too?

apps/bfd-pipeline/bfd-pipeline-idr/test_samples2/SYNTHETIC_BENE_HSTRY.csv
```diff
@@ -2,3 +2,25 @@ BENE_SK,BENE_XREF_EFCTV_SK,BENE_MBI_ID,BENE_LAST_NAME,BENE_1ST_NAME,BENE_MIDL_NA
 517782585,517782585,2B19C89AA35,Heeler,Alex,Joey,1959-03-14,,~,1,4,14152 ND-27,,,,,,Lisbon,58054,ND,~,2024-06-12 01:06:49.588129,2024-11-05 18:51:29.174693,9999-12-31 00:00:00.000000,9999-12-31 00:00:00.000000,Y
 848484848,517782585,2B19C89AA35,Heeler,Alex,Joey,1959-03-14,,~,1,4,14152 ND-27,,,,,,Lisbon,58054,ND,~,2024-06-12 01:06:49.588129,2024-11-05 18:51:29.174693,9999-12-31 00:00:00.000000,9999-12-31 00:00:00.000000,Y
 121212121,517782585,2B19C89AA35,Heeler,Alex,Joey,1959-03-14,,~,1,4,14152 ND-27,,,,,,Lisbon,58054,ND,~,2024-06-12 01:06:49.588129,2024-11-05 18:51:29.174693,9999-12-31 00:00:00.000000,9999-12-31 00:00:00.000000,Y
+612594477,612594477,4Z67VU7VN65,Jones,Chilli,Bingo,1910-09-18,,~,2,1,1970 S University Blvd,,,,,,Mobile,36609,AL,~,2025-01-19 15:22:02.803441,2025-06-29 03:28:29.508564,2025-07-19 06:28:51.647427,9999-12-31T00:00:00.000000+0000,Y
```
aschey-forpeople, line 5: Can we add the new test data into a sub-folder rather than keeping it all in the same file? That will make it easier to keep it organized.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +122,79 @@ public Bundle searchById(
...
+              var cp = e.getClaimProcedure();
+              return procedureMatchesDict(cp, dict, sys) || hcpcsCodeMatches(cl, dict, sys);
+            });
+  }
+
+  // cpt, Software and CodeSets.
+  private boolean hcpcsCodeMatches(
+      ClaimLine claimLine, Map<String, List<Map<String, Object>>> dict, List<String> sys) {
+    String hcpcs = claimLine.getHcpcsCode().getHcpcsCode().orElse("");
+    for (String link : sys) {
```
aschey-forpeople, line 170: It looks like this is checking every system for each code? We only want to check the system that matches the current type of code that we're checking against.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +122,79 @@ public Bundle searchById(
...
-    return claim.map(Claim::toFhir);
+  // Returns true if the given claim contains any procedure that matches a SAMHSA
+  // security label code from the provided dictionary and keys.
+  private boolean claimHasSamhsa(
+      Claim claim, Map<String, List<Map<String, Object>>> dict, List<String> sys) {
+
+    return claim.getClaimItems().stream()
+        .anyMatch(
+            e -> {
+              var cl = e.getClaimLine();
```
aschey-forpeople, line 160: There's four columns with SAMHSA data noted in the `mock-idr.sql` file and I only see three being checked here.

aschey-forpeople:
Some other ideas on things to test

- normalization: we should ensure our sample data matches the format of the codes available in prod and we should ensure we have tests that verify any required normalization logic
- claims with both one and multiple types of samhsa codes
- claims that would cause a false positive when there's a valid SAMHSA code for the wrong system

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimInstitutional.java
```diff
@@ -24,4 +24,21 @@ public class ClaimInstitutional {
...
   @Embedded private BenefitBalanceInstitutional benefitBalanceInstitutional;
+
+  /**
+   * Accessor for the institutional DRG (dgns_drg_cd) value. Kept here so callers in other packages
+   * do not need visibility into the internal supporting info embeddable types.
+   *
+   * @return optional DRG code (trimmed) if present and non-blank
+   */
+  public java.util.Optional<String> getDrgCode() {
+    if (supportingInfo == null || supportingInfo.getDiagnosisDrgCode() == null) {
```
aschey-forpeople, line 35: These checks will always be false. The only objects which can return null are the entities that are left joined. Everything else will return an `Optional`.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +111,105 @@ public Bundle searchById(
...
+    Bundle bundle = getFhirEobs(eobs, true);
+    return bundle.getEntry().stream()
+        .map(entry -> entry.getResource())
+        .filter(ExplanationOfBenefit.class::isInstance)
+        .map(ExplanationOfBenefit.class::cast)
+        .findFirst();
+  }
+
+  private String getDiagnosisCode(ClaimProcedure e) {
+    return e.getDiagnosisCode().orElse("").trim().replace(".", "").toLowerCase();
```
aschey-forpeople, line 124: Maybe we want to create a standard method to normalize the code since we're repeating the trim/replace logic a few times.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimInstitutional.java
np-complete0: yeah i was making sonarqube plugIn happy.

aschey-forpeople: We can ignore those warnings if they don't make sense.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +71,28 @@ public Bundle searchByBene(
...
         claimRepository.findByBeneXrefSk(
             beneXrefSk.get(), serviceDate, lastUpdated, count, startIndex, sourceIds);
+
+    return getFhirEobs(eobs, true);
+  }
+
+  protected Bundle getFhirEobs(List<Claim> eobs, boolean filterSamhsa) {
+
+    var filteredEobs = eobs;
+    var dict = SECURITY_LABELS_MAP;
```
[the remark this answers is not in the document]
aschey-forpeople: The one directly above this line. Seems like it's just assigning a local variable to a constant?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimInstitutional.java
```diff
@@ -24,4 +24,18 @@ public class ClaimInstitutional {
...
+
+  /**
+   * Accessor for the institutional DRG (dgns_drg_cd) value. Kept here so callers in other packages
+   * do not need visibility into the internal supporting info embeddable types.
+   *
+   * @return optional DRG code (trimmed) if present and non-blank
+   */
+  public java.util.Optional<String> getDrgCode() {
+    var code = supportingInfo.getDiagnosisDrgCode().getDiagnosisDrgCode();
+    if (code == null || code.isBlank()) {
```
aschey-forpeople, line 36: This check will always be false - the value will always be present and non-null.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +71,37 @@ public Bundle searchByBene(
...
+   *     SAMHSA-coded procedures
+   * @return a FHIR {@link Bundle} containing the converted Explanation of Benefits (EOBs)
+   */
+  protected Bundle getFhirEobs(List<Claim> claims, boolean filterSamhsa) {
+
+    var filteredClaims = claims;
+
+    if (filterSamhsa) {
+
+      filteredClaims =
```
aschey-forpeople, line 94: Can we trim some of this whitespace?
```suggestion
    var filteredClaims = claims;

    if (filterSamhsa) {
      filteredClaims =
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +71,37 @@ public Bundle searchByBene(
...
+              // keep claims that do NOT contain any SAMHSA-coded procedures
+              .filter(claim -> !claimHasSamhsa(claim, SECURITY_LABELS_DICTIONARY))
+              .toList();
+    }
+
+    var fhirEobs = filteredClaims.stream().map(Claim::toFhir).toList();
+
     return FhirUtil.bundleOrDefault(
-        eobs.stream().map(Claim::toFhir), claimRepository::claimLastUpdated);
+        fhirEobs.stream().map(Resource.class::cast), claimRepository::claimLastUpdated);
```
aschey-forpeople, line 104: We can get rid of the manual cast here if we avoid calling `toList` twice above and instead operate on the stream.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +120,94 @@ public Bundle searchById(
...
+    }
+
+    return false;
+  }
+
+  // Checks HCPCS and cpt.
+  private boolean hcpcsCodeMatches(
+      ClaimLine claimLine, Map<String, List<Map<String, Object>>> dict) {
+    String hcpcs = claimLine.getHcpcsCode().getHcpcsCode().map(String::toLowerCase).orElse("");
+    String claimSystem = FhirUtil.getHcpcsSystem(hcpcs);
```
aschey-forpeople, line 179: `getHcpcsSystem` is only meant to be used on the prefix of the HCPCS code. I can see why that's misleading and we should probably change that. But in this case, we can check both HCPCS systems here. (CPT is a subset of HCPCS).

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +120,94 @@ public Bundle searchById(
...
+                  var cl = e.getClaimLine();
+                  var cp = e.getClaimProcedure();
+
+                  return procedureMatchesDict(cp, dict) || hcpcsCodeMatches(cl, dict);
+                });
+  }
+
+  // Checks DRG.
+  private boolean drgCodeMatches(Claim claim, Map<String, List<Map<String, Object>>> dict) {
+    var entries = dict.getOrDefault(SystemUrls.CMS_MS_DRG, List.of());
```
aschey-forpeople, line 165: We can use `get` instead of `getOrDefault` since these keys will always be present in the list. If not, we have bigger problems.

```suggestion
    var entries = dict.get(SystemUrls.CMS_MS_DRG);
```

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -0,0 +1,193 @@
...
+
+  @Test
+  void eobReadIdNotFound() {
+    assertThrows(ResourceNotFoundException.class, () -> eobRead().withId("999").execute());
+  }
+
+  @Test
+  void eobReadInvalidIdBadRequest() {
+    assertThrows(InvalidRequestException.class, () -> eobRead().withId("abc").execute());
+  }
```
aschey-forpeople, line 105: These tests look like duplicates of what's in `EobReadIT`.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +120,94 @@ public Bundle searchById(
...
+    for (var entry : entries) {
+      if (getDictCode(entry).equals(procedureCode)) {
+        return true;
+      }
+      var icdIndicatorOpt = proc.getIcdIndicator();
+      if (icdIndicatorOpt.isPresent()
+          && icdIndicatorOpt.get().getDiagnosisSytem().equals(link)
+          && getDictCode(entry).equals(diagnosisCode)) {
+        return true;
+      }
```
brick-green, line 208: it took my a while to read and get this. So I took a stab at cleaning it up a bit 
```
private boolean procedureMatchesDict(
            ClaimProcedure proc) {

        String diagnosisCode = getDiagnosisCode(proc);
        String procedureCode = getClaimProcedureCode(proc);

        // icd indicator identifies the ICD code system
        var icdIndicatorOpt = proc.getIcdIndicator().orElse(IcdIndicator.DEFAULT);

        var procSystem = icdIndicatorOpt.getProcedureSystem();
        // filter the security labels down based on the procedure system
        var procSystemSecurityLabels = SECURITY_LABELS_DICTIONARY.getOrDefault(procSystem, List.of());

        for (var entry : procSystemSecurityLabels) {
            String dictCode = getDictCode(entry);
            
            // if the prod code matches return true
            if (dictCode.equals(procedureCode)) {
                return true;
            }
            
            // if the diag system is the same as the proc system and the diag code matches return true
            if (icdIndicatorOpt.getDiagnosisSytem().equals(procSystem)
                    && dictCode.equals(diagnosisCode)) {
                return true;
            }
        }

        return false;
    }

```

brick-green: Also specifically wanted to call out the change I'm suggesting to use the DEFAULT IcdIndicator instead of throwing an exception. That probably requires discussion.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +120,94 @@ public Bundle searchById(
...
+    }
+
+    return false;
+  }
+
+  // Checks HCPCS and cpt.
+  private boolean hcpcsCodeMatches(
+      ClaimLine claimLine, Map<String, List<Map<String, Object>>> dict) {
+    String hcpcs = claimLine.getHcpcsCode().getHcpcsCode().map(String::toLowerCase).orElse("");
```
brick-green, line 178: Consider checking the `hcpcs` var first and returning early:

```
if (hcpcs.isEmpty()) {
    return false;
  }
```

brick-green (commented):
A few comments. Also I think we should avoid the use of `dict` or `dictionary` in our java classes. It's a small thing but since we also use python where Dictionary structures are used I think referring to maps as dictionaries in our java is confusing. What do you think?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
np-complete0: @aschey-forpeople wants to throw exception..

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -0,0 +1,193 @@
...
+        bundle.getEntry().stream()
+            .map(Bundle.BundleEntryComponent::getResource)
+            .filter(resource -> resource instanceof ExplanationOfBenefit)
+            .map(resource -> (ExplanationOfBenefit) resource)
+            .collect(Collectors.toList());
+    return fhirClaims;
+  }
+
+  private boolean containsDiagnosisCode(List<ExplanationOfBenefit> fhirClaims, String code) {
+    return fhirClaims.stream()
```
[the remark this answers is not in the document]
np-complete0: I have checked the api response all the samhsa values was removed

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
aschey-forpeople: My previous comment about throwing an exception was for doing the lookup on the system since those are hardcoded and will always be present. For the ICD indicator, I think we can default to ICD 10 if there isn't a valid value present. For adjudicated data, this indicator should be present and correct, but for partially adjudicated data, it may be incorrect. In that case, it's not our job to correct it, but partially adjudicated data is new so it should be using ICD 10 99% of the time (there are a few exceptions).

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -25,6 +31,10 @@ public class EobHandler {
   private final BeneficiaryRepository beneficiaryRepository;
   private final ClaimRepository claimRepository;
 
+  // Cache the security labels dictionary to avoid repeated I/O and parsing
+  private static final Map<String, List<Map<String, Object>>> SECURITY_LABELS_DICTIONARY =
```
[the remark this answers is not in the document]
aschey-forpeople: I'm not sure I understand this one - what regex is allowing `SECURITY_LABELS_DICTIONARY` but not `SECURITY_LABELS_MAP`?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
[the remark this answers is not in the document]
np-complete0: updated.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +120,94 @@ public Bundle searchById(
...
+    String procedureCode = getClaimProcedureCode(proc);
+
+    var link = proc.getIcdIndicator().orElseThrow().getProcedureSystem();
+    var entries = dict.getOrDefault(link, List.of());
+    for (var entry : entries) {
+      if (getDictCode(entry).equals(procedureCode)) {
+        return true;
+      }
+      var icdIndicatorOpt = proc.getIcdIndicator();
+      if (icdIndicatorOpt.isPresent()
```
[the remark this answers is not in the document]
np-complete0: updated.

apps/bfd-pipeline/bfd-pipeline-idr/test_samples2/merged_benes/SYNTHETIC_BENE_HSTRY.csv
```diff
@@ -3,3 +3,4 @@ BENE_SK,BENE_XREF_EFCTV_SK,BENE_XREF_SK,BENE_MBI_ID,BENE_LAST_NAME,BENE_1ST_NAME
 848484848,517782585,517782585,2B19C89AA35,Heeler,Alex,Joey,1959-03-14,,~,1,4,14152 ND-27,,,,,,Lisbon,58054,ND,~,2024-06-12 01:06:49.588129,2024-11-05 18:51:29.174693,9999-12-31 00:00:00.000000,9999-12-31 00:00:00.000000,Y
 121212121,517782585,517782585,2B19C89AA36,Heeler,Alex,Joey,1959-03-14,,~,1,4,14152 ND-27,,,,,,Lisbon,58054,ND,~,2024-06-12 01:06:49.588129,2024-11-05 18:51:29.174693,9999-12-31 00:00:00.000000,9999-12-31 00:00:00.000000,Y
 232323232,517782585,517782585,2B19C89AA37,Heeler,Alex,Joey,1959-03-14,,~,1,4,14152 ND-27,,,,,,Lisbon,58054,ND,~,2024-06-12 01:06:49.588129,2024-11-05 18:51:29.174693,9999-12-31 00:00:00.000000,9999-12-31 00:00:00.000000,Y
+412457726,412457726,412457726,5E77A12BB01,Doe,John,Q,1945-02-10,,~,1,1,123 Main St,,,,,,Anytown,12345,CA,~,2024-01-01 00:00:00.000000,2024-01-01 00:00:00.000000,9999-12-31 00:00:00.000000,9999-12-31 00:00:00.000000,Y
```
aschey-forpeople, line 6: Can you put the new test data into a folder called "samhsa" or something instead of tacking it on to the existing files? Trying to keep the data organized so it's not impossible to figure out which records go with which tests.

apps/bfd-pipeline/bfd-pipeline-idr/mock-idr.sql
```diff
@@ -322,6 +322,7 @@ CREATE TABLE cms_vdm_view_mdcr_prd.v2_mdcr_clm_line (
     clm_line_bene_pd_amt NUMERIC,
     clm_line_cvrd_pd_amt NUMERIC,
     clm_line_blood_ddctbl_amt NUMERIC,
+    -- SAMHSA (ICD)
     clm_line_dgns_cd VARCHAR(7),
```
aschey-forpeople, line 326: It doesn't look like this column is being checked. You'll have to add it to the entity since it's not currently mapped to anything.

apps/bfd-pipeline/bfd-pipeline-idr/mock-idr.sql
```diff
@@ -322,6 +322,7 @@ CREATE TABLE cms_vdm_view_mdcr_prd.v2_mdcr_clm_line (
     clm_line_bene_pd_amt NUMERIC,
     clm_line_cvrd_pd_amt NUMERIC,
     clm_line_blood_ddctbl_amt NUMERIC,
+    -- SAMHSA (ICD)
```
aschey-forpeople, line 325:
```suggestion
    -- SAMHSA (ICD) -  redundant with the corresponding value in clm_prod
```

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -0,0 +1,253 @@
...
+  // Code: 99408 System: CPT [clm_line_hcpcs_cd]
+  private static final String CPT = "99408";
+
+  // Code: HZ2ZZZZ System: ICD10 [clm_prcdr_cd]
+  private static final String ICDX = "HZ2ZZZZ";
+
+  // Code: 522 System: DRG [dgns_drg_cd]
+  private static final int DRG = 522;
+
+  protected final List<String> codeToCheck = List.of(DRG + "", HCPCS, ICD_X, ICDX);
```
aschey-forpeople, line 60: Maybe `DRG` should be declared as a string so you don't have to cast it like this?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -0,0 +1,253 @@
...
+  private final FhirContext context = FhirContext.forR4(); // Create a FhirContext for R4
+  private static final String FHIR_JSON = "fhir+json";
+
+  protected long claimUniqueId = 566745788569L;
+  protected long claimUniqueId2 = 3233800161009L;
+  protected long claimWithHcpcsInIcd = 6288598524935L;
+  protected long beneSk = 412457726;
+
+  // Code: F10.10 System: icd-10-cm [clm_dgns_cd]
+  private static final String ICD_X = "F10.10";
```
aschey-forpeople, line 46: Instead of calling these `ICD_X` and `ICDX`, maybe it should be `ICD_DIAGNOSIS` and `ICD_PROCEDURE`?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/SecurityLabels.java
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * <p>This method loads the YAML file named {@code security_labels.yml} from the classpath, parses
+   * it into a map structure, and returns the top-level keys from the map. Each key typically
+   * represents a system or category defined in the YAML.
+   *
+   * @param sLabels Security labels dictionary
+   * @return a list of keys present in the security labels map
+   * @throws IllegalArgumentException if the YAML file is not found in the classpath
+   */
+  public static List<String> getSecurityLabelKeys(Map<String, List<Map<String, Object>>> sLabels) {
```
aschey-forpeople, line 89: I think this is unused now?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +118,122 @@ public Bundle searchById(
...
+
+  // Checks ICDs.
+  private boolean procedureMatchesSamhsaCode(
+      ClaimProcedure proc, Map<String, List<Map<String, Object>>> securityLabels) {
+    String diagnosisCode = getDiagnosisCode(proc);
+    String procedureCode = getClaimProcedureCode(proc);
+
+    // Fallback to DEFAULT (currently ICD-9) if absent; may adjust to ICD_10 per business decision.
+    var icdIndicator = proc.getIcdIndicator().orElse(IcdIndicator.ICD_10);
+    var procSystem = icdIndicator.getProcedureSystem();
```
aschey-forpeople, line 223: This doesn't look quite right - the diagnosis code will have a different system than the procedure code. Take a look at the `IcdIndicator` class.

apps/bfd-pipeline/bfd-pipeline-idr/mock-idr.sql
np-complete0: ?

apps/bfd-pipeline/bfd-pipeline-idr/mock-idr.sql
np-complete0: so we keep it or remove it 
image
?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
np-complete0: trivial

np-complete0:
@aschey-forpeople

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +118,133 @@ public Bundle searchById(
...
+    if (!procedureCode.isEmpty()) {
+      boolean procMatch =
+          procedureEntries.stream().anyMatch(e -> compare(getCode(e), procedureCode));
+      if (procMatch) {
+        return true;
+      }
+    }
+
+    // Only check diagnosis when indicator explicitly identifies ICD-9 or ICD-10.
+    if ((icdIndicator == IcdIndicator.ICD_9 || icdIndicator == IcdIndicator.ICD_10)
```
aschey-forpeople, line 240: Do we need this check? It's already defaulting to ICD 10 above so it should always be one of these two values here.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +118,132 @@ public Bundle searchById(
...
+  private boolean procedureMatchesSamhsaCode(
+      ClaimProcedure proc, Map<String, List<Map<String, Object>>> securityLabels) {
+    String diagnosisCode = getDiagnosisCode(proc);
+    String procedureCode = getClaimProcedureCode(proc);
+    var icdIndicator = proc.getIcdIndicator().orElse(IcdIndicator.ICD_10); // Fallback ICD10
+
+    // Procedure codes live under the CMS procedure system
+    // diagnosis codes under the HL7 CM system.
+    var procedureEntries =
+        securityLabels.getOrDefault(icdIndicator.getProcedureSystem(), List.of());
```
aschey-forpeople, line 227: This should be `get`

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +118,132 @@ public Bundle searchById(
...
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
+  // Checks ICDs.
+  private boolean procedureMatchesSamhsaCode(
+      ClaimProcedure proc, Map<String, List<Map<String, Object>>> securityLabels) {
```
aschey-forpeople, line 219: Can remove the second parameter since this is a constant

apps/bfd-server-ng/pom.xml
```diff
@@ -107,6 +109,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>2.2</version>
```
aschey-forpeople, line 115: The dependency versions should be set in the root POM. You can look at our other dependencies for examples.

apps/bfd-server-ng/pom.xml
```diff
@@ -224,7 +232,23 @@
                 <artifactId>spring-boot-maven-plugin</artifactId>
                 <version>${spring-boot.version}</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
aschey-forpeople, line 236: This is already configured in the root POM. I don't think we need to add it here too?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -42,7 +54,7 @@ public Optional<ExplanationOfBenefit> find(final Long fhirId) {
    * @param count record count
    * @param serviceDate service date
    * @param lastUpdated last updated
-   * @param startIndex start index
+   * @param startIndex start 160
```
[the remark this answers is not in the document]
np-complete0: that is not my line

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/SecurityLabels.java
```diff
@@ -0,0 +1,77 @@
...
+   * @param inputStream the path to the YAML file to be read
+   * @return a map where each key is a "system" value, and the value is a list of entries (maps)
+   *     associated with that system
+   */
+  private static Map<String, List<Map<String, Object>>> readYamlToMap(InputStream inputStream) {
+    var yaml = new Yaml();
+    var resultMap = new HashMap<String, List<Map<String, Object>>>();
+
+    try {
+      List<Map<String, Object>> yamlData = yaml.load(inputStream);
```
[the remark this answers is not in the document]
np-complete0: AC needs to be updated.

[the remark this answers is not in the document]
aschey-forpeople: Take a look at the last AC, it's there.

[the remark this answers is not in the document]
np-complete0: Also individual claim have `CLM_LINE_FROM_DT` and `CLM_LINE_THRU_DT`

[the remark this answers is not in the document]
aschey-forpeople: @alex-dzeda can weigh in on that ^ but I think in the past we just used the dates on the claim itself.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +76,34 @@ public Bundle searchByBene(
...
+   */
+  protected Bundle buildEobBundleExcludingSamhsa(List<Claim> claims, boolean excludeSamhsa) {
+
+    var filteredClaims = claims;
+    if (excludeSamhsa) {
+      filteredClaims =
+          claims.stream()
+              // keep claims that do NOT contain any SAMHSA-coded procedures
+              .filter(claim -> !claimHasSamhsa(claim))
+              .toList();
```
aschey-forpeople, line 101: I'd prefer to avoid casting the final results if we can help it since it's easy to cause runtime exceptions with that approach. If we avoid calling `toList` and streaming the data twice, we can do that.

```suggestion
    var claimStream = claims.stream();
    if (excludeSamhsa) {
      claimStream =
              claimStream
              // keep claims that do NOT contain any SAMHSA-coded procedures
              .filter(claim -> !claimHasSamhsa(claim));
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -61,8 +76,34 @@ public Bundle searchByBene(
...
+    }
+
     return FhirUtil.bundleOrDefault(
-        eobs.stream().map(Claim::toFhir), claimRepository::claimLastUpdated);
+        filteredClaims.stream().map(Claim::toFhir).map(Resource.class::cast),
+        claimRepository::claimLastUpdated);
```
aschey-forpeople, line 106:
```suggestion
        claimStream.map(Claim::toFhir), claimRepository::claimLastUpdated);
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +122,188 @@ public Bundle searchById(
 
   private Optional<ExplanationOfBenefit> searchByIdInner(
       Long claimUniqueId, DateTimeRange serviceDate, DateTimeRange lastUpdated) {
-    var claim = claimRepository.findById(claimUniqueId, serviceDate, lastUpdated);
+    var eobs = claimRepository.findById(claimUniqueId, serviceDate, lastUpdated).stream().toList();
+    var bundle = buildEobBundleExcludingSamhsa(eobs, true);
```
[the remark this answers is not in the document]
aschey-forpeople: If you look at the logic used in `bundleOrDefault` it's actually making a database query, so we definitely want to avoid calling that more than necessary.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +113,192 @@ public Bundle searchById(
...
+
+    // Only check diagnosis when indicator explicitly identifies ICD-9 or ICD-10.
+    if (!diagnosisCode.isEmpty()) {
+      boolean diagMatch =
+          diagnosisEntries.stream().anyMatch(e -> compare(getCode(e), diagnosisCode));
+      if (diagMatch) {
+        return true;
+      }
+    }
+    return false;
```
aschey-forpeople, line 270: The date check is missing here. That logic should probably be somewhere more central so we don't forget it.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -81,8 +122,188 @@ public Bundle searchById(
...
+  }
+
+  // Returns true if the given claim contains any procedure that matches a SAMHSA
+  // security label code from the dictionary.
+  private boolean claimHasSamhsa(Claim claim) {
+    var claimFromDate = claim.getBillablePeriod().getClaimFromDate();
+    var claimThroughDate = claim.getBillablePeriod().getClaimThroughDate();
+
+    // clm_from_dt of the claim comes after clm_thru_dt, indicates bad data.
+    var referenceDay = claimFromDate.isAfter(claimThroughDate) ? claimFromDate : claimThroughDate;
```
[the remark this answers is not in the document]
np-complete0: well this was the consensus from yesterday:

image

Also in Jira:
image

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -0,0 +1,399 @@
...
+    var eoBenefits = eobRead().withId(claimUniqueId).execute();
+    assertFalse(
+        codeToCheck.stream().anyMatch(e -> containsSamhsaCodeAnywhere(List.of(eoBenefits), e)));
+  }
+
+  @Test
+  void claimWithWrongCode() {
+    // Wrongly added HCPCS Samhsa code should not be removed
+    // even it is a Samhsa code it is in wrong system
+    // Samhsa code H0036 becomes H00.36 in ICD-10
```
aschey-forpeople, line 115: Where is H0036 from? I don't see that on the SAMHSA list.

np-complete0: it is the same for hcpcs code H0034 in icm field it becomes h0034. Updated.

PR #2779, opened 2025-08-15
BFD-4190: Refactor Coverage endpoints and add additional test cases

aschey-forpeople:
**JIRA Ticket:**
BFD-4190

### What Does This PR Do?

Replace me.

### Validation


apps/bfd-pipeline/bfd-pipeline-idr/load_synthetic.py
```diff
@@ -32,45 +32,54 @@
...
+            # skip empty files
+            if reader.fieldnames is None:
+                continue
+
+            # fetch the list of columns from the database and filter them out
+            # so we don't get errors trying to insert extra columns
+            db_columns = cur.execute(
+                f"""
+                    SELECT column_name FROM INFORMATION_SCHEMA.COLUMNS 
+                    WHERE table_name = '{sql_table}'
```
brick-green, line 61: Any thoughts on when to use f-string vs parameters? Looks like we lean f-string for the most part

brick-green (approved):
LGTM. Just one question

apps/bfd-pipeline/bfd-pipeline-idr/load_synthetic.py
aschey-forpeople: I prefer f-strings. It's a bit more readable in my opinion. But for log messages, `ruff` tells you to use format parameters instead of f-strings, see https://docs.astral.sh/ruff/rules/logging-f-string/

brick-green (approved):

apps/bfd-pipeline/bfd-pipeline-idr/load_synthetic.py
malessi: Hmm... is this correct? As far as I understand, `psycop3` recommends using [`execute` arguments plus placeholders](https://www.psycopg.org/psycopg3/docs/basic/params.html#execute-arguments) to add parameters to queries. They don't have an explicit f-string example, but their [example for SQL Injection dangers](https://www.psycopg.org/psycopg3/docs/basic/params.html#danger-sql-injection) shows them using the `%` binary string operator which is _roughly_ equivalent to f-strings.


There is [experimental support](https://www.psycopg.org/psycopg3/docs/basic/tstrings.html) for _template_ strings starting with `psycopg3` 3.3 and Python 3.14, but such strings start with `t"..."`, not `f"..."`.

This isn't something we need to address _now_, but maybe soon?

malessi (approved):

PR #2781, opened 2025-08-20
BFD-4193: Restructure bfd-model-idr, set up formatting/linting, and add CI

aschey-forpeople:
**JIRA Ticket:**
BFD-4193

### What Does This PR Do?

- Updates repo structure
  - Moves the location of `bfd-model-idr` from `apps/bfd-model/bfd-model-idr` to `apps/bfd-model-idr`
  - Flattens directory structure within `bfd-model-idr` to minimize nesting
  - These changes should make it easier to locate the correct paths and minimize the number of `cd` operations
- Formats all python code and fixes existing lint warnings
- Adds CI for the pipeline and model
  - The complete process of generating and loading synthetic data will be checked for both the pipeline and model
  - This will catch regressions caused by changes to the data model or pipeline
- Adds a new script `compile-all-resources.sh` to reduce the need to copy commands from the README
- Updates README to be a bit easier to follow

### Validation

Tested in CI, re-ran updated scripts


.github/workflows/ci-pipeline-synthetic.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
+      with:
+          version: "0.8.10"
+    - name: Install node
+      uses: actions/setup-node@v4
+      with:
+        node-version: '22'
+    - name: Install dependencies
+      run: |
```
aschey-forpeople, line 27: I was hoping to run the FHIR validator here as well, but it was causing out of memory exceptions. It seems to need a lot of resources for when it does the initial phase of pulling in all the code systems without any cached data.

apps/bfd-model-idr/gen_dd.py
```diff
@@ -0,0 +1,180 @@
+import json
+import os
+import subprocess
+from pathlib import Path
+
+import pandas as pd
+import yaml
+
+print("Generating Data Dictionary")
```
aschey-forpeople, line 9: There's no meaningful changes here. Seems like the formatter changed it enough that github doesn't realize it's the same file.

apps/bfd-model-idr/pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+    "FURB",
+    "RUF",
+]
+# Don't require docstrings
+ignore = ["D100", "D101", "D102", "D103", "D104", "D105", "D107"]
+
+[tool.ruff.lint.pydocstyle]
+convention = "pep257"
+
+[tool.ruff.format]
```
aschey-forpeople, line 59: I didn't configure pyright here because it will be a lot of work to figure out all the correct types to use. Anything involving pandas is a nightmare for static types. It may not be worth it in this case.

apps/bfd-model-idr/.gitignore
brick-green, line 1: looks like some of these were left in `/bfd-model/bfd-model-idr/`. When I checked out the branch I'm seeing 161 unversioned files. I think you meant `output/` instead of `out/` too.

brick-green (changes requested):

apps/bfd-model-idr/.gitignore
aschey-forpeople: I combined `out` and `output` so everything goes in `out` now. You might have to delete those files manually. Since they were untracked, git doesn't know about them, and now those files aren't part of `gitignore`.

brick-green: oh right, makes sense

.github/workflows/ci-pipeline-synthetic.yml
aschey-forpeople: Yeah, makes sense. I was mostly hoping to make sure it runs without blowing up for now since there have been a few issues with typos and things that caused it to fail that could've been easily caught with CI.

PR #2723, opened 2025-06-20
BFD-4143

alex-dzeda:
**JIRA Ticket:**
BFD-4143
BFD-4138 <--- incorporated due to timing. 

### What Does This PR Do?

This PR adds the remaining institutional mappings for BFD v3, as well as the sample-data generation. 

### What Should Reviewers Watch For?

If you're reviewing this PR, please check for these things in particular:

### What Security Implications Does This PR Have?

Please indicate if this PR does any of the following:  

* Adds any new software dependencies
* Modifies any security controls
* Adds new transmission or storage of data
* Any other changes that could possibly affect security? 

* [X] I have considered the above security implications as it relates to this PR. (If one or more of the above apply, it cannot be merged without the ISSO or team security engineer's (`@sb-benohe`) approval.) 
* [X] I have created tests to sufficiently ensure the reliability of my code, if applicable. If this is a modification to an existing piece of code, I have audited the associated tests to ensure everything works as expected.

### Validation

**Have you fully verified and tested these changes? Is the acceptance criteria met? Please provide reproducible testing instructions, code snippets, or screenshots as applicable.**
Outpatient + HHA + SNF + Hospice all appearing as expected using variations of the input sample data.

PR #2787, opened 2025-08-27
BFD-4227: Institutional EOB Mappings

brick-green:
**JIRA Ticket:**
BFD-4227

### What Does This PR Do?

Adds mapping as defined in this [PR](https://github.com/CMSgov/beneficiary-fhir-data/pull/2723). That includes introducing the FISS table along with fields in claim institutional.

- review the mapping definition [PR](https://github.com/CMSgov/beneficiary-fhir-data/pull/2723) and compare the implementation
- pay close attention the the code sets that are defined and the converters
- check the tests

### Validation


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/Claim.java
```diff
@@ -123,6 +132,12 @@ public ExplanationOfBenefit toFhir() {
               eob.addContained(p);
               eob.setProvider(new Reference(p));
             });
+    claimSourceId.toFhirOutcome().ifPresent(eob::setOutcome);
+    claimTypeCode.toFhirOutcome().ifPresent(eob::setOutcome);
```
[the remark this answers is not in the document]
brick-green: All three of these setOutcome (these two above and the one below) are mutually exclusive. But operate on different claim data elements. 

I could combine the logic in `claimTypeCode.toFhirOutcome()` into `claimFiss.toFhirOutcome()` since I'm already passing in the claimTypeCode there. But that feels like it crosses over the separation of concerns line to me. Then that doesn't really solve the problem as there are still two setOutcome left. I could move all the logic into claim but that goes against the pattern you've set up. Another option is to check if `outcome` is already set in claim. But then the priority order of which logic is used  is kind of arbitrary. But since they're mutually exclusive anyway I guess that's not an issue.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimObligatedToAcceptasFinalOneIndicatorCode.java
```diff
@@ -0,0 +1,23 @@
+package gov.cms.bfd.server.ng.claim.model;
+
+import gov.cms.bfd.server.ng.SystemUrls;
+import jakarta.persistence.Column;
+import java.util.Optional;
+import org.hl7.fhir.r4.model.Coding;
+import org.hl7.fhir.r4.model.Extension;
+
+public class ClaimObligatedToAcceptasFinalOneIndicatorCode {
```
aschey-forpeople, line 9:
```suggestion
public class ClaimObligatedToAcceptAsFinalOneIndicatorCode {
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimFiss.java
```diff
@@ -0,0 +1,33 @@
...
+  private long claimUniqueId;
+
+  @OneToOne(mappedBy = "claimFiss")
+  private Claim claim;
+
+  @Column(name = "clm_crnt_stus_cd")
+  private ClaimCurrentStatusCode claimCurrentStatusCode;
+
+  Optional<ExplanationOfBenefit.RemittanceOutcome> toFhirOutcome(int claimTypecode) {
+    if (claimTypecode >= 2000 && claimTypecode < 3000) {
```
aschey-forpeople, line 28: Might be nice to make a helper method in the `ClaimTypeCode` class called `isPacStage2` or something that performs this check, that way it gives a bit more context to what this range of code means.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/SystemUrls.java
```diff
@@ -324,15 +325,15 @@ public class SystemUrls {
       "http://hl7.org/fhir/us/carin-bb/CodeSystem/C4BBAdjudicationDiscriminator";
 
   /** URL. */
-  public static String CARIN_CODE_SYSTEM_ADJUDICATION =
+  public static final String CARIN_CODE_SYSTEM_ADJUDICATION =
```
aschey-forpeople, line 328: Thanks 👍

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/SystemUrls.java
```diff
@@ -456,6 +457,46 @@ public class SystemUrls {
   public static final String BLUE_BUTTON_CODE_SYSTEM_FINAL_ACTION_CODE =
       "https://bluebutton.cms.gov/fhir/CodeSystem/CLM-FI-ACTN-CD";
 
+  /** URL. */
+  public static final String BLUE_BUTTON_STRUCTURE_DEFINITION_REVENUE_DISCOUNT_INDICATOR_CODE =
```
aschey-forpeople, line 461: The reason these existing URLs didn't have proper comments is because I had one PR without 100 new URLs and I ran out of time to add proper comments. I took a stab at adding better comments to these in my [new PR](https://github.com/CMSgov/beneficiary-fhir-data/pull/2792/files#diff-84259481c88825ccc18c3e670e8ab1730f56cafe7e4a084fe88be6ccb70661b0R343)

apps/bfd-server-ng/codegen/main.py
```diff
@@ -36,12 +37,11 @@ def get_enum_val(concept, int_codes):
 
 
 if __name__ == "__main__":
```
aschey-forpeople, line 39: Sonar is currently complaining about the names of the enums this generates: https://sonarqube.cloud.cms.gov/project/issues?pullRequest=2787&issueStatuses=OPEN%2CCONFIRMED&id=bfd-parent

Can you add an annotation to the generated enums here so it will ignore [this rule](https://sonarqube.cloud.cms.gov/coding_rules?open=java%3AS115&rule_key=java%3AS115)? This is actually a bug in Sonar - the regex they're using isn't accounting for all valid cases. Not much we can do about that though.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimAnsiSignature.java
```diff
@@ -32,18 +32,33 @@ public class ClaimAnsiSignature {
...
-                revenueCenterAnsiReasonCode3,
-                revenueCenterAnsiReasonCode4)
-            .flatMap(Optional::stream);
-    return codes.map(this::toAdjudicationComponent).toList();
+    return Stream.of(
+            new Pair(revenueCenterAnsiGroupCode1, revenueCenterAnsiReasonCode1),
+            new Pair(revenueCenterAnsiGroupCode2, revenueCenterAnsiReasonCode2),
+            new Pair(revenueCenterAnsiGroupCode3, revenueCenterAnsiReasonCode3),
+            new Pair(revenueCenterAnsiGroupCode4, revenueCenterAnsiReasonCode4))
+        .flatMap(
```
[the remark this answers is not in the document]
brick-green: Found a way to simplify this section quite a bit. I don't think a comment is needed anymore

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/BenefitBalanceInstitutional.java
```diff
@@ -19,6 +22,12 @@ class BenefitBalanceInstitutional {
   @Column(name = "clm_instnl_per_diem_amt")
   private double perDiemAmount;
 
+  @Column(name = "clm_mdcr_hha_tot_visit_cnt")
+  private int totalHHAVisits;
+
+  @Column(name = "clm_mdcr_hospc_prd_cnt")
+  private int totalHospicePeriodTrailers;
```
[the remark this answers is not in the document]
aschey-forpeople: Huh, yeah I'm not familiar with that term. Might be something specific to the data format.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/RevenueCenterAnsiGroupCode.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.hl7.fhir.r4.model.Coding;
+
+/** Revenue center ANSI adjustment group codes for claims. */
+@Getter
+@AllArgsConstructor
```
alex-dzeda, line 13: @aschey-forpeople / @brick-green one thing we may want to look at in the future is just pulling these from the .fsh file, or the artifact, directly, instead of having them in multiple spots?

aschey-forpeople (approved):

PR #2796, opened 2025-09-04
BFD-4263: Add env vars to load benes/claims selectively + pipeline updates

aschey-forpeople:
**JIRA Ticket:**
BFD-4263

### What Does This PR Do?

- Adds two (temporary) variables that we can use to only load claims/benes. This won't have much utility once we've finalized all the claim filters.
- Adds a new progress tracking column to track the last loaded ID within a timestamp range. This was added because some of the historical beneficiary data has over 1 billion records with the same updated timestamp. If the pipeline restarts while loading that batch, it would have to reload all of those records.
- Handles some weird data from IDR with null bytes
- Fixes the transient exception handler
- Filters null MBI records and makes the MBI field mandatory in the model

I've set up the pipeline to load data hourly using a simple script for now. This isn't in the repo because it's a temporary thing, but here it is for posterity:

```sh
while true; do
  echo "starting pipeline"
  uv run pipeline.py 2&>test.log
  echo "sleeping 1 hour"
  sleep 3600
done
```

### Validation

Ran against prod data, all bene data loaded successfully


alex-dzeda (dismissed):

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/Configuration.java
```diff
@@ -97,12 +97,6 @@ private JdbcConnectionDetails getJdbcConfiguration() {
     return Objects.requireNonNullElseGet(jdbcConnectionDetails, () -> new JdbcConfiguration(this));
   }
 
-  private DatabaseOptions.AuthenticationType getAuthenticationType() {
```
aschey-forpeople, line 100: Thanks @meliGuzman for noticing this 👍

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/Configuration.java
```diff
@@ -97,12 +97,6 @@ private JdbcConnectionDetails getJdbcConfiguration() {
...
-    return useRds()
-        ? DatabaseOptions.AuthenticationType.JDBC
-        : DatabaseOptions.AuthenticationType.RDS;
```
aschey-forpeople, line 103: RDS authentication isn't implemented yet so this isn't necessary.

apps/bfd-pipeline/bfd-pipeline-idr/model.py
```diff
@@ -7,6 +7,8 @@
 
 from constants import CLAIM_TYPE_CODES, DEFAULT_MAX_DATE, DEFAULT_MIN_DATE
 
+type DbType = str | float | int | bool | date | datetime
```
np-complete0, line 10: I think this requires Python >= 3.12

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/patient/PatientHandler.java
```diff
@@ -68,7 +68,7 @@ private Patient toFhir(Beneficiary beneficiary) {
       // check for merged bene and if mbi identifier has already been added to the patient
       if (!beneficiary.isMergedBeneficiary()
           && patient.getIdentifier().stream()
-              .noneMatch(identifier -> identifier.getValue().equals(id.getMbi().orElse("")))) {
+              .noneMatch(identifier -> identifier.getValue().equals(id.getMbi()))) {
         id.toFhirIdentifier().ifPresent(patient::addIdentifier);
```
np-complete0, line 72: should we add unit test for
- null mbi
- empty string
- non-empty mbi

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/model/BeneficiaryIdentity.java
```diff
@@ -47,7 +47,7 @@ public Optional<Identifier> toFhirIdentifier() {
     if (mbi.isEmpty()) {
```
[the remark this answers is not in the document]
aschey-forpeople: Yep, good catch. This check isn't needed anymore - removed.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/patient/PatientHandler.java
aschey-forpeople: MBI will always be a valid, non-empty string now that we're filtering any null records.

PR #2802, opened 2025-09-11
BFD-4278: V3 server - Log pulled bene_sks

MahiFentaye:
**JIRA Ticket:**
BFD-4278

### What Does This PR Do?

V3 server - Log pulled bene_sks

### Validation

    Tested locally

### DIFF
--- a/apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/LogUtil.java
+++ b/apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/LogUtil.java
@@ -4,6 +4,12 @@ import org.slf4j.LoggerFactory;
 
 /** Utility class used for logging. */
 public class LogUtil {
+
+  // Private constructor to prevent instantiation
+  private LogUtil() {
+    // Intentionally empty
+  }
+
   private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(LogUtil.class);
 
   /**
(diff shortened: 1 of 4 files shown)


brick-green (commented):
The order of my comments might be kind of weird. But the main idea I have is to avoid passing in Optional to the `logBeneSkIfPresent()` methods and instead handle the null check before calling.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/Logger.java
brick-green, line 1: I think this class should be renamed to separate it conceptually from the slf4j.Logger class. Even something like `StructuredLogger` would be better. But maybe we can think of something more domain specific?

brick-green (commented):
I'd like to see the logging class renamed but the other suggestions are up to you

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/LogUtil.java
```diff
@@ -0,0 +1,21 @@
...
+/** Utility class used for logging. */
+public class LogUtil {
+  private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(LogUtil.class);
+
+  /**
+   * For logging bene_sk.
+   *
+   * @param beneSk the bene_sk
+   */
+  public static void logBeneSkIfPresent(Long beneSk) {
```
aschey-forpeople, line 14: `IfPresent` is redundant in the method name now since it's not checking if it's present.
```suggestion
  public static void logBeneSk(Long beneSk) {
```

aschey-forpeople (changes requested):
Can you address the sonar issues before merging? https://sonarqube.cloud.cms.gov/project/issues?pullRequest=2802&issueStatuses=OPEN%2CCONFIRMED&id=bfd-parent

[This one](https://sonarqube.cloud.cms.gov/coding_rules?open=java%3AS2272&rule_key=java%3AS2272) can be ignored since it's designed to be an infinite sequence.

PR #2820, opened 2025-09-25
BFD-4303: Format ICD-9 codes.

np-complete0:
**JIRA Ticket:**
BFD-4303

### What Does This PR Do?
Formats ICD-9 codes by the following logic.
If the code contains a '.' returns code the unchanged. 
If fully numeric and length > 3, inserts '.' after the 3rd char. 
If starts with 'E' and length > 4, insert '.' after the 4th char. 
If starts with 'V' and length > 3, insert '.' after the 3rd char. All other systems it returns the code as-is.

- Verify no production logic was altered.
- Implementation correctly formats ICD9 codes

   No
   No
   New snap file
   This PR should not affect security of the application.

### Validation


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/IcdIndicator.java
```diff
@@ -29,20 +29,76 @@ public static Optional<IcdIndicator> tryFromCode(String code) {
...
+   * '.' after the 3rd char. If starts with 'E' and length > 4, insert '.' after the 4th char. If
+   * starts with 'V' and length > 3, insert '.' after the 3rd char. All other systems it returns the
+   * code as-is.
    *
    * @param rawCode The raw diagnosis code string from the database.
    * @return The correctly formatted diagnosis code string for use in FHIR.
    */
   public String formatCode(String rawCode) {
-    if (this == ICD_10 && rawCode.length() > 3) {
+    if (rawCode == null || rawCode.isEmpty()) return rawCode;
```
aschey-forpeople, line 46: This isn't needed - null or empty codes are mapped as `Optional.empty`

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/IcdIndicator.java
```diff
@@ -29,20 +29,76 @@ public static Optional<IcdIndicator> tryFromCode(String code) {
...
+   * @return formatted procedure code
+   */
+  public String formatProcedureCode(String rawCode) {
+    if (rawCode == null || rawCode.isEmpty()) return rawCode;
+    if (rawCode.indexOf('.') >= 0) return rawCode;
+
+    if (this == ICD_9) {
+      if (rawCode.length() > 2) {
+        return rawCode.substring(0, 2) + "." + rawCode.substring(2);
+      }
```
aschey-forpeople, line 98: Probably makes sense to move this logic into a helper function since we're repeating it several times.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/IcdIndicatorIT.java
```diff
@@ -0,0 +1,81 @@
...
+            FORMATTED_ICD_9_CODE_25000,
+            FORMATTED_ICD_9_CODE_E8889,
+            FORMATTED_ICD_9_CODE_V1005);
+    assertEquals(expectedOrder, eobCodes, "ICD-9 codes should be formatted and ordered");
+    assertEquals(FORMATTED_ICD_9_CODE_25000, independentlyFormatted.get("25000"));
+    assertEquals(FORMATTED_ICD_9_CODE_E8889, independentlyFormatted.get("E8889"));
+    assertEquals(FORMATTED_ICD_9_CODE_V1005, independentlyFormatted.get("V1005"));
+
+    // Snapshot a compact payload instead of the entire EOB to reduce churn.
+    var snapshotPayload = new java.util.LinkedHashMap<String, Object>();
```
aschey-forpeople, line 56: I think we can just use a string comparison for this rather than saving it to a snapshot. The snapshots are really just for the large JSON payloads since it's not feasible to do a simple string comparison with those.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/IcdIndicator.java
```diff
@@ -29,19 +29,74 @@ public static Optional<IcdIndicator> tryFromCode(String code) {
...
+    if (rawCode.indexOf('.') >= 0) return rawCode;
+
+    if (this == ICD_10) return formatIcd10(rawCode);
+    if (this == ICD_9) return formatIcd9Diagnosis(rawCode);
+
+    return rawCode;
+  }
+
+  private String formatIcd10(String rawCode) {
+    if (rawCode.length() > 3) return insertDot(rawCode, 3);
```
aschey-forpeople, line 56: These length checks are redundant since you're already checking the length in `insertDot`.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/IcdIndicator.java
```diff
@@ -29,19 +29,73 @@ public static Optional<IcdIndicator> tryFromCode(String code) {
...
+   * code as-is.
    *
    * @param rawCode The raw diagnosis code string from the database.
    * @return The correctly formatted diagnosis code string for use in FHIR.
    */
   public String formatCode(String rawCode) {
-    if (this == ICD_10 && rawCode.length() > 3) {
-      return rawCode.substring(0, 3) + "." + rawCode.substring(3);
+    // If the code contains a dot, it's already formatted.
+    if (rawCode.contains(".")) return rawCode;
```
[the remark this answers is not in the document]
aschey-forpeople: That may be an optional rule that's not configured on our SonarQube server. It's saying that it doesn't like the conditional statement and return together without curly braces. I tend to prefer that we always use curly braces for conditionals, but I don't feel _too_ strongly about it.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/IcdIndicator.java
```diff
@@ -32,16 +65,59 @@ public static Optional<IcdIndicator> tryFromCode(String code) {
...
+    }
+    // Codes starting with 'V' . after 3rd char when long enough.
+    if (Character.toUpperCase(rawCode.charAt(0)) == 'V') {
+      return insertDot(rawCode, 3);
+    }
+
+    return rawCode;
+  }
+
+  private static String formatIcd9Procedure(String rawCode) {
```
aschey-forpeople, line 111: I noticed the AC says the dot should be after the second digit, but it looks like all of the ICD 9 procedure codes are numeric only, so I think this is fine.

PR #2832, opened 2025-10-02
BFD-4312: Implement Hamilton & Ray to Run IDR Pipeline

mel1-G:
**JIRA Ticket:**
BFD-4312

### What Does This PR Do?

- created parallelized workloads using the Ray adapter while maintaining dependency ordering of tables
> - Stage 1: Load auxiliary claim tables in parallel
> - Stage 2: Load main claim table
> - Stage 3: Load only overshared MBIs
> - Stage 4: Load auxiliary beneficiary tables in parallel
> - Stage 5: Load main beneficiary table last
- added worker logging configuration so we could see all logs from workers 
- refactored our timer logic since the old way wouldn't worked in our parallelized workflow
- added an initial commit for load_progress to better distinguish between "not started" vs "started but still working" since some tables are small and get loaded quickly but larger tables take much longer to extract/load. Before we wouldn't have visibility in the load_progress table on whether or not tables were being loaded into in parallel since we used to insert into load progress in the same transaction where we also loaded the data meaning if there was a lot of data to load, the initial insertion into load_progress would not be committed until the first batch was loaded in. 
- updated loader & extractor to return data that can be serialized. Before, some of our pipeline functions returned objects like (PostgresLoader, bool) and with the Ray adapter, hamilton nodes need to return data that can serialized so instead if returning live DB connections, I return serializable values. Each worker gets a connection string and builds its own PostgresLoader/SnowflakeExtractor locally.

Helpful resources to better understand Hamilton & Ray:
- Hamilton basics: https://www.tryhamilton.dev/intro 
- Quick notes on the Ray adapter: https://hamilton.apache.org/reference/graph-adapters/RayGraphAdapter/ 
- More detailed info on Ray: https://docs.ray.io/en/latest/ray-core/api/index.html

Metrics:
- Inserts take about 1 second, verified by observing the timer logs after running the pipeline with production data
- Each batch gets processed in a few seconds based off worker timer logs for all the phases (load next batch from IDR, copy data, insert)
- Checked memory and CPU usage using htop. RAM remains lows. CPU usage sometimes hits 100%. It happens infrequently with the larger claim tables so can benefit from bumping up cpu resources 
- Some htop screenshots:

Screenshot 2025-10-01 at 2 17 36 PM

Screenshot 2025-10-01 at 12 19 52 PM

Notes:
- I decided to create separate Hamilton versions of the pipeline and the related files to avoid merge conflicts
- Future improvements
> - Bump up the cpu resources since we observed some workers hitting 100% cpu usage sometimes
> - Use lifecycle hooks to implement a timer for Ray workers. We can replace our current timer utility for better monitoring 

- The main pipeline & related files like extractor and loader were being changed by others so please verify my hamilton versions are up to date. Should be about the same except for how we handle the connection due to the serialization 
- run locally and observe the logs and verify data is populated in your local db: uv run ./hamilton_pipeline.py local
- bonus: read up on hamilton & ray from the resources I listed above so you can better validate this implementation

* Adds any new software dependencies: YES, @sb-benohe
* Modifies any security controls: NO
* Adds new transmission or storage of data: NO

### Validation

- ran mvn clean verify
- ran locally and observed the logs and verified data was populated in my local db: uv run ./hamilton_pipeline.py local
- Created a separate database in prod and ran the pipeline in there. Verified data was populated, monitored timer logs, and used htop to collect metrics
- verified dependency ordering by printing registered nodes and their dependencies like so where dr is my driver (what executes the nodes):

> graph = dr.graph  
> for node_name, node in graph.nodes.items():  
>     print(f"Node: {node_name}")  
>     print(f"  Dependencies (incoming edges): {list(node.input_types.keys())}")


apps/bfd-pipeline/bfd-pipeline-idr/hamilton_extractor.py
```diff
@@ -0,0 +1,204 @@
...
+                    """,
+            ).replace("{ORDER_BY}", order_by),
+            {"timestamp": compare_timestamp},
+        )
+
+
+class PostgresExtractor(Extractor):
+    def __init__(self, connection_string: str, batch_size: int) -> None:
+        super().__init__()
+        self.connection_string = connection_string
```
mel1-G, line 117: only change here is that I removed conn to make PostgresExtractor serializable for Hamilton & Ray to work with. The connection gets created when needed so line 123 in extract_many for example

apps/bfd-pipeline/bfd-pipeline-idr/hamilton_loader.py
```diff
@@ -0,0 +1,276 @@
...
+    return f"host={os.environ['BFD_DB_ENDPOINT']} port={port} dbname={dbname} \
+        user={os.environ['BFD_DB_USERNAME']} password={os.environ['BFD_DB_PASSWORD']}"
+
+
+class PostgresLoader:
+    def __init__(
+        self,
+        connection_string: str,
+    ) -> None:
+        self.connection_string = connection_string
```
mel1-G, line 33: Same comment as PostgresExtractor

apps/bfd-pipeline/bfd-pipeline-idr/hamilton_loader.py
```diff
@@ -0,0 +1,276 @@
...
+class BatchLoader:
+    def __init__(
+        self,
+        conn: psycopg.Connection,
+        fetch_results: Iterator[list[T]],
+        model: type[T],
+        batch_start: datetime,
+        progress: LoadProgress | None,
+    ) -> None:
+        self.conn = conn
```
mel1-G, line 64: This is fine to remain as is since load returns the result of BatchLoader(...).load() only

apps/bfd-pipeline/bfd-pipeline-idr/hamilton_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+    mode = sys.argv[1] if len(sys.argv) > 1 else ""
+    if mode == "local":
+        connection_string = "host=localhost dbname=idr user=bfd password=InsecureLocalDev"
+    elif mode == "synthetic":
+        connection_string = get_connection_string()
+    else:
+        connection_string = get_connection_string()
+
+    dr.execute(
+        final_vars=["idr_beneficiary"],
```
mel1-G, line 35: You specify the last node "table" and the DAG will get executed in topological order based off the dependencies so it follow stages 1 through 5

apps/bfd-pipeline/bfd-pipeline-idr/pipeline_nodes.py
```diff
@@ -0,0 +1,128 @@
...
+        cls=cls,
+        connection_string=config_connection_string,
+        mode=config_mode,
+        batch_size=config_batch_size,
+    )
+
+
+# Stage 2: Load main claim table
+def idr_claim(
+    idr_claim_institutional: tuple[PostgresLoader, bool],
```
mel1-G, line 52: this is how you specify "load claim table AFTER loading the auxiliary claims tables"

apps/bfd-pipeline/bfd-pipeline-idr/hamilton_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+import ray
+from hamilton import base, driver
+from hamilton.plugins.h_ray import RayGraphAdapter
+
+import pipeline_nodes
+from hamilton_loader import get_connection_string
+from pipeline_utils import configure_logger
+
+
+def main() -> None:
```
aschey-forpeople, line 14: It's fine to create a copy of everything for testing, but if we're gonna merge this into master then we should merge this back into the main `pipeline.py` logic.

mel1-G: @aschey-forpeople I wasn't sure if we'd immediately switch to the parallelized workflow or wait until we verify with a real full pipeline run since we didn't let the the pipeline complete during our testing. I run a little paranoid so I like the idea of an initial backup. What do you think?

aschey-forpeople: We can run this against the main IDR database with just the bene data enabled and make sure it completes an incremental data load successfully. If that works then I'm good with this. It's fine if we find bugs later. I'm already pretty sure that this is gonna require some tweaks before we can load all the claims data successfully, even without the parallelism.

apps/bfd-pipeline/bfd-pipeline-idr/extractor.py
```diff
@@ -19,13 +19,7 @@
...
 transform_timer = Timer("transform")
 
-logger = logging.getLogger(__name__)
-
-
-def print_timers() -> None:
-    cursor_execute_timer.print_results()
-    cursor_fetch_timer.print_results()
-    transform_timer.print_results()
+logger = logging.getLogger("pipeline_worker")
```
aschey-forpeople, line 22: This is extremely minor, but it's conventional to use the `logging.getLogger(__name__)` setup so you know which file it's coming from.

aschey-forpeople (approved):
I'm a bit concerned that the large list of claim tables might be a bit unweildy to maintain with all the dependency ordering stuff, but I think we can save further improvements for later. Looks good!

np-complete0 (approved):

PR #2829, opened 2025-10-01
BFD-4308: Filter PAC data > 60 days old

brick-green:
**JIRA Ticket:**
BFD-4308

### What Does This PR Do?

Updates the pipeline to filter out PAC data that is older than 60 days based on the insert, update, or load dates in the clm table. The primary benefit of this filtering is for historical data loading.

### What Should Reviewers Watch For?

If you're reviewing this PR, please check for these things in particular:
- [ ] Review the updates to the test files
- [ ] Review implementation in model.py

### What Security Implications Does This PR Have?

Please indicate if this PR does any of the following:  

* Adds any new software dependencies
* Modifies any security controls
* Adds new transmission or storage of data
* Any other changes that could possibly affect security? 

* [X] I have considered the above security implications as it relates to this PR. (If one or more of the above apply, it cannot be merged without the ISSO or team security engineer's (`@sb-benohe`) approval.) 
* [X] I have created tests to sufficiently ensure the reliability of my code, if applicable. If this is a modification to an existing piece of code, I have audited the associated tests to ensure everything works as expected.

### Validation

**Have you fully verified and tested these changes? Is the acceptance criteria met? Please provide reproducible testing instructions, code snippets, or screenshots as applicable.**
    updated test for python

PR #2835, opened 2025-10-06
BFD-4334: Exclude part D claim codes & outdated claims based on temp flag"

mel1-G:
**JIRA Ticket:**
BFD-4334

### What Does This PR Do?

- I removed part D claim type codes from CLAIM_TYPE_CODES in constants.py
- I added back the temp flag "IDR_LATEST_CLAIMS" since on initial load we do only want to load in claims data where clm_ltst_clm_ind = 'N'. Later on, we'll have to make some adjustments for incremental loads since claims that were initially set as 'N' but later are updated to 'Y' will be missed

* Adds any new software dependencies - No
* Modifies any security controls - No
* Adds new transmission or storage of data - No
* Any other changes that could possibly affect security? - No

### Validation

I ran the pipeline locally and checked my local db to see if claims with claim type codes 1, 2, 3, 4 were no longer included in our claim table. I wasn't able to verify clm_ltst_clm_ind since are synthetic data always generates claims with clm_ltst_clm_ind set as 'Y'

Before: 

Screenshot 2025-10-06 at 1 45 51 PM

After:

Screenshot 2025-10-06 at 2 05 42 PM


apps/bfd-pipeline/bfd-pipeline-idr/constants.py
```diff
@@ -2,10 +2,6 @@
 DEFAULT_MIN_DATE = "0001-01-01"
 
 CLAIM_TYPE_CODES = [
-    1,
-    2,
-    3,
-    4,
```
aschey-forpeople, line 8: We need to keep these here (the list of allowed claim types), otherwise we won't get any part D claims at all. We'll just need to make an exception for these in the check for `clm_ltst_clm_ind`.

aschey-forpeople:
You can change the target branch to #2829 so the diff is clearer here.

apps/bfd-pipeline/bfd-pipeline-idr/model.py
```diff
@@ -616,10 +622,26 @@ def _current_fetch_query(start_time: datetime) -> str:  # noqa: ARG004
 
 
 def claim_type_clause(start_time: datetime) -> str:
+    fetch_latest_claims = "IDR_LATEST_CLAIMS" in os.environ and os.environ["IDR_LATEST_CLAIMS"] in (
+        "1",
+        "true",
+    )
+    add_latest_claim_ind = (
+        f""" AND {ALIAS_CLM}.clm_ltst_clm_ind = 'Y' """ if fetch_latest_claims else ""
```
aschey-forpeople, line 630: Can use single quotes here since it's not a multi-line string

```suggestion
        f" AND {ALIAS_CLM}.clm_ltst_clm_ind = 'Y' " if fetch_latest_claims else ""
```

apps/bfd-pipeline/bfd-pipeline-idr/model.py
```diff
@@ -616,10 +622,26 @@ def _current_fetch_query(start_time: datetime) -> str:  # noqa: ARG004
...
     start_time_sql = start_time.strftime("'%Y-%m-%d %H:%M:%S'")
     return f"""
     (
-        {ALIAS_CLM}.clm_type_cd IN ({",".join([str(c) for c in CLAIM_TYPE_CODES])})
+        {ALIAS_CLM}.clm_type_cd IN ({
+        ",".join(
+            [
+                str(c)
+                for c in CLAIM_TYPE_CODES
+                if not fetch_latest_claims or c not in EXCLUDED_CLAIM_TYPE_CODES
```
aschey-forpeople, line 640: I think the check for claim type code 1-4 needs to go with the `clm_ltst_clm_ind = 'Y'` check. It should pull every claim for part D, but only latest claims for the other claim types. The way it's written here, it will filter out all part D claims if `fetch_latest_claims = True`.

PR #2984, opened 2026-01-30
BFD-4500: Switch to using a fixed test date for integration and pipeline tests

valor4valkyrie:
**JIRA Ticket:**
BFD-4500

### What Does This PR Do?

- overall used a test controlled clock/date across both the server and pipeline to make time dependent tests deterministic 
- added a clock bean and modified DateUtil and its usage so it depends on it instead of calling system time directly so we can use a fixed date for our tests
- updated the pipeline to respect the test date (`BFD_TEST_DATE`) when set. It can override start_time and completion_time (if incremental load) during tests
- replaced usages of `CURRENT_DATE` in our models (`base_model` and `idr_contract_pbp_contact`) with the test date if set

Places where we are using *now* for any timestamps with data, especially queries and tests.

### Validation

- Updated our integ test configuration and test_pipeline to use a fixed date


apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -434,7 +434,8 @@ private static Stream<Arguments> ensureDrg() {
   @MethodSource
   @ParameterizedTest
   void ensureDrg(long claimId, String code, String system) {
-    var eob = eobHandler.find(claimId, SamhsaFilterMode.INCLUDE).get();
+    var eob =
```
aschey-forpeople: This probably shouldn't change. The intent is that the test data here should be returned.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/IntegrationTestConfiguration.java
```diff
@@ -63,11 +72,17 @@ public PostgreSQLContainer<?> postgres() throws IOException, InterruptedExceptio
         "testdb",
         "-c",
         "UPDATE cms_vdm_view_mdcr_prd.v2_mdcr_clm "
-            + "SET \"clm_idr_ld_dt\" = CURRENT_DATE,"
-            + "\"idr_insrt_ts\" = CURRENT_TIMESTAMP,"
-            + "\"idr_updt_ts\" = CURRENT_TIMESTAMP;");
+            + "SET \"clm_idr_ld_dt\" = '"
```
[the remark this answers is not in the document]
valor4valkyrie: Is there any date I should shoot for?

apps/bfd-pipeline-idr/pipeline.py
```diff
@@ -64,7 +64,10 @@ def run(load_mode: str) -> None:
         .build()
     )
 
-    start_time = datetime.now(UTC)
+    if not BFD_TEST_DATE:
```
aschey-forpeople: nit: I think we can invert this logic. I try to avoid inverting checks when possible because it's slightly more cognitive overhead.

```
    if BFD_TEST_DATE:
     ...
    else:
    ...
```

apps/bfd-pipeline-idr/pipeline.py
```diff
@@ -104,17 +110,25 @@ def run(load_mode: str) -> None:
...
         raise
 
     if idr_job_events:
         update_completion_times(
-            load_mode=load_mode, events=idr_job_events, completion_time=datetime.now(UTC)
+            load_mode=load_mode, events=idr_job_events, completion_time=resolve_time(load_mode)
         )
 
 
+def resolve_time(load_mode: str) -> datetime:
```
aschey-forpeople, line 124: Might make the intent a little clearer?
```suggestion
def resolve_test_date(load_mode: str) -> datetime:
```

PR #3067, opened 2026-03-24
BFD-4248: Send Release and Deploy BFD Workflow run status updates to Slack

jonathan-cantu-icf:
**JIRA Ticket:**
BFD-4248

### What Does This PR Do?
Enables and implements Slack Webhooks for BFD GitHub build and deploy workflow.

Watch the GitHub Actions for this PR during quality checks.

### Validation

The validation will be in seeing what comes out of GitHub quality checks for this PR once the changes are pushed.


mjburling:
> [..]
> `notify-startNode.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: cmsgov/cdap/actions/aws-params-env-action@main. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/ `
> 
> So we may want to address that before fully deploying and implementing this PR.

@mianava worked through these details last week: https://github.com/CMSgov/cdap/pull/444

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -77,3 +107,78 @@ jobs:
...
+            echo "status=SUCCESS" >> $GITHUB_OUTPUT
+            echo "failed_stage=none" >> $GITHUB_OUTPUT
+          fi
+
+      # TODO: Reevaluate after trying this out a couple of times whether we need to prune it or alter it.
+      # -------------------------
+      # Send Slack
+      # -------------------------
+      - name: Slack - FINAL SUMMARY
+        uses: rtCamp/action-slack-notify@v2
```
mjburling, line 163: @sb-benohe This is new, here and above. Adding you as a reviewer.

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -24,17 +24,44 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Load Slack webhook
+        uses: cmsgov/cdap/actions/aws-params-env-action@main
+        with:
+          params: |
+            SLACK_WEBHOOK_URL=/bfd/platform/alerting/sensitive/slack/bfd-notices/webhook
+
+      - name: Slack - STARTED
+        uses: rtCamp/action-slack-notify@v2
```
malessi: Since this is third-party, we should pin this to a SHA and add a semver comment that Dependabot understands.

.github/workflows/release-and-deploy-bfd.yml
sb-benohe: I have two thoughts here:

1. Could we use the official integration with GH for Slack? > https://github.com/integrations/slack 
1. if not can we follow @malessi recommendation of pinning to SHA versions (especially with lesser known actions developers)

sb-benohe (dismissed):

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -77,3 +107,78 @@ jobs:
...
+          elif [[ "${{ needs.deploy-to-sandbox.result }}" != "success" ]]; then
+            echo "status=FAILED" >> $GITHUB_OUTPUT
+            echo "failed_stage=deploy-to-sandbox" >> $GITHUB_OUTPUT
+          elif [[ "${{ needs.deploy-to-prod.result }}" != "success" ]]; then
+            echo "status=FAILED" >> $GITHUB_OUTPUT
+            echo "failed_stage=deploy-to-prod" >> $GITHUB_OUTPUT
+          else
+            echo "status=SUCCESS" >> $GITHUB_OUTPUT
+            echo "failed_stage=none" >> $GITHUB_OUTPUT
+          fi
```
malessi: nit: wonder if there's a good way to reduce the duplication here?

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -24,17 +24,44 @@ on:
...
   contents: write # This is required for actions/checkout
-
 defaults:
   run:
     shell: bash
 
+# -------------------------
+# START NOTIFICATION
+# -------------------------
 jobs:
```
malessi:
```suggestion
jobs:
```

*super* nit: header comments could probably be omitted given that the `job` names are fairly descriptive as-is

malessi (changes requested):
Just a few things!

.github/workflows/release-and-deploy-bfd.yml
malessi: This _nit_ applies to all of the header comments sprinkled throughout, by the way.

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: Thanks for your feedback @sb-benohe .  I will look into this today.

jonathan-cantu-icf: @sb-benohe  I did in fact switch to the GitHub slack action and also applied the full SHA. Thanks for the suggestion.

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: Agreed. Not sure why I didn't think it looked a bit silly back then but I do not. Thanks Mitch.

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: Done

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: Thanks Mitch. Yeah, agreed, there's a lot of repetition here.  I came up with an alternate pattern but I don't think it's as obvious what's going on, but I'll  push it and you can let me know what you think.

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -34,7 +38,57 @@ defaults:
...
+    steps:
+      - name: Setup node
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: 24.x
+
+      - name: Get role ARN
+        id: get-role-arn
+        run: |
+          role_arn="$(jq -r --arg account_type "${{ matrix.account }}" '.[$account_type]' <<<"$ACCOUNT_ROLE_MAP")"
```
malessi:
```suggestion
          role_arn="${{ secrets.NON_PROD_ACCOUNT_GHA_ROLE_ARN }}"
```

`ACCOUNT_ROLE_MAP` is not defined as an `env` in this Workflow, so this will result in an empty `role_arn`. _Fortunately_, `platform` SSM values exist in both BFD AWS Accounts, so it is only necessary to choose one of the two accounts to assume the GitHub Role in. In this case, I recommend assuming the GH Role in `non-prod`.

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -34,7 +38,57 @@ defaults:
...
+
+          echo "::add-mask::$role_arn"
+          echo "role-arn=$role_arn" >> "$GITHUB_OUTPUT"
+
+      - name: Assume role to AWS
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
+        env:
+          AWS_REGION: ${{ github.event.inputs.aws-region || 'us-east-1' }}
```
malessi:
```suggestion
          AWS_REGION: ${{ inputs.aws-region }}
```

The `github.events` prefix is unnecessary. We can just refer to `inputs` under the `inputs` context. There is also no need for the `false`y "coalesce" here since `aws-region` has a default value and is a `choice` so it can only ever be `us-east-1` _or_ `us-west-2`.

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -34,7 +38,57 @@ defaults:
...
+        with:
+          role-to-assume: ${{ steps.get-role-arn.outputs.role-arn }}
+          role-session-name: get-sops-slack-webhook-${{ github.run_id }}-${{ github.run_attempt }}
+          aws-region: ${{ env.AWS_REGION }}
```
malessi:
```suggestion
          aws-region: ${{ inputs.aws-region }}
```

There is no `AWS_REGION` `env` variable, but there _is_ an `input`. So, we can just refer to the region using `inputs.aws-region` here.

malessi (changes requested):
Just a few changes that should resolve the issues assuming the AWS Role to retrieve the SSM value.

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -17,24 +17,78 @@ on:
...
+          Default to `us-east-1`.
         default: us-east-1
         type: choice
         options:
           - us-east-1
           - us-west-2
         required: true
+  push:
+    branches:
+      - feature/BFD-4248_GitHub-BuildAndDeploy-Slack-Notification
```
mjburling: I assume you need to drop this ahead of merging, no?

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -17,24 +17,78 @@ on:
...
 
 permissions:
   id-token: write # This is required for requesting the AWS IAM OIDC JWT
   contents: write # This is required for actions/checkout
 
+env:
+  AWS_REGION: us-east-1
```
mjburling: Unclear if this is conflicting advice, but should this be `${{ inputs.aws-region }}` and... is this envvar meaningful?

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: Absolutely sir. Nice catch Burling!  Clearly, I missed that entirely.  I had just picked this back up after putting it on the back burner through the CAN effort (waiting on reviews and discussion around it) so kind of forgot about that.  Much appreciated.

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: Actually, no, I added that intentionally after experimenting with the AWS assumed role/region and the CDAP library which decrypts and pulls the SOPS value for our Webhook.  I couldn't find a way to explicitly pass that in as an input and as far as I could tell it would only work as an env var.  The env var was the only way I could get the AWS assumed role/credentials to work for that CDAP SOPS dependency.

jonathan-cantu-icf: Now I'm wondering if I could just replace the input in lieu of the env var throughout this code.  I do agree it's not ideal having two different sources for AWS_REGION.

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -29,12 +30,62 @@ permissions:
...
+
 defaults:
   run:
     shell: bash
 
 jobs:
+  notify-start:
+    runs-on: ubuntu-latest
```
mjburling: I assume @sb-benohe would raise exception with this (and elsewhere) and I'm also unsure of which platform makes the most sense. Going out on a limb and _guessing_ this should work fine:

```suggestion
    runs-on: ubuntu-24.04-arm
```

jonathan-cantu-icf: So just being a bit more deliberate and precise like we do with the SHAs now on GitHub actions etc?  OK, yes, by all means let's run that by Ben.  Thanks again Burling - much appreciated.

jonathan-cantu-icf: So it looks like we are already doing this in a couple of places in the BFD .github area of the repo, so I'll just make the change and we should be good. Thanks.

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -77,3 +128,83 @@ jobs:
...
+  workflow-summary:
+    name: Final Slack Summary
+    runs-on: ubuntu-latest
+    needs:
+      - build-release
+      - deploy-platform
+      - deploy-to-test
+      - deploy-to-sandbox
+      - deploy-to-prod
+    if: always()
```
mjburling: Unsure if you've already done your homework on this, but `always()` is somewhat controversial. Practitioners and GitHub recommend using `!cancelled()` for running on success **or** failure:
- https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always 
- https://github.com/orgs/community/discussions/26303

...still, `always()` _could_ be perfectly reasonable here ¯\\\_(ツ)_/¯

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -77,3 +128,83 @@ jobs:
       bfd-env: prod
       git-ref: refs/tags/${{ needs.build-release.outputs.bfd_release }}
     secrets: inherit
+
+  workflow-summary:
+    name: Final Slack Summary
+    runs-on: ubuntu-latest
```
mjburling:
```suggestion
    runs-on: ubuntu-24.04-arm
```

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: OK, I read those threads and I agree, let's go with !cancelled().

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: I checked the README in CDAPs repo for the action `aws-params-env-action` and there's no mention in there of how to explicitly pass the was-region so apparently you can only do it via an env var.  So a compromise here was to set an env var from the input since other areas of this code were already previously using that.

.github/workflows/release-and-deploy-bfd.yml
malessi: I've unresolved this conversation as this still needs a change as of the latest commit on this PR ([4cc6d8b](https://github.com/CMSgov/beneficiary-fhir-data/pull/3067/commits/4cc6d8b5d2050edcac60526cd0de66915f3d25e5)).

.github/workflows/release-and-deploy-bfd.yml
```diff
@@ -77,3 +128,83 @@ jobs:
...
+      - deploy-platform
+      - deploy-to-test
+      - deploy-to-sandbox
+      - deploy-to-prod
+    if: !cancelled()
+    steps:
+      - name: Get role ARN
+        id: get-role-arn
+        run: |
+          role_arn="${{ secrets.NON_PROD_ACCOUNT_GHA_ROLE_ARN }}"
```
mjburling, line 150: I'm only half-way paying attention to most things, as usual. This must **always** be non-prod here (and elsewhere)?

malessi: I [recommended this](https://github.com/CMSgov/beneficiary-fhir-data/pull/3067#discussion_r3155254922) as part of a prior review as common `platform` SSM parameters (such as Webhook URLs) exist in both accounts. So, _any_ BFD Account could be used here, but I suggested `non-prod` since it's, well, not `prod` 🙂

mjburling: Ah, yes. Half-way paying attention, indeed. In that case, this subtlety might be a good candidate for a brief comment.

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: Yep - missed the 2nd one. Got it.

.github/workflows/release-and-deploy-bfd.yml
jonathan-cantu-icf: OK. Done.

malessi (approved):

aschey-forpeople (approved):

sb-benohe (approved):

sb-benohe:
LGTM: thanks for masking the ARNs + the Pinning of the actions!

alex-dzeda (approved):

PR #3085, opened 2026-04-06
BFD-4348: Filter non-latest Claims from Server

mark-lewis-icf:
**JIRA Ticket:**
BFD-4348

### What Does This PR Do?

- ~~Adds a filter to the server to only return latest claims or Part D~~ - This work was concurrently done by @aschey-forpeople in a separate ticket)
- Adds a final step to the pipeline to purge any claim that is not the latest nor a Part D

- Verify hamilton configuration looks correct
- Why do we even need the server filter if we are already purging them in the pipeline?

### Validation

Run pipeline in bfd-pipeline-idr folder from master branch
`./run-db.sh test_samples2`

Observe a non-latest claim in target table, e.g. claim unique id 335126648140 in claim_professional_nch

Checkout BFD-4348 and rerun pipeline and observe the above the claim is no longer in target table


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/ClaimAsyncService.java
```diff
@@ -27,21 +28,22 @@ CompletableFuture<Optional<C>> findByIdInClaimType(
...
-    var filters = getFilters(filterBuilders, systemType);
-    var whereClause = buildWhereClause(filters, systemType);
+    var filters = getFilters(paramBuilders, systemType);
     var jpql =
         String.format(
             """
             %s
             WHERE c.claimUniqueId = :claimUniqueId
+            AND b.latestTransactionFlag = 'Y'
+            AND ( c.latestClaim = 'Y' OR c.claimTypeCode IN ( 1, 2, 3, 4 ) )
```
aschey-forpeople, line 40: I already made the changes on the server for this so we can remove all the java changes here.

mark-lewis-icf: Removed. I also removed the enum as I saw you just used boolean for Y / N latest claim field

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -100,3 +101,69 @@ def extract_and_load(
...
+    logger.info("Parent : %s", parent_table_name)
+    parentSQL = f"""
+        DELETE FROM {parent_table_name} AS p
+        WHERE p.clm_ltst_clm_ind  = 'N' 
+            AND p.clm_type_cd NOT IN ( 1, 2, 3, 4 )
+            AND {claim_range_filter}
+            AND {claim_type_code_filter}
+    """
+    try:
+        logger.info("Delete Parent SQL:")
```
aschey-forpeople: I don't think we want to do this try/catch here. The delete shouldn't fail and if it does, the failure should propagate.

apps/bfd-pipeline-idr/pipeline_nodes.py
```diff
@@ -240,3 +248,33 @@ def collect_stage4(
...
+        filter_tables([*_CLAIM_TABLES, IdrClaimRx], tables_to_load), load_type
+    )
+
+
+def do_stage5(
+    stage5_inputs: NodePartitionedModelInput,
+    load_mode: LoadMode,
+) -> bool:
+    model_type, partition = stage5_inputs
+    return purge_non_latest_claims(
```
aschey-forpeople, line 266: A test to verify this would be welcome

apps/bfd-pipeline-idr/pipeline_utils.py
mark-lewis-icf: removed. also removed debug log statements

apps/bfd-pipeline-idr/pipeline_nodes.py
mark-lewis-icf: Added relevant assertions to test_pipeline.py and updated relevant server-ng tests to expect non-latest to not exist in DB.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -100,3 +101,59 @@ def extract_and_load(
...
+        childSQL = f"""
+            DELETE FROM {child_table_name} AS c
+            WHERE EXISTS
+                (   SELECT NULL
+                    FROM {parent_table_name} AS p 
+                    WHERE p.clm_uniq_id = c.clm_uniq_id
+                        AND p.clm_ltst_clm_ind  = 'N' 
+                        AND p.clm_type_cd NOT IN ( 1, 2, 3, 4 )
+                        AND {claim_range_filter}
+                        AND {claim_type_code_filter}
```
aschey-forpeople, line 143: We don't need to filter on claim type code here because the claim tables are already separated out - `claim_rx` only has claim type code 1-4, `claim_professional_nch` only has 71,72,81,82, etc.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -100,3 +101,59 @@ def extract_and_load(
...
+    if child_table:
+        child_table_name = child_table.table()
+        logger.info("Child: %s", child_table_name)
+        childSQL = f"""
+            DELETE FROM {child_table_name} AS c
+            WHERE EXISTS
+                (   SELECT NULL
+                    FROM {parent_table_name} AS p 
+                    WHERE p.clm_uniq_id = c.clm_uniq_id
+                        AND p.clm_ltst_clm_ind  = 'N'
```
aschey-forpeople, line 141: I think we may want to create a [partial index](https://www.postgresql.org/docs/current/indexes-partial.html) on the claims where `clm_ltst_clm_ind  = 'N'` since that will be the vast minority of claims and should take up the least amount of space. Hopefully that should be the only index we need to make this work.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobReadIT.java
```diff
@@ -123,8 +123,8 @@ void eobReadNonLatestProfessionalIsNotReturned() {
                 ClaimProfessionalNch.class)
             .setParameter("claimId", Long.parseLong(CLAIM_ID_PROFESSIONAL_NON_LATEST))
             .getResultList();
-    // Precondition - claim should be available in the db
-    assertFalse(claims.isEmpty());
+    // Precondition - claim should be purged and not available
+    assertTrue(claims.isEmpty());
```
aschey-forpeople: I think we still want to preserve the behavior of this test. Since claims are purged async, we still want to verify the API is robust enough to not return non-latest claims since someone could request data before the purge is complete. We could pick a claim and manually update its `ltst_clm_ind` at the beginning of this test perhaps.

apps/bfd-db-migrator-ng/migrations/V1_31_0__claim_type_code_index.sql
```diff
@@ -0,0 +1,9 @@
+CREATE INDEX ON idr.claim_institutional_nch (clm_type_cd);
```
alex-dzeda: This may get to Austin's comment below about the query - do we need these indices?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobReadIT.java
mark-lewis-icf: Done. I had to manually control the transaction with Spring's TestTransaction, otherwise the API Async call wouldn't see the non-committed changes.

apps/bfd-pipeline-idr/pipeline_utils.py
mark-lewis-icf: Index created.

apps/bfd-db-migrator-ng/migrations/V1_31_0__claim_type_code_index.sql
mark-lewis-icf: Yes. Removed those indices and replaced with partial index he suggested.

apps/bfd-pipeline-idr/pipeline_utils.py
mark-lewis-icf: removed explicit filter on NOT 1 - 4, but kept other as that is part of partitioning. Do you want that removed as well? If so i think I would need to modify the partition generation to remove partitioning by claim type codes

aschey-forpeople:
Closing as stale

PR #3095, opened 2026-04-08
BFD-4660: v3 Deployment Readiness

malessi:
**JIRA Ticket:**
BFD-4660

### What Does This PR Do?

> [!IMPORTANT]
> The changes in this PR have [already been `apply`'d to all environments for v3](https://github.com/CMSgov/beneficiary-fhir-data/actions/runs/24683958467)

This PR makes a variety of changes that enable automated deployment and release of v3-related services and resources. Specifically:

- A new Workflow, `(TEMPORARY) v3 Release and Deployment`, has been introduced to "release" and deploy v3-related services and resources. This new Workflow requires that the _base reference_  from which its ran is a Git _Tag_ in the form of `v3.yyyy.MM.dd.<integer>`, where `<integer>` is a monotonically increasing number disambiguating releases on a given date. This format represents a particular "version" of a v3 release and will be the format in which those releases follow until v3 is ready to deploy with the existing process.
  - This Workflow deploys:
    - `config`
    - `cluster`
    - `database`
    - `migrator-ng`
    - `idr-pipeline`
    - `server-ng`
  - The Workflow builds the `server-ng` Image and the following other Images, tagging them with the Git Tag name and `version-suffix`:
    - `bfd-platform-server-fluent-bit`
    - `bfd-platform-idr-pipeline`
    - `bfd-platform-migrator-ng`
    - `bfd-platform-consume-idr-events`
    - `bfd-platform-run-idr-pipeline`
  - The Workflow can be ran in `BUILD_ONLY` or `DEPLOY_ONLY` mode to build/deploy, respectively
- Significant duplication has been reduced with respect to how "Tofuservices" (formally "Terraservices", but we use OpenTofu now) are deployed in GitHub Actions. Now, both `Deploy Environment Services` and `Deploy Platform Services` delegate to a single `Deploy Tofuservices` Workflow which delegates to further child-Workflows and Composite Actions to deploy/`apply` a set of Tofuservices. There is no functional change, but this should make making changes to our GitHub Actions Workflows simpler
  - This also makes the directory structure and naming of Tofuservices truly authoritative and important in the context of deployment via GHA. Tofuservices are "discovered" from either of `ops/services` or `ops/platform` and then filtered depending on the list of Tofuservices to deploy by-default per-Workflow
- `bfd-server-ng` Jib Image will now be built and pushed to ECR during the Release Pipeline
- The v3 Migrator is now built into an Image, `bfd-platform-migrator-ng`, that when ran will run the migrations built into it
- The aforementioned `bfd-platform-migrator-ng` Image will now be built and pushed to ECR during the Release Pipeline
- A new Tofuservice, `03-migrator-ng`, has been added to run the v3 Migrator Image as a Task during deployments
  - This "new" Tofuservice is 95% a copy/paste of `03-migrator` with some changes to the Task Definition to use the new Image, change the `environment` variables to configure the Image appropriately, inject the `FLYWAY_USER` and `FLYWAY_PASSWORD` env vars as `secrets` from SSM, and to ensure the Task is tagged appropriately. Everything else is largely the same
- Fluent Bit logging configuration has been updated for both the v1/v2 and v3 Servers to better support high-throughput logging scenarios such that performance is not impacted and logs are not lost
- Fluent Bit configuration for the v3 Server has been added such that irrelevant, noisy logs such as healthchecks and non-JSON logs are excluded from the main `messages` Log Group and pushed to other Log Groups instead
- Usage of `tofu` has been made more resilient to [ungraceful termination in GitHub Actions](https://github.com/orgs/community/discussions/26311)
- Duplication has been reduced in our GitHub Actions deployment Workflows
- Operators can now specify a JSON mapping of Tofu variables for Tofuservices deployed via the `Deploy Environment Services` Workflow

- See above

- Adds any new software dependencies
  - The new v3 Migrator Image is based on the [official Flyway Image](https://hub.docker.com/r/flyway/flyway)

- [x] I have created tests to sufficiently ensure the reliability of my code, if applicable. If this is a modification to an existing piece of code, I have audited the associated tests to ensure everything works as expected.

### Validation

- Deploying both `server` and `server-ng` to `4660-test` with the Fluent Bit changes and running the Locust tests against both, _verifying that_ their new logging configurations work as expected and do not degrade performance
- Building the new `bfd-platform-migrator-ng` Image locally, _verifying that_ it builds successfully and correctly
- Running the new `bfd-platform-migrator-ng` Image locally against a local database, _verifying that_ it migrates the database successfully
- Running the `migrate.sh` migration script _locally_ against a local database, _verifying that_ it works
- _Simulating_ running the `migrate.sh` migration script against a live environment, _verifying that_ it _would_ work as expected
- Running the v3 Release/Deploy Workflow, _verifying that_ [it succeeds](https://github.com/CMSgov/beneficiary-fhir-data/actions/runs/24423373639)


ops/services/04-server-ng/service.tf
```diff
@@ -108,16 +115,21 @@ resource "aws_ecs_task_definition" "server" {
...
         name              = "log_router"
         image             = data.aws_ecr_image.log_router.image_uri
         essential         = true
-        cpu               = 128
-        memoryReservation = 50
-        memory            = 100
-        user              = "0" # Default; reduces unnecessary terraform diff output
+        cpu               = max(min(1024, floor(0.05 * local.server_cpu)), 256)    # Max 1 CPU, min 1/4
+        memoryReservation = max(min(1024, floor(0.08 * local.server_memory)), 256) # Max 1 GB, min 256 MiB
+        memory            = max(min(1024, floor(0.10 * local.server_memory)), 312) # Max 1 GB, min 312 MiB
```
mjburling, line 131: Scaling this with guardrails against the overall allotted resources is interesting. This at least _hints_ at required minimums for the sdlc environment's definition for the server's cpu and memory allotment, doesn't it? 🤔

ops/services/README.md
```diff
@@ -12,13 +12,13 @@ Terraservices are opinionated, top-level, _flat_ modules that aim to describe th
 
 <!-- TODO: Expand this further -->
 
-- Terraservices are opionated, meaning that "hardcoding" invariant attribute values and resource definitions is _encouraged_. If a value changes depending on environment or cannot be computed, only then should it be elevated into our SSM configuration
+- Terraservices are opinionated, meaning that "hardcoding" invariant attribute values and resource definitions is _encouraged_. If a value changes depending on environment or cannot be computed, only then should it be elevated into our SSM configuration
 - Terraservices, excluding those that are `apply`'d prior to the Terraservice describing our SSM configuration (i.e. `base` or `config`), should expose as few variables as possible. Variables that _are_ exposed should be constrained to overrides only, and should be introduced sparingly
 - It should be possible for any operator, in the correct Terraform Workspace, to simply `terraform apply` the Terraservice and generate appropriate resources
```
mjburling: Strictly speaking, this isn't part of your change set, but the distinctions between OpenTofu/tofu and terraform readily confuses, what with the language being HCL2, the OpenTofu binary being `tofu`, the conventions remaining _terraform_, and both internal/external resources continuing to use _terraform_... even the _terraservice_ pattern is hard to separate from _terraform_, right? Some of this is subtle.

I'm not convinced of the wording, but... we can try to reduce potential confusion.

```suggestion
- It should be possible for any operator, in the correct OpenTofu Workspace, to simply `tofu apply` the Terraservice and generate appropriate resources
```

ops/services/04-server-ng/service.tf
mjburling, line 1: Looks like`04-server-ng` could use another execution of `terraform-docs` to capture the latest updates.

ops/services/04-idr-pipeline/main.tf
mjburling, line 1: `terraform-docs` for resource updates, etc.

While I think it's _OK_ to keep the minimum transaction dates hard-coded here (it may even be _good_. Then again, these things are so fundamental to a product like this... despite the low likelihood of them changing, maybe hoisting them out of the service would be something to consider), we really ought to make those values obvious and with specific call outs in the README, too.

mjburling (commented):
Ran out of time for this first pass–you'll find some in-line commentary, too, but this is lovely:
- wonderful that we're actually doing more than just explaining the service ordering with the prefixes, and actually using them to inform the deployment order 👍
- the differences between env-server and platform-service deployments are now _very_ minor 🤔  
- curious if you couldn't add a little more discussion/comments on the `all-vars-json` in the actions
- overall, terraform-docs need some attention

I'll give it another pass when you're ready!

ops/services/04-server-ng/service.tf
malessi: Indeed, it does. I added this as I felt that a static value wasn't necessarily correct, but I couldn't articulate _why_. Now that you've pointed this out, I believe having a _minimum_ resource allocation requirement for each container in a Task and the Task overall would make this feel more right.

Thinking about this, there's two ways to go about it:

- We "fail" silently: the Terraservice simply sets the minimum to some value without any warning if an operator tries to set a resource allocation too low
- We fail fast and loud: the Terraservice uses `precondition` checks to ensure that resource minimums are enforced and simply fails to `apply` if they're specified too low

I feel like I prefer failing fast and loud, but also... who cares? I'll think about it.

ops/services/README.md
malessi: You're right, our documentation isn't where I'd like it to be. Admittedly, I zeroed-in on the spelling mistakes here after switching to a new editor and felt that this was the place to clean them up. I didn't take a look at the rest of the document to see if there was any other improvements to be made.

Once v3 settles down we should take another pass at our repository documentation and really put some effort into to it. We have a lot of undocumented "convention" in our Tofu that _should_ be documented.

ops/services/04-server-ng/service.tf
malessi: I've done this in [866074c](https://github.com/CMSgov/beneficiary-fhir-data/pull/3095/commits/866074ca8b4b681ac3b56f9dd8cc0769fe2dc08c).

ops/services/04-idr-pipeline/main.tf
malessi: I've made the `terraform-docs` updates in [866074c](https://github.com/CMSgov/beneficiary-fhir-data/pull/3095/commits/866074ca8b4b681ac3b56f9dd8cc0769fe2dc08c).

I'll add specific `[!IMPORTANT]` callouts to each of the values in the README in a future commit.

.github/workflows/v3-release-deploy.yml
```diff
@@ -0,0 +1,416 @@
...
+          "idr-pipeline": {
+            "consume_idr_events_version_override": "${{ needs.setup.outputs.full-version-tag }}",
+            "pipeline_version_override": "${{ needs.setup.outputs.full-version-tag }}",
+            "run_idr_pipeline_version_override": "${{ needs.setup.outputs.full-version-tag }}"
+          },
+          "server-ng": {
+            "log_router_version_override": "${{ needs.setup.outputs.full-version-tag }}",
+            "server_version_override": "${{ needs.setup.outputs.full-version-tag }}"
+          }
+        }
```
malessi, line 404: @mjburling I've renamed `all-vars-json` in all Workflows/Actions to `per-service-vars-json` to better reflect its usage. Essentially, it's a mapping of Tofuservices to a mapping of its `variable`s to their values, e.g.:

```json
{
  "server": {
    "some_var": "some_val"
    "another_var": "another_val"
  },
  "idr-pipeline": {
    "one_more_var": "one_more_val"
  }
}
```

Where, for deploying `04-server`, the `tofu plan` command becomes:

```sh
tofu plan -var=some_var=some_val -var=another_var=another_val ...
```

And so on.

The logic for converting the per-Tofuservice mapping to command-line arguments already existed, but was never used. This PR adds the "plumbing" from the top-level Workflows down to the Composite Action that actually `plan`s/`apply`s to _allow_ for operators to specify variables if necessary. **This _is_ necessary for the v3 Release and Deploy Workflow because we must override the default Container Image Tag lookups to use the v3 style Tags**. Otherwise, the Tofuservices would try to use v2-style Tags and fail to find the appropriate Image.

You can see an example of the usage here (`.github/workflows/v3-release-deploy.yml:390-404`).

aschey-forpeople (dismissed):

ops/services/04-server-ng/service.tf
malessi: We'll leave this for another time.

ops/services/04-idr-pipeline/main.tf
malessi: I've added some callouts for those environment variables in [26ec7b8](https://github.com/CMSgov/beneficiary-fhir-data/pull/3095/commits/26ec7b8408959dbf590722a3f50ba28773e77b55).

aschey-forpeople (approved):

mjburling (approved):

PR #3098, opened 2026-04-09
BFD-4662: Claim query perf improvements

aschey-forpeople:
**JIRA Ticket:**
BFD-4662

### What Does This PR Do?

- Fixes query performance by applying filters on the claim table insert/update timestamps earlier in the query

### Validation

Ran queries against prod snowflake manually, verifying row counts look correct


apps/bfd-pipeline-idr/model/idr_claim_institutional_nch.py
```diff
@@ -470,16 +477,23 @@ def fetch_query(
...
-                    {clm}.geo_bene_sk,
-                    {clm}.clm_type_cd,
-                    {clm}.clm_num_sk,
-                    {clm}.clm_dt_sgntr_sk,
-                    {clm}.clm_idr_ld_dt
-                FROM cms_vdm_view_mdcr_prd.v2_mdcr_clm {clm}
-                WHERE {claim_filter(start_time, partition)}
+            WITH claim_base AS (
+                {clm_base_query(start_time, partition, cls.model_type())}
+            ),
```
aschey-forpeople, line 482: One thing to check here, we need to ensure we include all the child tables in the UNION here, otherwise updates could be missing

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -145,6 +145,182 @@ def provider_careteam_name_expr(alias: str, type: str | None) -> str:
...
+    clm = ALIAS_CLM
+    return f"""
+        SELECT
+            {clm}.clm_uniq_id,
+            {clm}.geo_bene_sk,
+            {clm}.clm_type_cd,
+            {clm}.clm_num_sk,
+            {clm}.clm_dt_sgntr_sk,
+            {clm}.clm_idr_ld_dt
+        FROM claim_base clm
```
mel1-G: Not that it matters much since they match but shouldn't this be {clm} not clm like in `clm_base_query`? Same thing for the rest

aschey-forpeople: Mmm yeah it doesn't technically matter here since the alias is local and not shared but why not.

apps/bfd-pipeline-idr/model/idr_claim_institutional_nch.py
```diff
@@ -470,16 +477,23 @@ def fetch_query(
...
+            claims AS (
+                {clm_query()}
+                UNION
+                {clm_dt_sgntr_query()}
+                UNION
+                {clm_child_query("v2_mdcr_clm_instnl")}
```
mel1-G, line 488: Might be a good idea to move these v2_tables into constants since it's repeated a lot? We already write out the names in load_synthetic.py too

aschey-forpeople: Probably... but I don't feel like it right now.

mel1-G (approved):

malessi (approved):

PR #3119, opened 2026-04-27
BFD-4626: Expose mapped claim line fields in v3 server

daveyeb:
**JIRA Ticket:**
BFD-4626

### What Does This PR Do?

Expose claim line fields in the v3 server that had existing IDR mappings                                                                                      but were not surfaced in the FHIR output.

- `item.serviced` changed from `servicedDate` to `servicedPeriod` when a thru date is present; verify no downstream consumers are broken
- Review snap diffs to confirm new fields appear only on expected claim types

### Validation

 - Ran FML maps through matchbox to confirm expected FHIR output
 - Loaded synthetic data into local PostgreSQL and started `bfd-server-ng`
 - Confirmed all 5 fields appear correctly in server output for type 71 professional and type 10 institutional claims
 - Integration tests (`EobReadIT`, `EobSearchIT`, `EobSamhsaFilterIT`) pass with updated snapshots


apps/bfd-model-idr/claims_adj.py
```diff
@@ -785,7 +785,7 @@ def gen_clm_line(
             clm_line[f.CLM_RNDRG_FED_PRVDR_SPCLTY_CD] = random.choice(
                 gen_utils.code_systems[f.CLM_PRVDR_SPCLTY_CD]
             )
-            if random.choice([0, 10]) == 7:
+            if random.randint(0, 10) == 7:
                 clm_line[f.CLM_LINE_ANSTHSA_UNIT_CNT] = random.uniform(0, 10)
             if random.choice([0, 15]) == 7:
```
daveyeb: @malessi out of scope. is this a bug or intentional? if it is, i can fix and put in my stuff.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineProfessionalBase.java
```diff
@@ -52,7 +56,16 @@ public Optional<ExplanationOfBenefit.ItemComponent> toFhirItemComponent() {
...
-    fromDate.map(d -> line.setServiced(new DateType(DateUtil.toDate(d))));
+
+    thruDate.ifPresentOrElse(
+        thru -> {
+          var period = new Period();
+          fromDate.ifPresent(d -> period.setStart(DateUtil.toDate(d)));
+          period.setEnd(DateUtil.toDate(thru));
+          line.setServiced(period);
+        },
+        () -> fromDate.ifPresent(d -> line.setServiced(new DateType(DateUtil.toDate(d)))));
```
daveyeb, line 67: @mel1-G @aschey-forpeople map doesn't handle null clm_line_thru_dt. should we remove the else clause to match the map or keep it as a fallback?

aschey-forpeople: EDIT: sorry, misunderstood. I would check with @alex-dzeda for any FHIR mapping questions.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobReadIT.snap
```diff
@@ -938,7 +938,10 @@ gov.cms.bfd.server.ng.EobReadIT.eobReadPhase1=[
       } ]
     },
     "sequence" : 1,
-    "servicedDate" : "2024-03-04"
+    "servicedPeriod" : {
+      "end" : "2024-04-01T00:00:00+00:00",
```
alex-dzeda: these are only ever going to be YYYY-MM-DD so we should trim to that

daveyeb: nicely trimmed ✂️

apps/bfd-model-idr/claims_adj.py
aschey-forpeople: Seems like a worthy fix, thanks!

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineProfessionalNchExtensions.java
```diff
@@ -36,11 +36,21 @@ public class ClaimLineProfessionalNchExtensions {
   @Column(name = "clm_prcsg_ind_cd")
   private Optional<ClaimProcessingIndicatorCode> processingIndicatorCode;
 
+  @Column(name = "clm_line_ansthsa_unit_cnt")
+  private BigDecimal anesthesiaUnitCount;
```
aschey-forpeople: Hmm do we want this one to be populated if it's missing?

daveyeb: True! I modeled it after totalUnitsCount, should that be changed too? I made a fix with Optional.ofNullable on line 51.

alex-dzeda: I would not populate either if it's missing. One good litmus test is if the field has a cclfMapping populated in bfd-model-idr/dictionary-support-files/ExplanationOfBenefit.yaml (or pharmacy). If no cclfMapping, and it's not part of a required field, that's a good indicator it doesn't need to be populated.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineProfessionalBase.java
alex-dzeda: Hopefully this helps:
http://hl7.org/fhir/R4/explanationofbenefit-definitions.html#ExplanationOfBenefit.item.serviced_x_

Image

Depending on if there is an end date, servicedDate or servicedPeriod is used, and the java methods do that automagically (as in FML). 
In FML:
src.CLM_LINE_FROM_DT as clmLineFromDt -> tgt.serviced = create('Period') as tgtPeriod then{
        clmLineFromDt -> tgtPeriod.start = clmLineFromDt "add start";
        src.CLM_LINE_THRU_DT as clmLineThruDt -> tgtPeriod.end = clmLineThruDt "add end";
    } "Add serviced period";
This means that when only CLM_LINE_FROM_DT is available (or non 9999-12-31, since we don't want to populate 9999-12-31), then it ends up as a servicedDate, otherwise when CLM_LINE_THRU_DT is present, we populate servicedPeriod, with both a start and end date.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobReadIT.snap
```diff
@@ -4991,6 +5020,18 @@ gov.cms.bfd.server.ng.EobReadIT.eobReadValidLong=[
...
       }
+    }, {
+      "amount" : {
+        "currency" : "USD",
+        "value" : 13.45
+      },
+      "category" : {
+        "coding" : [ {
+          "code" : "CLM_LINE_BLOOD_DDCTBL_AMT",
+          "display" : "Revenue Center Blood Deductible Amount",
```
alex-dzeda: @daveyeb can you re-run the codegen for this codesystem? The current name differs from this one. IMO once that's done, this ought to be good to go. I'm fine if we backlog the totalunitscount piece being discussed.

alex-dzeda: "category" : {
        "coding" : [{
          "system" : "https://bluebutton.cms.gov/fhir/CodeSystem/Adjudication",
          "code" : "CLM_LINE_BLOOD_DDCTBL_AMT",
          "display" : "Blood Deductible Amount"
        }]
      }

When using FML, the display is "Blood Deductible Amount". Revenue Centers are specific to institutional claims, so a while ago, I updated the code system to make this non-specific, since it is used by both professional + institutional claims.

In the readme for bfd-server-ng, check out ## Generating enum values

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineProfessionalNchExtensions.java
daveyeb: Leaving totalUnitsCount as-is per https://github.com/CMSgov/beneficiary-fhir-data/pull/3119#discussion_r3166646125 thread

alex-dzeda (dismissed):

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineProfessionalNchExtensions.java
aschey-forpeople: We shouldn't use `Optional.ofNullable` - we should make this `Optional<BigDecimal>` instead. The goal is to minimize the use of null wherever possible to avoid null reference exceptions. More on that in the ARCHITECTURE.md doc in this project.

daveyeb: Make sense. I added a converter for BigDecimal and updated field to `Optional<BigDecimal>`.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineAdjudicationChargeInstitutionalNch.java
```diff
@@ -29,6 +29,9 @@ class ClaimLineAdjudicationChargeInstitutionalNch {
   @Column(name = "clm_line_mdcr_ddctbl_amt")
   private BigDecimal deductibleAmount;
 
+  @Column(name = "clm_line_blood_ddctbl_amt")
+  private BigDecimal bloodDeductibleAmount;
```
aschey-forpeople, line 33: Looking at the map file, it looks like this should also be optional, along with the other `_amt` values. It's a bit of extra scope, but it shouldn't be too much to fix the other values in this file along with this one.

daveyeb: This would require reworking `toFhirAdjudication` in AdjudicationChargeType, touching 12+ files and 100+ `BigDecimal` fields. Happy to do it here or scope it to a follow-up ticket?

aschey-forpeople: Hmm I didn't realize there were that many. We can save it for later.

aschey-forpeople: Thanks for checking 👍

aschey-forpeople (approved):

alex-dzeda (approved):

PR #3126, opened 2026-05-05
BFD-4705: Expand v3 server metrics

mel1-G:
**JIRA Ticket:**
BFD-4705

### What Does This PR Do?

- Overall these changes include support for request telemetry logging for v3 and expands our use of Micrometer for v3 so that instead of only knowing that a db query took X time we can now correlate things like latency vs claim type, latency ver SAMHSA filter mode, response size vs processing time etc
- Added the micrometer-java21 dependency to collect virtual threads metrics
- Configured Micrometer to publish v3 metrics to CloudWatch (disabled for local). We’re leaving the configurations in our properties for now since we plan to switch to DataDog down the line.
- Expanded db query metrics:
    - Coverage
        - Timer along with tags that show whether a coverage query had part C, part D, and/or LIS since that grows the query
    - Claim
        - Fan-out so record how many claim-type queries are launched per request
        - Breakdown fetch latency by claim type so we can now determine which claim type dominated
    - Beneficiary
        - Whether there was a successful patient match which was a recent ask by BB2
        - Number of scenarios attempted which might be insightful in how it affects latency down the line when we're able to support a lot more of the scenarios
- Collect metrics at handler-level to measure total processing time to derive processing overhead & FHIR mapping transformation cost
    - Coverage
        - Timers added for searchByCoverageId & searchByBeneficiary
    - Eob
        - Timer with SAMHSA filter mode tag to understand how that affects our claims transformation/filtering
        - Capture result size 
    - Patient
        - Timers added for matchPatient & searchByBeneficiaryC4DIC since those are pretty complex
- For individual requests:
   - Added HAPI request lifecycle timestamp metrics & an interceptor for HAPI request lifecycle hooks
      - hapi_server_incoming_request_post_process_timestamp_in_millis
      - hapi_server_incoming_request_pre_handle_timestamp_in_millis
      - hapi_server_incoming_request_pre_process_timestamp_in_millis
      - hapi_server_outgoing_response_timestamp_in_millis
      - hapi_server_processing_completed_normally_timestamp_in_millis
      - hapi_server_processing_completed_timestamp_in_millis 
   - Captured request/response headers, status, duration, and resource counts
   - Extended ExtractMetadataFilter to record 

* Adds any new software dependencies --> Added micrometer-java21 dependency to collect metrics for virtual threads

### Validation

Tested by running the server locally, sending requests, and checking metrics at: http://localhost:8080/actuator/metrics
Example request:
xh \
  'http://localhost:8080/v3/fhir/ExplanationOfBenefit/_search?patient=412457726&_security=42CFRPart2' \
  'X-Amzn-Mtls-Clientcert:samhsa_allowed'
Collected these metrics:
 - "application.eob.handler.search_by_bene"
 - "application.beneficiary.search_xref_by_bene_sk"
 - "application.claim.search_by_bene"
 - "application.claim.search_by_bene.fan_out"
 - "application.claim.fetch_claims_with_claim_type"
 - "application.eob.handler.results.size"

With these metrics we can now see latency broken down by stages similar to traces. We have orchestration timing vs database timing through handler-level and query-level metrics. We take advantage of dynamic tags so in this request we have metrics with SAMHSA mode tagged so we can later slice performance to compare. Claim type is also tagged so we can tell which claim type dominated latency in this request. I tested specifically that by querying with a tag filter here: http://localhost:8080/actuator/metrics/application.claim.fetch_claims_with_claim_type?tag=claim_type:ClaimRx

Also, I verified that we now capture virtual thread metrics. You can now see the following two under http://localhost:8080/actuator/metrics:
    - jvm.threads.virtual.submit.failed which is "the number of events when starting or unparking a virtual thread failed"
    - jvm.threads.virtual.pinned which is "the duration while the virtual thread was pinned without releasing its platform thread"
    
As for logging individual requests, I tested by running the server locally and sending some requests. This is an example of what gets records:

```
23:19:19.146Z  INFO 59664 g.c.b.s.ng.log.RequestTelemetryLogger    : Request Completed
 logType="requestTelemetry" 
mdc.hapi.server_incoming_request_pre_handle_timestamp_in_millis = 1779117590925
mdc.certificateAlias = samhsa_allowed
mdc.clientName = null
mdc.clientIp = null
mdc.requestId = /[0:0:0:0:0:0:0:1]:52925@28FA6350#0#0
mdc.hapi.server_processing_completed_timestamp_in_millis = 1779117591310
mdc.http_access_request_header_Accept-Encoding = gzip, deflate, br, zstd
mdc.hapi.server_incoming_request_post_process_timestamp_in_millis = 1779117590925
mdc.hapi.server_processing_completed_normally_timestamp_in_millis = 1779117591310
mdc.http_access_response_duration_milliseconds = 402
mdc.hapi.server_incoming_request_pre_process_timestamp_in_millis = 1779117590921
mdc.hapi.server_outgoing_response_timestamp_in_millis = 1779117591254
mdc.http_access_response_header_Content-Encoding = gzip
mdc.http_access_request_uri = /v3/fhir/ExplanationOfBenefit/_search
mdc.http_access_response_header_X-Request-ID = G2wou9605XFjyHMC
mdc.http_access_request_url = http://localhost:8080/v3/fhir/ExplanationOfBenefit/_search
mdc.clientId = null
mdc.http_access_request_http_method = GET
mdc.uri = /v3/fhir/ExplanationOfBenefit/_search
mdc.remoteAddress = [0:0:0:0:0:0:0:1]
mdc.resources_returned_count = 1
mdc.http_access_response_status = 200
```

```
Database query metrics:
15:19:51.095Z  INFO 25645 g.c.b.server.ng.log.QueryTelemetryUtil   : Database query
Completed logType="databaseQueryTelemetry"
query_name = getXrefSkFromBeneSk
duration_in_milliseconds = 29
row_count = 1
mdc.hapi.server_incoming_request_pre_handle_timestamp_in_millis = 1779117590925
mdc.certificateAlias = samhsa_allowed
mdc.clientName = null
mdc.clientIp = null
mdc.requestId = /[0:0:0:0:0:0:0:1]:52925@28FA6350#0#0
mdc.http_access_request_header_Accept-Encoding = gzip, deflate, br, zstd
mdc.hapi.server_incoming_request_post_process_timestamp_in_millis = 1779117590925
mdc.hapi.server_incoming_request_pre_process_timestamp_in_millis = 1779117590921
mdc.http_access_request_uri = /v3/fhir/ExplanationOfBenefit/_search
mdc.http_access_request_url = http://localhost:8080/v3/fhir/ExplanationOfBenefit/_search
mdc.clientId = null
mdc.http_access_request_http_method = GET
mdc.uri = /v3/fhir/ExplanationOfBenefit/_search
mdc.remoteAddress = [0:0:0:0:0:0:0:1]
```


mel1-G:
@sb-benohe tagging ya since I'm adding the micrometer-java21 dependency so we can collect virtual thread metrics.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/Application.java
```diff
@@ -84,6 +85,19 @@ public TimedAspect timedAspect(MeterRegistry registry) {
...
 
+  /**
+   * Configures Micrometer to collect virtual thread metrics.
+   *
+   * @param registry meter registry
+   * @return virtual thread metrics
+   */
+  @Bean
+  public VirtualThreadMetrics virtualThreadMetrics(MeterRegistry registry) {
+    VirtualThreadMetrics metrics = new VirtualThreadMetrics();
```
aschey-forpeople:
```suggestion
    var metrics = new VirtualThreadMetrics();
```

apps/bfd-server-ng/pom.xml
```diff
@@ -234,6 +234,13 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>dynamodb</artifactId>
         </dependency>
+
+        <dependency>
+            <!-- Used to collect metrics for virtual threads. -->
```
aschey-forpeople: I was worried that we might have a JDK mismatch here so it might be good to clarify this.
```suggestion
            <!-- Used to collect metrics for virtual threads. 
            Note that java21 refers to the minimum required java version
            for this package, not the exact version.
            -->
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/LoggerConstants.java
```diff
@@ -70,6 +70,58 @@ private LoggerConstants() {
    */
   public static final String FINAL_DETERMINATION = "finalDetermination";
 
+  /** Name of the timestamp when the HAPI request reached pre-handle processing. */
+  public static final String HAPI_INCOMING_PRE_HANDLE =
```
jadudm, line 74: Per convo, adding some "why/what are these" comments to these metrics may be useful for both now and future us.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/BeneficiaryRepository.java
```diff
@@ -136,45 +142,64 @@ public Optional<Long> getXrefSkFromMbi(String mbi) {
...
-        LogUtil.logBeneSk(matchedBene.getBeneSk());
-        var finalDetermination =
-            new FinalDetermination(combinationIndex, matchedRecords.getFirst());
-        return new PatientMatchResult(
-            combinationResults, Optional.of(finalDetermination), Optional.of(matchedBene));
+    var timer = new MetricTimer(meterRegistry);
+    var scenariosAttempted = 0;
+    var success = false;
+
+    try {
```
aschey-forpeople: I'm not sure if we want to wrap all of these in `try/finally`. If some of these error out, then it could skew the metrics in a way that we don't want.

aschey-forpeople: Oh hmm I see, without this it could result in the timer starting but never stopping I guess. We may want to look at using `record(() -> {})` and `recordCallable(() -> {})` instead? I don't know if the nesting would be unwieldy, but I tend to prefer wrapping these types of things so you don't forget to call `stop` if we add some branching logic in the future.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryLogger.java
```diff
@@ -0,0 +1,169 @@
...
+   *
+   * @param dbQueries the database query telemetry collected
+   */
+  public void recordDatabaseQueries(List<DbQueryTelemetry> dbQueries) {
+    if (dbQueries.isEmpty()) {
+      return;
+    }
+    try {
+      MDC.put(logKey(MDC_PREFIX, DB_QUERIES), objectMapper.writeValueAsString(dbQueries));
+    } catch (JsonProcessingException _) {
```
aschey-forpeople: Is this exception expected to happen sometimes or is this just because it requires you to explicitly handle this one (checked exception)?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryLogger.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  /** Emits a structured log containing all metrics and attributes of a request. */
+  public void logRequestComplete() {
+
+    var logBuilder =
+        LOGGER.atInfo().setMessage("Request Completed").addKeyValue(LOG_TYPE, "requestTelemetry");
+
+    logBuilder = addMdcIfPresent(logBuilder, HAPI_INCOMING_PRE_HANDLE);
```
aschey-forpeople: Could shorten this a bit by creating a `static List` of all the key names and then looping through them instead of calling `addMdcIfPresent` repeatedly.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryContext.java
```diff
@@ -0,0 +1,67 @@
...
+ * is needed so database query timings can be accumulated amd emitted in the final request telemetry
+ * log.
+ */
+public class RequestTelemetryContext {
+
+  private static final ThreadLocal<RequestTelemetryContext> CONTEXT = new ThreadLocal<>();
+  private final List<DbQueryTelemetry> dbQueries = new ArrayList<>();
+
+  /** Creates a context for the current thread. */
+  public static void createContext() {
```
aschey-forpeople: Rather than having a separate `current` and `createContext`, I think you could have `current` lazily initialize the context if it's `null`. Simplifies things a bit.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/MetricTimer.java
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public <T> CompletableFuture<T> recordMetricAsync(
+      String metricName,
+      Supplier<Iterable<Tag>> tagsSupplier,
+      Supplier<CompletableFuture<T>> supplier) {
+
+    var timer = Timer.start(meterRegistry);
+
+    try {
+      return supplier.get().whenComplete((_, _) -> stop(metricName, timer, tagsSupplier.get()));
```
aschey-forpeople, line 56: Did `Timer.builder(metricName).tags(tags).register(meterRegistry).record(callable)` not work here? Seems like that would be less boilerplate but I could be wrong.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/MdcAwareThreadPoolExecutor.java
```diff
@@ -12,14 +13,22 @@ public class MdcAwareThreadPoolExecutor extends ThreadPoolTaskExecutor {
   static Runnable wrapWithMdcContext(Runnable task) {
     // save the current MDC context
     var contextMap = MDC.getCopyOfContextMap();
+    var telemetryContext = RequestTelemetryContext.current();
```
aschey-forpeople: I wonder if it makes sense to centralize this `RequestTelemetryContext` stuff and `MDC` context into a single class like `ThreadLocalContext` or something? There we could ensure that any kind of thread-local context gets updated in the same place.

aschey-forpeople: Alternatively, I wonder if you could use a [request-scoped bean](https://www.baeldung.com/spring-bean-scopes#1-request-scope) to avoid some of this thread-local stuff?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/LoggerConstants.java
aschey-forpeople: Open to suggestions, but I think the comment here explains what the constant is sufficiently. If you look where the constants are referenced, there are additional explanations.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryLogger.java
mel1-G: It wasn't that it required me to explicitly handle a JsonProcessingException. I just thought back to how we handled it for the DynamoDB audit logger since this logger at the time similarly serialized a list of objects for dbQueries. I probably could have just kept it general though.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/MetricTimer.java
mel1-G: It didn't work because of the dynamic tags. Tags like hasPartC and the patient match outcome for example depend on the operation result so it doesn't work as cleanly which is why I ended up refactoring MetricTimer a bit.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryLogger.java
aschey-forpeople: For the dynamodb one we had it rethrow the exception with the message which I think would be okay 

```
 catch (JsonProcessingException e) {
      throw new IllegalArgumentException("Failed to serialize patient match audit record", e);
    }
```

I would just avoid silently swallowing errors like this:

```
catch (JsonProcessingException _)
```

because it makes it difficult to see what's going on if there is an error there.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/MetricTimer.java
aschey-forpeople: Oh I see, because it needs the result. Alright, that makes sense.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryLogger.java
```diff
@@ -0,0 +1,117 @@
...
+    if (value != null) {
+      MDC.put(logKey(MDC_PREFIX, key), value);
+    }
+  }
+
+  private String replaceAllMbis(String input) {
+    if (input == null) {
+      return input;
+    }
+    Matcher matcher = MBI_REGEX.matcher(input);
```
aschey-forpeople:
```suggestion
    var matcher = MBI_REGEX.matcher(input);
```

aschey-forpeople (dismissed):

aschey-forpeople (approved):

alex-dzeda (approved):

PR #3129, opened 2026-05-12
BFD-4716: Update Matchbox Settings

alex-dzeda:
**JIRA Ticket:**
BFD-4716

### What Does This PR Do?

Updates the settings for the matchbox docker compose + adds a gitignore to temp files. Also adds a delay after the healthcheck on matchbox is active. 

Test out ./compile-all-resources.sh 

@sb-benohe switching to alpine postgres, FYI.

### Validation

"ImplementationGuideProviderR4:251: Finished engines during startup  Memory used 2Gb"
./compile-all-resources continues to function, and no OOMs while the docker compose itself guarantees a higher minimum heap size.

### DIFF
--- a/apps/bfd-model-idr/docker-compose.yml
+++ b/apps/bfd-model-idr/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       - "8080:8080"
     depends_on:
       - matchbox-db
-    entrypoint: ["java", "-Xms512m", "-Xmx6G", "-XX:+UseZGC", "-XX:ZUncommitDelay=30", "-XX:MaxMetaspaceSize=512m", "-Dfhir.settings.path=/config/fhir-settings.json", "-Dspring.config.additional-location=optional:file:/config/application.yaml", "-jar", "/matchbox.jar"]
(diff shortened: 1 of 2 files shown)


apps/bfd-model-idr/docker-compose.yml
```diff
@@ -26,9 +26,9 @@ services:
     environment:
       - UV_PYTHON=3.14.3
       - UV_PROJECT_ENVIRONMENT=/tmp/.venv
-    command: uv run setup_matchbox.py --url http://matchbox:8080/matchboxv3/fhir
+    command: /bin/sh -c "sleep 10 && uv run setup_matchbox.py --url http://matchbox:8080/matchboxv3/fhir"
```
aschey-forpeople: I'd prefer to have some kind of wait loop with a timeout instead of hardcoding a sleep here.

alex-dzeda: The block above (health check) is effectively that, the problem is sometimes it's ready but unable to accept requests, which is a race condition, and there are no other hooks we can latch on to. 
The alternative would be retries on setup_matchbox 🤔.

alex-dzeda: Scratch that, found a better hook -> verifiable that it doesn't even attempt to start the container that pushes data until matchbox is ready.

aschey-forpeople (approved):

sb-benohe (approved):
good use of alpine!

mel1-G (approved):

PR #3131, opened 2026-05-12
bfd-4719: uv dependancy updates and floor pin

sb-benohe:
**JIRA Ticket:**
BFD-4719

### What Does This PR Do?

- updates uv dependancies based on complaints from Dependabot
- pins versions of unpinned dependancies
- fixes tests/lint that broke as result

 - that these changes are logical for the usecases of these files
 - tests pass

### Validation

- ran tests and lint locally and on GH Actions
- asked claude for recommended fixes, confirmed this was the way


apps/utils/locust_tests/common/bfd_user_base.py
```diff
@@ -189,7 +189,7 @@ def get_by_url(
             catch_response=True,
         ) as response:
             if response.status_code != 200:
-                if isinstance(response, ResponseContextManager):
+                if isinstance(response, ResponseContextManager):  # pyright: ignore[reportUnnecessaryIsInstance]
```
sb-benohe, line 192: Happy to remove this check and re-write this as pyright is reporting this is ALWAYS an instance of ResponseContextManager

apps/utils/locust_tests/common/bfd_user_base.py
```diff
@@ -227,7 +227,7 @@ def post_by_url(
...
```
sb-benohe, line 230: Same here as previous comment

apps/bfd-pipeline-idr/pipeline.py
```diff
@@ -30,8 +30,6 @@
     bfd_test_date,
 )
 
-telemetry.disable_telemetry()
```
aschey-forpeople, line 33: Removed because this is a no-op in newer versions of hamilton (telemetry was removed)

apps/utils/locust_tests/common/bfd_user_base.py
aschey-forpeople: pyright may be correct here but this is legacy code and I don't want to risk breaking it so I think it's fine for now

aschey-forpeople (approved):

malessi (approved):

PR #3139, opened 2026-05-19
BFD-4732: v3 server config/performance updates

aschey-forpeople:
**JIRA Ticket:**
BFD-4732

### What Does This PR Do?

- Fixes parallel db query execution by creating a new `EntityManager` per task instead of using the shared one
  - `@PersistenceContext` _does_ manage `@EntityManager` scoping accordingly, but since we were passing the same bean instance across multiple tasks, it was still being shared inappropriately
- Fixes task executor configuration
  - We require a custom `Executor` to appropriately propagate the MDC context, but we were using the incorrect executor type, causing virtual threads to not be enabled as expected.
- Adds `@PersistenceContext` to the remaining `EntityManager` instances since it seems like you're supposed to include this for Spring-managed instances 

### Validation

Deployed changes to prod with a noticeable performance increase on db queries due to increase in parallelism


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/beneficiary/BeneficiaryRepository.java
```diff
@@ -30,7 +31,7 @@
 @AllArgsConstructor
 @SuppressWarnings("java:S2077")
 public class BeneficiaryRepository {
-  private final EntityManager entityManager;
+  @PersistenceContext private EntityManager entityManager;
```
aschey-forpeople, line 34: Adding `@PersistenceContext` on a `final` variable results in this warning:

```
WARNING: Final field entityManager in class gov.cms.bfd.server.ng.coverage.CoverageRepository has been mutated reflectively by class org.springframework.beans.factory.annotation.InjectionMetadata$InjectedElement in unnamed module @5af97850
```

Seems like Spring modifies these at startup using reflection so I removed `final` 🤷

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSearchIT.java
```diff
@@ -38,13 +35,8 @@
 import org.junit.jupiter.params.provider.EnumSource;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.junit.jupiter.params.provider.ValueSource;
-import org.mockito.Mock;
-import org.springframework.beans.factory.annotation.Autowired;
 
 class EobSearchIT extends IntegrationTestBase {
-  @Autowired private EobResourceProvider eobResourceProvider;
-  @Autowired private EntityManager entityManager;
-  @Mock HttpServletRequest request;
```
aschey-forpeople, line 47: Unrelated change but these three variables were unused

alex-dzeda (approved):

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/Application.java
```diff
@@ -104,7 +104,7 @@ public DynamoDbClient dynamoDbClient(Configuration configuration) {
    */
   @Bean
   public SimpleAsyncTaskExecutor taskExecutor(SimpleAsyncTaskExecutorBuilder builder) {
-    return builder.virtualThreads(true).taskDecorator(Application::wrapWithMdcContext).build();
+    return builder.taskDecorator(Application::wrapWithMdcContext).build();
```
alex-dzeda, line 107: How were you able to verify that it was redundant?

aschey-forpeople: Ran it locally with the debugger and inspected the properties on that object to make sure it was already `true`.

malessi (approved):

PR #3152, opened 2026-06-01
BFD-4743: IDR Pipeline performance improvements V2

malessi:
**JIRA Ticket:**
BFD-4743

### What Does This PR Do?

This PR refactors the IDR Pipeline's `BatchLoader` to load batches of 100,000 rows of data 1000 (by default) rows at a time _concurrently_ using asynchronous `psycopg`. After considerable testing, this does seem to meaningfully improve the performance of loading these batches. More testing is necessary to determine the magnitude of improvement, though.

This PR also adds some additional timing logs, and updates the logging configuration to use the non-blocking `QueueHandler` that works well with asynchronous operations.

### Validation

- Running the modified IDR Pipeline against `4686-prod`, _verifying that_ the IDR Pipeline does run and is able to load tables as expected

### DIFF
--- a/apps/bfd-pipeline-idr/loader.py
+++ b/apps/bfd-pipeline-idr/loader.py
@@ -358,7 +358,7 @@ class BatchLoader:
             # We want to immediately terminate the transaction if there is already a lock on
             # the table so that we avoid extraneous waits because if there is a lock this table
             # is being updated concurrently and that existing update will have the same result
-            await cur.execute("SET LOCAL statement_timeout=500")
+            await cur.execute("SET LOCAL statement_timeout=3000")
             await cur.execute(
                 f"""
                 WITH current_ts AS (
--- a/apps/bfd-pipeline-idr/loader.py
+++ b/apps/bfd-pipeline-idr/loader.py
@@ -4,6 +4,7 @@ import logging
 from collections.abc import Iterator, Sequence
 from datetime import UTC, date, datetime
 
+import anyio
 import psycopg
 import psycopg_pool
 from psycopg.abc import Params, QueryNoTemplate
@@ -49,10 +50,15 @@ class PostgresLoader:
         load_type: LoadType,
         load_mode: LoadMode,
     ) -> bool:
-        return asyncio.run(
-            self._async_load(
-                fetch_results, model, job_start, partition, progress, load_type, load_mode
-            )
+        return anyio.run(
+            self._async_load,
+            fetch_results,
+            model,
+            job_start,
+            partition,
+            progress,
+            load_type,
+            load_mode,
         )
 
     async def _async_load(
@@ -187,14 +193,18 @@ class BatchLoader:
             num_rows += len(results)
 
             self.insert_batch_timer.start()
-            await asyncio.gather(
-                *(
-                    self._load_batch_part(idx, part, timestamp)
-                    for idx, part in enumerate(
-                        itertools.batched(results, PER_BATCH_CONCURRENT_ROWS, strict=False)
+            async with anyio.create_task_group() as tg:
+                for idx, part in enumerate(
+                    itertools.batched(results, PER_BATCH_CONCURRENT_ROWS, strict=False)
+                ):
+                    tg.start_soon(
+                        self._load_batch_part,
+                        idx,
+                        part,
+                        timestamp,
+                        name=f"{self.table}-{self.partition.name}-{idx}",
                     )
-                )
-            )
+
             self.insert_batch_timer.stop()
 
             async with self.pool.connection() as conn, conn.cursor(binary=True) as cur:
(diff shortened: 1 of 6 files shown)


apps/bfd-pipeline-idr/loader.py
```diff
@@ -50,122 +49,191 @@ def load(
...
+        partition: LoadPartition,
+        progress: LoadProgress | None,
+        load_type: LoadType,
+        load_mode: LoadMode,
+    ) -> bool:
+        async with psycopg_pool.AsyncConnectionPool(
+            conninfo=get_connection_string(load_mode),
+            min_size=PER_BATCH_MIN_CONNECTIONS if not LoadMode.LOCAL else 1,
+            max_size=PER_BATCH_MAX_CONNECTIONS if not LoadMode.LOCAL else 1,
+            timeout=600,
```
malessi, line 85: The timeout must be this large as during the asynchronous upserts _most_ of the time is actually spent  acquiring a connection from the pool, not actually executing the `COPY` + `INSERT ... FROM ... ON CONFLICT`. We were unable to determine why this was the case, but we saw this behavior exhibited by both `psycopg` _and_ `asyncpg` so it must be some quirk of Postgres.

apps/bfd-pipeline-idr/loader.py
```diff
@@ -50,122 +50,200 @@ def load(
         load_type: LoadType,
         load_mode: LoadMode,
     ) -> bool:
-        return BatchLoader(
-            self.conn,
+        return anyio.run(
+            self._async_load,
```
malessi, line 53: `BatchLoader.load` simply delegates to the asynchronous `_async_load` so that asynchronous code is self-contained within the class.

apps/bfd-pipeline-idr/loader.py
```diff
@@ -50,122 +50,200 @@ def load(
...
+            else (
+                f"DO UPDATE SET {self.update_set_str}, bfd_updated_ts=%(timestamp)s "
+                f"{self.where_clause}"
+            )
+        )
+        self.last_updated_set_clause = ", ".join(
+            f"{col} = %(timestamp)s" for col in self.model.last_updated_date_column()
+        )
+        self.timestamp_placeholders = ", ".join("%(timestamp)s" for _ in self.meta_keys)
+
```
malessi, line 153: Most of these changes were to remove any sort of string manipulation from the "hot loop" (the concurrent upserts) as they are all static. A _micro_ optimization, but beneficial over a long-running data load.

apps/bfd-pipeline-idr/loader.py
```diff
@@ -50,122 +50,200 @@ def load(
...
+                for idx, part in enumerate(
+                    itertools.batched(results, PER_BATCH_CONCURRENT_ROWS, strict=False)
+                ):
+                    tg.start_soon(
+                        self._load_batch_part,
+                        idx,
+                        part,
+                        timestamp,
+                        name=f"{self.table}-{self.partition.name}-{idx}",
+                    )
```
malessi, line 213: This is the **most relevant** change. This will spawn many concurrent tasks for each "batch" (`part`) of 1000 rows generated by the `itertools.batched` generator that will upsert those rows into the target `self.table` table. While _all_ of the tasks are spawned at once, most will be blocked on waiting for a connection to be acquired from the connection pool at any given moment in time.

apps/bfd-pipeline-idr/loader.py
aschey-forpeople: Maybe put a `TODO` comment explaining that we may want to investigate this later?

apps/bfd-pipeline-idr/loader.py
```diff
@@ -50,122 +50,200 @@ def load(
...
+            async with self.pool.connection() as conn, conn.cursor(binary=True) as cur:
+                if self.load_type == LoadType.INCREMENTAL and self.model.last_updated_date_table():
+                    self.last_updated_timer.start()
+                    full_temp_table = await self._setup_temp_table(cur, self.partition.name)
+                    await self._copy_data(cur, full_temp_table, results)
+                    await self._last_updated(cur, full_temp_table, timestamp)
+                    await conn.commit()
+                    self.last_updated_timer.stop()
+
+                await self._calculate_load_progress(cur, results)
```
malessi, line 228: Last updated was already problematic, and trying to _concurrently_ updated the relevant last updated columns only made the problems worse. This was the best solution I could come up with that ensured the upsert performance remained consistent. However, for _most_ batches, last updated will still timeout.

In short, we need a better solution for last updated.

malessi: After some discussion with @aschey-forpeople and @mel1-G, we have decided that a half-working solution that only _sometimes_ updates the relevant last updated columns is worse than having no last updated solution at all. Instead, a follow-up PR will introduce a more comprehensive solution for last updated and these lines have been commented-out.

mel1-G (dismissed):

apps/bfd-pipeline-idr/loader.py
malessi: Good idea. Done in [61dad85](https://github.com/CMSgov/beneficiary-fhir-data/pull/3152/commits/61dad8552d0f50ad84ce856d719da496b941ab6b).

aschey-forpeople (approved):

mel1-G (approved):

PR #3153, opened 2026-06-02
BFD-4474: add clm rndrg prvdr tax num

patrickgmayday:
**JIRA Ticket:**
BFD-4474

### What Does This PR Do?

Adds CLM_RNDRG_PRVDR_TAX_NUM to bfd-server-ng, and sets it as an Extension on ExplanationOfBenefits.Line if the header "IncludeTaxNumbers" is passed in to the search call.

Please follow standard Git commit message guidelines:
* First line should be a capitalized, short (50 chars or less) summary.
* The rest of the message should be in standard Markdown format, wrapped to 72 characters.
* Describe your changes in imperative mood, e.g. "make xyzzy do frotz" instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy to do frotz", as if you are giving orders to the codebase to change its behavior.
* List all relevant Jira issue keys, one per line at the end of the message, per: <https://confluence.atlassian.com/jirasoftwarecloud/processing-issues-with-smart-commits-788960027.html>.

Reference: <https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project>.
-->

1) Add support for IncludeTaxNumbers header
2) Expose CLM_RNDRNG_PRVDR_TAX_NUM when IncludeTaxNumbers is set to true
3) Reject incorrectly repeating headers via parseBooleanHeader
4)Claim security state moved to ClaimState class from user-supplied filter options

### Validation

Code tested under the following conditions:

 echo "TEST 1: no IncludeTaxNumbers header — expected no output"   
curl -s -H "Accept: application/fhir+json" \
"http://localhost:8080/v3/fhir/ExplanationOfBenefit/1028512039960" \
| jq | grep "912834729"
TEST 1: no IncludeTaxNumbers header — expected no output

 echo "TEST 2: IncludeTaxNumbers false — expected no output"
curl -s -H "Accept: application/fhir+json" \
-H "IncludeTaxNumbers: false" \
"http://localhost:8080/v3/fhir/ExplanationOfBenefit/1028512039960" \
| jq | grep "912834729"
TEST 2: IncludeTaxNumbers false — expected no output

 echo "TEST 3: IncludeTaxNumbers true — expected tax number output"
curl -s -H "Accept: application/fhir+json" \
-H "IncludeTaxNumbers: true" \
"http://localhost:8080/v3/fhir/ExplanationOfBenefit/1028512039960" \
| jq | grep "912834729"
TEST 3: IncludeTaxNumbers true — expected tax number output
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"
            "value": "912834729"

and finally, a test that shows the full message:

 curl -s -H "Accept: application/fhir+json" \
-H "IncludeTaxNumbers: true" \
"http://localhost:8080/v3/fhir/ExplanationOfBenefit/1028512039960" \
| jq

Found in the output:

       {
          "url": "https://bluebutton.cms.gov/fhir/StructureDefinition/CLM-PRVDR-TAX-NUM",
          "valueIdentifier": {
            "type": {


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/FhirInputConverter.java
```diff
@@ -453,4 +454,26 @@ private static MetaSourceSk parseSourceToken(TokenParam token) {
...
+
+    if (headerValues == null || headerValues.isEmpty()) {
+      return false;
+    } else if (headerValues.size() > 1) {
+      // we do not know which to honor
+      throw new InvalidRequestException(
+          "Unsupported " + headerName + " header value: " + headerValues);
+    } else {
+      // if not true is false and doesn't need an error
+      return Boolean.parseBoolean(headerValues.getFirst());
```
clewellyn-nava: Just commenting that I like this 👍

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobReadIT.java
```diff
@@ -26,13 +28,31 @@
 @ExtendWith(MockitoExtension.class)
 @MockitoSettings(strictness = Strictness.LENIENT)
 class EobReadIT extends IntegrationTestBase {
+  private static final String INCLUDE_TAX_NUMBERS = "IncludeTaxNumbers";
```
clewellyn-nava: Do we have a universal place for constants?

patrickgmayday: > Do we have a universal place for constants?

There is an IdrConstants class that doesn't look like it's used very much, we could use that

clewellyn-nava: Hmmm, would like the team to weigh in on this before we make a decision.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/FhirInputConverter.java
```diff
@@ -453,4 +454,26 @@ private static MetaSourceSk parseSourceToken(TokenParam token) {
...
   }
+
+  /**
+   * Checks a RequestDetails object for headers, returns True if 1 exists.
+   *
+   * @param requestDetails requestDetails object from call
+   * @param headerName name of header to check for existence
+   * @return True if the header exists and there is only 1, False otherwise
+   */
+  public static Boolean parseBooleanHeader(RequestDetails requestDetails, String headerName) {
```
aschey-forpeople: Might be good to make this `Optional<Boolean>` just in case we want to change the default value for another use case.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/FhirInputConverter.java
```diff
@@ -453,4 +454,26 @@ private static MetaSourceSk parseSourceToken(TokenParam token) {
...
+    List<String> headerValues = requestDetails.getHeaders(headerName);
+
+    if (headerValues == null || headerValues.isEmpty()) {
+      return false;
+    } else if (headerValues.size() > 1) {
+      // we do not know which to honor
+      throw new InvalidRequestException(
+          "Unsupported " + headerName + " header value: " + headerValues);
```
mel1-G: Omit user input from the log here

clewellyn-nava (dismissed):
LGTM

marshallweekley-eng (dismissed):

mel1-G:
General feedback: 
Make sure you fully fill out the PR check list so moving the description of your changes to where it says "Replace me," marking any security implications, checking off that you considered the security implications and created tested if needed, and how you validated your changes.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/FhirInputConverter.java
```diff
@@ -453,4 +454,26 @@ private static MetaSourceSk parseSourceToken(TokenParam token) {
...
+  /**
+   * Checks a RequestDetails object for headers, returns True if 1 exists.
+   *
+   * @param requestDetails requestDetails object from call
+   * @param headerName name of header to check for existence
+   * @return True if the header exists and there is only 1, False otherwise
+   */
+  public static Optional<Boolean> parseBooleanHeader(
+      RequestDetails requestDetails, String headerName) {
+    List<String> headerValues = requestDetails.getHeaders(headerName);
```
aschey-forpeople:
```suggestion
    var headerValues = requestDetails.getHeaders(headerName);
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/FhirInputConverter.java
```diff
@@ -453,4 +454,26 @@ private static MetaSourceSk parseSourceToken(TokenParam token) {
...
+
+    if (headerValues == null || headerValues.isEmpty()) {
+      return Optional.of(false);
```
aschey-forpeople: This should probably return `Optional.empty()` if no header values are supplied.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/FhirInputConverter.java
```diff
@@ -453,4 +454,26 @@ private static MetaSourceSk parseSourceToken(TokenParam token) {
...
+    } else if (headerValues.size() > 1) {
+      // we do not know which to honor
+      throw new InvalidRequestException("Multiple values supplied for header: " + headerName);
+    } else {
+      // if not true is false and doesn't need an error
```
aschey-forpeople: This comment hurts my brain a little... what's the intent here?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobReadIT.java
```diff
@@ -114,12 +158,12 @@ void eobReadNonLatestProfessionalIsNotReturned() {
         entityManager
             .createQuery(
                 """
-                SELECT c
-                  FROM ClaimProfessionalNch c
-                  JOIN FETCH c.beneficiary b
-                  JOIN FETCH c.claimItems cl
-                  WHERE c.claimUniqueId = :claimId
-                """,
+                            SELECT c
```
aschey-forpeople: nit: can we revert the extra indentation?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobReadIT.java
aschey-forpeople: You could put this in `IntegrationTestBase` to avoid defining it separately in each test class.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/ClaimFilterOptions.java
```diff
@@ -0,0 +1,17 @@
...
+/** Claim Filter options. */
+@Value
+@Builder
+public class ClaimFilterOptions {
+
+  @Builder.Default SamhsaFilterMode samhsaFilterMode = SamhsaFilterMode.EXCLUDE;
+
+  @Builder.Default boolean includeTaxNumber = false;
+
+  @With @Builder.Default ClaimSecurityStatus securityStatus = ClaimSecurityStatus.NONE;
```
aschey-forpeople: I'm not sure if `ClaimSecurityStatus` necessarily belongs here. Since this is `ClaimFilterOptions`, I think this should only include things that the _user_ specifies should modify the claim. `ClaimSecurityStatus` is a value computed for the internal state, so maybe it belongs in another class called `ClaimState` or something. We'd then have to pass two params through the call stack instead of one, but I think that's okay.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -180,7 +179,9 @@ private ExplanationOfBenefit transformToFhir(ClaimBase claim, SamhsaFilterMode s
     var securityStatus =
         isSamhsa ? ClaimSecurityStatus.SAMHSA_APPLICABLE : ClaimSecurityStatus.NONE;
 
-    return claim.toFhir(securityStatus);
+    var newOptions = options.withSecurityStatus(securityStatus);
```
aschey-forpeople: Building off my previous comment, this is what makes me think we need to separate this out. Constructing an instance of `ClaimFilterOptions` without a known value for `ClaimSecurityStatus` is a bit risky, because what if we add some more code higher up in the call stack that erroneously believes the value for `ClaimSecurityStatus` is already computed, but it's actually left at the default value? We would have a pretty severe bug at that point.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobReadIT.java
```diff
@@ -26,13 +28,31 @@
...
+  private IReadTyped<ExplanationOfBenefit> eobReadWithIncludeTaxNumbersHeader(String headerValue) {
+    final var fhirClient = getFhirClient();
+    final var headersInterceptor = new AdditionalRequestHeadersInterceptor();
+    headersInterceptor.addHeaderValue(INCLUDE_TAX_NUMBERS, headerValue);
+    fhirClient.registerInterceptor(headersInterceptor);
+
+    return fhirClient.read().resource(ExplanationOfBenefit.class);
+  }
+
+  private boolean hasTaxNumberExtension(ExplanationOfBenefit eob) {
```
aschey-forpeople: This and any other setup logic could also be moved to `IntegrationTestBase` to avoid duplicating code between the different test classes.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/FhirInputConverter.java
patrickgmayday: The idea was really: returns true only for "true"; other values return false. Will update the comments to make more sense

patrickgmayday:
> General feedback: Make sure you fully fill out the PR check list so moving the description of your changes to where it says "Replace me," marking any security implications, checking off that you considered the security implications and created tested if needed, and how you validated your changes.

Made sure to fill those in with a description and testing notes, indicating how the changes were tested

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobResourceProvider.java
```diff
@@ -49,14 +52,25 @@ public Class<ExplanationOfBenefit> getResourceType() {
...
-    var eob =
-        eobHandler.find(
-            FhirInputConverter.toLong(fhirId),
-            getFilterModeForRequest(request, SamhsaSearchIntent.UNSPECIFIED));
+  public ExplanationOfBenefit find(
+      @IdParam final IdType fhirId,
+      final HttpServletRequest request,
+      final RequestDetails requestDetails) {
+
+    Optional<Boolean> includeTaxNumbers =
```
aschey-forpeople:
```suggestion
    var includeTaxNumbers =
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobResourceProvider.java
```diff
@@ -87,12 +102,21 @@ public Bundle searchByPatient(
       @OptionalParam(name = TYPE) final TokenAndListParam type,
       @OptionalParam(name = Constants.PARAM_SOURCE) final TokenAndListParam source,
       @OptionalParam(name = Constants.PARAM_SECURITY) final TokenAndListParam security,
+      final RequestDetails requestDetails,
       final HttpServletRequest request) {
 
+    Optional<Boolean> includeTaxNumbers =
```
aschey-forpeople:
```suggestion
    var includeTaxNumbers =
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobResourceProvider.java
```diff
@@ -122,12 +147,23 @@ public Bundle searchById(
       @OptionalParam(name = SERVICE_DATE) final DateRangeParam serviceDate,
       @OptionalParam(name = ExplanationOfBenefit.SP_RES_LAST_UPDATED)
           final DateRangeParam lastUpdated,
+      final RequestDetails requestDetails,
       final HttpServletRequest request) {
+
+    Optional<Boolean> includeTaxNumbers =
```
aschey-forpeople:
```suggestion
    var includeTaxNumbers =
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimInstitutionalBase.java
```diff
@@ -114,10 +114,12 @@ private void addInsurance(ExplanationOfBenefit eob) {
   }
 
   private void addClaimItems(ExplanationOfBenefit eob) {
+    var options = ClaimFilterOptions.builder().build();
```
aschey-forpeople: Should this use the `ClaimFilterOptions` object passed in from the request instead?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineBase.java
```diff
@@ -9,7 +10,7 @@ interface ClaimLineBase {
 
   Optional<Observation> toFhirObservation(int bfdRowId);
 
-  Optional<ExplanationOfBenefit.ItemComponent> toFhirItemComponent();
+  Optional<ExplanationOfBenefit.ItemComponent> toFhirItemComponent(ClaimFilterOptions options);
```
aschey-forpeople, line 13: It doesn't look like any of the implementations of this method use the new parameter so it can probably be removed for now.

patrickgmayday: > It doesn't look like any of the implementations of this method use the new parameter so it can probably be removed for now.

I was taking a look at refactoring this back to the normal implementation, and it seems like we hit a snag in 
ClaimLineProfessionalBase => line 92 -    ` getFhirExtensions(options).forEach(line::addExtension);`
expects options -

This is from ClaimLineProfessionalExtensions.java =>

toFhir - 

```
List<Extension> toFhir(ClaimFilterOptions options) {
    return Stream.of(
            supplierTypeCode.map(ClaimSupplierTypeCode::toFhir),
            federalTypeOfServiceCode.map(ClaimFederalTypeOfServiceCode::toFhir),
            paymentCode.map(ClaimPaymentCode::toFhir),
            providerSpecialtyCode.map(ProviderSpecialtyCode::toFhirExtension),
            serviceDeductibleCode.map(ClaimServiceDeductibleCode::toFhir),
            options.isIncludeTaxNumber()
                ? claimTaxNumberCode.toFhir()
                : Optional.<Extension>empty())
        .flatMap(Optional::stream)
        .toList();
  }
```
I think it may require some heavier refactoring at that point.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimInstitutionalBase.java
patrickgmayday: > Should this use the `ClaimFilterOptions` object passed in from the request instead?

good catch!

apps/bfd-db-migrator-ng/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>gov.cms.bfd</groupId>
         <artifactId>bfd-parent</artifactId>
-        <version>2.249.0-SNAPSHOT</version>
+        <version>2.248.0-SNAPSHOT</version>
```
marshallweekley-eng: Was there a bad merge into the branch? This number if changing most likely should be going up not down. If a new version is needed, it should probably by 2.250.0-SNAPSHOT. There are no changes in this module. I previous PR probably changed it and this change needs to be reverted. Look at the other poms also.  It looks like its reverting a few changes.

patrickgmayday: > Was there a bad merge into the branch? This number if changing most likely should be going up not down. If a new version is needed, it should probably by 2.250.0-SNAPSHOT. There are no changes in this module. I previous PR probably changed it and this change needs to be reverted. Look at the other poms also. It looks like its reverting a few changes.

Let me try and figure out how that happened and undo all of those unrelated changes.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimLineBase.java
aschey-forpeople: Oh that's fine then. We don't need to refactor that, I just missed that part.

PR #3156, opened 2026-06-04
BFD-4707: Eob updates for previous and next links on patient searches

marshallweekley-eng:
**JIRA Ticket:**
BFD-4707

### What Does This PR Do?

- Applies links to ExplanationOfBenefit for patient searches that that have more records that the limit allows
- Adopting _offset as the primary parameter for offsets/startindex
   - We will still respect startIndex if _offset is not provided to avoid a breaking change
   - All offsets in links will use _offset expect for self which will return whatever the server was given.

Take note of the comments to how it is determined to add the links

Look at the method used to determine if there is next. A case could be made to make it a little more explicit. I chose not to for performance reasons to avoid processing streams more than needed.

There are no security concerns with the PR. It only adds additional model detail to request reponses.

### Validation

The links in the FHIR json are now present
    
```
{
  "resourceType": "Bundle",
  "id": "06fa4107-507e-4d3a-9b1e-f4b8ff9822a1",
  "meta": {
    "lastUpdated": "2026-06-04T10:03:39.072-05:00"
  },
  "type": "searchset",
  "total": 2,
  "link": [
    {
      "relation": "next",
      "url": "http://localhost:8080/v3/fhir/ExplanationOfBenefit?patient=284934751&_count=2&startIndex=4"
    },
    {
      "relation": "previous",
      "url": "http://localhost:8080/v3/fhir/ExplanationOfBenefit?patient=284934751&_count=2"
    },
    {
      "relation": "self",
      "url": "http://localhost:8080/v3/fhir/ExplanationOfBenefit?_count=2&patient=284934751&startIndex=2"
    }
  ],
  "entry": []
 }
```


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
```diff
@@ -128,4 +134,69 @@ public static Bundle defaultBundle(Supplier<ZonedDateTime> batchLastUpdated) {
     bundle.setMeta(new Meta().setLastUpdated(DateUtil.toDate(batchLastUpdated.get())));
     return bundle;
   }
+
+  /**
+   * Returns a bundle updated with previous and next links. This method assumes you have at least
+   * one addition item in your bundle than the limit to know to add the updated next link
```
clewellyn-nava: addition -> additional?

marshallweekley-eng: Corrected comment

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
```diff
@@ -128,4 +134,69 @@ public static Bundle defaultBundle(Supplier<ZonedDateTime> batchLastUpdated) {
...
+          .addLink()
+          .setRelation(Constants.LINK_PREVIOUS)
+          .setUrl(buildLinkURL(requestDetails, previousOffset));
+    }
+    return bundle;
+  }
+
+  private static String buildLinkURL(RequestDetails requestDetails, Integer offset) {
+    String baseUrl = requestDetails.getCompleteUrl();
+    String[] urlParts = baseUrl.split("\\?", 2);
```
aschey-forpeople: Is there a URL parser or something we could maybe use? Doing manual string manipulation like this always makes me nervous about edge cases around escape characters and such.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
```diff
@@ -128,4 +134,69 @@ public static Bundle defaultBundle(Supplier<ZonedDateTime> batchLastUpdated) {
...
+      bundle
+          .addLink()
+          .setRelation(Constants.LINK_PREVIOUS)
+          .setUrl(buildLinkURL(requestDetails, previousOffset));
+    }
+    return bundle;
+  }
+
+  private static String buildLinkURL(RequestDetails requestDetails, Integer offset) {
+    String baseUrl = requestDetails.getCompleteUrl();
```
aschey-forpeople: Our code style is to prefer `var` for variable declarations.
```suggestion
    var baseUrl = requestDetails.getCompleteUrl();
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -99,7 +99,8 @@ public Bundle searchByBene(ClaimSearchCriteria criteria, SamhsaFilterMode samhsa
             () ->
                 filterSamhsaClaims(claims, samhsaFilterMode)
                     .skip(repositoryCriteria.resolveOffset())
-                    .limit(repositoryCriteria.resolveLimit())
+                    // we need to do this to know if we need include a link down stream
+                    .limit(repositoryCriteria.resolveLimit() + 1L)
```
aschey-forpeople: Would this result in one more result returned than specified in the request?

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSearchIT.java
```diff
@@ -58,6 +58,10 @@ void eobSearchById(SearchStyleEnum searchStyle) {
             .execute();
     assertEquals(1, eobBundle.getEntry().size());
     expectFhir().scenario(searchStyle.name()).toMatchSnapshot(eobBundle);
+
+    if (searchStyle == SearchStyleEnum.POST) {
```
aschey-forpeople: What's the purpose of this?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
marshallweekley-eng: Addressed in change to spring url parser

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
marshallweekley-eng: Changed to move to the Spring url parser

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
marshallweekley-eng: No. The FhirUtil.applyLinks method that builds the links looks for that extra record removes it and the apps the next link. It was done this way to try to limit the number of times a stream needs to be processed.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSearchIT.java
marshallweekley-eng: This might be an educational thing on for me. I ran 

```
mvn -pl bfd-server-ng -Dtest=EobSearchIT#eobSearchByPatientLimitOffset test -DupdateSnapshot=
```
That regenerated the snapshot and then added that new part to the snapshot test code. Being it was generated I did not removed it feeling that it could upset the tests and generator.

marshallweekley-eng: I have resolved this and removed it

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
marshallweekley-eng: I made some changes to make this a little more explicit to what is going on here to make it a little clearer to the intent

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -98,11 +103,18 @@ public Bundle searchByBene(ClaimSearchCriteria criteria, ClaimFilterOptions opti
             () ->
                 filterSamhsaClaims(claims, options.getSamhsaFilterMode())
                     .skip(repositoryCriteria.resolveOffset())
-                    .limit(repositoryCriteria.resolveLimit())
+                    .limit(repositoryCriteria.resolveLimit(true))
```
aschey-forpeople: This is more of a nitpick, but I try to avoid boolean params in method calls like this because it's not possible to know what `true` means in `resolveLimit(true)` since there's no context in the method name. Perhaps something like `resolveLimitWithExtra(1)` might help with readability?

marshallweekley-eng: I made this change

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/ClaimSearchCriteria.java
```diff
@@ -42,7 +42,17 @@ public Integer resolveOffset() {
...
+  }
+
+  /**
+   * Returns the limit or the default.
+   *
+   * @param extra extra to add for pagination checking than the requested limit
+   * @return limit
+   */
+  public Integer resolveLimitWithExtra(Integer extra) {
+    return limit.orElse(5000) + (extra == null ? 0 : extra);
```
aschey-forpeople: If you use `int` instead of `Integer` here, we should be able to prevent the possibility of a null value sneaking in here. In general, we try to use the type system to prevent null values outside of a few special places like input validation and left joins - you can see `DESIGN.md` for more info.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
```diff
@@ -112,9 +164,37 @@ public static Bundle bundleWithFullUrls(
...
+        resources
+            .map(r -> new Bundle.BundleEntryComponent().setResource(r))
+            .collect(
+                Collectors.teeing(
+                    Collectors.toList(),
+                    Collectors.counting(),
+                    // collecting and counting to see if we do have a next
+                    // limits the stream to only return the requested limit
+                    (list, count) ->
+                        new Page(
```
aschey-forpeople, line 185: Might be good to break this logic into a helper function since there's quite a few repeated method calls here and having all of this inline makes it a bit difficult to figure out what these two arguments are doing.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/ClaimSearchCriteria.java
marshallweekley-eng: Makes sense. Updated

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
marshallweekley-eng: Created two new helper methods to make this a little more clear

trimEntriesToLimit
determineHasMore

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobSearchIT.snap
```diff
@@ -15283,7 +15286,7 @@ gov.cms.bfd.server.ng.EobSearchIT.eobSearchByPatientLimitOffset[offset1]=[
       }, {
         "amount" : {
           "currency" : "USD",
-          "value" : 0.0
+          "value" : 8101.52
```
aschey-forpeople, line 15289: Hmm why are these changing now?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/FhirUtil.java
aschey-forpeople: Looks good, thanks 👍

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobSearchIT.snap
marshallweekley-eng: That is from the offset now working so that one test is now a different resource

aschey-forpeople: Ah yes, forgot about that. Thanks!

aschey-forpeople (dismissed):

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobResourceProvider.java
```diff
@@ -123,12 +120,14 @@ public Bundle searchByPatient(
             FhirInputConverter.toDateTimeRange(serviceDate),
             FhirInputConverter.toDateTimeRange(lastUpdated),
             Optional.ofNullable(count),
-            FhirInputConverter.toIntOptional(startIndex),
+            // we will support both offset and startIndex for now, but they can't be used together.
+            // If both are provided, offset will take precedence
+            offset != null ? Optional.of(offset) : FhirInputConverter.toIntOptional(startIndex),
```
mel1-G: This is just an opinion since I don't like null checks but you can also do `Optional.ofNullable(offset).or(() -> FhirInputConverter.toIntOptional(startIndex));` Feel free to disregard though.

marshallweekley-eng: The code based has been written in a way that prefers to use optionals over null checks. I will make the change to stick with the same style.

mel1-G (approved):

aschey-forpeople (approved):

PR #3164, opened 2026-06-09
BFD-4761: Replace CW log metric filters with custom Micrometer metrics

mel1-G:
**JIRA Ticket:**
BFD-4761

### What Does This PR Do?

Some context: If we decide to move our metrics to DataDog, we won't be able to transfer/replicate our log metric filters since we can't push up logs. Hence:

- Removed all CW log metric filters (request counts + latency & response status codes 4xx + 5xx) and replaced them with custom micrometer metrics
- Consolidated more of the metrics collection. See RequestTelemetryLogger & MetricRecorder
- Fixed a bug that Austin caught. The metric `application.eob.handler.samhsa_filter` wasn't actually measuring the time that the SAMHSA filtering takes since I had not included the actual execution, just the stream :pensive:

### Validation

Applied to test and verified that the log metric filters are gone and the dashboard is able to pull in the new custom metrics


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -93,28 +90,26 @@ public Bundle searchByBene(ClaimSearchCriteria criteria, SamhsaFilterMode samhsa
...
-        metricTimer.recordMetric(
-            "application.eob.handler.search_by_bene",
-            () ->
-                filterSamhsaClaims(claims, samhsaFilterMode)
-                    .skip(repositoryCriteria.resolveOffset())
-                    .limit(repositoryCriteria.resolveLimit())
-                    .map(claim -> transformToFhir(claim, samhsaFilterMode)),
+    var bundle =
+        metricRecorder.recordMetric(
+            "application.eob.handler.samhsa_filter",
```
aschey-forpeople: Maybe something like `application.eob.handler.transform` would be more appropriate? Since this encompasses not just the SAMHSA filter but all of the FHIR transform logic.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryLogger.java
```diff
@@ -169,23 +176,52 @@ public void logRequestException(BaseServerResponseException exception) {
...
+
+      metricRecorder.recordDuration(
+          OVERALL_REQUEST_LATENCY_METRIC, duration, TimeUnit.MILLISECONDS);
+
+      var uri = getMdcValue(HTTP_ACCESS_REQUEST_URI);
+      var certificateAlias = getMdcValue(CERTIFICATE_ALIAS);
+
+      if (uri.isPresent() && certificateAlias.isPresent()) {
+        var resourceEndpoint =
+            uri.get().startsWith("/v3/fhir/") ? uri.get().substring(9).split("/", 2)[0] : "unknown";
```
aschey-forpeople: I believe Spring Boot has some kind of URL parser that might be a bit more robust than hardcoded substrings here.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/MetricRecorder.java
```diff
@@ -33,8 +31,79 @@ public class MetricTimer {
...
+
+  /**
+   * Records a duration in a timer metric.
+   *
+   * @param metricName the metric name
+   * @param duration the recorded duration of an operation
+   * @param unit the timed unit
+   * @param tags the tags to dimension the metric by
+   */
+  public void recordDuration(String metricName, long duration, TimeUnit unit, String... tags) {
```
aschey-forpeople, line 88: Nice, I think it's a good idea to keep all the micrometer-specific stuff in a single place 👍

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/RequestTelemetryLogger.java
mel1-G: I checked it out and I saw `ServletUriComponentsBuilder` but then I realized that we're already storing the requested resource in the MDC so I ended switching to that. `metadata` isn't a FHIR resource which means it doesn't get put in the MDC so I accounted for that. Lmk what you think

aschey-forpeople: Hmm are you sure? `SERVER_INCOMING_REQUEST_PRE_HANDLED` should fire for every endpoint managed by HAPI FHIR which means it should be added there.

aschey-forpeople: Oh it's because it's Metadata is an operation but not a resource. You could use `requestDetails.getOperation()` instead maybe?

aschey-forpeople: Oof okay, this is complicated actually. Resource and operation are both optional, but at least one should be populated. In the case of `Patient/$generate-insurance-card`, both are populated though. I think we want to combine both in a single key. Something like this maybe?

```java
Stream.of(requestDetails.getResourceName(), requestDetails.getOperation())
        .filter(Objects::nonNull)
        .collect(Collectors.joining("/"));
```

mel1-G: Done in [ceca5c1](https://github.com/CMSgov/beneficiary-fhir-data/pull/3164/commits/ceca5c174aa5ce6a12266092b5365309664414a6) Copying Mitch's style :sparkles:

ops/services/05-server-ng-metrics/templates/test.json.tftpl
```diff
@@ -0,0 +1,473 @@
+{
```
aschey-forpeople: Was this supposed to be committed?

mel1-G: oppsie :pensive:

aschey-forpeople (dismissed):

aschey-forpeople (approved):

marshallweekley-eng (approved):

malessi (approved):

PR #3173, opened 2026-06-16
BFD-4641: Add AuditEvent Resource Endpoint

marshallweekley-eng:
**JIRA Ticket:**
BFD-4641

### What Does This PR Do?

Add AuditEvent Resource Endpoint

- Create DynamoDB model for AuditEvent
- Create Repository Object to query and save from dynamoDB
- Create Handler for lookups for the AuditEvent endpoint
- Create Java mapping for Audit DB to FHIR

- Pay special attention to DynamoDB models and interactions.
- Id generation for the resources.

Make sure we are good with the Dynamo query patterns as it is new to the code base.

This is used to query dynamoDB. It has no known CVEs
```
<dependency>
    <!-- Used for object mapping between DynamoDB items and Java models -->
    <groupId>software.amazon.awssdk</groupId>
    <artifactId>dynamodb-enhanced</artifactId>
</dependency>
``` 

* [ x] I have considered the above security implications as it relates to this PR. (If one or more of the above apply, it cannot be merged without the ISSO or team security engineer's (`@sb-benohe`) approval.) 
* [ x] I have created tests to sufficiently ensure the reliability of my code, if applicable. If this is a modification to an existing piece of code, I have audited the associated tests to ensure everything works as expected.

### Validation

A mapped audit event resource.

```
{
  "resourceType": "AuditEvent",
  "id": "353816020-20260611204033828277000",
  "meta": {
    "profile": [
      "http://hl7.org/fhir/StructureDefinition/AuditEvent"
    ]
  },
  "type": {
    "system": "http://terminology.hl7.org/CodeSystem/iso-21089-lifecycle",
    "code": "transmit"
  },
  "subtype": [
    {
      "system": "urn:cms:hte",
      "code": "patient-match-request"
    }
  ],
  "action": "E",
  "recorded": "2026-06-11T15:40:33.828-05:00",
  "outcome": "0",
  "purposeOfEvent": [
    {
      "coding": [
        {
          "system": "http://terminology.hl7.org/CodeSystem/v3-ActCode",
          "code": "PATRQT",
          "display": "patient requested"
        }
      ],
      "text": "patient requested"
    }
  ],
  "agent": [
    {
      "altId": "bbapi",
      "name": "bbapi",
      "requestor": true,
      "network": {
        "address": "192.168.4.27",
        "type": "2"
      }
    }
  ],
  "source": {
    "observer": {
      "display": "CMS Beneficiary FHIR Data Server"
    }
  },
  "entity": [
    {
      "what": {
        "reference": "Patient/353816020"
      },
      "type": {
        "system": "http://hl7.org/fhir/resource-types",
        "code": "Patient",
        "display": "Patient"
      },
      "role": {
        "system": "http://terminology.hl7.org/CodeSystem/object-role",
        "code": "1",
        "display": "Patient"
      },
      "name": "Match Info",
      "description": "The detail elements refer to the matching algorithm + version combination that resulted in a unique match.",
      "detail": [
        {
          "type": "matchAlgorithm",
          "valueString": "08"
        },
        {
          "type": "matchAlgorithmVersion",
          "valueString": "3.2.2"
        }
      ]
    }
  ]
}
```


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
```diff
@@ -0,0 +1,97 @@
...
+  private static void validateTimestampToken(String timestampToken) {
+    if (!timestampToken.chars().allMatch(Character::isDigit)
+        || timestampToken.length() != TIMESTAMP_TOKEN_LENGTH) {
+      throw new IllegalArgumentException("Invalid timestamp token format");
+    }
+  }
+
+  private static Instant parseTimestampTokenToInstant(String timestampToken) {
+    try {
+      var year = Integer.parseInt(timestampToken.substring(0, 4));
```
aschey-forpeople: Would it be possible to use the timestamp string directly as it is coming out of dynamo rather than parsing it to/from an `Instant`? Looks like the only place that calls this immediately converts it back to a string again.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/PatientMatchIT.java
```diff
@@ -541,6 +546,175 @@ void requestWithMissingHeadersFails() {
     assertThrows(InvalidRequestException.class, res::execute);
   }
 
+  private IQuery<Bundle> searchAuditBundle() {
+    return getFhirClient().search().forResource(AuditEvent.class).returnBundle(Bundle.class);
+  }
+
+  @Test
+  void testAuditEventByBene() {
```
aschey-forpeople, line 557: Might be nice to have one or two snapshot tests here just to get a record of what the endpoint response looks like.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
marshallweekley-eng: This issue is a timestamp cannot go to the id node of FHIR. It will want pass the regex for validation. They only allow . and - as valid on letter/number chars. The key in dyanmo is {beneSk,timestamp}. I think we are stuck transforming it in some way. 
```
[A-Za-z0-9\-\.]{1,64}
```
https://hl7.org/fhir/R4/datatypes.html#id

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/PatientMatchIT.java
marshallweekley-eng: I will work on this. Currently we let calls to the endpoint populate the dynamo image for tests. I will inflate some known records we can snap off of.

alex-dzeda: Thank you! when reviewing changes to the server they're often the first thing I instinctively look at.

marshallweekley-eng: Added snaps.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventBase.java
```diff
@@ -0,0 +1,201 @@
...
+    return eventId;
+  }
+
+  /**
+   * Builds a FHIR AuditEvent from the Dynamo Object.
+   *
+   * @return maps to a FHIR AuditEvent resource for API response
+   */
+  public final AuditEvent toFhir() {
+    AuditEvent ae = new AuditEvent();
```
aschey-forpeople: Our convention is to prefer `var` where possible
```suggestion
    var ae = new AuditEvent();
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventRepository.java
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * @param id ID of resource to pull
+   * @return Audit Event
+   */
+  public AuditEvent findById(AuditEventId id) {
+    try {
+      var key = Key.builder().partitionValue(id.beneId()).sortValue(id.toDynamoSortKey()).build();
+      var item = getTable().getItem(key);
+
+      return item != null ? item.toFhir() : null;
```
aschey-forpeople: We try our best to avoid `null`. This could return `Optional<AuditEvent>` instead I think.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventBase.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   * Generates AuditEvent Resource ID.
+   *
+   * @return resource id
+   */
+  public final AuditEventId getAuditId() {
+    if (eventId == null) {
+      eventId = AuditEventId.fromDynamoTimestamp(getMatchedBeneSk(), getTimestamp());
```
aschey-forpeople: Would it be possible to set this in the constructor?

marshallweekley-eng: The way dynamo builds these it calls the setters on the rows from the table mapping definition. Those values are not set yet. We could just return a new object every time or calculate it whether we can whenever ever the other setters are called.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventRepository.java
marshallweekley-eng: I have removed the null check.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/PatientMatchIT.snap
```diff
@@ -96,4 +96,559 @@ gov.cms.bfd.server.ng.PatientMatchIT.patientMatchSnapshot=[
...
+  "id" : "00000000-0000-0000-0000-000000000000",
+  "link" : [ {
+    "relation" : "self",
+    "url" : "http://localhost/v3/fhir/AuditEvent/_search"
+  } ],
+  "meta" : {
+    "lastUpdated" : "9999-12-31T00:00:00.000+00:00"
+  },
+  "resourceType" : "Bundle",
+  "total" : 1,
```
alex-dzeda, line 179: This seems interesting, no? I suspect it may not be directly related to this PR. I'd expect this to be 0? (Especially contrasted to the next example, where there is actually an entry)

marshallweekley-eng: Good catch let me check on that.  I was providing a last update of now() because we do not have a last updated like we do on other resources. They will never be edited for grabbing them from then.

marshallweekley-eng: I did update the test for one of them.  That timestamp thing is in all the test snapshots. It does not do it in reality. Not sure how much effort we would want to spend here.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventBase.java
```diff
@@ -0,0 +1,201 @@
...
+    ae.setSource(
+        new AuditEvent.AuditEventSourceComponent()
+            .setObserver(new Reference().setDisplay("CMS Beneficiary FHIR Data Server")));
+    ae.addPurposeOfEvent()
+        .setText("patient requested")
+        .addCoding()
+        .setDisplay("patient requested")
+        .setCode("PATRQT")
+        .setSystem(SystemUrls.SAMHSA_ACT_CODE_SYSTEM_URL);
+    AuditEvent.AuditEventEntityComponent entityComponent = ae.addEntity();
```
aschey-forpeople:
```suggestion
    var entityComponent = ae.addEntity();
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventBase.java
```diff
@@ -0,0 +1,201 @@
...
+        .setValue(new StringType(getFinalDetermination()));
+    entityComponent
+        .addDetail()
+        .setType("matchAlgorithmVersion")
+        .setValue(
+            new StringType(
+                Optional.ofNullable(getMatchAlgorithmVersion())
+                    .orElse(CURRENT_MATCH_ALGORITHM_VERSION)));
+
+    AuditEvent.AuditEventAgentComponent agent = ae.addAgent();
```
aschey-forpeople:
```suggestion
    var agent = ae.addAgent();
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventRepository.java
```diff
@@ -0,0 +1,153 @@
...
+                  QueryConditional.keyEqualTo(
+                      Key.builder().partitionValue(criteria.beneSkId()).build()))
+              .scanIndexForward(false)
+              .limit(criteria.resolveLimitWithExtra(1));
+
+      criteria
+          .lastIndex()
+          .ifPresent(
+              lastIndex -> {
+                AuditEventId id = AuditEventId.parse(lastIndex);
```
aschey-forpeople:
```suggestion
                var id = AuditEventId.parse(lastIndex);
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventBase.java
marshallweekley-eng: Done

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventRepository.java
marshallweekley-eng: Done

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventBase.java
aschey-forpeople: Yeah I think it'd be fine to just recalculate it on the fly since this is only called once per event.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @param beneId patient ID
+ * @param timestampToken canonical UTC timestamp token (yyyyMMddHHmmssnnnnnnnnn)
+ */
+public record AuditEventId(Long beneId, String timestampToken) {
+
+  private static final DateTimeFormatter TIMESTAMP_TOKEN_FORMATTER =
+      DateTimeFormatter.ofPattern("yyyyMMddHHmmssnnnnnnnnn");
+
+  private static final int TIMESTAMP_TOKEN_LENGTH = 23;
```
aschey-forpeople: nit: we could use the length of the pattern here instead of hardcoding this.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
```diff
@@ -0,0 +1,97 @@
...
+   *
+   * @param id string
+   * @return id object
+   */
+  public static AuditEventId parse(String id) {
+    var isNegative = id.startsWith("-");
+    if (isNegative) {
+      id = id.replaceFirst("-", "");
+    }
+    String[] split = id.split("-", 2);
```
aschey-forpeople:
```suggestion
    var split = id.split("-", 2);
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventResourceProvider.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Returns a {@link AuditEvent} by its ID.
+   *
+   * @param fhirId FHIR ID
+   * @return AuditEvent
+   */
+  @Read
+  public AuditEvent find(@IdParam final IdType fhirId) {
```
aschey-forpeople, line 45: Do we actually need this endpoint? Seems like we could get rid of a bit of parsing code without it.

marshallweekley-eng: The jira does not specifically call for it. I think the FHIR standard says if we support an endpoint we are suppose to have read by ID. @alex-dzeda  Thoughts?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
marshallweekley-eng: Good suggestion. Done.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
marshallweekley-eng: Done. I will work on using var more. A new habit I need to form.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventBase.java
marshallweekley-eng: Adjusted to recalualting.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventResourceProvider.java
aschey-forpeople: Yeah, that does seem to be true. It's a bit annoying because historically the read endpoints seem to go unused so we end up adding a good bit of extra code that no one uses.

marshallweekley-eng: Completely agree. In this case it is probably a low used to never used endpoint for sure. With certifications I know we generally were required to support /{id} and ?_id={id}. I supported them out of habit.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/PatientMatchIT.snap
aschey-forpeople: The timestamps are replaced with the sentinel value here because otherwise they'd be changed every time we regenerated the snapshots.

aschey-forpeople: I believe the `total` discrepancy is because calling `setEntry` doesn't automatically update the `total` field. You have to call `setTotal` as well.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventHandler.java
```diff
@@ -0,0 +1,86 @@
...
+      // DynamoDB's exclusiveStartKey starts *after* this key, so the next request
+      // will correctly begin on the very next record.
+      bundle.setEntry(bundle.getEntry().subList(0, limit));
+      bundle
+          .addLink()
+          .setRelation(Constants.LINK_NEXT)
+          .setUrl(
+              buildLinkURL(requestDetails, bundle.getEntry().getLast().getResource().getIdPart()));
+    }
+    bundle.setTotal(bundle.getEntry().size());
```
aschey-forpeople: Maybe it would be good to have a method in `FhirUtil` that calls both `setEntry` and `setTotal` to ensure we don't repeat this issue in the future?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * Parse an Audit id string to an AuditId object.
+   *
+   * @param id string
+   * @return id object
+   */
+  public static AuditEventId parse(String id) {
+    var isNegative = id.startsWith("-");
```
aschey-forpeople: Rather than having special logic for negative IDs, maybe we could use `lastIndexOf("-")` to split the timestamp from the ID?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
```diff
@@ -0,0 +1,96 @@
...
+    if (!timestampToken.chars().allMatch(Character::isDigit)
+        || timestampToken.length() != TIMESTAMP_FORMAT.length()) {
+      throw new IllegalArgumentException("Invalid timestamp token format");
+    }
+  }
+
+  private static Instant parseTimestampTokenToInstant(String timestampToken) {
+    try {
+      var year = Integer.parseInt(timestampToken.substring(0, 4));
+      var month = Integer.parseInt(timestampToken.substring(4, 6));
```
aschey-forpeople: We could maybe use a `StringBuffer` to reduce the dependence on hardcoded character indexes. Something like this?

```java
  private static Integer parseNextChars(StringBuffer buffer, int numChars) {
    var res = Integer.parseInt(buffer.substring(0, numChars));
    buffer.delete(0, numChars);
    return res;
  }

  private static Instant parseTimestampTokenToInstant(String timestampToken) {
    try {
      var buffer = new StringBuffer(timestampToken);
      var year = parseNextChars(buffer, 4);
      var month = parseNextChars(buffer, 2);
      ...
}
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventHandler.java
marshallweekley-eng: I moved this to the bundle. I left part in the wrapper than is there if there is a next link. This did improve bundle total counts on other resources but it did cause the regeneration of quite a few snaps.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
marshallweekley-eng: Done.  Good suggestion. Code is much cleaner now.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventId.java
marshallweekley-eng: Done. Cleaner approach.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventHandler.java
marshallweekley-eng: This has been changed to omit bundles totals when we have a next link to avoid bad totals on those.

aschey-forpeople (approved):

patrickgmayday (approved):

clewellyn-nava (approved):
LGTM, the little ae variable that could.

PR #3177, opened 2026-06-17
BFD-4763: Support pulling prior auth data in pipeline

mel1-G:
**JIRA Ticket:**
BFD-4763

### What Does This PR Do?

- Modifies the pipeline to now pull in prior auth data based off a feature flag since real prior auth data cannot be pulled into prod just yet
- Update schema and models for the new fields

### Validation

- Updated test data and pipeline tests + ran locally with the feature flag enabled/disabled to verify if data was loaded/not loaded. 
- Generated and pushed the synthetic prior auth data to all environments


apps/bfd-pipeline-idr/model/idr_prior_auth.py
```diff
@@ -0,0 +1,98 @@
...
+    @override
+    @staticmethod
+    def model_type() -> ModelType:
+        return ModelType.PRIOR_AUTH
+
+    @override
+    @classmethod
+    def fetch_query(cls, partition: LoadPartition, start_time: datetime, source: Source) -> str:
+        # Prior auth data older than 2 years should be filtered
+        prior_auth_cutoff_date = start_time - relativedelta(years=2)
```
aschey-forpeople: Might be worth putting this lookback period with the rest of the constants somewhere.

apps/bfd-pipeline-idr/mock-idr.sql
```diff
@@ -805,3 +820,42 @@ CREATE TABLE cms_vdm_view_mdcr_prd.v2_mdcr_cntrct_pbp_sgmt(
     cntrct_pbp_sgmt_num VARCHAR(3) NOT NULL,
     PRIMARY KEY(cntrct_pbp_sk, cntrct_pbp_sgmt_num)
 );
+
+CREATE TABLE cms_edp_view_cvm_prau_prd.prauc (
```
aschey-forpeople, line 824: We'll need to make a new migration in `bfd-db-migrator-synthetic` to add this to our snowflake dev instance too

apps/bfd-pipeline-idr/extractor.py
```diff
@@ -52,7 +52,7 @@ def cols_str(self) -> str:
         return ",".join(self.cols)
 
     def full_table(self) -> str:
-        return f"{IDR_PREFIX}.{self.table}"
+        return f"{self.table}"
```
aschey-forpeople: This seems redundant now

apps/bfd-pipeline-idr/pipeline_stages.py
```diff
@@ -106,18 +109,15 @@ def _stage1_do_bene_overshare_mbi(self) -> Stage[bool]:
...
-                *_CLAIM_TABLES,
-                *_BENE_TABLES,
-            ]
-            if self.load_type == LoadType.INITIAL
-            else [*_CLAIM_AUX_TABLES, *_BENE_AUX_TABLES]
-        )
+        tables = [*_CLAIM_AUX_TABLES, *_BENE_AUX_TABLES]
+        if self.load_type == LoadType.INITIAL:
+            tables.extend([*_CLAIM_TABLES, *_BENE_TABLES])
+            if enable_prior_auth_ingestion():
```
aschey-forpeople: Looks like this will only load prior auth if load type is `initial`. Is that intended?

mel1-G: Isn't that true since we're holding off on modifying the incremental load for prior auth? I still have the separate flag for prior auth that's under here to check

apps/bfd-pipeline-idr/extractor.py
mel1-G: What do you mean? I had to remove `IDR_PREFIX` since that's only `cms_vdm_view_mdcr_prd` and the new view that we're pulling in is instead under `cms_edp_view_cvm_prau_prd`. I got away with that by getting rid of this hard coded `IDR_PREFIX` here and modifying `apps/bfd-pipeline-idr/load_synthetic.py` to use `full_table` instead of `sql_table` since `full_table` will already have the corresponding prefix.

apps/bfd-pipeline-idr/pipeline_stages.py
aschey-forpeople: It should still work for incremental loads since we have the insert/update timestamp included. We may have to adjust later for prod, but we still want to be able to use it for synthetic data/testing.

apps/bfd-pipeline-idr/extractor.py
aschey-forpeople: I mean this method is redundant since it just returns `self.table`. You can just access the field directly. Also no need for a f string when there's nothing to format.

apps/bfd-pipeline-idr/model/idr_prior_auth.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @override
+    @staticmethod
+    def model_type() -> ModelType:
+        return ModelType.PRIOR_AUTH
+
+    @override
+    @classmethod
+    def fetch_query(cls, partition: LoadPartition, start_time: datetime, source: Source) -> str:
+        # Prior auth data older than 2 years should be filtered
```
aschey-forpeople: Might be good to keep this generic since we're not hardcoding 2 here
```suggestion
        # Prior auth data older than the lookback period should be filtered
```

apps/bfd-pipeline-idr/load_events.py
```diff
@@ -84,6 +85,7 @@
         IdrBeneficiary,
         IdrBeneficiaryMaPartDEnrollment,
         IdrBeneficiaryMaPartDEnrollmentRx,
+        IdrPriorAuth,
```
aschey-forpeople, line 90: Can we add a TODO comment noting that this is likely a temporary solution until we figure out how to load this in prod?

apps/bfd-pipeline-idr/pipeline.py
```diff
@@ -58,13 +58,28 @@
     help="Load type - affects claim filtering",
 )
 @click.option("--seed-from", type=click.Path(exists=True, resolve_path=True))
-def main(source: Source, load_mode: LoadMode, load_type: LoadType, seed_from: str | None) -> None:
+@click.option(
+    "--truncate-mode",
```
aschey-forpeople: Maybe this could just be a boolean flag since there are only two choices? `--truncate` maybe

apps/bfd-pipeline-idr/pipeline.py
```diff
@@ -58,13 +58,28 @@
...
+    type=click.Choice(["reset", "append"]),
+    default="reset",
+    show_default=True,
+    help="reset = truncate tables before loading (default), append = do not truncate",
```
aschey-forpeople: I would make the default behavior to not truncate. Data loss should be opt-in generally.

apps/bfd-pipeline-idr/load_synthetic.py
aschey-forpeople, line 170: We should probably support the new `truncate` argument here as well

apps/bfd-pipeline-idr/model/idr_prior_auth.py
```diff
@@ -0,0 +1,96 @@
...
+    @staticmethod
+    def model_type() -> ModelType:
+        return ModelType.PRIOR_AUTH
+
+    @override
+    @classmethod
+    def fetch_query(cls, partition: LoadPartition, start_time: datetime, source: Source) -> str:
+        # Prior auth data older than the lookback period should be filtered
+        prior_auth_cutoff_date = start_time - relativedelta(years=PRIOR_AUTH_LOOKBACK_YEARS)
+        start_time_sql = prior_auth_cutoff_date.strftime("'%Y-%m-%d %H:%M:%S'")
```
aschey-forpeople: I think this should probably use an absolute date rather than a relative one here since we're not going to be pruning old data (for now).

apps/bfd-pipeline-idr/load_synthetic.py
```diff
@@ -166,6 +172,11 @@ def _load_file(extractor: DbExecutor, src_folder: str, file: str, full_table: st
         default=default_dir,
         help="base directory to load files from",
     )
+    parser.add_argument(
+        "--truncate",
+        action="store_true",
```
aschey-forpeople, line 177: This should probably have a default set to be safe
```suggestion
        action="store_true",
        default=False,
```

apps/bfd-pipeline-idr/pipeline.py
```diff
@@ -58,13 +58,27 @@
     help="Load type - affects claim filtering",
 )
 @click.option("--seed-from", type=click.Path(exists=True, resolve_path=True))
-def main(source: Source, load_mode: LoadMode, load_type: LoadType, seed_from: str | None) -> None:
+@click.option(
```
aschey-forpeople, line 61: We should probably update the `README` to let people know about this new behavior

apps/bfd-pipeline-idr/load_synthetic.py
mel1-G: It does already[ default to false](https://docs.python.org/3/library/argparse.html#:~:text=%27store_true%27%20and%20%27store_false%27%20%2D%20These%20are%20special%20cases%20of%20%27store_const%27%20that%20respectively%20store%20the%20values%20True%20and%20False%20with%20default%20values%20of%20False%20and%20True%3A) but I'll add that so it's clear.

aschey-forpeople: Oh okay, I wasn't sure if it would do that or default to `None`. We should probably replace `argparse` with `click` like we do in `pipeline.py` but meh, I don't expect people will be invoking this file directly very often.

mel1-G: Hm "meh" is compelling enough. Alright, deferred :slay:

aschey-forpeople (approved):

marshallweekley-eng (approved):

PR #3178, opened 2026-06-18
BFD-4349: Prune older shared system phase 1 claims

JFisk42:
**JIRA Ticket:**
BFD-4349

### What Does This PR Do?
Prunes Phase 1 claims from shared systems that are older than 60 days. 

> [!CAUTION]
> **Remember, _before merging_, we need to update `idr.flyway_schema_history` in established environments to ensure Flyway migrations can run.**

The DELETE query in pipeline_utils for accuracy and performance
Ignore that the changes to pipeline_nodes do not yet reflect on @malessi 's changes

### Validation
This was tested via an initial load (after commenting out the INCREMENTAL guard clauses). 
Command used: `./run-db.sh ./test_samples1`
Logs snippet:
```
...
[INFO] 2026-06-17 17:09:37,040 INFO:hamilton.execution.state:Enqueuing 1 task: collect-stage2_inputs
[INFO] 2026-06-17 17:09:37,076 INFO:hamilton.execution.state:Enqueuing 1 task: expand-stage5_inputs
[INFO] 2026-06-17 17:09:37,114 INFO:hamilton.execution.state:Enqueuing 5 tasks: expand-stage5_inputs.0.block-stage5_inputs, expand-stage5_inputs.1.block-stage5_inputs, ...
[INFO] 2026-06-17 17:09:37,165 INFO:pipeline_utils:pruned 203 rows from idr.claim_item_institutional_ss
[INFO] 2026-06-17 17:09:37,166 INFO:pipeline_utils:pruned 14 rows from idr.claim_institutional_ss
[INFO] 2026-06-17 17:09:37,170 INFO:pipeline_utils:pruned 0 rows from idr.claim_item_professional_ss
[INFO] 2026-06-17 17:09:37,170 INFO:pipeline_utils:pruned 0 rows from idr.claim_professional_ss
[INFO] 2026-06-17 17:09:37,215 INFO:hamilton.execution.state:Enqueuing 1 task: collect-stage5_inputs
[INFO] 2026-06-17 17:09:37,246 INFO:hamilton.execution.executors:Graph is done, graph state is TaskState.SUCCESSFUL
Done loading full tables from ./test_samples1
```


apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -101,3 +110,46 @@ def extract_and_load(
         except Exception as ex:
             logger.error("error loading %s", cls.table(), exc_info=ex)
             raise ex
+
+# todo, add some logs
```
JFisk42: Todo: Remove this todo now that it is todone.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -101,3 +110,45 @@ def extract_and_load(
...
+        CLAIM_PROFESSIONAL_SS_TABLE: CLAIM_PROFESSIONAL_ITEM_SS_TABLE,
+    }
+
+    claim_table = cls.table()
+    item_table = shared_claim_tables.get(claim_table)
+    if item_table is None:
+        return True
+
+    with psycopg.connect(get_connection_string(load_mode)) as conn, conn.transaction():
+        res = conn.execute(
```
aschey-forpeople: If you look in `base_model.py`, the `claim_filter` function already has logic for determining which PAC phase 1 claims are out of meet the cutoff threshold. It would be good to reuse that logic here since the criteria should be identical.

JFisk42: I've updated it to include the `clm_src_id` logic. I didn't include the the idr timestamps or partition type from pac_filter. My understanding is that since this prune targets our db that those aren't applicable/usable to the query at all. Or are you saying that I need to bring in other parts like the bene_sk check and the deceased_bene_filter?

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -110,3 +122,49 @@ def extract_and_load(
...
+        CLAIM_INSTITUTIONAL_SS_TABLE: CLAIM_INSTITUTIONAL_ITEM_SS_TABLE,
+        CLAIM_PROFESSIONAL_SS_TABLE: CLAIM_PROFESSIONAL_ITEM_SS_TABLE,
+    }
+
+    claim_table = cls.table()
+    item_table = shared_claim_tables.get(claim_table)
+    if item_table is None:
+        return True
+
+    prune_cutoff_date = job_start - timedelta(days=60)
```
aschey-forpeople: The 60 day constant here should probably be lifted into `constants.py` since it's shared between here and `base_model.py`

apps/bfd-pipeline-idr/pipeline_utils.py
aschey-forpeople: That's true, I think it's fine as it is. I forgot that it's coming from a different db so the SQL is a bit different. I do think we're gonna need an index on `clm_idr_ld_dt` probably or else this query is gonna be really slow.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -110,3 +122,49 @@ def extract_and_load(
...
+    logger.info("pruning phase 1 ss claims older than {}", prune_cutoff_date)
+
+    with psycopg.connect(get_connection_string(load_mode)) as conn, conn.transaction():
+        res = conn.execute(
+            f"""
+                DELETE FROM {item_table}
```
aschey-forpeople: I think it'd probably be good to do these deletes in batches so we're not acquiring a lock on the table for too long without giving it a chance to be released and picked up by another process. And so we don't have to restart the transaction from scratch if it gets interrupted partway through.

apps/bfd-pipeline-idr/pipeline_utils.py
JFisk42: Moved: [Move 60 day pac cutoff to constants](https://github.com/CMSgov/beneficiary-fhir-data/pull/3178/commits/8ac93ce4f0e1170ef2314f5c4e3335ff67af470b)

apps/bfd-pipeline-idr/pipeline_utils.py
JFisk42: Added a simple LIMIT clause: https://github.com/CMSgov/beneficiary-fhir-data/pull/3178/commits/23e81e75a230a089129be9ea6f8b1f5cb828f073

apps/bfd-pipeline-idr/pipeline_utils.py
JFisk42: Added a simple index https://github.com/CMSgov/beneficiary-fhir-data/pull/3178/commits/a7c4e0990f18c740fcb8db3b3a914d397feb2af3. I'm not sure if we need to add any extra magic for our db size.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -110,3 +123,54 @@ def extract_and_load(
         except Exception as ex:
             logger.opt(exception=True).error("error loading {}", cls.table())
             raise ex
+
+
+def prune_phase_1_ss_claims(
```
marshallweekley-eng, line 127: With this method I see we limit the number of rows for the call. Is there something in the infrastructure that will make sure that it keeps deleting until all are gone. Example checking to make sure the number of rows deleted is < 1000 and breaking.

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX ON idr.claim_professional_ss (clm_idr_ld_dt);
```
marshallweekley-eng: This may need a veteran opinion but to avoid locks on an existing table I would recommend CREATE INDEX CONCUURENTLY

apps/bfd-pipeline-idr/pipeline_utils.py
aschey-forpeople: Yeah we should loop through this until the `DELETE` returns no new rows.

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
aschey-forpeople: Yep, you're correct. These tables are massive so we definitely want to prevent locks while creating indexes.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -110,3 +123,54 @@ def extract_and_load(
...
+
+        res = conn.execute(
+            f"""
+                DELETE FROM {claim_table}
+                WHERE clm_uniq_id IN (
+                    SELECT clm_uniq_id FROM {claim_table}
+                    WHERE clm_type_cd BETWEEN {PHASE_1_SS_MIN} AND {PHASE_1_SS_MAX}
+                    AND clm_src_id IN ('{FISS_CLM_SOURCE}', '{MCS_CLM_SOURCE}', '{VMS_CLM_SOURCE}')
+                    AND clm_idr_ld_dt < %s
+                    LIMIT 1000
```
aschey-forpeople: I'd probably make this limit configurable in `settings.py`

aschey-forpeople: We may want to default to something larger like 100,000

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
JFisk42: Turns out by default flyway doesn't allow for concurrent index creation by default, instead they will just quietly hang indefinitely. The fix for this involved setting the flyway transaction lock to false. My understanding is that this won't be an issue when running flyway in different environments. https://github.com/CMSgov/beneficiary-fhir-data/pull/3178/commits/07c45c80c41fb8b13563b3572cfc4c76f272a34d

apps/bfd-pipeline-idr/pipeline_utils.py
JFisk42: Added the option to supply as an envvar or else it will default to 1000.

apps/bfd-pipeline-idr/pipeline_utils.py
JFisk42: This required a few adjustments to the query. I adjusted the item query to a join so that it's the item rows being limited and not the header rows going into that query. This also ensures it's grabbing fresh header rows if needed. https://github.com/CMSgov/beneficiary-fhir-data/pull/3178/commits/5012549e6537b524a2bf631676546de3b2236c01

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
JFisk42: Summary of events/discussion:
The actions that relied on mvn-verify were failing for the same reason I had to update the other config (flyway never closing out the index creation so it just hangs). The fix _would have been_ to update the place where we do flyway config for maven (IntegrationTestConfiguration.java). 

During a huddle to find the fix for the maven issue, a different path was chosen. The consensus was that altering our flyway config was not worth any tradeoffs. So the index creation will remain as the simple version it was before. This will only be an issue for the dataset in prod. For that environment we will run the index creation manually.

marshallweekley-eng (dismissed):

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
```diff
@@ -0,0 +1,3 @@
...
+
+CREATE INDEX ON idr.claim_institutional_ss (clm_idr_ld_dt);
```
marshallweekley-eng:
```suggestion
CREATE INDEX IF NOT EXISTS ON idr.claim_institutional_ss (clm_idr_ld_dt);
```

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
marshallweekley-eng:
```suggestion
CREATE INDEX IF NOT EXISTS ON idr.claim_professional_ss (clm_idr_ld_dt);
```

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
aschey-forpeople: We can leave this change off - we'll fake the migration when we run it in prod

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX ON idr.claim_professional_ss(clm_type_cd, clm_idr_ld_dt) WHERE clm_type_cd BETWEEN 1000 and 1999
+
+CREATE INDEX ON idr.claim_institutional_ss(clm_type_cd, clm_idr_ld_dt) WHERE clm_type_cd BETWEEN 1000 and 1999
```
aschey-forpeople: Missing semicolons - also I think we can omit `clm_type_cd` from the column list since it's already in the predicate

```suggestion
CREATE INDEX ON idr.claim_professional_ss(clm_idr_ld_dt) WHERE clm_type_cd BETWEEN 1000 and 1999;

CREATE INDEX ON idr.claim_institutional_ss(clm_idr_ld_dt) WHERE clm_type_cd BETWEEN 1000 and 1999;
```

JFisk42:
Update: After discussion the logic for prunes needs to be updated. Currently the prune looks for candidates by their created date `clm_idr_ld_dt`, however it is more beneficial to prune them by their last updated timestamp `bfd_updated_ts`. This last updated check should also check the last update for claim lines.

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX ON idr.claim_professional_ss(clm_idr_ld_dt) WHERE clm_type_cd BETWEEN 1000 and 1999;
+
+CREATE INDEX ON idr.claim_institutional_ss(clm_idr_ld_dt) WHERE clm_type_cd BETWEEN 1000 and 1999;
```
marshallweekley-eng: Add a new line at the bottom.

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
```diff
@@ -0,0 +1,5 @@
+CREATE INDEX ON idr.claim_professional_ss(bfd_updated_ts) WHERE clm_type_cd BETWEEN 1000 and 1999;
+CREATE INDEX ON idr.claim_item_professional_ss(clm_uniq_id, bfd_updated_ts);
```
aschey-forpeople: I think we might be able to get away with omitting the indexes on the item table since we're joining against `clm_uniq_id` which should already be fast - adding an unconditional index on the item table will be quite large so it's best to avoid if possible.

JFisk42: Good point. And we can always test it in prod to see how efficient it is.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -115,6 +127,89 @@ def extract_and_load(
...
+    claim_table = cls.table()
+    item_table = shared_claim_tables.get(claim_table)
+    if item_table is None:
+        return True
+
+    prune_cutoff_date = job_start - timedelta(days=PHASE_1_CUTOFF)
+    logger.info("pruning phase 1 ss claims older than {}", prune_cutoff_date)
+
+    with psycopg.connect(get_connection_string(load_mode)) as conn:
+        totalRowCount = 0
```
aschey-forpeople: Variables should be `snake_case`
```suggestion
        total_row_count = 0
```

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -115,6 +127,89 @@ def extract_and_load(
...
+        while True:
+            with conn.transaction():
+                res = conn.execute(
+                    f"""
+                        DELETE FROM {item_table}
+                        WHERE (clm_uniq_id, bfd_row_id) IN (
+                            SELECT item.clm_uniq_id, item.bfd_row_id
+                            FROM {item_table} item
+                            JOIN {claim_table} clm ON clm.clm_uniq_id = item.clm_uniq_id
```
aschey-forpeople: We should be able to speed this up a good bit by removing this `JOIN`. We don't need the `bfd_row_id` since we want to delete all rows for the claim if we're pruning it.

aschey-forpeople: With this change I think the two queries will be identical aside from the `DELETE FROM {table}` part so that might be a good way to reduce some duplication here.

apps/bfd-pipeline-idr/settings.py
```diff
@@ -100,6 +100,10 @@ def enable_prior_auth_ingestion() -> bool:
 """Number of minimum connections to hold in the pool concurrently per-batch for non-LOCAL loads.
 Defaults to 20."""
 
+PHASE_1_PRUNE_BATCH_LIMIT = int(getenv("PHASE_1_PRUNE_BATCH_LIMIT", "1000"))
```
aschey-forpeople: I think this can be bumped to 10,000 at least. The query seems to be pretty quick.

apps/bfd-pipeline-idr/pipeline_utils.py
JFisk42: Ope, Java-brain leaking through 😅

apps/bfd-pipeline-idr/pipeline_utils.py
JFisk42: Removed the join clause and shuffled things around to clean up the duplication. @aschey-forpeople let me know if any placement or style issues still remain.

aschey-forpeople (approved):

marshallweekley-eng (approved):

apps/bfd-db-migrator-ng/migrations/V1_33_0__add_clm_idr_ld_dt_index.sql
aschey-forpeople: ^ you can configure your editor to do this for you btw, there's a setting in VS Code somewhere.

JFisk42: Ooops this is the older version of this migration. If you refresh GitHub you should be able to see the new "V1_35" version

JFisk42: Ohhhh this might have been an older review comment that popped up when @marshallweekley-eng added his review.

PR #3181, opened 2026-06-18
BFD-4091: Add class-value query param to Coverage Resource

marshallweekley-eng:
**JIRA Ticket:**
BFD-4091

### What Does This PR Do?
Add class-value query param to Coverage Resource.

This is straight forward. I will note that I did not implement it on _id and only on beneficiary based queries. The coverage value for BFD is tightly coupled to the class-value. It really didn't add any value and just code complexity to add it.

To filter the coverage type(class-value) this can be done simply by delimiting what type of coverage we want to filter in the coverage handler. We iterate across the coverage records to determine class type as is. It is easy enough to just limit that iteration on the stream to strip them out.

N/A 

* [ x] I have considered the above security implications as it relates to this PR. (If one or more of the above apply, it cannot be merged without the ISSO or team security engineer's (`@sb-benohe`) approval.) 
* [ x] I have created tests to sufficiently ensure the reliability of my code, if applicable. If this is a modification to an existing piece of code, I have audited the associated tests to ensure everything works as expected.

### Validation

A unit test has been added that validates it


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/coverage/CoverageRepository.java
```diff
@@ -96,23 +91,41 @@ latestLis AS (
...
                           AND (
                               CAST(:lowerBound AS ZonedDateTime) IS NULL
-                              OR GREATEST(
-                                    b.meta.partACoverageUpdatedTs,
-                                    b.meta.partBCoverageUpdatedTs,
-                                    b.meta.partCCoverageUpdatedTs,
-                                    b.meta.partDCoverageUpdatedTs,
-                                    b.meta.partDualCoverageUpdatedTs
-                                ) %s :lowerBound
+                              OR (:lowerBoundInclusive = TRUE
```
aschey-forpeople: What's the reason for this change? Is there something that makes the explicit check for inclusive/exclusive necessary now?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/CoveragePart.java
```diff
@@ -62,6 +62,28 @@ public static Optional<CoveragePart> fromExactRawPrefix(String rawUrlPrefix) {
...
+   *
+   * @param param The query param from.
+   * @return An {@link Optional} containing the matching {@link CoveragePart}, or {@link
+   *     Optional#empty()} if no exact match.
+   */
+  public static Optional<CoveragePart> parseFromQueryParam(String param) {
+    if (param == null || param.isBlank()) {
+      return Optional.empty();
+    }
+    param = param.trim().toLowerCase().replace(" ", "-").replace("_", "-");
```
aschey-forpeople: I think it might make sense to keep this string manipulation logic in `FhirInputConverter` to keep the unsanitized inputs away from the rest of the codebase. Open to other thoughts though.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/coverage/CoverageRepository.java
marshallweekley-eng: This is the change caused by SonarQube. It did not like the String.format and was throwing a critical code smell. It was sensing it as a potential SQL injection attack.

aschey-forpeople: It's not a potential SQL attack since we don't include any user inputs in the SQL string. We use dynamic query builders all over the place because of the various filter params. We can ignore sonar warnings if they're false positives.

marshallweekley-eng: I will revert the change and mark to to be ignored.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/CoveragePart.java
marshallweekley-eng: I moved this to the controller layer and the FhirInputConverter.

aschey-forpeople (dismissed):

aschey-forpeople (approved):

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/CoverageSearchCriteria.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Optional;
+
+/**
+ * Coverage Criteria to Search By.
+ *
+ * @param beneSk beneficiary id
+ * @param lastUpdated date last updated
+ * @param coveragePart Coverage Part filter
+ */
+public record CoverageSearchCriteria(
```
clewellyn-nava, line 12: Is there a generalized solution to search criteria that we could have? 

Unrelated, noticed in ClaimSearchCriteria that hasLastUpdated is mispelled "hasLasUpdated" 😢

clewellyn-nava (approved):
One comment about search criteria in general, but looks good to me otherwise.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/input/CoverageSearchCriteria.java
marshallweekley-eng: No they are per ResourceType and sometimes per Search type.

PR #3193, opened 2026-07-01
BFD-4791: Set hibernate defaults as read only

alex-dzeda:
**JIRA Ticket:**
BFD-4791

### What Does This PR Do?

Tells hibernate to treat our session / queries as read only to reduce unnecessary cpu utilization.

### Validation

Compiles, no snapshot changes, still functions.

### DIFF
--- a/apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/ClaimAsyncService.java
+++ b/apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/ClaimAsyncService.java
@@ -62,7 +62,6 @@ public class ClaimAsyncService {
                 DbFilterParam.withParams(
                         entityManager.createQuery(jpql, claimClass), filters.params())
                     .setParameter("claimUniqueIds", claimUniqueIds);
-            query.setHint("org.hibernate.readOnly", true);
             var result = queryTelemetryUtil.executeAndTrack("findByIdsInClaimType", query);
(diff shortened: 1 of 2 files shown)


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/ClaimAsyncService.java
```diff
@@ -57,10 +57,12 @@ CompletableFuture<List<C>> findByIdsInClaimType(
         () -> Tags.of(CLAIM_TYPE, claimClass.getSimpleName()),
         () -> {
           try (var entityManager = entityManagerFactory.createEntityManager()) {
+            entityManager.unwrap(org.hibernate.Session.class).setDefaultReadOnly(true);
```
aschey-forpeople: Do we need to modify both the entity manager and the query? Seems redundant.

alex-dzeda: we don't! I did the latter first, and then didn't remove it. I see the same performance with just the entity manager as entity manager + query, so removing the query

aschey-forpeople:
@alex-dzeda I moved your changes to a helper function and added the readonly annotation to the bene respository.

aschey-forpeople (approved):

marshallweekley-eng (approved):

PR #3174, opened 2026-06-16
BFD-4593 Add EOB Outcome Search

marshallweekley-eng:
**JIRA Ticket:**
BFD-4593

### What Does This PR Do?

Add EOB outcome search support

Add V3 ExplanationOfBenefit outcome search support by parsing the
outcome query parameter and applying it through claim search criteria.

### What Should Reviewers Watch For?

This is expected to address the user item by adding V3 EOB outcome
search support and validating the behavior with integration coverage.

No additional documentation updates are expected.

Test coverage includes supported outcome values, case insensitive input,
multiple outcome scenarios, and invalid outcome handling.

User input is parsed through the existing FHIR token parameter handling and
invalid outcome values return an invalid request response.

If you're reviewing this PR, please check for these things in particular:

### What Security Implications Does This PR Have?

Please indicate if this PR does any of the following:  

* Adds any new software dependencies
* Modifies any security controls
* Adds new transmission or storage of data
* Any other changes that could possibly affect security? 

This change does not add any new software dependencies, modify existing
security controls, introduce new data transmission or storage paths, or
otherwise affect the security posture of the application.

* [x] I have considered the above security implications as it relates to this PR. (If one or more of the above apply, it cannot be merged without the ISSO or team security engineer's (`@sb-benohe`) approval.) 
* [x] I have created tests to sufficiently ensure the reliability of my code, if applicable. If this is a modification to an existing piece of code, I have audited the associated tests to ensure everything works as expected.

### Validation

CURL testing - 

curl -s -H "Accept: application/fhir+json" \
"http://localhost:8080/v3/fhir/ExplanationOfBenefit?patient=$PATIENT&outcome=CoMpLeTe&_count=50" \
| jq '{entryCount: (.entry // [] | length), outcomes: ([.entry[]?.resource.outcome] | unique)}'
{
  "entryCount": 5,
  "outcomes": [
    "complete"
  ]
}

 curl -s -H "Accept: application/fhir+json" \
"http://localhost:8080/v3/fhir/ExplanationOfBenefit?patient=Patient/$PATIENT&outcome=partial&_count=50" \
| jq '{entryCount: (.entry // [] | length), outcomes: ([.entry[]?.resource.outcome] | unique)}'

{
  "entryCount": 1,
  "outcomes": [
    "partial"
  ]
}

curl -s -H "Accept: application/fhir+json" \
[...]

PR #3201, opened 2026-07-07
BFD-4764: Prior Auth Server Implementation

mel1-G:
**JIRA Ticket:**
BFD-4764

### What Does This PR Do?

- Separates prior auth into two tables: idr.prior_auth and idr.prior_auth_item
- Implements the prior authorization mappings on the server
- Supports filtering by _source=CWF. Prior Auth eobs will always be returned by default otherwise.

_MISC:_
- cleans up ClaimAuditStatusCode since we stopped relying on it to determine the outcome on an EOB in 4593: https://github.com/CMSgov/beneficiary-fhir-data/pull/3174 
- Implements the missing mappings for ClaimAuditStatusCode and ClaimPaidStatusCode
- Adds support to collect metrics on our executor
- renames the samhsa_filter metric to transform so our dashboard uses the correct metric

_Notes:_
- The data quality of prior auth is...not great. We still have some open questions and future meetings with IDR. This is why prior auth is behind a feature flag for prod. We'll only be serving synthetic data for now.
- Searching for prior auth's by id currently is not supported. This is why I short circuit relying on our SourceFilterParam. I left it out since it turns out to have a large blast radius. It can be done but as a future effort. 

Make sure that I implemented the mappings correctly!

### Validation

Integration tests


apps/bfd-model-idr/maps/ExplanationOfBenefit-PriorAuth.map
```diff
@@ -236,7 +236,7 @@ group addSvcRenderSt(source src: String, target tgt: Extension){
 
 group addMRCountInd(source src: String, target tgt: Extension){
     src -> tgt.url = "https://bluebutton.cms.gov/fhir/StructureDefinition/MR-COUNT-IND" "Add extension url";
-    src as val -> tgt.value = cast(val, "decimal") "set value";
+    src as val -> tgt.value = cast(val,'unsignedInt') "set value";
```
aschey-forpeople:
```suggestion
    src as val -> tgt.value = cast(val, 'unsignedInt') "set value";
```

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -143,17 +143,33 @@ def provider_careteam_name_expr(alias: str, type: str | None) -> str:
...
-                THEN {_normalize(f"{alias}.prvdr_lgl_name")}
-                ELSE {_normalize(f"{alias}.prvdr_last_name")}
-                    || COALESCE(', ' || {_normalize(f"{alias}.prvdr_1st_name")}, '')
-            END
+            {normalize_provider_careteam_name_expr(alias)}
         )
     """
 
 
+def normalize_provider_careteam_name_expr(alias: str) -> str:
```
aschey-forpeople: I think we can use the existing `provider_careteam_name_expr` function here rather than make a new one

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -143,17 +143,33 @@ def provider_careteam_name_expr(alias: str, type: str | None) -> str:
...
+            ELSE {_normalize(f"{alias}.prvdr_last_name")}
+                || COALESCE(', ' || {_normalize(f"{alias}.prvdr_1st_name")}, '')
+        END
+    """
+
+
+def provider_npi_type_expr(alias: str) -> str:
+    return f"""
+        CASE 
+            WHEN {alias}.prvdr_lgl_name IS NOT NULL THEN 2
```
aschey-forpeople, line 167: Is a null check adequate here or do we also need to check for `''` and `~`?

apps/bfd-pipeline-idr/pipeline_stages.py
```diff
@@ -113,7 +114,7 @@ def _stage2_do_claims_and_benes_tbls(self) -> Stage[bool]:
         if self.load_type == LoadType.INITIAL:
             tables.extend([*_CLAIM_TABLES, *_BENE_TABLES])
         if enable_prior_auth_ingestion():
-            tables.append(*_PRIOR_AUTH_TABLES)
+            tables.extend([*_PRIOR_AUTH_TABLES])
```
aschey-forpeople: This should work since this is already an array
```suggestion
            tables.extend(_PRIOR_AUTH_TABLES)
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimInstitutionalSharedSystems.java
```diff
@@ -90,7 +89,25 @@ public static SystemType getSystemType() {
...
+            claimRecordType.toFhir(supportingInfoFactory),
+            claimPaidStatusCode.map(cd -> cd.toFhir(supportingInfoFactory)),
+            buildAuditStatusSupportingInfo())
+        .flatMap(Optional::stream)
+        .toList();
+  }
+
+  private Optional<ExplanationOfBenefit.SupportingInformationComponent>
+      buildAuditStatusSupportingInfo() {
+    // since audit trail status codes can overlap between the different shares systems, we must
```
aschey-forpeople:
```suggestion
    // since audit trail status codes can overlap between the different shared systems, we must
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimProfessionalSharedSystems.java
```diff
@@ -76,12 +76,26 @@ public class ClaimProfessionalSharedSystems extends ClaimProfessionalBase {
...
+                claimPaidStatusCode.map(cd -> cd.toFhir(supportingInfoFactory)),
+                buildAuditStatusSupportingInfo()),
             buildRxSupportingInfo())
         .flatMap(Optional::stream)
         .toList();
   }
 
+  private Optional<ExplanationOfBenefit.SupportingInformationComponent>
+      buildAuditStatusSupportingInfo() {
+    // since audit trail status codes can overlap between the different shares systems, we must
```
aschey-forpeople:
```suggestion
    // since audit trail status codes can overlap between the different shared systems, we must
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimTypePriorAuth.java
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Maps the display type to a ClaimContext. It is temporarily being used in addCareTeam from
+   * ProviderHistoryBase to determine the NPI type for non-prior authorization providers. This will
+   * no longer be needed once we have separate NPI type fields for each npi in BFD-4661.
+   *
+   * @return a ClaimContext or an empty Optional if the display did not match
+   */
+  default Optional<ClaimContext> toContext() {
+    if (getDisplay().equals("institutional")) {
```
aschey-forpeople: Might be nice to use `ClaimType.INSTITUTIONAL.getCode()` here (and in the enum definition) to avoid comparing against duplicated string constants.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/OrderingOrReferringCareTeam.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  protected CareTeamType getCareTeamType() {
+    return CareTeamType.REFERRING;
+  }
+
+  @Override
+  @Transient
+  protected ProviderHistoryBase.NpiType getNpiType() {
+    if (npiType.isPresent() && npiType.get().equals(2)) {
```
aschey-forpeople: This feels like something that should live in `NpiType` so it can be reused.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/ClaimAsyncService.java
```diff
@@ -57,15 +56,13 @@ CompletableFuture<List<C>> findByIdsInClaimType(
...
+          try (var entityManager = entityManagerFactory.createEntityManager()) {
             var query =
                 DbFilterParam.withParams(
                         entityManager.createQuery(jpql, claimClass), filters.params())
                     .setParameter("claimUniqueIds", claimUniqueIds);
             var result = queryTelemetryUtil.executeAndTrack("findByIdsInClaimType", query);
-            result.stream()
-                .findFirst()
-                .ifPresent(claim -> LogUtil.logBeneSk(claim.getBeneficiary().getBeneSk()));
+            result.forEach(claim -> LogUtil.logBeneSk(claim.getBeneficiary().getBeneSk()));
```
aschey-forpeople: Probably want to dedupe these first otherwise we could get 100 of the same ID

aschey-forpeople: Maybe make this into a helper function that we could use for all the repository methods as well just in case we support querying multiple benes at a time for other endpoints in the future.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/PriorAuthorizationRepository.java
```diff
@@ -0,0 +1,37 @@
...
+   * @return all prior authorizations
+   */
+  public List<PriorAuthorization> getPriorAuthorizationFromMbi(String mbi) {
+    return entityManager
+        .createQuery(
+            """
+                 SELECT p
+                 FROM PriorAuthorization p
+                 JOIN FETCH p.beneficiary b
+                 LEFT JOIN FETCH p.items i
```
aschey-forpeople: Do we need a left join here? Every prior auth record should have at least one detail record right?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/eob/EobHandler.java
```diff
@@ -111,13 +137,23 @@ public Bundle searchByBene(
...
+                      filteredClaims,
+                      loadProgressRepository::lastUpdated,
+                      requestDetails,
+                      // we want the raw limit
+                      Optional.of(repositoryCriteria.resolveLimit()),
+                      Optional.of(repositoryCriteria.resolveOffset()));
+              var includePriorAuth =
+                  criteria.sources().isEmpty() || priorAuthSourceEvaluation.hasPriorAuthSource();
+              if (includePriorAuth) {
+                var priorAuths =
```
aschey-forpeople: I think we'll want to query prior auth in parallel with claims here rather than sequentially so we don't introduce additional latency.

apps/bfd-pipeline-idr/model/base_model.py
mel1-G: Just a null check. I looked and `prvdr_lgl_name` is never `''` nor `~ `

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimPaidStatusCode.java
```diff
@@ -1,62 +1,67 @@
...
 
   /**
    * Convert from a database code.
    *
    * @param code database code
    * @return claim paid status code
    */
-  public static Optional<ClaimPaidStatusCode> tryFromCode(String code) {
+  static Optional<ClaimPaidStatusCode> tryFromCode(String code) {
     if (code == null) {
```
aschey-forpeople: I think we need to handle empty codes here too
```suggestion
    if (code == null || code.isBlank()) {
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimTypePriorAuth.java
```diff
@@ -0,0 +1,149 @@
...
+   */
+  String getInsuranceType();
+
+  /**
+   * Convert from a database code.
+   *
+   * @param code database code
+   * @return claim type prior auth code or empty Optional if code is null or blank
+   */
+  static ClaimTypePriorAuth tryFromCode(String code) {
```
aschey-forpeople: Should this return `Optional<ClaimTypePriorAuth>` with a null/empty check?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimPaidStatusCode.java
mel1-G: I thought so too at first but looking at the codes, an empty string has meaning. It indicates a partial outcome

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimTypePriorAuth.java
mel1-G: Oof you're right. It can't be null but it can be empty.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/ClaimPaidStatusCode.java
aschey-forpeople: Ah you're right. Well we normalize null to empty string anyway so I don't think we need a valid/invalid switch here.

aschey-forpeople: Confirmed in prod that there are no nulls - only empty values

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/ClaimRepository.java
```diff
@@ -166,7 +166,7 @@ public List<ClaimBase> findByIds(
    * @return claims
    */
   @Timed(value = "application.claim.search_by_bene")
-  public List<ClaimBase> findByBeneXrefSk(
+  public ClaimAndAuthResult findByBeneXrefSk(
       @MeterTag(key = "hasClaimThroughDate", expression = "hasClaimThroughDate()")
           @MeterTag(key = "hasLastUpdated", expression = "hasLasUpdated()")
```
aschey-forpeople: Can we fix all the references to `hasLasUpdated`? I keep seeing warnings in the logs about it

aschey-forpeople: Nevermind, I'll fix it in my other PR https://github.com/CMSgov/beneficiary-fhir-data/pull/3204/changes/c5f0ba66f57b23504bf9f1461c25d65de6e839d5

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/HcpcsOrCptOrHippsCode.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    return systems;
+  }
+
+  CodeableConcept toFhir() {
+    return new CodeableConcept()
+        .setCoding(
+            getCodingSystems().stream()
+                .map(system -> new Coding().setSystem(system).setCode(hcpcsOrCptOrHipps))
```
aschey-forpeople: Am I remembering correctly that there can be multiple codes in the same string field here? Do we need to separate them out so each code has its own entry?

mel1-G: I thought about it but that doesn't seem the case at least based off the mapping:

```
paItem.HCPCS_OR_CPT_OR_HIPPS as itemCode where(itemCode.length()=5 and itemCode.substring(0,1).convertsToInteger()) -> tgtItem.productOrService = cc("http://www.ama-assn.org/go/cpt", itemCode) "Add item code";
paItem.HCPCS_OR_CPT_OR_HIPPS as itemCode where(itemCode.length()=5 and itemCode.substring(0,1).convertsToInteger().not()) -> tgtItem.productOrService = cc("https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets", itemCode) "Add item code";
paItem.HCPCS_OR_CPT_OR_HIPPS as itemCode where(itemCode.length()=5 and itemCode.substring(1,2).convertsToInteger().not()) -> tgtItem.productOrService = cc("https://www.cms.gov/Medicare/Medicare-Fee-for-Service-Payment/ProspMedicareFeeSvcPmtGen/HIPPSCodes", itemCode) "Add item code";
paItem.HCPCS_OR_CPT_OR_HIPPS as itemCode where(itemCode.length()!=5) -> tgtItem.productOrService = cc("http://terminology.hl7.org/CodeSystem/data-absent-reason","not-applicable") "add DAR";
```

aschey-forpeople: I feel like that might be a typo. It doesn't seem very user-friendly to return them smooshed together like that. Also the SAMHSA filter won't work correctly if the codes are like that because it won't match what's on the sensitive list.

apps/bfd-pipeline-idr/model/idr_prior_auth_item.py
```diff
@@ -0,0 +1,83 @@
...
+    pa_req_rec_dt: date
+    pa_decision_dt: Annotated[date, BeforeValidator(transform_null_date_to_min)]
+    pa_decision_exp_dt: Annotated[date, BeforeValidator(transform_null_date_to_max)]
+    service_cnts: int
+    svc_render_st: str
+    mr_count_ind: int
+    mr_count_st_dt: Annotated[date, BeforeValidator(transform_null_date_to_min)]
+    mr_count_end_dt: Annotated[date, BeforeValidator(transform_null_date_to_max)]
+    rrb_excl_ind: Annotated[str, BeforeValidator(transform_default_string)]
+    # TODO: ask if we need these ts sine same as header table
```
clewellyn-nava: TODO

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/PriorAuthorizationRepository.java
clewellyn-nava, line 1: Just a general question, why does PriorAuth need its own @Repository annotation / entity manager? Is this because it makes sense structurally, or it is necessary and this code couldn't exist inside the ClaimRepository?

apps/bfd-pipeline-idr/model/idr_prior_auth_item.py
mel1-G: oppsie no longer needed. I'll remove that comment!

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/PriorAuthorizationRepository.java
mel1-G: That's a good question. It doesn't but I did it this way because it made sense to me structurally wise. Prior authorization is different from ClaimBase. It doesn't fit the claim type patterns and most of the filters don't even apply to it which is why ClaimRepository treats it as something external and just merges it into ClaimAndAuthResult

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/filter/SourceFilterParam.java
```diff
@@ -56,4 +56,22 @@ public boolean matchesSystemType(@NotNull SystemType systemType) {
...
+    for (var orList : metaSourceSk) {
+      if (orList.isEmpty()) {
+        continue;
+      }
+
+      if (!orList.contains(MetaSourceSk.CWF)) {
+        return false;
+      }
+    }
+    return true;
```
aschey-forpeople: Double check me here but I think this is semantically equivalent? (allMatch returns true for an empty stream)
```suggestion
    return metaSourceSk.stream()
            .filter(orList -> !orList.isEmpty())
            .allMatch(orList -> orList.contains(MetaSourceSk.CWF));
```

mel1-G: You're right. I'll also update `TagCriteriaFilterParam` with this catch for the empty stream!

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -164,7 +164,9 @@ private List<ExplanationOfBenefit> getClaimsByBene(long beneSk, ClaimFilterOptio
             List.of(),
             Collections.emptyList());
     var claims = eobHandler.searchByBene(criteria, options, Optional.empty());
-    return getEobFromBundle(claims);
+    return getEobFromBundle(claims).stream()
+        .filter(c -> c.getUse() != ExplanationOfBenefit.Use.PREAUTHORIZATION)
```
aschey-forpeople: Instead of filtering these, I think we should extend these tests to ensure prior auth is respecting the filter properly. Snapshots aren't really adequate for this.

mel1-G: I have samhsa tests with prior auth in EobPriorAuthorizationIT. I wanted to just keep all of the tests for prior auth there. I only touched this file since I added prior auth eobs for this bene and the tests ended up failing due to prior auth resource IDs being uuids so this check breaks without the filter I added: `!skipBundleVerification.contains(Long.parseLong(c.getIdPart()))` We use that to filter for regular claims that have samhsa.

aschey-forpeople: Hmm I'll have to think about this a bit. `EobSamhsaFilterIT` has a lot of infrastructure set up to ensure the preconditions for the SAMHSA claims are set up appropriately so the tests don't produce false positives. It might be better to keep it there but I'll look some more.

apps/bfd-pipeline-idr/model/idr_prior_auth_item.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @override
+    @classmethod
+    def fetch_query(cls, partition: LoadPartition, start_time: datetime, source: Source) -> str:
+        # Prior auth data older than the lookback period should be filtered
+        prior_auth = ALIAS_PRIOR_AUTH
+        return f"""
+            WITH distinct_prior_auths AS (
+                SELECT *, ROW_NUMBER() 
+                    OVER (PARTITION BY mbi_num, utn, current_segment ORDER BY mbi_num) as row_order
```
marshallweekley-eng, line 76: This partition statement is different from the one in [apps/bfd-pipeline-idr/model/idr_prior_auth.py‎](https://github.com/CMSgov/beneficiary-fhir-data/pull/3201/changes#diff-a43e99f5d57458ce86d9c499a5d9052883b94f1eae8f5e5dc3da56ea6723448b). I do not know this data just double checking.

clewellyn-nava (dismissed):
I can find no faults and I understand the changes I'll need to incorporate into the profile work.

LGTM

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
mel1-G: Refactored in [440d86d](https://github.com/CMSgov/beneficiary-fhir-data/pull/3201/changes/440d86d28acf958f8a1e7108a24c0c00d5289830)

apps/bfd-pipeline-idr/model/idr_prior_auth_item.py
mel1-G: Answered in open topics but for posterity, in IDR there is only one view for prior auth `PRAUC` but they merged the line items into it so we split it out here. You can think of current segment like line number.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/util/IdrConstants.java
```diff
@@ -19,9 +19,17 @@ private IdrConstants() {
...
+  /** The date used by IDR to indicate a missing or non-applicable value in a date column. */
   public static final LocalDate DEFAULT_DATE = LocalDate.of(9999, 12, 31);
 
+  /**
+   * If there's any date with the year before 1582, Java will switch to the Julian calendar and
+   * misinterpret and convert the date. IDR has the alternate date 1000-01-01 to indicate a missing
+   * or non-applicable value in a date column so we use this check to convert those dates to
+   * optionals.
+   */
+  public static final LocalDate ALTERNATE_DEFAULT_DATE = LocalDate.of(1900, 1, 1);
```
aschey-forpeople: I think this name might make more sense.
```suggestion
  public static final LocalDate OLD_DATE_CUTOFF = LocalDate.of(1900, 1, 1);
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/converter/OptionalDateConverter.java
```diff
@@ -18,7 +18,9 @@ public LocalDate convertToDatabaseColumn(Optional<LocalDate> maybeDate) {
 
   @Override
   public Optional<LocalDate> convertToEntityAttribute(LocalDate value) {
-    if (value == null || value.equals(IdrConstants.DEFAULT_DATE)) {
+    if (value == null
+        || value.equals(IdrConstants.DEFAULT_DATE)
+        || value.isBefore(IdrConstants.ALTERNATE_DEFAULT_DATE)) {
```
aschey-forpeople: A bit more context as to why we're doing this may be helpful for posterity
```suggestion
        // Old/garbage dates should be normalized to null in the pipeline, but we'll add a failsafe here in case we forget any
        || value.isBefore(IdrConstants.ALTERNATE_DEFAULT_DATE)) {
```

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
```diff
@@ -71,6 +67,17 @@ class EobSamhsaFilterIT extends IntegrationTestBase {
   private static final long BENE_SK4 = 167719446;
   private static final long BENE_SK5 = 27590072;
 
+  private static final PriorAuthIdentifier NO_SAMHSA_PRIOR_AUTH_BENE_SK2 =
+      new PriorAuthIdentifier(String.valueOf(BENE_SK2), "-0Y2K7HP0Y71BO");
+  private static final PriorAuthIdentifier SAMHSA_PRIOR_AUTH_BENE_SK2 =
```
aschey-forpeople, line 72: A few more things to ensure we have covered here:
- at least one code from each system (HCPCS, CPT, HIPPS)
- Each SAMHSA code used should be covered in the `samhsaCodesToCheck` list
- A test that ensures the expected codes exist in the test data - see `ensureHcpcs` for an example. This is necessary to prevent false positives if the test data is changed accidentally.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobReadIT.snap
```diff
@@ -2540,6 +2559,45 @@ gov.cms.bfd.server.ng.EobReadIT.eobReadPhase2=[
...
+        "system" : "https://bluebutton.cms.gov/fhir/CodeSystem/Supporting-Information"
+      }, {
+        "code" : "info",
+        "display" : "Information",
+        "system" : "http://terminology.hl7.org/CodeSystem/claiminformationcategory"
+      } ]
+    },
+    "code" : {
+      "coding" : [ {
+        "code" : "A",
```
alex-dzeda: this isn't right

mel1-G: oppsie. Fixed the system [here](https://github.com/CMSgov/beneficiary-fhir-data/pull/3201/changes/a107871dcc15ed3d9e573261628ac5eedaa640c6#diff-5df8a37b5e6a4e2bf1a14aa4e419f11ed0f3c4d9294a4f5a3240bec0e40b7457R2597)

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/EobSamhsaFilterIT.java
mel1-G: Thanks for this! Fyi, even tho HIPPS aren't SAMHSA codes I still added it to check in `ensureHcpcsCptHipps`. Covered in [a107871](https://github.com/CMSgov/beneficiary-fhir-data/pull/3201/changes/a107871dcc15ed3d9e573261628ac5eedaa640c6)

aschey-forpeople (dismissed):

clewellyn-nava (dismissed):

JFisk42 (dismissed):

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobPriorAuthorizationIT.snap
```diff
@@ -0,0 +1,335 @@
...
+        }, {
+          "url" : "https://bluebutton.cms.gov/fhir/StructureDefinition/SVC-RENDER-ST",
+          "valueCoding" : {
+            "code" : "FL",
+            "system" : "https://www.usps.com"
+          }
+        } ],
+        "modifier" : [ {
+          "coding" : [ {
+            "system" : "https://www.cms.gov/Medicare/Coding/HCPCSReleaseCodeSets"
```
alex-dzeda: seems off

alex-dzeda: (duplicate)

alex-dzeda: or perhaps this is the default for when it's not populated? I'd expect it to just skip modifier entirely if not present.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobPriorAuthorizationIT.snap
```diff
@@ -0,0 +1,335 @@
...
+          "coding" : [ {
+            "code" : "32X",
+            "system" : "https://www.nubc.org/CodeSystem/TypeOfBill"
+          } ]
+        },
+        "sequence" : 2
+      } ],
+      "type" : {
+        "coding" : [ {
+          "code" : "H",
```
alex-dzeda: code = institutional for this coding

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobReadIT.snap
```diff
@@ -3391,6 +3449,45 @@ gov.cms.bfd.server.ng.EobReadIT.eobReadProfessionalClaim=[
...
+        "system" : "https://bluebutton.cms.gov/fhir/CodeSystem/Supporting-Information"
+      }, {
+        "code" : "info",
+        "display" : "Information",
+        "system" : "http://terminology.hl7.org/CodeSystem/claiminformationcategory"
+      } ]
+    },
+    "code" : {
+      "coding" : [ {
+        "code" : "00",
```
alex-dzeda: this may still be off. If I look at display = Void/Entry Code 3 Claim on the codesystem, I see the code should be V0008, but here I only see 00.

mel1-G: You're right. I brought it back in and cleaned up the other mapping issues that you and validator pointed out here [97d152c](https://github.com/CMSgov/beneficiary-fhir-data/pull/3201/changes/97d152cbab6d0967aebeaf5a9dc9e01191b22e3d)

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobPriorAuthorizationIT.snap
mel1-G: Good catch! I had that wonky so it populated even if they were empty. Fixed neow.

clewellyn-nava (dismissed):

alex-dzeda (dismissed):

alex-dzeda (approved):

aschey-forpeople (approved):

PR #3203, opened 2026-07-09
BFD-4348: Prune Older Claim Versions Excluding Part D

patrickgmayday:
**JIRA Ticket:**
BFD-4348

### What Does This PR Do?

Non part D claims pruning structured closely after Phase 1 claims pruning.
Introduction of associated indexes script

 Must ensure safe deletion of data.

* N/A
     Removal of old non part D claims 
    Be mindful of performance when running on DB

### Validation

    Validation driven by additions to test_pipline.py based on associated logic to this item.


apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -110,3 +127,127 @@ def extract_and_load(
...
+                    )
+                """,
+            (prune_cutoff_date,),
+        )
+
+        logger.info("pruning non-latest non-Part-D ss claims older than {}", prune_cutoff_date)
+
+        # Claim items can exist even when the non-latest parent claim was filtered
+        # before final claim-table load, so use the source claim table.
+        _prune_table_in_batches(
```
aschey-forpeople: Rather than running all of these sequentially, each pruning job should probably be `yield`ed independently so they can run concurrently.

apps/bfd-pipeline-idr/src/idr_pipeline/pipeline_stages.py
```diff
@@ -175,6 +177,18 @@ def _stage5_prune_obsolete_rows(self) -> Stage[bool]:
...
+                model,
+                self.load_mode,
+                self.start_time,
+            )
+            yield functools.partial(
+                prune_non_latest_non_part_d_ss_parent_claims,
+                model,
+                self.load_mode,
+                self.start_time,
+            )
```
aschey-forpeople, line 190: These two should run sequentially like we did with the phase 1 prune, otherwise we'll end up with orphaned records in the item tables.

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -865,3 +865,44 @@ def stale_phase_1_claims_query(
...
+    cutoff_date: datetime,
+) -> tuple[str, tuple[datetime]]:
+    part_d_codes = ",".join(str(code) for code in PART_D_CLAIM_TYPE_CODES)
+
+    return (
+        f"""
+            SELECT clm.clm_uniq_id
+            FROM {claim_table} clm
+            WHERE clm.clm_ltst_clm_ind = 'N'
+            AND clm.clm_type_cd NOT IN ({part_d_codes})
```
aschey-forpeople: I don't think we need the `clm_type_cd` filter here - all the part d claims are in the same table so we should just skip running this against the `claim_rx` table.

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -865,3 +865,44 @@ def stale_phase_1_claims_query(
...
+    item_table: str,
+    cutoff_date: datetime,
+) -> tuple[str, tuple[datetime]]:
+    part_d_codes = ",".join(str(code) for code in PART_D_CLAIM_TYPE_CODES)
+
+    return (
+        f"""
+            SELECT item.clm_uniq_id, item.bfd_row_id
+            FROM {item_table} item
+            JOIN {IDR_CLAIM_TABLE} clm ON clm.clm_uniq_id = item.clm_uniq_id
```
aschey-forpeople: We shouldn't use `IDR_CLAIM_TABLE` here - that refers to the source table in Snowflake

apps/bfd-pipeline-idr/src/idr_pipeline/pipeline_stages.py
patrickgmayday: Would it be feasible to do the following:

pipeline_stagles.py 


```
 yield functools.partial(
                prune_non_latest_non_part_d_ss_claims,
                model,
                self.load_mode,
                self.start_time,
            )


```


and then pipeline_utils.py

```
def prune_non_latest_non_part_d_ss_claims(
    cls: type[T],
    load_mode: LoadMode,
    job_start: datetime,
) -> bool:

    items_pruned = prune_non_latest_non_part_d_ss_claim_items(cls, load_mode, job_start)
    if not items_pruned:
        return False

    return prune_non_latest_non_part_d_ss_parent_claims(cls, load_mode, job_start)

```

to make sure that child items are deleted before parent? 

thank you!

aschey-forpeople: It could work but it would be better to delete each batch from both the item table and the header table sequentially like we do with the phase 1 prune, otherwise you leave a lot of orphaned header entries lying around for a while before they get deleted and it would leave the db in a somewhat inconsistent state if the pipeline gets stopped halfway through.

apps/bfd-db-migrator-ng/migrations/V1_38_0__add_non_latest_claim_prune_idx.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX ON idr.claim_professional_ss(clm_idr_ld_dt, clm_uniq_id) WHERE clm_ltst_clm_ind = 'N';
```
alex-dzeda: Is this being run out of band with the pipeline being suppressed while the index is created?

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -165,6 +169,54 @@ def prune_phase_1_ss_claims(
     return True
 
 
+def prune_non_latest_non_part_d_ss_claims(
```
JFisk42: Cosmetic suggestion: `stale` might work in place of `non latest`, at least that's what I finally settled on for the one I created. I only added it to the query name, but I don't have a strong defense for that pattern. My main thought here is that it would be nice to have unified verbiage in these method names.

apps/bfd-db-migrator-ng/migrations/V1_38_0__add_non_latest_claim_prune_idx.sql
```diff
@@ -0,0 +1,3 @@
...
+
+CREATE INDEX ON idr.claim_institutional_ss(clm_idr_ld_dt, clm_uniq_id) WHERE clm_ltst_clm_ind = 'N';
```
JFisk42: You should be able to remove `clm_uniq_id` from the index because your query is only filtering on `clm_idr_ld_dt` and `clm_ltst_clm_ind`.

apps/bfd-pipeline-idr/pipeline_utils.py
```diff
@@ -165,6 +169,54 @@ def prune_phase_1_ss_claims(
...
+    cls: type[T],
+    load_mode: LoadMode,
+    job_start: datetime,
+) -> bool:
+    claim_table = cls.table()
+    item_table = _SHARED_SYSTEM_CLAIM_ITEM_TABLES.get(claim_table)
+    if item_table is None:
+        return True
+
+    prune_cutoff_date = job_start - timedelta(days=PHASE_1_CUTOFF)
```
JFisk42: Not sure about the specifics of terminology here, but should `PHASE_1_CUTOFF` be given a more generic name?

apps/bfd-db-migrator-ng/migrations/V1_38_0__add_non_latest_claim_prune_idx.sql
aschey-forpeople: These indexes need to be created on the nch tables too

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/ClaimAsyncService.java
```diff
@@ -108,7 +113,9 @@ protected <T extends ClaimBase> CompletableFuture<List<T>> fetchClaims(
 
   private String buildWhereClause(DbFilter filter, SystemType systemType) {
     var latestClaimFilter =
-        systemType.filterLatestClaims() ? "AND c.latestClaimIndicator = 'Y'" : "";
+        systemType.filterLatestClaims()
```
aschey-forpeople: We don't need the changes in this file, this exclusion is already headed by `SystemType`

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -865,3 +865,24 @@ def stale_phase_1_claims_query(
...
+            WITH claims AS (
+                SELECT clm.clm_uniq_id
+                FROM {claim_table} clm
+                WHERE clm.clm_ltst_clm_ind = 'N'
+                AND clm.clm_idr_ld_dt < %s
+                ORDER BY clm.clm_idr_ld_dt, clm.clm_uniq_id
+            )
+            SELECT clm.clm_uniq_id
+            FROM claims clm
+            LIMIT {PHASE_1_PRUNE_BATCH_LIMIT}
```
JFisk42: I can't remember if this was discussed, but shouldn't the `LIMIT` be inside of the cte? I think it's still deterministic because of the `ORDER BY`, but unless the cte results are cached, the cte is going to pull all results every loop through. @aschey-forpeople 

This would also affect the phase 1 query.

aschey-forpeople: This one's a bit different since we don't have to join against the item table. It's fine to sort the original result set, the issue with the other one was that it was performing the join and sorting at the same time. The CTE is unnecessary here, it's not doing anything different than a single query block would.

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -870,3 +870,24 @@ def stale_phase_1_claims_query(
...
+    claim_table: str,
+    cutoff_date: datetime,
+) -> tuple[str, tuple[datetime]]:
+    return (
+        f"""
+            WITH claims AS (
+                SELECT clm.clm_uniq_id
+                FROM {claim_table} clm
+                WHERE clm.clm_ltst_clm_ind = 'N'
+                AND clm.clm_idr_ld_dt < %s
```
aschey-forpeople: We can omit this `AND` - it's not part of the prune criteria

apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -870,3 +870,24 @@ def stale_phase_1_claims_query(
...
+                ORDER BY clm.clm_idr_ld_dt, clm.clm_uniq_id
```
aschey-forpeople: Only ordering by clm_uniq_id would be better here since it's the pk

apps/bfd-db-migrator-ng/migrations/V1_38_0__add_non_latest_claim_prune_idx.sql
aschey-forpeople: Since we're removing the `clm_idr_ld_dt` filter, I think it should be this `CREATE INDEX ON idr.claim_institutional_ss(clm_ltst_clm_ind) WHERE clm_ltst_clm_ind = 'N';` - we already have an index on clm_uniq_id so no need to include it here.

apps/bfd-db-migrator-ng/migrations/V1_38_0__add_non_latest_claim_prune_idx.sql
patrickgmayday: Responding to this here based on discussion in scrum - yes, we won't want to have the tables locked while these are created.

apps/bfd-pipeline-idr/pipeline_utils.py
patrickgmayday: Removed since we aren't doing date based processing any longer (my logic was too coupled to the phase 1 claims)

apps/bfd-pipeline-idr/src/idr_pipeline/constants.py
```diff
@@ -9,9 +9,11 @@
 BENEFICIARY_TABLE = "idr.beneficiary"
 CLAIM_RX_TABLE = "idr.claim_rx"
 CLAIM_PROFESSIONAL_NCH_TABLE = "idr.claim_professional_nch"
+CLAIM_PROFESSIONAL_ITEM_NCH_TABLE = "idr.claim_item_professional_nch"
```
aschey-forpeople, line 12: While we're at it, can we change the model files (`idr_claim_item_professional_nch.py` etc) to use these constants as well?

apps/bfd-pipeline-idr/src/idr_pipeline/model/base_model.py
```diff
@@ -870,3 +870,13 @@ def stale_phase_1_claims_query(
         """,
         (cutoff_date, cutoff_date),
     )
+
+
+def stale_non_part_d_claims_query(claim_table: str) -> str:
+    return f"""
+        SELECT clm.clm_uniq_id
+        FROM {claim_table} clm
+        WHERE clm.clm_ltst_clm_ind = 'N'
```
aschey-forpeople, line 879: We need to omit synthetic data here
```suggestion
        WHERE clm.clm_ltst_clm_ind = 'N'
        AND clm.clm_uniq_id > 0
```

JFisk42 (dismissed):

apps/bfd-pipeline-idr/src/idr_pipeline/model/base_model.py
```diff
@@ -870,3 +870,14 @@ def stale_phase_1_claims_query(
...
+        AND clm.clm_uniq_id > 0
+        ORDER BY clm.clm_uniq_id
+        LIMIT {PHASE_1_PRUNE_BATCH_LIMIT}
```
aschey-forpeople: Should probably name this constant something more generic if we're gonna reuse it like this

apps/bfd-pipeline-idr/src/idr_pipeline/pipeline_stages.py
```diff
@@ -176,6 +177,13 @@ def _stage5_prune_obsolete_rows(self) -> Stage[bool]:
                 self.start_time,
             )
 
+        for model in self._filter_tables(CLAIM_TABLES):
```
aschey-forpeople: If `CLAIM_TABLES` includes everything except part D, we might want to give this a more descriptive name

aschey-forpeople: Oh I see, it's because part D is in `CLAIM_AUX_TABLES`. Seems a little funky though, I'll ask about that.

apps/bfd-pipeline-idr/src/idr_pipeline/settings.py
```diff
@@ -105,9 +105,8 @@ def enable_prior_auth_ingestion() -> bool:
 )
 """Maximum rows to delete per prune statement for Part D beneficiary records."""
 
-PHASE_1_PRUNE_BATCH_LIMIT = int(getenv("PHASE_1_PRUNE_BATCH_LIMIT", "10_000"))
-"""The maximum batch size for pruning old claims (phase 1 claims from shared systems) on
-INCREMENTAL loads. Defaults to 10000."""
+CLAIM_PRUNE_BATCH_LIMIT = int(getenv("CLAIM_PRUNE_BATCH_LIMIT", "10_000"))
+"""The maximum batch size for pruning claims on INCREMENTAL loads. Defaults to 10000."""
 
 BENEFICIARY_PRUNE_BATCH_LIMIT = int(getenv("IDR_BENEFICIARY_PRUNE_BATCH_LIMIT", "1000"))
```
aschey-forpeople: I think we can probably combine all of these batch prune limits into a single setting for now unless we have a need to use separate numbers. Right now 10,000 seems fine for all of them.

aschey-forpeople (approved):

JFisk42 (approved):

PR #3217, opened 2026-07-21
BFD-4788: Switch from filter out to pruning for part d obsolete records

marshallweekley-eng:
**JIRA Ticket:**
BFD-4788

### What Does This PR Do?

Switch from filter out to pruning for part d obsolete records

Check prune logic and new stage. It will need to reconcile with other prune work out there based on commit order.

Take note of the repeatable script for flyway. To date we have just created a new script to update a view. This works a little differently. Flyway checks the check sum of the repeatable file and if it changes it runs the script again. It is useful for it to be easy as an engineer to see the most up to date schema of a given view. They are typically helpful for concepts like views and functions and lets source control give revision history. They run last always in the flyway deployment. If we do not like the repeatable script its contents can be moved to an versioned script.

N/A

### Validation

Tests are updated to check for this scenario. I tested the view update and it prevents the server-ng from ever seeing the unpruned records.


mel1-G:
I'm sold on the repeatable script here and for future changes. I didn't know that existed. This is going to reduce a lot of the clutter of scripts. Nice find :) Happy to approve once your merge conflicts are resolved.

apps/bfd-db-migrator-ng/migrations/V1_35_0__add_bene_ma_part_d_obs_date_idx.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX ON idr.beneficiary_ma_part_d_enrollment(idr_trans_obslt_ts);
```
aschey-forpeople: We might be able to pare this down by making this conditional since there should be very few rows matching the condition here
```suggestion
CREATE INDEX ON idr.beneficiary_ma_part_d_enrollment(idr_trans_obslt_ts) WHERE idr_trans_obslt_ts < '9999-12-31';
```

marshallweekley-eng: I added it to this and the _rx table also.

mel1-G (dismissed):

apps/bfd-pipeline-idr/pipeline_stages.py
```diff
@@ -137,6 +138,20 @@ def _stage4_do_beneficiary(self) -> Stage[bool]:
             self._gen_partitioned_node_inputs(self._filter_tables(BENE_TABLES))
         )
 
+    def _stage5_do_phase_1_prune(self) -> Stage[bool]:
```
aschey-forpeople: Should this be renamed? Something like `_stage5_do_ma_part_d_prune`?

marshallweekley-eng: That is the same method that others are calling their prunes from. The util prun method that it calls is specific to the job. I need to resolve conflicts looking at this it will be more clear.

aschey-forpeople: Ah alright, I'm fine with renaming this once conflicts are fixed.

marshallweekley-eng: Resolved. If it needs renaming or broken to a new stage let me know

aschey-forpeople: I think it's fine to keep it in the same stage (we want these to run in parallel), but we should probably name it something more generic since it's not specific to phase 1 anymore.`_stage5_prune_obsolete_rows` or something.

marshallweekley-eng: Renamed. IT was very specific to a more broad concept

aschey-forpeople (approved):

mel1-G (approved):

PR #3227, opened 2026-07-27
BFD-4810: Implement Prior Auth Sync Strategy

mel1-G:
**JIRA Ticket:**
BFD-4810

### What Does This PR Do?

- Added support for a full sync and diff load strategy for prior auth and any other future tables that don't have reliable update timestamps so we can track changes and upstream deletions 
- Added a flag on the models to opt in to this strategy without changing how any other table is loaded
- I tried to reuse as much of the exisiting code as possible so I reused our extraction logic and detection of any updates + used some inheritance to build the new loading logic on top of the existing loader 
- Removed the IDR timestamps from the prior auth models since it's confirmed that IDR won't be providing those fields :(
- Added a safeguard so the deletion logic doesn't run on our synthetic data in prod

### Validation

Added prior auth update and delete tests to our pipeline tests


apps/bfd-pipeline-idr/model/base_model.py
```diff
@@ -486,6 +486,15 @@ def should_replace() -> bool:
         """Whether to merge or replace data when loading this table."""
         return False
 
+    @staticmethod
+    def should_fully_sync_delete_diff() -> bool:
```
aschey-forpeople: Maybe something like `should_delete_missing` is a more readable name?

apps/bfd-pipeline-idr/extractor.py
```diff
@@ -192,6 +192,14 @@ def extract_idr_data(
             {"timestamp": compare_timestamp},
         )
 
+    def extract_full_idr_data(self, source: Source) -> Iterator[list[T]]:
+        start_time = self.cls.model_type().min_transaction_date
+        fetch_query = self.get_query(start_time, source)
+        logger.info("extracting full {}", self.cls.table())
+        return self.extract_many(
+            fetch_query.replace("{LAST_TS}", "%(timestamp)s"), {"timestamp": start_time}
```
aschey-forpeople: I think `{MIN_TS}` is probably a better name for this placeholder since it's the minimum date, not the most recent one.

apps/bfd-pipeline-idr/src/idr_pipeline/loader.py
```diff
@@ -418,6 +422,77 @@ async def _copy_data(
                 )
 
 
+class FullSyncBatchLoader(BatchLoader):
+    async def load(self) -> bool:
```
aschey-forpeople, line 446: We should use `@override` on any methods that we override.

```suggestion
    @override
    async def load(self) -> bool:
```

apps/bfd-pipeline-idr/loader.py
```diff
@@ -418,6 +422,77 @@ async def _copy_data(
...
+            "{}-{}: finished full sync",
+            self.table,
+            self.partition.name,
+        )
+        return True
+
+    async def _delete_missing(self, cur: psycopg.AsyncCursor[Any], temp_tablename: str) -> int:
+        # We have to exclude our synthetic data that also exists in prod from deletion
+        synthetic_data_filter = (
+            "" if self.load_mode == LoadMode.SYNTHETIC else "WHERE utn NOT LIKE '-%'"
```
aschey-forpeople: I think it'd be best to avoid hardcoding a column that's specific to prior auth in a method that's shared like this

apps/bfd-pipeline-idr/src/idr_pipeline/loader.py
```diff
@@ -418,6 +422,77 @@ async def _copy_data(
...
+        timestamp = datetime.now(UTC)
+        self.full_load_timer.start()
+        num_rows = 0
+
+        async with self.pool.connection() as conn, conn.cursor(binary=True) as cur:
```
aschey-forpeople, line 451: A lot of this is copypasta'd from the parent method. Might be nice to break some of it into a helper function.

apps/bfd-pipeline-idr/loader.py
```diff
@@ -418,6 +422,77 @@ async def _copy_data(
...
+            )
+            await self._mark_batch_complete(cur)
+
+        self.full_load_timer.stop()
+        logger.info(
+            "{}-{}: finished full sync",
+            self.table,
+            self.partition.name,
+        )
+        return True
```
aschey-forpeople: I think we can track whether data was loaded here just like the parent method instead of hardcoding `True`

apps/bfd-pipeline-idr/loader.py
aschey-forpeople: Also, a string search on the secondary column from the primary key like this is gonna have perf issues. It might be better to make a separate index just on this column.

aschey-forpeople: Keep in mind that you might need to do a bit of research to make sure the index works with `NOT LIKE text%`

apps/bfd-pipeline-idr/test_pipeline.py
```diff
@@ -545,6 +545,77 @@ def _do_test_pipeline(conn: Connection[DictRow], load_type: LoadType) -> None:
         assert updated_ss_job.completion_time >= ss_clm_ts
 
 
+def _do_test_prior_auth_update_and_delete(conn: Connection[DictRow], load_type: LoadType) -> None:
+    if not enable_prior_auth_ingestion():
```
aschey-forpeople: I think we should default the setting for this to `true` now, and probably remove it entirely in a subsequent PR if it doesn't cause issues.

mel1-G: Just for the test pipeline, right?

aschey-forpeople: No, we want to enable this in prod now right? So we shouldn't need to default it to false.

apps/bfd-pipeline-idr/loader.py
```diff
@@ -417,6 +409,105 @@ async def _copy_data(
...
+class FullSyncBatchLoader(BatchLoader[T]):
+    @override
+    async def load(self) -> bool:
+        timestamp = datetime.now(UTC)
+        self.full_load_timer.start()
+        data_loaded = False
+
+        async with self.pool.connection() as conn, conn.cursor(binary=True) as cur:
+            await self._record_batch_start(conn, cur, commit=False)
+            full_temp_table = await self._setup_temp_table(cur, "full_temp", True)
```
aschey-forpeople: nit: I like to use named args for boolean parameters because just seeing `True` in an argument list doesn't tell you anything about what that param does.
```suggestion
            full_temp_table = await self._setup_temp_table(cur, "full_temp", copy_indexes=True)
```

apps/bfd-pipeline-idr/src/idr_pipeline/loader.py
```diff
@@ -417,6 +413,105 @@ async def _copy_data(
...
+        logger.info(
+            "{}-{}: finished full sync",
+            self.table,
+            self.partition.name,
+        )
+        return data_loaded
+
+    async def _delete_missing(self, cur: psycopg.AsyncCursor[Any], temp_tablename: str) -> int:
+        # We have to exclude our synthetic data that also exists in prod from deletion
+        synthetic_data_filter = self.model.synthetic_data_filter()
```
aschey-forpeople, line 494: I think this whole method should probably be a no-op for a synthetic load, otherwise we'll end up deleting all of the non-synthetic data when this runs

aschey-forpeople (dismissed):

apps/bfd-pipeline-idr/src/idr_pipeline/model/base_model.py
```diff
@@ -490,6 +490,20 @@ def should_replace() -> bool:
...
+
+        Upstream data can be deleted with no indicator like an obsolete timestamp, requiring
+        us to delete it on our end.
+        """
+        return False
+
+    @staticmethod
+    def synthetic_data_filter() -> str:
+        """Expression used to exclude synthetic data from being deleted in FullSyncBatchLoader."""
+        return ""
```
thomaswilson-eng, line 510: Are we really saying `WHERE ` when we want to filter synthetic data? Maybe I misunderstand

apps/bfd-db-migrator-ng/migrations/V1_42_0__add_utn_pattern_idx.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX ON idr.prior_auth (utn) WHERE utn NOT LIKE '-%';
```
thomaswilson-eng, line 1: Nice index this should work.

apps/bfd-pipeline-idr/src/idr_pipeline/model/base_model.py
mel1-G: @thomaswilson-eng what do you mean? I'll explain the use if that helps clarifies things. This method only gets overridden by IdrPriorAuth and IdrPriorAuthItem
```
@override
@staticmethod
def synthetic_data_filter() -> str:
    return "utn NOT LIKE '-%'"
```
and we use that in `_delete_missing` under loader.py to filter out synthetic data so it doesn't get deleted in prod or in a synthetic load:
```
synthetic_data_filter = self.model.synthetic_data_filter()
synthetic_where_clause = (
    f"WHERE {synthetic_data_filter}" if synthetic_data_filter and not test_mode() else ""
)
```

thomaswilson-eng: Okay figures its a form of python abstraction and I just didn't detect the context. Very nice. I saw the two other implementations, so that tracks. Thank you!

thomaswilson-eng (dismissed):

aschey-forpeople (dismissed):

thomaswilson-eng (dismissed):

aschey-forpeople (approved):

thomaswilson-eng (approved):

PR #3229, opened 2026-07-28
BFD-4576: Make IDR->BFD validation script run automatically on a schedule

malessi:
**JIRA Ticket:**
BFD-4576

### What Does This PR Do?

This PR refactors the original `idr_bfd_comparer.py` script into its own Python utility: `idr-bfd-validator`. To support this, the IDR Pipeline application (`bfd-pipeline-idr`/`idr-pipeline`) has been refactored into a [packaged application](https://docs.astral.sh/uv/concepts/projects/init/#applications) project such that it can be imported as a package by other Python code; in this case, `idr-bfd-validator`. Additionally, a new container Image (named `bfd-platform-idr-validator`) is now built from the `idr-bfd-validator` and a new Tofuservice, `04-idr-bfd-validator`, has been introduced to automatically run said container as an ECS Task automatically every hour.

- See above

### Validation

- Running the `idr-bfd-validator` locally, _verifying that_ it works as expected
- Running the `idr-pipeline` locally, _verifying that_ it still works as expected
- Running the `idr-pipeline` tests, _verifying that_ they still pass 
- `tf apply`ing `idr-bfd-validator` in `prod`, _verifying that_ it `apply`s successfully and the Task runs as expected and works


apps/bfd-pipeline-idr/Dockerfile.dockerignore
```diff
@@ -2,6 +2,6 @@
 !pyproject.toml
 !uv.lock
 !.python-version
-!model/*.py
-!matching/*.py
-!*.py
+!src/idr_pipeline/model/*.py
```
aschey-forpeople, line 5: Could we use a recursive wildcard here instead? Like `!src/idr_pipeline/**/*.py`? I know it supports recursive wildcards to some extent.

malessi: Doesn't look like either `!src/idr_pipeline/**.py` or `!src/idr_pipeline/**/*.py` work. The former seems to just be invalid glob syntax, and the latter just seems to unignore the `.py` files in `idr_pipeline` and none of the subdirectories like `model` or `matching`.

marshallweekley-eng (dismissed):

apps/bfd-pipeline-idr/Dockerfile.dockerignore
aschey-forpeople: Hmm, that's annoying. I think there's a high probability that we forget to update this when we add a new folder. What about just un-ignoring `**/*.py`? I'd rather have a few stray python files than a production failure due to missing files.

malessi: So, this is possible with Docker but apparently not with Podman (as we discovered yesterday—Podman just doesn't understand `!**/*.py` for whatever reason). Personally, I find the safety that an _opt-in_ approach for passing files to the build context to be worth the potential risk we forget to add a sub-directory to the `Dockerfile.dockerignore`.

I am going to still spend some time thinking about how to better handle this. At the very least, I'll try and think of some way to make it very obvious that new modules within `idr-pipeline` _must_ be added to the `Dockerfile.dockerignore`, if I can't come up with another workaround.

ops/services/04-idr-bfd-validator/main.tf
```diff
@@ -0,0 +1,252 @@
...
+            value = local.env
+          },
+          {
+            name  = "BFD_DB_ENDPOINT"
+            value = data.aws_rds_cluster.main.reader_endpoint
+          },
+          # TODO: Unexclude these tables when they work correctly
+          {
+            name  = "IDR_EXCLUDE_TABLES"
+            value = join(",", ["idr.prior_auth", "idr.beneficiary_low_income_subsidy_cmbnd"])
```
mel1-G: Also `idr.prior_auth_item` since we extracted out the line item fields from the original `idr.prior_auth`

malessi: Good catch! I'll add that one, too.

malessi: Done in [6960d7f](https://github.com/CMSgov/beneficiary-fhir-data/pull/3229/commits/6960d7f5c23ebce9e53e77d6195eec8dd1173b72).

apps/bfd-pipeline-idr/Dockerfile.dockerignore
malessi: In the interest of getting this in, I believe we have agreed to accept this for now and come back to this at another point.

If not, we can unresolve this conversation and we can continue to figure this out.

aschey-forpeople (approved):

marshallweekley-eng (approved):

PR #2341, opened 2024-06-06
BFD-3481: Flatten parameter hierarchy source

mjburling:
**JIRA Ticket:**
[BFD-3481](https://jira.cms.gov/browse/BFD-3481)

**User Story or Bug Summary:**
Flatten parameter hierarchy source.

---

### What Does This PR Do?

While working on related changes for better SSM Parameter Store management, we came to realize just how much more complicated things are with the nested hierarchies in source code. Before addressing other simplification and standardization, we seek to re-flatten the hierarchies. While the result looks somewhat busier, each line provides just about all the context that's needed for an operator or reviewer to orient themselves to the configuration at issue.

### How was this evaluated?

The following criteria were used to evaluate the correctness of the flattened structure:
1. Path-to-production environments exhibit no relevant changes when running terraform on this branch against the environments
2. Ephemeral environments created with the latest changes in the trunk/master branch exhibit no changes when terraform is run against that environment using the logic from this branch, and vice versa

The first case holds true for all path-to-production environments: there is no diff.
The second case largely holds true, but while working through these changes, I've addressed a minor bug for two values that should have been marked as UNDEFINED but were instead producing the string "null." Both parameter store values will be identified with a self-reviewed comment here.

### What Should Reviewers Watch For?
If you're reviewing this PR, please check for these things in particular:
* This change set is as small as possible and no smaller–reviewers will need to evaluate by cloning this branch and attempting to apply changes to existing environments with terraform:
  * for path-to-production environments, there should be no substantive changes
  * ephemeral environments should exhibit a diff on two unused variables:
    * `/bfd/${env}/pipeline/nonsensitive/rda/grpc/inprocess_server/s3/dir`
    * `/bfd/${env}/pipeline/nonsensitive/rda/grpc/inprocess_server/s3/bucket` 
* Verify all PR security questions and checklists have been completed and addressed.

### What Security Implications Does This PR Have?

Submitters should complete the following questionnaire:

* If the answer to any of the questions below is **Yes**, then **you must** supply a link to the associated Security Impact Assessment (SIA), security checklist, or other similar document in Confluence here: **N/A**

    * Does this PR add any new software dependencies? 
      * [ ] Yes
      * [x] No
    * Does this PR modify or invalidate any of our security controls?
      * [ ] Yes
      * [x] No
    * Does this PR store or transmit data that was not stored or transmitted before?
      * [ ] Yes
      * [x] No

[...]

PR #3230, opened 2026-07-28
BFD-4672: Create mappings to certificates to the partner

marshallweekley-eng:
**JIRA Ticket:**
BFD-4672

### What Does This PR Do?

Create mappings to certificates to the partner api that made the request for audit events.

Checking how we pull the cert to know what app is the orginator.

Note the mapping and make sure we agree with them.

### Validation

Unit test has been added that tests this condition in the DynamoDB.


apps/bfd-server-ng/src/main/resources/application.properties
```diff
@@ -1,6 +1,7 @@
 spring.application.name=bfd-server
 # Default to local configuration (overridden on deployment)
 spring.profiles.active=local
+spring.config.import=optional:classpath:certs.yml
```
aschey-forpeople: Should this be optional? I'm fine with making this required since we control the input here.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/PatientMatchIT.java
```diff
@@ -43,9 +44,12 @@ private Bundle searchBundle(Patient patient) {
     var ctx = FhirContext.forR4();
     var params = ParametersUtil.newInstance(ctx);
     ParametersUtil.addParameterToParameters(ctx, params, "IDIPatient", patient);
-
+    var headersInterceptor = new AdditionalRequestHeadersInterceptor();
+    headersInterceptor.addHeaderValue("X-Amzn-Mtls-Clientcert", "good_cert");
```
aschey-forpeople: Since we're hardcoding this header name a few places now, we may just want to make `CertificateUtil.LEAF_CERT_HEADER` public so we can reference that.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/DynamoDbAuditLogger.java
```diff
@@ -30,6 +32,8 @@ public void log(PatientMatchAuditRecord auditRecord) {
         auditEvent.setTimestamp(auditRecord.timestamp().toString());
         auditEvent.setClientId(auditRecord.clientId());
         auditEvent.setClientName(auditRecord.clientName());
+        auditEvent.setPartnerAppName(
```
aschey-forpeople, line 35: We may want to default to "unknown" for the actual app name and just log a warning or something if the name isn't found, that way we don't publicly leak internal details if we end up surfacing this on the audit log response.

apps/bfd-server-ng/src/main/resources/certs.yml
malessi, line 1: Hate to come in and derail this, but this should be subsumed into our existing configuration in each environment's `sopsw.yaml` file. Something like (for `prod`):

```yml
/bfd/${env}/server-ng/nonsensitive/certificate_aliases_json/BBAPI: '["bluebutton_backend_prod_data_server_client_test", "bluebutton_root_ca", "bluebutton_backend_test_samhsa"]'
# ... another for AB2D, BCDA, etc.
```

With this, we keep our configuration centralized and avoid the possibility of forgetting to update the certificate list whenever (really, if ever) we change trusted certificates.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/PatientMatchIT.java
marshallweekley-eng: Done

apps/bfd-server-ng/src/main/resources/application.properties
marshallweekley-eng: Done

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/log/DynamoDbAuditLogger.java
marshallweekley-eng: Done

apps/bfd-server-ng/src/main/resources/certs.yml
aschey-forpeople: That's a good point, I was thinking that this might be overkill to put in SSM since it's largely a constant and shouldn't change often, if at all, but keeping the cert aliases in one place is a good idea.

marshallweekley-eng: It is injectable via the mapping now. I can update the branch soon. It is json for all the partners. I did not do it partner specific because that is also hard coded to an extent. so now it will be a

`bfd.nonsensitive.cert_partners.partner_certificate_json`
 
and it will take that json and build the pairings.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/CertPartnersConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+  private CertPartnerProperties certPartners = new CertPartnerProperties();
+  private List<Partner> partners;
+
+  /**
+   * Map for certificate alias mapped to their partner name.
+   *
+   * @return Map
+   */
+  public Map<String, String> getPartnerNamesByCertificateAlias() {
+    if (partners == null) {
```
aschey-forpeople: You can automatically make this lazy load using `@Getter(lazy = true)` instead of the null check here. Same trick we use for `clientCertsToAliases`

marshallweekley-eng: Done

ops/services/01-config/values/test.sopsw.yaml
```diff
@@ -233,6 +233,7 @@
 /bfd/${env}/server/nonsensitive/500_errors_alerter_log_lookback_seconds: "330"
 /bfd/${env}/server/nonsensitive/500_errors_alerter_slack_webhook: bfd_test
 /bfd/${env}/server/nonsensitive/samhsa_allowed_certificate_aliases_json: '["bluebutton_root_ca", "bluebutton_backend_test_data_server_client_test", "test_client_samhsa_cert", "regression_tests"]'
+/bfd/${env}/server-ng/nonsensitive/cert_partners/partner_certificate_json: '[{"name":"BBAPI","certificate_aliases":["bluebutton_backend_prod_data_server_client_test","bluebutton_root_ca","bluebutton_root_ca_sbx","bb2_local_client","bluebutton_backend_dpr_data_server_client_test","bluebutton_root_ca_test","bluebutton_backend_test_samhsa"]},{"name":"AB2D","certificate_aliases":["ab2d_cms_gov","ab2d_cms_gov_non_prod","ab2d_prod_client","ab2d_prod_validation_client","ab2d_dev_prod_sbx_client","ab2d_sbx_client","ab2d_imp_client","ab2d_cms_gov_non_prod"]},{"name":"BCDA","certificate_aliases":["bcda_prod_client","bcda_sbx_client","bcda_dev_client","bcda_test_client","bcda_local_client"]},{"name":"DPC","certificate_aliases":["dpc_local_client","dpc_prod_client","dpc_prod_sbx_client","dpc_test_client","dpc_dev_client"]}]'
```
alex-dzeda: Can we go to "Blue Button API" instead of BBAPI? easier to understand if they eventually show to end users.

ops/services/01-config/values/prod.sopsw.yaml
```diff
@@ -749,6 +749,7 @@
   -----END CERTIFICATE-----
 /bfd/${env}/server-ng/nonsensitive/samhsa_allowed_certificate_aliases_json: '["bluebutton_backend_test_samhsa", "bluebutton_root_ca", "regression_tests"]'
 /bfd/${env}/server-ng/nonsensitive/internal_certificate_aliases_json: '["bluebutton_backend_prod_data_server_client_test", "bluebutton_backend_test_samhsa"]'
+/bfd/${env}/server-ng/nonsensitive/cert_partners/partner_certificate_json: '[{"name":"BBAPI","certificate_aliases":["bluebutton_backend_prod_data_server_client_test","bluebutton_root_ca","bluebutton_root_ca_sbx","bb2_local_client","bluebutton_backend_dpr_data_server_client_test","bluebutton_root_ca_test","bluebutton_backend_test_samhsa"]},{"name":"AB2D","certificate_aliases":["ab2d_cms_gov","ab2d_cms_gov_non_prod","ab2d_prod_client","ab2d_prod_validation_client","ab2d_dev_prod_sbx_client","ab2d_sbx_client","ab2d_imp_client","ab2d_cms_gov_non_prod"]},{"name":"BCDA","certificate_aliases":["bcda_prod_client","bcda_sbx_client","bcda_dev_client","bcda_test_client","bcda_local_client"]},{"name":"DPC","certificate_aliases":["dpc_local_client","dpc_prod_client","dpc_prod_sbx_client","dpc_test_client","dpc_dev_client"]}]'
```
malessi: nit: being _real_ nitpicky here, because this _works_, but the typical pattern for our configuration values is to avoid storing _objects_ in a stringified value. The idea behind this is that an object structure can be represented in the key itself, as we (and AWS, actually) treat `/` as denoting the next level in the hierarchy of the "object" tree.

E.g.

```yml
/bfd/test/server-ng/nonsensitive/cert_partners/partner_certificate_json: ...value...
```

can be thought of as _semantically_ equivalent to:

```yml
bfd:
  test:
    server-ng:
      nonsensitive:
        cert_partners:
          partner_certificate_json: ...value...
```

(and, for a little bit [a few years ago](https://github.com/CMSgov/beneficiary-fhir-data/pull/2341) we did structure our configuration with nested YAML instead of flat keys).

Thinking about it this way, separate keys for each partner alias (`BBAPI`, `DPC`, etc.) make sense:

```
/bfd/${env}/server-ng/nonsensitive/cert_partners/partner_certificate_json/BBAPI: ...just BBAPI certs...
/bfd/${env}/server-ng/nonsensitive/cert_partners/partner_certificate_json/DPC: ...just DPC certs...
```

now, the server-ng code can simply look at `partner_certificate_json` and inverse map the set of keys under that object to their values. This is the same exact strategy we use for `client_certiifcate` values; each key under `.../client_certificates/...` is automatically resolved in the server-ng code and an inverse map is built.

This makes the mappings more obvious because the mapping isn't _stringified_ in the value but rather is truly a YAML mapping to a value. Also, we avoid making our values _too_ large to store as SSM parameters, and we additionally get more granular deltas between updates for each partner's list.

I'm leaving this as a nit because the existing structure works fine and I'm OK with accepting it as-is given that this particular pattern was probably not obvious from reading the `01-config/README.md`.

malessi: The values for each partner in each environment should only contain the certificate aliases that exist in said environment. E.g. `bcda_test_client` should only be defined for BCDA in `test` and `sandbox` because it's trusted in those environments, but not `prod`.

Running the following ripgrep (`brew install rg`) in the `ops/services/01-config/values` directory will give you a list, per-environment, of the certificate aliases trusted in each environment:

```sh
rg --replace='$1' --only-matching '/bfd/.*/server-ng/.*/client_certificates/(.*):'
```

_This is less of a nit, so I'm leaving it as a separate comment on this line._

marshallweekley-eng: I did this as one config to not be locked in to requiring a code change if a new partner ever happened. I might be overthinking this and being too flexible and if that is the case I can break it down to a config per partner.

aschey-forpeople: It's very unlikely that we'll get a new partner in the foreseeable future, but even so, it shouldn't require a code change. If you look at `clientCertificates` in `Configuration.java` you can see how we did it last time - each leaf node gets automatically deserialized to an entry in a hashmap rather than a separate variable in the code.

ops/services/01-config/values/test.sopsw.yaml
marshallweekley-eng: Done

ops/services/01-config/values/prod.sopsw.yaml
marshallweekley-eng: Updated to allow a config per app partner

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/CertPartnersConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+  private static class CertPartnerProperties {
+    private Map<String, String> partnerCertificate = new HashMap<>();
+  }
+
+  /** Partner certificate configuration from cert_partners properties. */
+  @Data
+  private static class Partner {
+    private String name;
+
+    @SerializedName("certificate_aliases")
```
aschey-forpeople: Is this required? It should be able to convert from snake case to camel case automatically. We don't use this annotation elsewhere.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/CertPartnersConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Getter(lazy = true)
+  private final Map<String, String> partnerNamesByCertificateAlias =
+      getPartnerNamesByCertificateAliasInternal();
+
+  private Map<String, String> getPartnerNamesByCertificateAliasInternal() {
+    return certPartners.getPartnerCertificate().entrySet().stream()
+        .flatMap(
+            entry -> {
+              Partner partner = new Gson().fromJson(entry.getValue(), Partner.class);
```
aschey-forpeople:
```suggestion
              var partner = new Gson().fromJson(entry.getValue(), Partner.class);
```

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/CertPartnersConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+              return partner.getCertificateAliases().stream()
+                  .filter(Objects::nonNull)
+                  .map(alias -> Map.entry(alias, partner.getName()));
+            })
+        .collect(
+            Collectors.toMap(
```
aschey-forpeople: I think we can omit the third argument here. The default is to throw an exception if there's a duplicate key and I think that's preferred over silently ignoring it since we control this configuration.

ops/services/01-config/values/prod.sopsw.yaml
marshallweekley-eng: Done

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/CertPartnersConfiguration.java
marshallweekley-eng: It was because the value was in json and not the properties spring mapping.  I adjusted the name to be able to remove it.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/CertPartnersConfiguration.java
marshallweekley-eng: Removed

aschey-forpeople (approved):

patrickgmayday (approved):

PR #3244, opened 2026-08-05
BFD-4841: Improve IDR Pipeline error handling

malessi:
**JIRA Ticket:**
BFD-4841

### What Does This PR Do?

This PR fixes a variety of issues with the way the IDR Pipeline manages _unhandled `Exception`s_ between background/subprocess workers (e.g. `LoadingBatchWorker` or any of the `ParallelStageExecutor` jobs). 

The most notable improvements are:

- _All_ `Exception`s retain their full context and stack trace regardless of which process they were `raise`d in allowing us to more easily see where an unhandled error occurred
   - This may seem strange as why _wouldn't_ an `Exception` not have all of its context? This occurred because `Exception` "tracebacks" (a.k.a stack traces) are not preserved when an `Exception` is "pickled" (serialized) across processes. We now do this serialization manually using `tblib`, so all context is preserved
- Extraneous noise caused when sibling processes are forcibly stopped by `raise`ing `Exception`s in a signal handler  has been suppressed making it much easier to see the _real_ `Exception` in the logs
- `Exception`s `raise`d  when extracting/loading a table/`BaseModel`-derived model are now wrapped by a custom `Exception` that is explicit about _which_ table failed to load

Altogether, the error handling is significantly improved and we should be able to determine the cause of issues much more easily.

- See above

   * Yes,  `tblib`. I will not merge until @sb-benohe approves.

### Validation

- `raise`ing various `Exception`s in multiple different parts of the IDR Pipeline that run in subprocesses, _validating that_:
   - The full stack trace and `Exception` context is logged upon exit
   - No extraneous noise from the subprocess stopping is logged
   - The IDR Pipeline exits completely and near-immediately, as expected

### DIFF
--- a/apps/bfd-pipeline-idr/src/idr_pipeline/batch_worker.py
+++ b/apps/bfd-pipeline-idr/src/idr_pipeline/batch_worker.py
@@ -553,7 +553,7 @@ class LoadingBatchWorkerManager:
         try:
             async with anyio.create_task_group() as tg:
                 tg.start_soon(watch_queue)
-        except* Exception:
+        except BaseException:
             cancel_signal.set()
             raise
 
--- a/apps/bfd-pipeline-idr/src/idr_pipeline/batch_worker.py
+++ b/apps/bfd-pipeline-idr/src/idr_pipeline/batch_worker.py
@@ -3,6 +3,7 @@ import os
 import random
 import signal
 import string
+import sys
 import threading
 import uuid
 from collections import Counter, OrderedDict
@@ -13,6 +14,7 @@ from itertools import batched
 from math import ceil
 from multiprocessing import Manager, Process
 from multiprocessing.managers import SyncManager
+from pathlib import Path
 from queue import Empty, Queue
 from threading import Event
 from types import FrameType
@@ -28,6 +30,11 @@ from psycopg import sql
 from psycopg.errors import DeadlockDetected, InFailedSqlTransaction
 from psycopg_pool.abc import ACT
 
+from .exception_utils import (
+    SerializedExceptionChain,
+    rebuild_exception_chain,
+    serialize_exception_chain,
+)  # type: ignore
 from .load_partition import LoadPartition
 from .model.base_model import DbType, IdrBaseModel
 from .model.load_progress import LoadProgress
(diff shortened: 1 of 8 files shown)


apps/bfd-pipeline-idr/src/idr_pipeline/batch_worker.py
```diff
@@ -200,24 +207,31 @@ def __init__(
         self._running_tasks: set[_Task] = set()
 
     def run(self) -> None:
-        self._logger.reinstall()
+        sys.unraisablehook = lambda _: None
+        sys.excepthook = lambda _, __, ___: None
+        with Path(os.devnull).open("w") as devnull:
+            sys.stderr = devnull
```
malessi, line 216: These lines (and the equivalent in `parallel_executor`) suppress any default unhandled-`Exception` logging subproceses do that have been causing significant log bloat when the IDR Pipeline stops due to an unhandled `Exception`.

I don't believe setting `sys.unraisablehook` or `sys.excepthook` really does much, but it doesn't _hurt_ either. The real solution seems to have been directing each subprocesses' `stderr` to `/dev/null`; we don't want them to write to `stderr` anyways, so this is fine.

apps/bfd-pipeline-idr/src/idr_pipeline/exception_utils.py
```diff
@@ -0,0 +1,84 @@
+from dataclasses import dataclass
+from types import TracebackType
+from typing import Any, cast
+
+from tblib import Traceback  # type: ignore
```
malessi, line 5: Unfortunately, `tblib` has no real type-hinting and does not ship any type stubs nor does it mark its modules with `py.typed`, so we have to have `# type: ignore` whenever we use it.

apps/bfd-pipeline-idr/src/idr_pipeline/batch_worker.py
aschey-forpeople: Might be worth adding something like this as a comment in the code. I assume whoever looks at this might be a bit confused as to why we're overwriting these.

malessi: Good point. I've added comments in [a754885](https://github.com/CMSgov/beneficiary-fhir-data/pull/3244/commits/a754885c3683fcba52c182558cb40a27c83fb57b).

aschey-forpeople (approved):

JFisk42 (approved):

PR #3248, opened 2026-08-07
BFD-4492: Shared Systems Parity CLM_VAL_CD

JFisk42:
**JIRA Ticket:**
BFD-4492

### What Does This PR Do?
This PR adds several new fields to be routed through CLM_VAL and consolidates the previously existing ones to use the same path. 
* Old columns have been removed if their values are stored in CLM_VAL (**NOT** in the DB, but elsewhere)
  * A SQL script to migrate the old values will need to be created.
* Structure Definition added for ClaimValues, mappings for CLM_VAL fields updated accordingly
* Generator code updated to handle CLM_VAL fields
* test_samples CSVs have had old columns removed and that data migrated to CLM_VAL_AMT/CLM_VAL_CD
* server-ng ClaimValues refactored so that ClaimValues own's it's related values

Anything and everything. There's a lot going on here.
In particular: 
* How the ClaimsValue structure was mapped in bfd-model-idr
* How the files were placed within the new bfd-server-ng hierarchy

### Validation
# todo

### Related Tickets
BFD-4862: This ticket will drop the obsolete fields from the db.


apps/bfd-model-idr/dictionary-support-files/ExplanationOfBenefit.yaml
```diff
@@ -1936,7 +1914,7 @@
     - Basis
     - Regular
     - CMS
-- inputPath: ExplanationOfBenefit-Base.CLM_BLOOD_PT_FRNSH_QTY
+- inputPath: ExplanationOfBenefit-Base.CLM_BLOOD_PT_FRNSH_QTY #TODO
```
JFisk42: In progress, I think I have all the info for these now

apps/bfd-model-idr/sushi/input/fsh/Extension-CLM-OPRTNL-IME-AMT.fsh
```diff
@@ -1,6 +1,6 @@
 Extension: CLM_OPRTNL_IME_AMT
 Title: "Operating Indirect Medical Education Amount"
-Description: "This is one component of the total amount that is payable on PPS claims, and reflects the IME (indirect medical education) payments for operating expenses (such as labor) for the claim."
+Description: "# TODO FOR EACH NEW CLAIM VALUE. This is one component of the total amount that is payable on PPS claims, and reflects the IME (indirect medical education) payments for operating expenses (such as labor) for the claim."
```
JFisk42: I'm not 100% sure on if we want an extension for each of these new fields? I think if they do get a CLM_VAL code system then it feels a little redundant, but I'm not sure how much information we want to pass along.
Example of what this might look like:
``` json
{
  "supportingInfo" : [{
    "sequence" : 1,
    "category" : {
      "coding" : [{
        "system" : "http://terminology.hl7.org/CodeSystem/claiminformationcategory",
        "code" : "info",
        "display" : "Information"
      },
      {
        "system" : "https://bluebutton.cms.gov/fhir/CodeSystem/Supporting-Information",
        "code" : "CLM_BLOOD_PT_FRNSH_QTY",
        "display" : "Blood Pints Furnished Quantity"
      },
      {
        "system" : "https://bluebutton.cms.gov/fhir/CodeSystem/CLM-VAL-CD",
        "code" : "37",
        "display" : "CLM_BLOOD_PT_FRNSH_QTY"
      }    ]
    },
    "valueQuantity" : {
      "value" : 15,
      "unit" : "pint",
      "system" : "http://unitsofmeasure.org",
      "code" : "[pt_us]"
    }
  }
]}
```

apps/bfd-model-idr/sushi/input/fsh/StructureDefinition-CLM-VAL.fsh
```diff
@@ -0,0 +1,9 @@
+Logical: ExplanationOfBenefit-CLM-VAL
```
JFisk42: How this should be implemented (in terms of how to structure) here and in `ExplanationOfBenefit-Base.fsh` is my biggest point of confusion.

apps/bfd-model-idr/sushi/input/fsh/CodeSystem-Adjudication.fsh
```diff
@@ -41,7 +41,7 @@ Description: "This code system includes a selection of Adjudication Value Codes
 * #CLM_LINE_DMERC_SCRN_SVGS_AMT "Screen Savings Amount" "The amount of savings attributable to the coverage screen for this DMERC line item."
 * #CLM_BENE_PD_AMT "Amount Paid To Beneficiary Amount" "An amount paid to a beneficiary for a submitted claim."
 * #CLM_OPRTNL_DSPRTNT_AMT "Operating Disproportionate Share Amount" "This is one component of the total amount that is payable on prospective payment system (PPS) claims and reflects the DSH (disproportionate share hospital) payments for operating expenses (such as labor) for the claim."
-* #CLM_OPRTNL_IME_AMT "Operating Indirect Medical Education Amount" "This is one component of the total amount that is payable on PPS claims, and reflects the IME (indirect medical education) payments for operating expenses (such as labor) for the claim."
+* #CLM_OPRTNL_IME_AMT "TODO? Operating Indirect Medical Education Amount" "This is one component of the total amount that is payable on PPS claims, and reflects the IME (indirect medical education) payments for operating expenses (such as labor) for the claim."
```
JFisk42: No longer a `TODO?` but a `TODO`, the new adjudication related fields need to be added here.

apps/bfd-model-idr/maps/ExplanationOfBenefit-Base.map
```diff
@@ -17,7 +17,8 @@ uses "http://hl7.org/fhir/StructureDefinition/ExplanationOfBenefit" alias BFDEOB
 
 imports "https://bfd.cms.gov/MappingLanguage/Maps/ExplanationOfBenefit-Helper" 
 imports "https://bfd.cms.gov/MappingLanguage/Maps/ExplanationOfBenefit-SupportingInfo-Helper" 
-imports "https://bfd.cms.gov/MappingLanguage/Maps/ExplanationOfBenefit-Item-Helper" 
+imports "https://bfd.cms.gov/MappingLanguage/Maps/ExplanationOfBenefit-Item-Helper"
+// imports "https://bfd.cms.gov/MappingLanguage/Maps/CLM_VAL"
```
JFisk42: Ignore, leftover from a different attempt

apps/bfd-model-idr/maps/ExplanationOfBenefit-Base.map
```diff
@@ -190,6 +190,9 @@ group createEOBBase(source src: ExplanationOfBenefitBase, target tgt: BFDEOB){
 
     src.profComponents as profComponents -> tgt as profTgt then addProfessionalItems(src, profTgt) "Add professional components";
     
+    src.claimValues as claimValues -> tgt as cvSiTgt then addSupportingClmVal(src, cvSiTgt) "Add Supporting info claim vals";
+    src.claimValues as claimValues -> tgt as cvAdTgt then addAdjudicationClmVal(src, cvAdTgt) "Add Adjudication claim vals";
```
JFisk42: The question for the .map files is if this is being structured correctly, i.e. should they live here in a simple group like this.

apps/bfd-model-idr/sushi/input/fsh/Extension-CLM-OPRTNL-IME-AMT.fsh
alex-dzeda: I don't think we need a CLM-VAL-CD codesystem for now. For example, the token https://bluebutton.cms.gov/fhir/CodeSystem/Supporting-Information| CLM_BLOOD_PT_FRNSH_QTY conveys the meaning of what this element is for. Likewise, for those that end up in adjudication elements, you can just update/add the descriptions in the adjudication codesystem.

apps/bfd-model-idr/maps/ExplanationOfBenefit-Base.map
alex-dzeda: since we're only planning to populate some of them, this is right, and you can do 

src.claimValues as claimValues where(claimValues.CLM_VAL_CD in ('1234' | '4567')) -> 

And then these will only execute on the relevant ones (syntax may need refinement but that's the gist.

apps/bfd-model-idr/sushi/input/fsh/StructureDefinition-CLM-VAL.fsh
thomaswilson-eng: Instinct says you are on the right track with what you have so-far...someone can check that.

@alex-dzeda is this another one of these situations were we would want to extend apps/bfd-model-idr/maps/ExplanationOfBenefit-Helper.map adding another conceptmap for CLM_VAL_CD mapping to some StructureDefinitionUrl? Then can simplify the actual mapping implementation by calling transform() instead of all the specific handler methods like addBloodPintsFurnished all for this field CLM_BLOOD_PT_FRNSH_QTY?  and removing the explicit calls like  from SupportingInfo.map for the more generic translate implementation with CLM_VAL_CD and CLM_VAL_AMT?

apps/bfd-model-idr/maps/ExplanationOfBenefit-Base.map
```diff
@@ -573,6 +552,67 @@ group addProcedures(source src: ProcedureComponent, target tgt: BackboneElement)
...
+group addSupportingClmVal(source src: ExplanationOfBenefitBase, target tgt: BackboneElement){
+  src.claimValues as pintsFurnished where claimValues="73" -> tgt then addBloodPintsFurnished(src, tgt) "Add blood pints furnished";
+  
+  // more if necessary
+}
+
+group addAdjudicationClmVal(source src: ExplanationOfBenefitBase, target tgt: BackboneElement){
+  src.claimValues as claimValues -> tgt then {
+
+    claimValues.CLM_VAL_CD as valCode where valCode="18" -> tgt.adjudication = create('BackboneElement') as tgtBBElement then {
```
thomaswilson-eng: Fairly certain you can use a concept map to make all of these CLM_VAL_CD rules condense into one by mapping the CLM_VAL_CD to the "CML_OPRTNL_IME_AMT" as a random example. See ExplanationOfBenefit-Helper.map for some examples. I'm learning translate() right now too, so is good timing.

JFisk42: Thanks for that tip, I had glossed over how that was being used. This will be super helpful!

apps/bfd-model-idr/out/ExplanationOfBenefit-HHA.json
```diff
@@ -1847,15 +1844,86 @@
...
     "amount" : {
       "value" : 12342.14,
       "currency" : "USD"
     }
   },
+  {
+    "category" : {
+      "coding" : [{
+        "system" : "https://bluebutton.cms.gov/fhir/CodeSystem/Adjudication",
+        "code" : "CLM_SQSTRTN_RDCTN_AMT"
```
JFisk42, line 1859: The display field for the adjudicated claim values has been dropped. This is only for these sample jsons and the server does output the display string. 

I'm unsure of the importance of these json files being faithful to our server's output. A simple fix would be to use a translate mapping and have the displays live inside `ExplanationOfBenefit-Helper.map`. However, if it's not necessary then it's one less thing to track when these are updated.

apps/bfd-model-idr/sushi/input/fsh/StructureDefinition-SupportingInfoComponent.fsh
```diff
@@ -13,8 +13,9 @@ Id: SupportingInfoComponent
 * DGNS_DRG_CD 0..1 string "Related Diagnosis Group Code" ""
 * BENE_PTNT_STUS_CD 0..1 string "Patient Discharge Status Code" ""
 * CLM_ADMSN_TYPE_CD 0..1 string "Claim Admission Type Code" ""
-* CLM_BLOOD_PT_FRNSH_QTY 0..1 string "Blood Pints Furnished Quantity" ""
 * CLM_MDCR_INSTNL_MCO_PD_SW 0..1 string "MCO Paid Switch" ""
+* claimValues 0..* BFDClaimValues "Claim Values List" "A list of claim values filtered by CLM_VAL_CD"
+  * ^type.targetProfile = "https://bfd.cms.gov/fhir/StructureDefinition/BFDClaimValues"
```
JFisk42, line 18: `claimValues` is present here but not in the adjudication fsh in part to allow the suppInfo mapping to set a sequence for the blood pints.

apps/bfd-model-idr/sample-data/EOB-Inpatient-FISS-Sample.json
```diff
@@ -172,6 +171,52 @@
         "CLM_MDCR_NPMT_RSN_CD": "P",
         "CLM_FI_ACTN_CD": "1"
     },
+    "claimValues": [
```
JFisk42, line 174: I couldn't find what json is generated from the sample data in this file, but still updated it.

apps/bfd-pipeline-idr/test_samples1/SYNTHETIC_CLM_VAL.csv
```diff
@@ -83,3 +83,104 @@ CLM_DT_SGNTR_SK,CLM_NUM_SK,GEO_BENE_SK,CLM_TYPE_CD,CLM_VAL_CD,CLM_VAL_AMT,CLM_VA
 159777952560,1,75220,40,19,11443.97,3,2024-05-22 09:31:05.458675,
 746386566763,1,90844,1013,18,11670.22,14,2025-02-08 15:29:01.262826,
 746386566763,1,90844,1013,19,11443.97,3,2024-05-22 09:31:05.458675,
+659052754348,1,11121,71,37,16,5,2024-09-20 06:28:48.128184,
```
JFisk42: The values for the CLM_VAL CSVs were taken from the eliminated columns on the other tables.

apps/bfd-server-ng/src/main/resources/application-local.properties
```diff
@@ -17,7 +17,7 @@ bfd.nonsensitive.client_certificates.samhsa_allowed=samhsa_allowed
 bfd.nonsensitive.client_certificates.samhsa_not_allowed=samhsa_not_allowed
 bfd.nonsensitive.samhsa_allowed_certificate_aliases_json=["samhsa_allowed"]
 bfd.nonsensitive.cert_partners.partner_certificate.bbapi={"name":"Blue Button API","certificateAliases":["bluebutton_backend_prod_data_server_client_test","bluebutton_root_ca","bluebutton_root_ca_sbx","bb2_local_client","bluebutton_backend_dpr_data_server_client_test","bluebutton_root_ca_test","bluebutton_backend_test_samhsa"]}
-bfd.nonsensitive.cert_partners.partner_certificate.ab2d={"name":"AB2D","certificateAliases":["ab2d_cms_gov","ab2d_cms_gov_non_prod","ab2d_prod_client","ab2d_prod_validation_client","ab2d_dev_prod_sbx_client","ab2d_sbx_client","ab2d_imp_client","ab2d_cms_gov_non_prod"]}
+bfd.nonsensitive.cert_partners.partner_certificate.ab2d={"name":"AB2D","certificateAliases":["ab2d_cms_gov","ab2d_cms_gov_non_prod","ab2d_prod_client","ab2d_prod_validation_client","ab2d_dev_prod_sbx_client","ab2d_sbx_client","ab2d_imp_client"]}
```
JFisk42, line 20: This popped up as an error at one point while I was working on the snapshots. There was an issue because `ab2d_cms_gov_non_prod` is in this list twice.

apps/bfd-db-migrator-ng/migrations/V1_43_0__drop_clm_val_fields.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE idr.claim_institutional_ss
```
aschey-forpeople: I think we're gonna want to move this to a separate PR. We don't want to drop these columns until after the pipeline and server changes have been deployed since it's a breaking change.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/professional/entities/ClaimProfessionalCmsNch.java
```diff
@@ -74,7 +72,6 @@ public class ClaimProfessionalCmsNch extends ClaimProfessionalBase {
             claimDispositionCode.map(c -> c.toFhir(supportingInfoFactory)),
             claimQueryCode.map(c -> c.toFhir(supportingInfoFactory)),
             nchWeeklyProcessingDate.toFhir(supportingInfoFactory),
-            bloodPints.toFhir(supportingInfoFactory),
             claimPaymentDenialCode.map(c -> c.toFhir(supportingInfoFactory)))
```
JFisk42, line 77: BloodPints has been removed as I could not find any professional clm_type_cd values for it in snowflake. 

Query so you can double check my math:
``` sql 
SELECT clm_type_cd, count(clm_type_cd)
FROM "IDRC_PRD"."CMS_VDM_VIEW_MDCR_PRD"."V2_MDCR_CLM_VAL" 
WHERE clm_val_cd IN ('37')
GROUP BY clm_type_cd
ORDER BY clm_type_cd;
```

apps/bfd-db-migrator-ng/migrations/V1_43_0__drop_clm_val_fields.sql
JFisk42: Ah you beat me to it! I was going to bring this up to discuss which path is preferred. Agreed that this makes handling this PR more straightforward. Logged a jira https://jira.cms.gov/browse/BFD-4862

apps/bfd-model-idr/claims_generator.py
```diff
@@ -984,18 +980,17 @@ def generate(
             ]
             adj_clms_tbls[CLM_DCMTN].extend(clm_dcmtns)
 
+            # todo verify clm val fields are appropriate for these clm type cds
```
JFisk42: Ooops, I lost this round of "spot the todo before opening the PR". Fixed https://github.com/CMSgov/beneficiary-fhir-data/pull/3248/changes/3f9e489b11ca7c0e0413262620c2675e45f56691

apps/bfd-db-migrator-ng/migrations/V1_43_0__drop_clm_val_fields.sql
JFisk42: Removed https://github.com/CMSgov/beneficiary-fhir-data/pull/3248/commits/7dfabdac7634f20ffd77a227a5f9048e0b6440b9

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/converter/OptionalBigDecimalToAmountConverter.java
```diff
@@ -0,0 +1,25 @@
...
+
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.util.Optional;
+
+/** Converts null BigDecimal values to an empty Optional and rounds to two decimal places. */
+@Converter
+public class OptionalBigDecimalToAmountConverter
```
aschey-forpeople, line 11: Nit: maybe call this `OptionalBigDecimalRoundingConverter`? Since that would describe the differences from the other `OptionalBigDecimalConverter` a bit clearer

apps/bfd-server-ng/src/main/resources/application-local.properties
aschey-forpeople: I think this is fixed in master now too

apps/bfd-model-idr/claims_generator.py
```diff
@@ -985,17 +986,17 @@ def generate(
...
-                ime_clm_val = adj_util.gen_ime_clm_val(
-                    clm=clm,
-                    init_clm_val=ime_clm_val_per_fpk.get(four_part_key(clm)),
-                )
-                adj_clms_tbls[CLM_VAL].append(ime_clm_val)
+                for sqnc_num, value_code in enumerate(AVAILABLE_CLM_VAL_CDS, start=1):
+                    # 'QM' is the only clm_val_cd with a limited clm_type_cd range
+                    if value_code == "QM" and clm_type_cd != 40:
+                        continue
+                    clm_val_row = adj_util.gen_clm_val(
```
JFisk42, line 993: There should be some probabilistic guard here otherwise all possible codes will be generated for all claims. That is too much.

JFisk42:
Update: Moving this back to draft as it was discovered that the relationship between CLM_INSTNL_PRFNL_AMT and the CLM_VAL table is more complex than CLM_VAL_CD == 05. This will be de-scoped and removed before the PR is reopened.

JFisk42:
Closing this ticket due to two additional changes in scope: 1) Noted above, the removal of CLM_INSTNL_PRFNL_AMT 2) The splitting of the server implementation into it's own ticket to reduce PR review burden. See https://jira.cms.gov/browse/BFD-4492?focusedId=4648311&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4648311 for more details.

PR #3257, opened 2026-08-13
BFD-4661: Use derived NPI type for claim care team providers

patrickgmayday:
**JIRA Ticket:**
BFD-4661

### What Does This PR Do?

Map NPI Type columns and us provider_npi_type_expr from the pipeline.
Replace GetNPI logic to honor the logic from the pipeline.

If it's not an NPI -> Organization
If it is an npi -> check if type 1 or type 2

     N/A
     N/A
     Carries additional NPI Type field for provider classes from pipeline to server
     N/A

### Validation

Tested against existing unit and integration tests with updated snapshots.


apps/bfd-db-migrator-ng/migrations/V1_39_0__add_claim_care_team_provider_npi_type.sql
```diff
@@ -0,0 +1,37 @@
...
+ADD COLUMN bfd_prvdr_rfrg_npi_type integer;
+
+ALTER TABLE idr.claim_item_professional_nch
+ADD COLUMN bfd_prvdr_rndrng_npi_type integer;
+
+ALTER TABLE idr.claim_item_professional_ss
+ADD COLUMN bfd_prvdr_rndrng_npi_type integer;
+
+ALTER TABLE idr.claim_rx
+ADD COLUMN bfd_prvdr_prscrbng_npi_type integer;
```
mel1-G: There's a few missing here:

1. idr.claim_professional_ss
- bfd_blg_prvdr_npi_type

2. idr.claim_professional_nch
- bfd_blg_prvdr_npi_type

3. idr.claim_institutional_nch
- bfd_blg_prvdr_npi_type

4. idr.claim_institutional_ss
- bfd_blg_prvdr_npi_type

5. idr.claim_rx
- bfd_srvc_prvdr_npi_type
- bfd_srvc_prvdr_gnrc_id_npi_type

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/rx/ServiceProviderPharmacy.java
```diff
@@ -36,13 +41,9 @@ public CareTeamType getCareTeamType() {
...
   @Override
-  @Transient
   public ProviderHistoryBase.NpiType getNpiType() {
-    if (providerFirstName.isEmpty() || getProviderNpiNumber().isEmpty()) {
-      return ProviderHistoryBase.NpiType.ORGANIZATION;
-    } else {
-      return ProviderHistoryBase.NpiType.INDIVIDUAL;
-    }
+    return ProviderHistoryBase.NpiType.fromNpiTypeCode(
+        npiType.isPresent() ? npiType : genericIdNpiType);
```
mel1-G: I checked in prod and there are cases where records have both `prvdr_srvc_prvdr_npi_num` and `clm_srvc_prvdr_gnrc_id_num` populated so setting the npi type for just one isn't correct.

patrickgmayday: I just noticed I'm doing the following:

```
   bfd_srvc_prvdr_npi_type: Annotated[
        int | None,
        {EXPR: provider_npi_type_expr(ALIAS_PRVDR_SRVC)},
    ]
    bfd_srvc_prvdr_gnrc_id_npi_type: Annotated[
        int | None,
        {EXPR: provider_npi_type_expr(ALIAS_PRVDR_SRVC)},
    ]

```

should bfd_srvc_prvdr_gnrc_id_npi_type be using the same alias or have its' own?

patrickgmayday: I have one more follow up question: In cases where we have both provider Npi and generic Npi populated, 

would we expect the following logic to change (with callers) to accommodate both, since this only uses `getNpiType` 

```
public Optional<DomainResource> toFhirNpiType() {
    return getProviderNpiNumber()
        .map(
            npi ->
                (getNpiType() == NpiType.ORGANIZATION
                    ? createOrganization(npi)
                    : createBillingPractitioner(npi)));
  }
```

patrickgmayday: From discussion:

```
If it's not an NPI -> Organization
If it is an npi -> check if type 1 or type 2
```

apps/bfd-db-migrator-ng/migrations/V1_40_0__add_claim_provider_npi_type.sql
```diff
@@ -0,0 +1,43 @@
...
+ALTER TABLE idr.claim_item_professional_nch
+ADD COLUMN bfd_prvdr_rndrng_npi_type integer;
+
+ALTER TABLE idr.claim_item_professional_ss
+ADD COLUMN bfd_prvdr_rndrng_npi_type integer;
+
+ALTER TABLE idr.claim_rx
+ADD COLUMN bfd_prvdr_prscrbng_npi_type integer,
+ADD COLUMN bfd_srvc_prvdr_gnrc_id_npi_type integer,
+ADD COLUMN bfd_srvc_prvdr_npi_type integer;
```
mel1-G: `bfd_srvc_prvdr_npi_type` can be removed from here and in the rx model since we don't need it. We'll only be using `bfd_srvc_prvdr_gnrc_id_npi_type` for the service provider,

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/common/ProviderHistoryBase.java
mel1-G, line 36: I think this should be reworked because if `npiTypeCode` is empty then that maps to `UNKNOWN` not `ORGANIZATION`

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/common/CareTeamBase.java
```diff
@@ -10,27 +10,19 @@
 @MappedSuperclass
 public abstract class CareTeamBase extends ProviderHistoryBase {
   private Optional<ProviderSpecialtyCode> specialtyCode;
+  private Optional<Integer> npiType;
+
+  @Override
+  protected NpiType getNpiType() {
+    return NpiType.fromNpiTypeCode(npiType);
+  }
```
mel1-G: Since this is the same for each class now I think you can just get rid of these overrides and create the following under `ProviderHistory`:
```
public NpiType getNpiType() {
    return NpiType.fromNpiTypeCode(npiType);
  }
```
so this means you'd also need npiType in `ProviderHistory` and @AttributeOverrides for all classes that inherit it. Only ServiceProviderPharmacy will override it since it has that special logic with the qualifier code.

mel1-G (changes requested):

mel1-G (dismissed):

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/__snapshots__/EobPriorAuthorizationIT.snap
```diff
@@ -14,8 +14,7 @@ gov.cms.bfd.server.ng.EobPriorAuthorizationIT.eobSearchPriorAuthorizationByPatie
               } ]
             },
             "value" : "1570418214"
-          },
-          "type" : "Practitioner"
```
aschey-forpeople, line 18: Hmm why are these being removed?

mel1-G: I checked this npi and it doesn't exist in the synthetic provider history data under `test_samples2` so it wasn't able to pull in the names which caused the NPI type to be null. @patrickgmayday I think you can regenerate prior auth using that test data by just deleting the 3 related NPI columns and running the generator with it. The provider history generator might be in a funky state tho IIRC so you might end up having to also generate provider history from scratch and the NPIs in the synthetic claims data but try out my first suggestion and lmk how that goes

patrickgmayday: > I checked this npi and it doesn't exist in the synthetic provider history data under `test_samples2` so it wasn't able to pull in the names which caused the NPI type to be null. @patrickgmayday I think you can regenerate prior auth using that test data by just deleting the 3 related NPI columns and running the generator with it. The provider history generator might be in a funky state tho IIRC so you might end up having to also generate provider history from scratch and the NPIs in the synthetic claims data but try out my first suggestion and lmk how that goes

I seem to be having some pretty weird issues with my local data generation, if you have steps on how best to manage running that regen fix I can take care of it. Apologies in the delay.

mel1-G (approved):

aschey-forpeople (approved):

PR #3260, opened 2026-08-14
BFD-4830: Only update lastUpdated for NPI type backfill if NPI type actually changes

mel1-G:
**JIRA Ticket:**
BFD-4830

### What Does This PR Do?

- Added an opt-in process so backfilling the new npi_type columns onto previously loaded claims doesn't bump the last updated column for all claims. Bumps only if the legacy npi_type != actual npi_type. Legacy npi_type is just what we computed in the server in the past. Actual npi_type will now be calculated in the pipeline as done in 4661. 
- Added `NPI_TYPE_BACKFILL_COMPARE` and `npi_type_backfill_compare_cols()` to the base model. `npi_type_backfill_compare_cols()`  builds a {legacy npi type column: real npi type column} map per model
- Added `legacy_*_npi_type_expr` helpers that reproduce the server logic so we can determine the legacy npi type and compare with the actual npi type
- Extended the temp table and COPY to carry the legacy npi type vs actual npi type comparison columns. Also modified `on_conflict_where_clause` to also run if there's only an npi_type mismatch and to use the correct ts for `bfd_udated_ts` depdending on match/mismatch

MISC
- fix a tiny bug with the claims generate. Handle empty `clm[f.CLM_RLT_COND_SGNTR_SK]` so we avoid an integer conversion errors
- Regenerated the synthetic data under `test_samples1` since it's very outdated and the NPIs did not match those in the claims
- Removes duplicate certificate alias from `bfd.nonsensitive.cert_partners.partner_certificate.ab2d` so we can run the app locally
- Fixed some server mapping issues. 
  - Claim professional NCH was missing the service care team member. Previously it was implemented with just the referring care team.
  - Fixed `OtherProfessionalSharedSystemsCareTeam` and changed the implementation of `getCareTeamType()` so it takes in the claim type code. I did this because the care team type under `OtherProfessionalSharedSystemsCareTeam` can be `OTHER` if the claim type code falls under DME otherwise the care team type is actually `SUPERVISOR`

### Validation

Added pipeline tests that verified that the last updated timestamp was updated if there was an npi type change compared to the legacy version and also checked that the last updated timestamp was NOT updated if the npi types matched


apps/bfd-pipeline-idr/src/idr_pipeline/model/base_model.py
```diff
@@ -867,18 +881,57 @@ def stale_phase_1_claims_query(
...
+
+
+def legacy_service_npi_type_expr_for_context(is_institutional: bool) -> str:
+    return "1" if is_institutional else "NULL"
+
+
+def legacy_rx_prescribing_npi_type_expr(provider_qualifier_code: str) -> str:
+    return f"""
+        CASE
+            WHEN {provider_qualifier_code} <> '' THEN 1
```
aschey-forpeople: nit: `!=`

apps/bfd-db-migrator-ng/migrations/V1_40_0__add_derived_npi_types.sql
```diff
@@ -0,0 +1,42 @@
+ALTER TABLE idr.claim_institutional_nch
```
mel1-G: Note this script will be removed once 4661 is merged in

thomaswilson-eng:
Java changes looking great.

apps/bfd-model-idr/claims_adj.py
```diff
@@ -131,7 +131,7 @@ def gen_clm(
             clm[f.CLM_FINL_ACTN_IND] = "N"
 
         clm_ltst_clm_ind = "N"
-        if clm_type_cd in (1, 2, 3, 4, 10, 20, 30, 40, 50, 60, 61, 62, 63, 71, 72, 81, 82):
+        if clm_type_cd in (1, 2, 3, 4, 10, 20, 30, 40, 50, 60, 61, 62, 63, 71, 72, 81, 82) and probability(0.10):
```
thomaswilson-eng, line 134: Only 10% of the time? Why?

mel1-G: This is mainly arbitrary. I just needed a low percentage. For our synthetic data we don't want the latest claim indicator to be set to N for most of the claims since they will get filtered out and a lot of the synthetic beneficiaries won't have claims. This came up from a [user ask](https://cmsgov.slack.com/archives/CMT1YS2KY/p1786483572136839).

thomaswilson-eng: Awesome. Its to curate better more useful data. I can totally understand that.

apps/bfd-pipeline-idr/src/idr_pipeline/model/idr_claim_institutional_nch.py
```diff
@@ -33,8 +33,10 @@
     COLUMN_MAP,
     EXPR,
     HISTORICAL_BATCH_TIMESTAMP,
+    INSERT_EXCLUDE,
```
thomaswilson-eng, line 36: I'm close to comfortable with what's going on here overall. What is this INSERT_EXCLUDE for?

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/PriorAuthRenderingCareTeam.java
```diff
@@ -1,9 +1,6 @@
 package gov.cms.bfd.server.ng.claim.model;
 
-import gov.cms.bfd.server.ng.claim.model.common.CareTeamType;
-import gov.cms.bfd.server.ng.claim.model.common.ClaimContext;
-import gov.cms.bfd.server.ng.claim.model.common.ProviderFhirHelper;
-import gov.cms.bfd.server.ng.claim.model.common.ProviderHistoryBase;
+import gov.cms.bfd.server.ng.claim.model.common.*;
```
thomaswilson-eng: This looks like the intellij replace imports with * auto-import setting for java needs to be bumped up to 99.

thomaswilson-eng (changes requested):

apps/bfd-pipeline-idr/src/idr_pipeline/model/idr_claim_institutional_nch.py
mel1-G: @thomaswilson-eng  `INSERT_EXCLUDE` is when you don't want a column to be inserted into a table like my legacy npi type fields. I only use those to determine if we need to update the last updated fields if there's a change for the npi type.

apps/bfd-pipeline-idr/src/idr_pipeline/loader.py
```diff
@@ -125,6 +131,58 @@ def __init__(
             f"WHERE ({', '.join(f't.{v}' for v in self.update_set)}) IS "
             f"DISTINCT FROM ({', '.join(f'EXCLUDED.{v}' for v in self.update_set)})"
         )
+        self.npi_type_backfill_map = model.npi_type_backfill_compare_cols()
+        self.npi_type_backfill_enabled = bool(self.npi_type_backfill_map)
```
aschey-forpeople: I think we should probably make this configurable via a setting as well so we can be intentional about when we enable it. We only want to enable it when we run the backfill.

apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/claim/model/PriorAuthRenderingCareTeam.java
mel1-G: Thanks for catching this! :) I changed my setting and brought back the explicit imports.

apps/bfd-pipeline-idr/src/idr_pipeline/loader.py
mel1-G: Done [here](https://github.com/CMSgov/beneficiary-fhir-data/pull/3260/changes/2fd2bb1603981da8330fa0d493030c8a42b3f5b7). I wish this supported :salute-slay: but you can use your imagination.

aschey-forpeople (approved):

JFisk42 (approved):

thomaswilson-eng (approved):

PR #3265, opened 2026-08-20
BFD-4796: Investigate possible claims data inconsistencies between IDR and BFD

malessi:
**JIRA Ticket:**
BFD-4796

### What Does This PR Do?

This PR updates the `idr-bfd-validator` to resolve a number of false-negatives for several tables that were previously failing validation. Additionally, a reporting feature has been introduced for local usage to help an operator diagnose validation failures. Also, the matching process has been refactored to enable matching even when row counts differ and `click` has been introduced to better manage options.

- See above

### Validation

- Running the updated validator locally, _verifying that_:
   - new click options are respected as expected
   - previously erroneously failing tables no longer fail to validate
   - reports are generated as expected


ops/services/04-idr-bfd-validator/main.tf
```diff
@@ -153,10 +153,9 @@ resource "aws_ecs_task_definition" "this" {
           # TODO: Unexclude these tables when they work correctly
           {
             name = "IDR_EXCLUDE_TABLES"
-            value = join(",", [
+            value = join(" ", [
```
malessi, line 160: `click` splits non-`Path`/`File`, `multiple=True` options' environment variables values by whitespace rather than by comma, so we need to update this.

apps/utils/idr-bfd-validator/main.py
```diff
@@ -63,6 +85,50 @@
...
+        dict[tuple[type[IdrBaseModel]], tuple[str, list[DbType]]],
+        {
+            (
+                IdrClaimProfessionalSs,
+                IdrClaimItemProfessionalSs,
+                IdrClaimInstitutionalSs,
+                IdrClaimItemInstitutionalSs,
+            ): (
+                f"{ALIAS_CLM}.idr_updt_ts > {{}}",
+                [(datetime.now(UTC) - timedelta(days=PHASE_1_CUTOFF)).date().isoformat()],
```
JFisk42: Just now familiarizing myself with the validator so I may answer this question myself later... does any of the other new prune job logic need to be accounted for?

apps/utils/idr-bfd-validator/main.py
```diff
@@ -62,6 +85,50 @@
...
+_ADDITIONAL_WHERE_CLAUSES_PER_MODEL = {
+    k: v
+    for keys, v in cast(
+        dict[tuple[type[IdrBaseModel]], tuple[str, list[DbType]]],
+        {
+            (
+                IdrBeneficiaryMaPartDEnrollment,
+                IdrBeneficiaryMaPartDEnrollmentRx,
+                IdrBeneficiaryLowIncomeSubsidyCmbnd,
+            ): ("(idr_trans_obslt_ts = {} OR idr_trans_obslt_ts = null)", [DEFAULT_MAX_DATE]),
```
JFisk42: The new bene prune queries also filter on `WHERE idr_trans_obslt_ts < %s`, does that need to be added here?

JFisk42, line 106: The new bene prune queries also filter on `WHERE idr_trans_obslt_ts < %s`, does that need to be added here?

Mostly this is a question of intent. Are you trying to eliminate as many rows as possible that would fail the check? 

Edit: There's also a couple discrepancies between the claim table clauses below and the prune queries (ex: using `idr_updt_ts` instead of `bfd_updated_ts`)

malessi: > The new bene prune queries also filter on WHERE idr_trans_obslt_ts < %s, does that need to be added here?

Nope. The prune queries filter on that to _remove_ rows, so we need to the filter on the inverse of that condition (that being `idr_trans_obslt_ts = '9999-12-31'`) to ensure we don't get back any rows that would be pruned.

> Mostly this is a question of intent. Are you trying to eliminate as many rows as possible that would fail the check?

Exactly. We need to remove rows that would be pruned, otherwise the validator will fail to match them in the BFD DB and will erroneously report that there were validation failures.

apps/utils/idr-bfd-validator/main.py
```diff
@@ -62,6 +85,50 @@
...
+    for keys, v in cast(
+        dict[tuple[type[IdrBaseModel]], tuple[str, list[DbType]]],
+        {
+            (
+                IdrClaimProfessionalSs,
+                IdrClaimItemProfessionalSs,
+                IdrClaimInstitutionalSs,
+                IdrClaimItemInstitutionalSs,
+            ): (
+                f"{ALIAS_CLM}.idr_updt_ts > {{}}",
```
aschey-forpeople: The prune query is based on `bfd_update_ts` so it might be better to keep that consistent. In practice it shouldn't matter _too_ much.

aschey-forpeople: Also I think the `Ss` tables also need the `clm_ltst_clm_ind` check like you have for the NCH ones below.

malessi: > The prune query is based on bfd_update_ts so it might be better to keep that consistent. In practice it shouldn't matter too much.

That isn't possible because this `WHERE` condition is added to the IDR query's `clm_base_query` part, not the BFD query, because IDR is the source of truth in this validation model. So, there's no such `bfd_updated_ts` column.

> Also I think the Ss tables also need the clm_ltst_clm_ind check like you have for the NCH ones below.

Ah, you're right. Good catch. I'll add that check to those tables, too.

malessi: And, I just realized the latest claim check is wrong 🤦. It needs to be inverted, right now it's just pulling back rows that would be pruned. I'll fix that, too.

malessi: Both issues fixed in [d569d16](https://github.com/CMSgov/beneficiary-fhir-data/pull/3265/commits/d569d16088aaf61ad671364e914826b063d15ab3).

apps/utils/idr-bfd-validator/main.py
malessi: > Edit: There's also a couple discrepancies between the claim table clauses below and the prune queries (ex: using idr_updt_ts instead of bfd_updated_ts)

I answered this in my response to Austin's comment:

> That isn't possible because this WHERE condition is added to the IDR query's clm_base_query part, not the BFD query, because IDR is the source of truth in this validation model. So, there's no such bfd_updated_ts column.

apps/utils/idr-bfd-validator/main.py
```diff
@@ -227,33 +375,119 @@ def _compare_table(
...
+                # Some rows in the BFD DB still have `1001-01-01` when the equivalent IDR row
+                # returns `null`. We need to replace those values with null as those are treated
+                # as null
+                return transform_default_date_to_null(d)
+            case _:
+                pass
+
+    match val:
+        case datetime() as d:
+            # Some primery key columns (specifically bene_cmbnd_deemd_efctv_dt for
```
aschey-forpeople: If we notice things like these, we should probably fix them in the model. Guessing these are just mistakes because it shouldn't be a datetime in the model if it's a date in the db.

malessi: Yeah, that's fair. Should I make the change to that column in this PR, then?

aschey-forpeople: Yeah I think that's a good idea, there should be no additional changes required besides just changing the datatype on the model class.

malessi: Sounds good to me. I'll do that and get rid of this fixup logic on the validator side.

malessi: Fixed the column in [129194a](https://github.com/CMSgov/beneficiary-fhir-data/pull/3265/commits/129194a4b49e80a6f436838d625d643147c7a5aa) and verified that the table still validates successfully without the workaround.

aschey-forpeople (approved):

JFisk42 (approved):

PR #3301, opened 2026-09-15
BFD-4890: Fix null pointer for not found audit events

marshallweekley-eng:
**JIRA Ticket:**
BFD-4890

### What Does This PR Do?

Fix null pointer for not found audit events

N/A

Moving the not found up to the resource provider cause hapi to format the response correctly. When done from the Repository or data provider it does not know the path for a proper error message.

### Validation

I have tested  via rest call and unit test.


apps/bfd-server-ng/src/main/java/gov/cms/bfd/server/ng/audit/AuditEventRepository.java
```diff
@@ -56,9 +55,7 @@ public AuditEvent findById(AuditEventId id) {
       var key = Key.builder().partitionValue(id.beneId()).sortValue(id.toDynamoSortKey()).build();
       var item = getTable().getItem(key);
 
-      return Optional.of(item)
-          .map(AuditEventBase::toFhir)
-          .orElseThrow(() -> new ResourceNotFoundException(id.getIdAsString()));
+      return Optional.ofNullable(item).map(AuditEventBase::toFhir).orElse(null);
```
aschey-forpeople: I think this should return `Optional<AuditEvent>` rather than propagate the null value here. That way we contain the chance for a NPE to just the local function.

aschey-forpeople: It would be good to add a test case for this as well.

marshallweekley-eng: Done on both fronts.

apps/bfd-server-ng/src/test/java/gov/cms/bfd/server/ng/PatientMatchIT.java
```diff
@@ -728,8 +729,9 @@ void testAuditEventByIdInvalid() {
     assertThrows(ResourceNotFoundException.class, res::execute);
   }
 
-  @Test
-  void testAuditEventByIdNotFound() {
+  @ParameterizedTest
+  @ValueSource(strings = {"999999999999-20260101000000000", "-731271436-20260914164040787646024"})
+  void testAuditEventByIdNotFound(String badId) {
     var res =
         getFhirClient().read().resource(AuditEvent.class).withId("999999999999-20260101000000000");
```
aschey-forpeople: I think you forgot to change the test to actually use the input param here

marshallweekley-eng: Yup good catch.

aschey-forpeople (approved):

patrickgmayday (approved):

PR #3303, opened 2026-09-16
BFD-4893: Initial prek config

aschey-forpeople:
**JIRA Ticket:**
BFD-4893

### What Does This PR Do?

This is the first step in making our development workflow a bit more automated and streamlined.

To use this, simply run `./bootstap.sh` from the root.

Details:

- Adds initial prek config for running pre-commit hooks
- Additive-only change for now (except that it replaces the existing hook)
- Everything is configured to only run on modified files so we can adopt it incrementally (we can run it to format every file in the repo if we want)
- The java format step now runs on pre-commit rather than pre-push since it only needs to run on the modified files and should complete much faster
- Supports the following checks
  - Formatting for all common filetypes we use
  - Automatically fixing common things such as newlines at end of file, CRLF/LF, etc
  - Runs our existing checks - shellcheck, sops, gitleaks, etc.
- Removes the need to manually run commands like `terraform-docs` and `mvn fmt` 
- All checks are also checked in CI in case of local misconfiguration

Python is notably absent from the initial config here due to the complexity of running in a monorepo. That will come later if this initial config seems to work well. I've also decided not to remove any of our other configs that this could replace so this is easy to revert if it ends up not working well.

I've also added a very simple `Justfile` here. `just` is a command runner that works well for multi-lingual environments. This can serve as the basis to add more common commands and automated setup steps.

### Validation

Ran config locally + verified CI job passes


patrickgmayday (dismissed):

prek.toml
```diff
@@ -0,0 +1,63 @@
...
+rev = "v8.24.2"
+hooks = [{ id = "gitleaks" }]
+
+[[repos]]
+repo = "local"
+hooks = [
+  { id = "java", name = "java", language = "system", types = [
+    "java",
+  ], entry = "./hooks/java-format.sh" },
+  { id = "sops", name = "sops", language = "system", files = "\\.sopsw.yaml$", entry = "./hooks/sops.sh" },
```
clewellyn-nava, line 55: Should it be `files = "\.sopsw\.yaml$"`?

aschey-forpeople: Nope, it has to escape the slash so it passes `\.` along to the regex engine to escape the period.

marshallweekley-eng (dismissed):

prek.toml
clewellyn-nava: Right, then wouldn't it be `files = "\\.sopsw\\.yaml$"`

aschey-forpeople: Ah yep good point. It still worked when I tested it since the files still match the regex but it would be better to change that.

aschey-forpeople:
```suggestion
  { id = "sops", name = "sops", language = "system", files = "\\.sopsw\\.yaml$", entry = "./hooks/sops.sh" },
```

clewellyn-nava (approved):

patrickgmayday (approved):

thomaswilson-eng (approved):
working for me.

mel1-G (approved):

